PR #2, opened 2016-09-09
add cells from paper http://jmlr.org/papers/volume17/14-203/14-203.pdf

gundramleifert:
These cells are tested in 2-dimensional RNNs and some of them yield to much better results. For 1-dimensional RNNs there are no good test - maybe one of you RWTH guys can substitute the LSTMs by these cells and check the performance. Especially for long sequences they maybe outperform the LSTM.

### DIFF
--- a/NetworkRecurrentLayer.py
+++ b/NetworkRecurrentLayer.py
@@ -130,6 +130,142 @@ class LSTME(Unit):
     # return: next outputs (# unit.n_act, y_t, c_t, ...)
     return (y_t * i_output, c_t * i_h + c_p * (1 - i_h)) + tuple(other_outputs)
 
+class Leaky(Unit):
+  """
+  A 1D cell proposed in http://jmlr.org/papers/volume17/14-203/14-203.pdf
+  The simplified equations can be seen in Table 7, page 36.
+  Type A with gamma_3==0.
+  This cell has 3 units instead of 4 like LSTM
+  """
+  def __init__(self, n_units, **kwargs):
+    super(LSTME, self).__init__(
+      n_units=n_units,
+      n_in=n_units * 3,  # forget gate (FG), output gate (OG), net input (IN)
+      n_out=n_units,
+      n_re=n_units * 3,
+      n_act=2  # output, cell state
+    )
+    self.o_output = T.as_tensor(numpy.ones((n_units,), dtype='float32'))
+    self.o_h = T.as_tensor(numpy.ones((n_units,), dtype='float32'))
+
+  def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
+    # See Unit.scan() for seqs.
+    # args: seqs (x_t = unit.xc, z_t, i_t), outputs (# unit.n_act, y_p, c_p, ...), non_seqs (none)
+    other_outputs = []
+    if self.recurrent_transform:
+      state_vars = other_args[:len(self.recurrent_transform.state_vars)]
+      self.recurrent_transform.set_sorted_state_vars(state_vars)
+      z_r, r_updates = self.recurrent_transform.step(y_p)
+      z_t += z_r
+      for v in self.recurrent_transform.get_sorted_state_vars():
+        other_outputs += [r_updates[v]]
+    z_t += T.dot(y_p, self.W_re)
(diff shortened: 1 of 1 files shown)


doetsch:
That was fast, thanks a lot! I will try it asap and will let you know about the results.

NetworkRecurrentLayer.py
```diff
@@ -130,6 +130,142 @@ def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
     # return: next outputs (# unit.n_act, y_t, c_t, ...)
     return (y_t * i_output, c_t * i_h + c_p * (1 - i_h)) + tuple(other_outputs)
 
+class Leaky(Unit):
```
doetsch: Units need to be specified in all capital letters (the specification in the config file is case insensitive and will be capitalized to create the object). I.e. Leaky -> LEAKY, and maybe let's say LEAKYLSTM since the same idea could also be applied to GRUs for example.

NetworkRecurrentLayer.py
```diff
@@ -130,6 +130,142 @@ def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
...
+  """
+  A 1D cell proposed in http://jmlr.org/papers/volume17/14-203/14-203.pdf
+  The simplified equations can be seen in Table 7, page 36.
+  Type A with gamma_3==0.
+  This cell has 3 units instead of 4 like LSTM
+  """
+  def __init__(self, n_units, **kwargs):
+    super(LSTME, self).__init__(
```
doetsch: Please provide the correct class name in the call of the super constructor (LSTME -> LEAKY)

NetworkRecurrentLayer.py
```diff
@@ -130,6 +130,142 @@ def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
...
+    c_t = (1-forgetgate) * input + forgetgate * c_p
+    # y(t) = OG(t) * c(t) HINT: There can be added an additional nonlinearity (substitute c_t:=T.tanh(x_t))
+    y_t = outgate * c_t
+    i_output = T.outer(i_t, self.o_output)
+    i_h = T.outer(i_t, self.o_h)
+    # return: next outputs (# unit.n_act, y_t, c_t, ...)
+    return (y_t * i_output, c_t * i_h + c_p * (1 - i_h)) + tuple(other_outputs)
+
+
+class LeakyLP(Unit):
```
doetsch: same here (capitalization required)

NetworkRecurrentLayer.py
```diff
@@ -130,6 +130,142 @@ def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
...
+  """
+  A 1D cell proposed in http://jmlr.org/papers/volume17/14-203/14-203.pdf
+  The simplified equations can be seen in Table 7, page 36.
+  Type A.
+  This cell has 4 units like the LSTM
+  """
+  def __init__(self, n_units, **kwargs):
+    super(LSTME, self).__init__(
```
doetsch: same here (super constructor)

NetworkRecurrentLayer.py
```diff
@@ -130,6 +130,142 @@ def step(self, i_t, x_t, z_t, y_p, c_p, *other_args):
...
+  A 1D cell proposed in http://jmlr.org/papers/volume17/14-203/14-203.pdf
+  The simplified equations can be seen in Table 7, page 36.
+  Type E. This cell works as a dynamic PID filter of the input. The forget gate
+  determines if it has PD od PI characteristic, the Proportional gate gates the P/I part,
+  the Difference gate the D/P part. It can have advantages if there is no subsampling in
+  the layer.
+  This cell has 4 units like the LSTM
+  """
+  def __init__(self, n_units, **kwargs):
+    super(LSTME, self).__init__(
```
doetsch: ... and here :)

gundramleifert:
I changed names to capitalized and changed the constructors. I wouldn't append "LSTM" to all structrues, I suggest LSTM, GRU and LEAKY are on the same hierarchy - they are all a fixed combination of basic components (a specific number of unit and update equations). I changed the name, but let's check first, if they work - then we can discuss about names ;-)

PR #48, opened 2018-06-21
Functions for supporting server mode, including vocabs specified in the config file

e-matusov:


GeneratingDataset.py
```diff
@@ -1549,7 +1561,7 @@ def get_seq(self, sentence):
     :rtype: list[int]
     """
     segments = self._segment_sentence(sentence)
-    seq = self.get_seq_indices(segments)
+    seq = super(BytePairEncoding, self).get_seq(segments)
```
albertz: This is wrong. It should be `get_seq_indices`.

GeneratingDataset.py
```diff
@@ -1339,12 +1344,19 @@ def init_vocab_var(session):
...
     """
     :param list[str] seq:
     :rtype: list[int]
     """
     return [self.vocab.get(k, self.unknown_label_id) for k in seq]
+  
+  def get_seq_labels(self, seq):
+    """
+    :param list[int] seq:
+    :rtype: string
```
albertz: The rtype is `str`, not `string`.

LmDataset.py
```diff
@@ -1018,6 +1023,7 @@ def _data_str_to_numpy(self, vocab, s):
       unknown_label_id = vocab[self._unknown_label]
       words_idxs = [vocab.get(w, unknown_label_id) for w in words]
     return numpy.array(words_idxs, dtype=numpy.int32)
+    '''
```
albertz: There should never be commented out code.
Why is the commented out code not correct?

LmDataset.py
```diff
@@ -1009,6 +1009,11 @@ def _data_str_to_numpy(self, vocab, s):
     :rtype: numpy.ndarray
     """
     words = s.split()
+
+
+    word_to_idx = [vocab[word] if word in vocab else vocab['<UNK>'] for word in words]
```
albertz: You should not use a hardcoded constant `'<UNK>'` here, but rather use `self._unknown_label`. Actually, this is exactly the code you commented out. What was wrong with it?

TFEngine.py
```diff
@@ -554,6 +555,8 @@ def __init__(self, config=None):
     self.use_search_flag = config.value("task", None) == "search"
     self.use_eval_flag = config.value("task", None) != "forward"
     self._const_cache = {}  # type: dict[str,tf.Tensor]
+    self.source_voc = None
+    self.target_voc = None
```
albertz: That should not be part of the `Engine` class. But see my comment below.

TFEngine.py
```diff
@@ -1746,25 +1753,37 @@ def search_single_seq(self, source, output_layer_name=None):
...
     :return: list of all hyps, which is a tuple of score and string
     :rtype: list[(float,str)]
     """
-    assert source_bpe.num_labels == self.network.extern_data.data["data"].dim
-    assert target_bpe.num_labels == self.network.extern_data.data["classes"].dim
-    source_seq_list = source_bpe.get_seq(source)
+    assert self.source_voc != None
+    assert self.target_voc != None
+    assert self.source_voc.num_labels == self.network.extern_data.data["data"].dim
+    assert self.target_voc.num_labels == self.network.extern_data.data["classes"].dim
```
albertz: Do not use `self.source_voc`/`self.target_voc` here but use the parameters you are passing (which you currently ignore).

TFEngine.py
```diff
@@ -1746,25 +1753,37 @@ def search_single_seq(self, source, output_layer_name=None):
...
 
+  def translate(self, source): 
+    """
+    :param str source: source as a string. 
+    :return: tuple of (best score, best hyp)
+    :rtype: (float,str)
+    """   
+    self._checked_uninitialized_vars = True
+    results = self.search_single_string_to_string_seq(source)
+    return results[0]
```
albertz: This function should be removed.
* It does not add really anything over `search_single_string_to_string_seq`.
* The naming bad. `Engine` works with any kind of task, on any kind of data. This is nothing MT related.
* `self._checked_uninitialized_vars = True` could be moved to `search_single_string_to_string_seq`.

LmDataset.py
e-matusov: This was a hack change by Julian which I just took over. I guess he wanted to have UNK in even if it was not used in training or where self._unknown_label is not set. Could you please check with him?

TFEngine.py
e-matusov: I strongly disagree here. The vocs should be part of the config file and semantically are part of the model, not of the Dataset. The dataset can change, but vocs stay same, it is a global voc based on training data. Since we don't have training data here, it has to be part of the model. I suggest to leave this like this, the two parameters are not going to be there anymore.

TFEngine.py
e-matusov: Ok, I agree here, will change.

TFEngine.py
albertz: `Engine` works for any kind of dataset, any kind of task, and any kind of model. It must not know about anything specific like a vocabulary.
The model is defined by `TFNetwork`. There is the specific class `ExternData` which holds information about the kind of data (source, targets, anything). There would maybe be the right place for this. Maybe it could be added there.

albertz: `Engine` is a wrapper over TensorFlow. TensorFlow also does not have an argument for a vocabulary in its `tf.Session`, because `tf.Session` is just generic, just like `Engine`.

albertz: I'm adding it now to `Data`.

albertz: I added it now, and pushed it. You can set it now via `num_outputs`. Sth like:

    num_outputs = {'data': {'dim': ..., 'sparse': True, 'vocab': 'file'}, 'classes': {...}}

In this function here, you could access it via:

    self.network.extern_data.data["data"].vocab

Or btw, this would be a bit better:

    self.network.extern_data.get_default_input_data().vocab

And then for the targets, this:

    self.network.extern_data.get_default_target_data().vocab

TFEngine.py
```diff
@@ -795,6 +798,10 @@ def init_network_from_config(self, config=None):
       except tf.errors.NotFoundError:
         print("Exiting now because model cannot be loaded.", file=log.v1)
         sys.exit(1)
+    if config.has("source_voc"):
+      self.source_voc = Vocabulary(config.value("source_voc", ""), "<UNK>")
+    if config.has("target_voc"):
+      self.target_voc = Vocabulary(config.value("target_voc", ""), "<UNK>")
```
albertz: This must be removed here. See my other comments:
https://github.com/rwth-i6/returnn/pull/48/files/c0aa865b600430eae3d272d022def324f1964107..d5cc0cafdc3de214e0c7ef71834bf7f11d022097

Also `"<UNK>"` is a bad default. In most other cases, it was `"UNK"`. And this is already the default. This should be consistent. Also, this default must not be changed anymore, because it would break lots of setups.

TFEngine.py
albertz: See my comments here: https://github.com/rwth-i6/returnn/pull/48/files/c0aa865b600430eae3d272d022def324f1964107..d5cc0cafdc3de214e0c7ef71834bf7f11d022097

TFEngine.py
```diff
@@ -1746,22 +1753,23 @@ def search_single_seq(self, source, output_layer_name=None):
...
     """
-    assert source_bpe.num_labels == self.network.extern_data.data["data"].dim
-    assert target_bpe.num_labels == self.network.extern_data.data["classes"].dim
-    source_seq_list = source_bpe.get_seq(source)
+    assert self.source_voc != None
+    assert self.target_voc != None
+    self._checked_uninitialized_vars = True
+    assert self.source_voc.num_labels == self.network.extern_data.data["data"].dim
+    assert self.target_voc.num_labels == self.network.extern_data.data["classes"].dim
+    source_seq_list = self.source_voc.get_seq_indices(source.split())
```
albertz: See my comments here:
https://github.com/rwth-i6/returnn/pull/48/files/c0aa865b600430eae3d272d022def324f1964107..d5cc0cafdc3de214e0c7ef71834bf7f11d022097

Also, `source.split()`, i.e. to split by whitespace, is very `Vocabulary` specific. I think this logic should not be in `Engine`, but in `Vocabulary`.

albertz:
Btw, when you fix the remaining things, maybe in the end, some stashing / cleanup of the commits would be nice.

TFEngine.py
e-matusov: So num_outputs would then correspond to the target side? What about the source side? Can you maybe send me an example of what should be in the config file instead of the following?
```
source_voc = "dataset/source.vocab.pkl"
target_voc = "dataset/target.vocab.pkl"
```
I'll attach the current config file.
[crnn-deploy.config.txt](https://github.com/rwth-i6/returnn/files/2140337/crnn-deploy.config.txt)

albertz: Because of historical reasons, `num_outputs` is a dict containing the information for all data (source, targets, and whatever).

    num_outputs = {
        'data': {'dim': ..., 'sparse': True, 'vocab': "dataset/source.vocab.pkl"}, 'classes': {...}},
        'targets': {'dim': ..., 'sparse': True, 'vocab': "dataset/target.vocab.pkl"}, 'classes': {...}},
    }

e-matusov:
Ok, I addressed all the changes, it seems to be working fine now for both dataset and single line translations. I had to add unknown_label to extern_data in TFUtil.py so that a different one than the default can be specified with the vocab. The default is now UNK. In LmDataset.py, I removed Julian's hack and restored the previously commented out code - again, if unknown_label is specified as part of TranslationDataset in config file, everything works.  So if you are fine with everything now, I will merge.

GeneratingDataset.py
```diff
@@ -1589,7 +1609,7 @@ def get_seq(self, sentence):
     :rtype: list[int]
     """
     segments = self._segment_sentence(sentence)
-    seq = self.get_seq_indices(segments)
+    seq = super(BytePairEncoding, self).get_seq_indices(segments)
```
albertz: Why is the `super` needed here?

LmDataset.py
```diff
@@ -1009,6 +1009,7 @@ def _data_str_to_numpy(self, vocab, s):
     :rtype: numpy.ndarray
     """
     words = s.split()
+
```
albertz: Why?

TFUtil.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, name,
                available_for_inference=True,
                auto_create_placeholders=False,
                vocab=None,
+               unknown_label = "UNK",
```
albertz: Please remove this. This is redundant. You can already pass it via the `vocab` argument. There is no need for an additional parameter.

TFEngine.py
```diff
@@ -27,6 +27,7 @@
 from tensorflow.python.client import timeline
 
 from Dataset import Dataset, Batch, BatchSetGenerator
+from GeneratingDataset import Vocabulary
```
albertz: I think this is not needed, right?

TFUtil.py
```diff
@@ -91,6 +92,7 @@ def __init__(self, name,
       The size is always a tensor of shape (batch,), i.e. the size can be different for each sequence in a batch.
     :param bool available_for_inference: e.g. the extern data "classes" is usually not available for inference
     :param str|dict[str]|GeneratingDataset.Vocabulary|None vocab:
+    :param str|"UNK" unknown_label: unknown word label for the optional vocabulary
```
albertz: "`"UNK"`" is not a type. Descriptions for default values or so, should go into the description (after the `:`). But anyway, this argument should be removed.

TFUtil.py
albertz: (Btw, also this code formatting is not PEP conform. There should not be whitespaces around the `=`. Also it's inconsistent to all the other code.)

TFUtil.py
```diff
@@ -150,7 +152,7 @@ def __init__(self, name,
     if vocab is not None:
       from GeneratingDataset import Vocabulary
       if isinstance(vocab, str):
-        vocab = Vocabulary(vocab)
+        vocab = Vocabulary(vocab, unknown_label=unknown_label)
```
albertz: As said, not needed. If you want to specify this, you can just use the case which you find just in the next code line.

GeneratingDataset.py
e-matusov: Because get_seq_indices() only exists in the parent Vocabulary class. If just writing "self" would also work, I can remove it.

LmDataset.py
e-matusov: You told me to do it! To put splitting into words inside the vocabulary class!

albertz: I don't understand. The only change here is adding a newline. I was asking, why is adding a newline needed here. Better remove this newline.

TFUtil.py
e-matusov: This is not redundant.  Otherwise I can't specify the vocabulary as file directly in the config file, I would need to load it somewhere else. I need it to be loaded by this class, and for this I need the unknown_label which may be different from the default one.

GeneratingDataset.py
albertz: Yes of course, just `self` works. Anything related to this code also has not changed, `get_seq_indices` was always part of the parent class.

TFUtil.py
e-matusov: I can't use  Vocabulary(**vocab) because my requirement is that the vocab is specified just as a vocab filename in the config file, and is loaded exactly in this line here.

TFUtil.py
albertz: Sure you can. Why do you think you can not?

Just like `... vocab={'vocab_file': ..., 'unknown_label': ...) ...`. Just see the code, and the documentation for the argument `vocab`. It can also be a `dict`.

TFUtil.py
albertz: I don't understand. You can specify the vocab filename + the unknown label in the config. Where is the problem?

TFUtil.py
e-matusov: Sorry, I am not so good in python. Yes, it works this way, but the specification in the config file is already pretty complicated. Ideally, it should be as simple as possible, just name-value pairs without nesting. But I will revert this anyway, it is not too late, we haven't decided to freeze the config file conventions.

albertz: I prefer it to be explicit and cleanly structured. The `Data` class should not know anything about the unknown label. This is specific to `Vocabulary`. Otherwise, imagine that other things get added to `Vocabulary`. It would be a nightmare to then add all further options also to lots of other unrelated classes.

TFEngine.py
```diff
@@ -1752,22 +1752,25 @@ def search_single_seq(self, source, output_layer_name=None):
...
-    :param GeneratingDataset.BytePairEncoding source_bpe:
-    :param GeneratingDataset.BytePairEncoding target_bpe:
     :param str|None output_layer_name: e.g. "output". if not set, will read from config "search_output_layer"
     :return: list of all hyps, which is a tuple of score and string
     :rtype: list[(float,str)]
     """
-    assert source_bpe.num_labels == self.network.extern_data.data["data"].dim
-    assert target_bpe.num_labels == self.network.extern_data.data["classes"].dim
-    source_seq_list = source_bpe.get_seq(source)
+    self._checked_uninitialized_vars = True
```
albertz: A final question: Why was this needed here? I could imagine that this could cause troubles in some cases.

TFEngine.py
```diff
@@ -1752,22 +1752,25 @@ def search_single_seq(self, source, output_layer_name=None):
...
+    source_voc = self.network.extern_data.data["data"].vocab
+    target_voc = self.network.extern_data.data["targets"].vocab
+    assert source_voc.num_labels == self.network.extern_data.data["data"].dim
+    assert target_voc.num_labels == self.network.extern_data.data["classes"].dim
+
+    source_seq_list = source_voc.get_seq(source)
+
```
albertz: The two empty lines, do you think it makes the code cleaner here? I'm a bit confused about them. I think I would remove them.

PR #50, opened 2018-07-12
Fix windowed_nd in TFUtil for padding valid

dthulke:
Calling `windowed_nd` in `TFUtil.py` with `source=[[[1]], [[2]], [[3]]]`, `window='valid'`, `window_size=2`, `time_axis=0` and `new_window_axis=1` results in the following error:

ValueError: Cannot reshape a tensor with 8 elements to shape [2,5,1,1] (10 elements) for 'windowed_batch/Reshape_1' (op: 'Reshape') with input shapes: [8], [4] and with input tensors computed as partial shapes: input[1] = [2,5,1,1].

The expected result would be `[[[[1]], [[2]]], [[[2]], [[3]]]]`

The reason for this is that the final reshaping does not consider the smaller size of the tensor due to the missing padding.


albertz:
Thanks

PR #51, opened 2018-07-27
Fix output.size_placeholder of SliceLayer and ConvLayer

dthulke:
Fixes two errors in the calculation of the output size in the Slice and Conv Layer.


albertz:
Thanks.

PR #58, opened 2018-08-15
added orth_vocab to SprintDataset

julian90:


GeneratingDataset.py
```diff
@@ -1316,18 +1316,20 @@ def create_vocab(cls, **opts):
       clz = BytePairEncoding
     return clz(**opts)
 
-  def __init__(self, vocab_file, unknown_label="UNK", num_labels=None):
+  def __init__(self, vocab_file, unknown_label="UNK", num_labels=None, seq_postfix=[]):
```
albertz: If you add it here, then you should delete it in `BytePairEncoding`. Also, never use a list (`[]`) as default param (see PyCharm comment why that is a bad idea). Anyway, use the same default param as in `BytePairEncoding`. Basically, just make it exactly the same as in `BytePairEncoding`.

SprintDataset.py
```diff
@@ -79,6 +80,10 @@ def __init__(self, target_maps=None, str_add_final_zero=False, input_stddev=1.,
     if bpe:
       from GeneratingDataset import BytePairEncoding
       self.bpe = BytePairEncoding(**bpe)
+    if orth_vocab:
```
albertz: Something like `self.orth_vocab = None` is missing before the `if`. Also see the failed test case.

SprintDataset.py
```diff
@@ -293,6 +300,13 @@ def addNewData(self, features, targets=None, segmentName=None):
       assert isinstance(orth, (str, unicode))
       assert "bpe" not in targets
       targets["bpe"] = numpy.array(self.bpe.get_seq(orth.strip()), dtype="int32")
+    if self.orth_vocab:
+      assert not self.orth_post_process
+      assert "orth" in targets
+      orth = targets["orth"]
+      assert isinstance(orth, (str, unicode))
+      assert "orth_classes" not in targets
+      targets["orth_classes"] = numpy.array(self.orth_vocab.get_seq(orth.strip()))
```
albertz: Like for `bpe`, I think you should explicitly specify `dtype="int32"`.

julian90:
not sure what will happen if this is merged, since this pull request contains now two commits with commits from other guys in-between...

PR #60, opened 2018-08-19
Add linter to the Travis

nerox8664:
It's necessary to keep all the code in the same and proper style that satisfy PEP8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ dist: trusty
 # quiet (-qq) to keep log smaller
 # |cat to avoid interactive banner (which blows up the log)
 install:
+  - pip install flake8
   - pip2 install -qq --user --upgrade pip setuptools wheel six | cat  # Python2<->Python3
   - pip2 install --user -r requirements.txt | cat  # need for Python2<->Python3 communication tests
   - pip install -qq --upgrade pip setuptools wheel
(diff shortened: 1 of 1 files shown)


albertz:
Yes, it's a good idea to have that. However, some issues with it right now:

- It should not be added as a test case for now if it does not pass. Otherwise all the Travis test results would be screwed for a while before all of it is fixed. This means that we should fix code formatting problems before we have the test enabled.
- Maybe we should relax the formatting rules a bit. E.g. I think PEP8 requires 4 spaces for indentation, but we have 2. And maybe there are some more things which we should relax. Let's collect maybe some good references here. E.g. what does Google usually use?
- I don't think that the test should be added as `before_script` in Travis. I think this rather should be just an additional separate test. I want the other tests to run as well.

@curufinwe @doetsch @pavelgolik Comments on this?

albertz:
Google uses [pylint](https://www.pylint.org/) in their [official Python style guide](https://github.com/google/styleguide/blob/gh-pages/pyguide.md). However, in some other places, e.g. their internal Python style guide and [Chromium](https://www.chromium.org/chromium-os/python-style-guidelines), they have some extra rules.
Google also has [YAPF](https://github.com/google/yapf/).

albertz:
I'm closing this now. Please update/reopen when you have addressed my points.

PR #63, opened 2018-09-21
fix: partition_epoch in TranslationDataset

patrick-wilken:
The index offset due to partition_epoch was still added in _get_line_nr(). This is wrong, as partition_epoch is now handled by get_sequence_order_for_epoch() and thus self._seq_order contains only elements for the current sub-epoch.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -252,7 +252,7 @@ class Dataset(object):
     :return: the order for the given epoch. such that seq_idx -> underlying idx
     :rtype: list[int]
     """
-    partition_epoch = self.partition_epoch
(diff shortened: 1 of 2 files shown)


patrick-wilken:
Also, the call of get_seq_order_for_epoch in LmDataset seems to be wrong, but I have never worked with it.

albertz:
Yes, I think `partition_epoch` in `LmDataset` was also broken. I pushed another fix for that. I didn't really tested it though, and I think we also do not have good test cases which cover this.
@kazuki-irie So keep this in mind, and next time you use it, make sure that it works correct. Who else uses the `LmDataset` currently?

PR #64, opened 2018-09-21
AmbigInputTranslationDataset, e.g. used for confustion network input

patrick-wilken:
AmbigInputTranslationDataset allows for multiple (weighted) options for each word in the source sequence. This was used for confusion network input in the AppTek submission for IWSLT 2018.


albertz:
There seems to be a conflict in the file. Can you rebase?
Also, the explanation will get lost here. Just put this into the documentation in the code, in the docstring.

patrick-wilken:
Should be done. I also renamed to ConfusionNetworkDataset, this is clearer, I think.

PR #65, opened 2018-09-24
MetaDataset fixes

patrick-wilken:
Some fixes for the MetaDataset.

I chose to create sequence tags for the MetaDataset by joining the tags of the sub-datasets with "<->", maybe someone has a better idea.


albertz:
Looks good (from a quick glance).
Btw, we do not have tests for any of these (`LmDataset`, `TranslationDataset`, `MetaDataset`, etc), which is bad. This is also why the partition epoch change broke a few things and we did not recognize this immediately. And with all these further changes, I'm a bit afraid that other things might have broken. We really should have some test cases (with toy data of course, which you can generate automatically on the fly within the test case, and then check things like partition epoch logic, etc). If you have some time, would be nice if you could add such.

patrick-wilken:
Seems I would have time, I will look into this.

PR #66, opened 2018-09-25
some more misc. fixes

patrick-wilken:
Should rather be 3 separate PRs...

I forgot some last changes to the CN dataset that were necessary for the ensembling experiments, this is the first commit.

Then I solved a crash occurring when max_seq_length is set in search. Translations for the filtered sentences were expected, causing a key error in out_cache. With this solution the filtered sentences do not appear in the output at all. Maybe we should add some line or enforce max_seq_length=0 instead? You might want to check this (although I have), an error here would be critical ;) 

Finally, get_dep_layers had to be implemented for the SwitchLayer to be able to use it in a recurrent unit.

### DIFF
--- a/TFEngine.py
+++ b/TFEngine.py
@@ -1792,12 +1792,16 @@ class Engine(object):
       else:
         print("Dataset have_corpus_seq_idx == False, i.e. it will not be sorted for optimal performance.", file=log.v3)
         dataset.seq_ordering = "default"  # enforce order as-is, so that the order in the written file corresponds
+
+    max_seq_length=self.config.typed_value('max_seq_length', None) or self.config.float('max_seq_length', 0)
+    assert not max_seq_length, "Set max_seq_length = 0 for search (i.e. no maximal length). We want to keep all source sentences."
+
     dataset.init_seq_order(epoch=self.epoch)
     batches = dataset.generate_batches(
       recurrent_net=self.network.recurrent,
       batch_size=self.config.int('batch_size', 1),
       max_seqs=self.config.int('max_seqs', -1),
-      max_seq_length=self.config.typed_value('max_seq_length', None) or self.config.float('max_seq_length', 0),
+      max_seq_length=max_seq_length,
       used_data_keys=self.network.used_data_keys)
 
     output_layer = self.network.layers[output_layer_name]
--- a/LmDataset.py
+++ b/LmDataset.py
@@ -877,7 +877,7 @@ class TranslationDataset(CachedDataset2):
       target.vocab.pkl
   """
 
-  MapToDataKeys = {"target": "classes"}  # just by our convention
+  MapToDataKeys = {"source": "data", "target": "classes"}  # just by our convention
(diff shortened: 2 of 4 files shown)


TFEngine.py
```diff
@@ -1889,15 +1892,13 @@ def extra_fetches_callback(seq_idx, seq_tag, output, targets=None, beam_scores=N
       runner.num_steps, self.format_score(runner.score), self.format_score(runner.error)), file=log.v1)
     if output_file:
       assert out_cache
-      assert 0 in out_cache
-      assert len(out_cache) - 1 in out_cache
```
albertz: Why? Esp for the txt output format, if some seqs are missing, there is no way to see that, or see which seqs are missing. Also, usually all further tools would expect that the seqs are complete.

TFEngine.py
```diff
@@ -1793,11 +1793,14 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
         print("Dataset have_corpus_seq_idx == False, i.e. it will not be sorted for optimal performance.", file=log.v3)
         dataset.seq_ordering = "default"  # enforce order as-is, so that the order in the written file corresponds
     dataset.init_seq_order(epoch=self.epoch)
+    max_seq_length = self.config.typed_value('max_seq_length', None) or self.config.float('max_seq_length', 0)
+    if max_seq_length > 0:
+      print("WARNING: Sentences of length greater than max_seq_length = {:.0f} will not be translated. ".format(max_seq_length))
```
albertz: `print` should go to some of the log streams, e.g. `file=log.v2` or so.

TFEngine.py
patrick-wilken: Ok, the question is whether there is a use case for filtering sentences by length when doing search. If no, we should assert max_seq_length == 0. If yes, then the output file inevitably will not contain all sequences.

As I said, the behaviour right now is that filtered source sequences are not put into any batch, therefore not translated, therefore not appearing in out_cache.  When we then loop from 0 to len(out_cache), there will be a key error. (This is not caught by the asserts.)

albertz: Yes. I think an `assert max_seq_length == 0` might be better (for now). Or leave the warning as you did, and then move the `assert 0 in out_cache`  and  the other for only the txt output format. But maybe just do `assert max_seq_length == 0`. I cannot really think of a case where it would be a good idea to ignore some seqs.

patrick-wilken: Ok, I'll keep the assertions.
How about setting max_seq_length = 0 and printing a warning that max_seq_length from config was ignored? Because I know that some people use max_seq_length = 1000 or so for search, which would fail if only 0 is allowed. Or should we force them to use the right value? ;)

albertz: Hm, not sure. Maybe just force them. Because maybe at some point we want to support that (`max_seq_length`), and then do some more clever thing (whatever that could be, maybe configurable, like translating a subseq only, or just split it, or so).

PR #71, opened 2018-10-10
Adding NCHW support

squarenabla:
get_placeholder_flattened works if time_dim != 1. copy_with_feature_dim_axis is fixed. Now it works for all feature_dim_axis 0

### DIFF
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -2712,7 +2712,7 @@ def flatten_with_seq_len_mask(x, seq_lens, batch_dim_axis=0, time_dim_axis=1, ti
   :param tf.Tensor seq_lens: shape (batch,) of int32
   :param int batch_dim_axis: index of batch_dim in x
   :param int time_dim_axis: index of time_dim in x
-  :param: bool time_major: whether time axis is 0 (redundant, kept for compatibility)
+  :param bool time_major: whether time axis is 0 (redundant, kept for compatibility)
   :return: tensor of shape (time', ...s...s'...) where time' = sum(seq_len) <= batch*time
   :rtype: tf.Tensor
   """
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -952,11 +952,11 @@ class Data(object):
       dyn_axes = [(i if (i < removed_axis) else (i - 1))
                   for i in dyn_axes]
       ndim -= 1
+
     if len(dyn_axes) > 1:
-      assert 0 in dyn_axes, "would need some transpose, not supported at the moment"
-      for i in dyn_axes:
-        if i > 0:
-          assert i - 1 in dyn_axes, "would need some transpose, not supported at the moment"
+      # Transpose x to get dyn axes first
+      perm = [i for i in sorted(dyn_axes)] + [i for i in range(ndim) if i not in dyn_axes]
+      x = tf.transpose(x, perm=perm)
       shape = tf.shape(x)
       x = tf.reshape(
         x,
@@ -965,8 +965,10 @@ class Data(object):
       dyn_axes = [0]
     assert dyn_axes == [0]
     if keep_dims and orig_num_dyn_axes >= 2:
-      for i in range(orig_num_dyn_axes - 1):
-        x = tf.expand_dims(x, axis=1)
+      exp_axes = sorted(self.get_spatial_batch_axes() + [self.batch_dim_axis])
+      for i in exp_axes:
+        if i != 0:
+          x = tf.expand_dims(x, axis=i)
     return x
 
   def get_axes(self, exclude_time=False, exclude_batch=False):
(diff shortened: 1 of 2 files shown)


albertz:
Next time: Choose a better title for the pull request ("The changes in the TFUtil.py" doesn't tell me too much...)

albertz:
You are changing multiple things here (`get_placeholder_flattened` and `copy_with_feature_dim_axis`). This should be multiple separate commits (and if possible also multiple separate pull requests).
Also, I explained that you should also add some test cases for these (for the old behavior/functionality which already worked before, and for the newly added functionality).

TFUtil.py
```diff
@@ -906,10 +921,11 @@ def get_placeholder_with_specific_batch_dim_axis(self, batch_dim_axis):
   def get_placeholder_time_flattened(self):
     assert self.placeholder is not None
     assert self.have_time_axis()
-    # flatten_with_seq_len_mask only works for these two cases at the moment:
-    assert (self.time_dim_axis, self.batch_dim_axis) == (0, 1) or (self.time_dim_axis, self.batch_dim_axis) == (1, 0)
+    # flatten_with_seq_len_mask only works for these cases at the moment:
+    assert (self.time_dim_axis, self.batch_dim_axis) in [(0, 1), (1, 0), (0, 2), (2, 0)]
```
albertz: This is too specific. If you want to support any axes for batch/time, why are you checking that it must be 0,2 or 2,0?

squarenabla:
How should I do that? I'm testing everything in the debug shell and calling Data methods with different parameters. Do you have any examples of tests?

TFUtil.py
```diff
@@ -2688,24 +2706,33 @@ def reversed(x):
...
+def flatten_with_seq_len_mask(x, seq_lens, batch_dim_axis=0, time_dim_axis=1, time_major=False):
   """
-  :param tf.Tensor x: shape (batch,time,...s...) with time_major=False or otherwise shape (time,batch,...s....)
+  :param tf.Tensor x: shape (batch,...s..., time, ...s'...) or shape (time,...s...., batch, ...s'...)
   :param tf.Tensor seq_lens: shape (batch,) of int32
-  :param bool time_major: if the time-dim is the first dimension in x
-  :return: tensor of shape (time', ...s...) where time' = sum(seq_len) <= batch*time
+  :param int batch_dim_axis: index of batch_dim in x
+  :param int time_dim_axis: index of time_dim in x
+  :param: bool time_major: whether time axis is 0 (redundant, kept for compatibility)
```
albertz: The doc format syntax is wrong. Please check PyCharm/PEP, or just see the other code.

TFUtil.py
```diff
@@ -2688,24 +2706,33 @@ def reversed(x):
   return y
 
 
-def flatten_with_seq_len_mask(x, seq_lens, time_major=False):
+def flatten_with_seq_len_mask(x, seq_lens, batch_dim_axis=0, time_dim_axis=1, time_major=False):
```
albertz: You added some conflicting params (`time_dim_axis`, `batch_dim_axis` are conflicting with `time_major`). You cannot have the default params like that. Keep the default params all as `None`, extend the documentation about the default param, and add checks in the code, that only either `time_major` or `batch_dim_axis` and `time_dim_axis` can be set.

albertz:
Look at the tests in Returnn.

albertz:
Ah, and about commit message: Keep the message in the first line limited to 80 chars.
If you want to add further description for the commit: Add another empty line after the first line, and then you can add the description.
(Btw, if you use PyCharm, it will tell you already that. Are you ignoring this?)
(When you split the commit into multiple, as I explained earlier, you probably will not have this problem anymore.)

TFUtil.py
squarenabla: Actually, it supports other cases, but they have to be checked. The only requirement is that either batch_dim_axis or time_dim_axis is 0.

TFUtil.py
squarenabla: Ok

TFUtil.py
albertz: Then check only that.

albertz:
As explained, please do separate pull requests for these other things (channel first for conv layer etc).
Also, please follow my comments (about e.g. commit message, etc).

squarenabla:
I made this commits before you message, that's why they I big and have a bulky message( Can I keep old commits as they are? That will be time consuming to split them into different commits.

albertz:
They have to be changed anyway (see all my comments). I don't see why it's difficult then to also change the commit message.

squarenabla:
> They have to be changed anyway (see all my comments). I don't see why it's difficult then to also change the commit message.

Done

albertz:
As I said, please make separate pull requests, and cleanup the commits.

PR #72, opened 2018-10-10
LayerNormVariantsLSTMCell

mmz33:
Fixes include:
1. Class name changed from "LayerNormLSTMCell" to "LayerNormVariantsLSTMCell" just to make sure not to face override problems with TF official implementation.
2. Removed tf.nn.dropout and use Returnn dropout implementation for speedup.
3. Added "name" param to "_norm" function to create different TF variables.
4. In "_linear" function, the bias is not added if there is layer norm. In addition, "var_creation_scope()" is used when creating the bias.
5. Added overview descriptions about different LSTM layer norm variants.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -4177,3 +4177,207 @@ class BlocksparseMultiplicativeMultistepLSTMCell(_WrapBaseCell):
     return y
 
 
+class LayerNormVariantsLSTMCell(BaseRNNCell):
+  """LSTM unit with layer normalization and recurrent dropout
+
+  This LSTM cell can apply different variants of layer normalization:
+
+  1. Layer normalization as in the original paper:
+  Ref: https://arxiv.org/abs/1607.06450
+  This can be applied by having:
+    all default params
+
+  2. Layer normalization for RNMT+:
+  Ref: https://arxiv.org/abs/1804.09849
+  This can be applied by having:
+    all default params except
+   - global_norm = False,
+   - per_gate_norm = True
+
+  3. TF official `LayerNormBasicLSTMCell`
+  Ref: https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/LayerNormBasicLSTMCell
+  This can be reproduced by having:
+    all default params except
+  - global_norm = False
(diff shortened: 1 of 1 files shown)


TFNetworkRecLayer.py
```diff
@@ -4177,3 +4177,141 @@ def call(self, *args, **kwargs):
     return y
 
 
+class LayerNormLSTMCell(BaseRNNCell):
+  """LSTM unit with layer normalization and recurrent dropout
+
+  Layer normalization implementation is based on:
+        https://arxiv.org/abs/1607.06450
+
+  Recurrent dropout l2p
```
albertz: What is `l2p`?

TFNetworkRecLayer.py
```diff
@@ -4177,3 +4177,141 @@ def call(self, *args, **kwargs):
...
+      computation output
+    :param with_concat: A boolean, if True then the input and prev hidden state
+      is concatenated for the computation
+    :param global_norm: A boolean, if True then layer normalization is applied
+      for the forward and recurrent outputs
+    :param per_gate_norm: A boolean, if True then layer normalization is applied
+      per lstm gate
+    :param cell_norm: A boolean, if True then layer normalization is applied
+      to the LSTM new cell output
+    :param hidden_norm: A boolean, if True then layer normalization is applied
```
albertz: Write the type (`int`, `float`, `bool`) in the common PEP/PyCharm format if possible (e.g. `:param int num_inits: number of LSTM units`).

TFNetworkRecLayer.py
```diff
@@ -4177,3 +4177,141 @@ def call(self, *args, **kwargs):
     return y
 
 
+class LayerNormLSTMCell(BaseRNNCell):
+  """LSTM unit with layer normalization and recurrent dropout
+
+  Layer normalization implementation is based on:
+        https://arxiv.org/abs/1607.06450
```
albertz: Actually, this is only one of the variants you cover, right?
You should explain that in the description.
E.g. say sth like: With `global_norm=True`, `per_gate_norm=...`, ... (etc) (and other params left as default) you get the original layer norm LSTM as presented here (link).

PR #74, opened 2018-10-15
Adding NCHW support to TFNetworkLayer.py

squarenabla:


albertz:
Please do separate pull requests for each separate change.
Also, please follow the conventions for commit messages. E.g. there should be no dot at the end (why are you doing this? see at other commit messages. this is very inconsistent).
Also, please follow Python / PEP code style conventions. Follow what e.g. PyCharm tells you (why do I need to repeat that so often?).
There are some questions I need to ask about each commit/change individually also. But it will be easier to discuss when you make separate pull requests.

PR #75, opened 2018-10-15
Added onedirectional 2D-LSTM

ChristopherBrix:
I've added support for a onedirectional 2D-LSTM in TensorFlow.

The code is based on the Theano implementation for 4 directions, and is not as general as I would like. But it might be helpful nevertheless.

The inputs to the 2D-LSTM layer have to be two sequences (eg. source and target). The output is a sequence, not a 2D grid, so it can easily be combined with other layers.

During inference, it is assumed that the shape of the second (target) dimension is always 1 (the previous target word).

This might be relevant for issue #8.

### DIFF
--- a/NativeOp.cpp
+++ b/NativeOp.cpp
@@ -190,6 +190,80 @@ static void tf_cuda_sgemm(
 #endif  // GOOGLE_CUDA
 }
 
+// This method can be used to perform multiple matrix-matrix multiplications in parallel
+// One possible use-case is the 2D-LSTM layer: There, for performance reasons, a whole diagonal of states are computed
+// in parallel.
+template<typename T>
+static void tf_cuda_sgemm_batched(
+        OpKernelContext* context,
+        char transa, char transb,
+        int m, int n, int k,
+        const T* alpha_, const T** in_x, int lda,
+        const T** in_y, int ldb, const T* beta_,
+        const T** out,
+        int ldc,
+        int batchSize,
+        bool finalize_stream = 0) {
+    T alpha = *alpha_;
+    T beta = *beta_;
+// https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/rnn/kernels/blas_gemm.cc
+#if GOOGLE_CUDA
+    typedef perftools::gputools::DeviceMemory<float> DeviceMemoryType;
+    std::vector<DeviceMemoryType> a_device_memory;
+    std::vector<DeviceMemoryType> b_device_memory;
+    std::vector<DeviceMemoryType> c_device_memory;
+    std::vector<DeviceMemoryType*> a_ptrs;
+    std::vector<DeviceMemoryType*> b_ptrs;
+    std::vector<DeviceMemoryType*> c_ptrs;
+    a_device_memory.reserve(batchSize);
+    b_device_memory.reserve(batchSize);
+    c_device_memory.reserve(batchSize);
+    a_ptrs.reserve(batchSize); // (b, m, k)
+    b_ptrs.reserve(batchSize); // (b, k, n)
+    c_ptrs.reserve(batchSize); // (b, m, n)
+
+    for (int64 i = 0; i < batchSize; ++i) {
+      a_device_memory.push_back(AsDeviceMemory(in_x[i]));
+      b_device_memory.push_back(AsDeviceMemory(in_y[i]));
+      c_device_memory.push_back(AsDeviceMemory(out[i]));
+      a_ptrs.push_back(&a_device_memory.back());
+      b_ptrs.push_back(&b_device_memory.back());
+      c_ptrs.push_back(&c_device_memory.back());
+    }
+
+    cudaStream_t cuda_stream = context->eigen_gpu_device().stream();
+
+    // cublasCreate, http://docs.nvidia.com/cuda/cublas/#cublascreate
+
+    auto dev_ctx = context->op_device_context();
+    perftools::gputools::Stream* dev_stream = dev_ctx->stream();
+    OP_REQUIRES(context, dev_stream, errors::Internal("No GPU stream available."));
+
+    bool blas_launch_status =
+        dev_stream
+             ->ThenBlasGemmBatched(get_transpose(transa), get_transpose(transb),
+                            (uint64)m, (uint64)n, (uint64)k, alpha, a_ptrs,
+                            lda, b_ptrs, ldb, beta, c_ptrs, ldc, batchSize)
+             .ok();
+    OP_REQUIRES(context, blas_launch_status, errors::Aborted("CuBlasGemm failed!"));
+
+    // The above call to ThenBlasGemmBatched allocates temporary memory on the GPU
+    // This memory is only freed once the stream is "finalized", which might not happen automatically
+    // Therefore, if the flag is set, we force a synchronization. This is bad for the performance, so it should be
+    // done as seldom as possible
+    if(finalize_stream) {
+        blas_launch_status =
+            dev_stream
+                ->BlockHostUntilDone()
+                .ok();
+        OP_REQUIRES(context, blas_launch_status, errors::Aborted("BlockHostUntilDone failed!"));
+    }
+#else  // GOOGLE_CUDA
+    context->SetStatus(errors::InvalidArgument("CuBlasGemm needs CUDA."));
+#endif  // GOOGLE_CUDA
+}
+
+
 #if CUDA
 #if !GOOGLE_CUDA
 #error "GOOGLE_CUDA not defined"
@@ -200,6 +274,12 @@ static void tf_cuda_sgemm(
 	transpose_A, transpose_B, \
 	m, n, k, alpha, A, lda, B, ldb, beta, C, ldc) \
     tf_cuda_sgemm<float>(context, transpose_A, transpose_B, m, n, k, alpha, A, lda, B, ldb, beta, C, ldc);
+
+#define Ndarray_sgemm_batched( \
+	transpose_A, transpose_B, \
+	m, n, k, alpha, A, lda, B, ldb, beta, C, ldc, batchSize, finalize_stream) \
+    tf_cuda_sgemm_batched<float>(context, transpose_A, transpose_B, m, n, k, alpha, A, lda, B, ldb, beta, C, ldc, batchSize, finalize_stream);
+
 
 #else  // CUDA
 /*
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -1162,7 +1162,8 @@ class TFNetwork(object):
     """
     from TFNetworkRecLayer import RecStepInfoLayer, _SubnetworkRecCell
     rec_layer = self.get_rec_parent_layer()
-    if not rec_layer:
+    # the second condition is true if all layers have been optimized out of the rec layer
+    if not rec_layer or len(rec_layer.cell.layers_in_loop) == 0:
       assert not must_exist, "%s: We expect to be the subnet of a RecLayer, but we are not." % self
       return None
     assert isinstance(rec_layer.cell, _SubnetworkRecCell)
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -4618,3 +4618,279 @@ class LayerNormVariantsLSTMCell(BaseRNNCell):
       new_h = self._norm(new_h, name='new_h')
 
     return new_h, rnn_cell.LSTMStateTuple(new_c, new_h)
+
+
+class TwoDLSTMLayer(LayerBase):
+  layer_class = "twod_lstm"
+  recurrent = True
+
+  def __init__(self,
+               pooling='last',
+               unit_opts=None,
+               forward_weights_init=None, recurrent_weights_init=None, bias_init=None,
+               **kwargs):
+    """
+    :param str pooling: defines how the 1D return value is computed based on the 2D lstm result. Either 'last' or 'max'
+    :param None|dict[str] unit_opts: passed to RNNCell creation
+    :param str forward_weights_init: see :func:`TFUtil.get_initializer`
+    :param str recurrent_weights_init: see :func:`TFUtil.get_initializer`
+    :param str bias_init: see :func:`TFUtil.get_initializer`
+    """
+    super(TwoDLSTMLayer, self).__init__(**kwargs)
+    import re
+    from TFUtil import is_gpu_available
+    from tensorflow.contrib import rnn as rnn_contrib
+    from tensorflow.python.util import nest
+    if is_gpu_available():
+      from tensorflow.contrib import cudnn_rnn
+    else:
+      assert False, "currently, there's no CPU support"
+    self.pooling = pooling
+    # On the random initialization:
+    # For many cells, e.g. NativeLSTM: there will be a single recurrent weight matrix, (output.dim, output.dim * 4),
+    # and a single input weight matrix (input_data.dim, output.dim * 4), and a single bias (output.dim * 4,).
+    # The bias is by default initialized with 0.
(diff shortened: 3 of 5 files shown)


albertz:
Can you rebase and resolve the conflicts? (Also remove the merge commit. Just rebase it.)

TFNetwork.py
```diff
@@ -1120,7 +1120,8 @@ def get_rec_step_info(self, must_exist=True):
     """
     from TFNetworkRecLayer import RecStepInfoLayer, _SubnetworkRecCell
     rec_layer = self.get_rec_parent_layer()
-    if not rec_layer:
+    # the second condition is true if all layers have been optimized out of the rec layer
+    if not rec_layer or len(rec_layer.cell.layers_in_loop) == 0:
```
albertz: I don't understand.
If `rec_layer` is None, the code below will lead to an exception (any code below which access `rec_layer`).
Also, `must_exist` specifically says that it must exists. So if it does not exist, it should throw an exception (that is what `must_exists` is for).

TFNetworkRecLayer.py
```diff
@@ -4383,3 +4383,273 @@ def __call__(self, inputs, state, scope=None):
       new_h = self._norm(new_h, name='new_h')
 
     return new_c, new_h
+
+class TwoDLSTMLayer(LayerBase):
```
albertz: Two empty lines. See my earlier comment.

TFNetworkRecLayer.py
```diff
@@ -4383,3 +4383,273 @@ def __call__(self, inputs, state, scope=None):
...
+        self.output.batch_dim_axis = 1
+
+      # Very generic way to collect all created params.
+      # Note that for the TF RNN cells, there is no other way to do this.
+      # Also, see the usage of :func:`LayerBase.cls_layer_scope`, e.g. for initial vars.
+      params = tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope=re.escape(scope_name_prefix))
+      self._add_params(params=params, scope_name_prefix=scope_name_prefix)
+
+  @classmethod
+  def _post_init_output(cls, output, network, target=None, size_target=None, sources=(), **kwargs):
```
albertz: Why do you need this function? Everything what you do here should be done in `__init__`, if possible.

TFNetworkRecLayer.py
```diff
@@ -4383,3 +4383,273 @@ def __call__(self, inputs, state, scope=None):
...
+    if unit_opts is None:
+      unit_opts = {}
+
+    assert not self.sources[0].output.sparse
+    n_input_dim_parts = [self.sources[0].output.dim, self.sources[1].output.dim]
+    cell = rnn_cell_class(
+      n_hidden=n_hidden, n_input_dim=sum(n_input_dim_parts), n_input_dim_parts=n_input_dim_parts,
+      input_is_sparse=self.sources[0].output.sparse,
+      **unit_opts,
+      pooling=self.pooling)
```
albertz: I think this is Python 3 syntax, right? The code should also work (or at least load without syntax error) in Python 2 (at least for a while).

TFNetworkRecLayer.py
```diff
@@ -4383,3 +4383,273 @@ def __call__(self, inputs, state, scope=None):
...
+    """
+    if network.train_flag is not False:
+      # TODO: In training, this is ok. Maybe as well as for eval but not clear.
+      # In forward, mark_data_key_as_used=False should be used and anyway that target value is not available.
+      output.size_placeholder = cls._static_get_target_value(
+        target=target or size_target, network=network,
+        mark_data_key_as_used=network.train_flag is not False).size_placeholder.copy()
+    else:
+      with tf.control_dependencies(None):
+        output.size_placeholder = {} # Not sure why this has to be defined. Is it ever used?
```
albertz: Follow Python code style conventions, ie. two spaces before `#`.
Also, remove the comment anyway here. The `output.size_placeholder` is used in many places. (Let's discuss offline if you wonder where...)

TFNetworkRecLayer.py
```diff
@@ -4383,3 +4383,273 @@ def __call__(self, inputs, state, scope=None):
...
+      y = self._get_output_native_rec_op(self.cell)
+
+      self.output.placeholder = y
+      if self.network.have_rec_step_info():
+        self.output.shape = (self.output.shape[-1],)
+        self.output.time_dim_axis = None
+        self.output.batch_dim_axis = 0
+      else:
+        self.output.time_dim_axis = 0
+        self.output.batch_dim_axis = 1
```
albertz: The layer output format should not be defined in `__init__`. This should be defined in `get_out_data_from_opts`, which you need to overload. See other layers for example.

TFNetwork.py
ChristopherBrix: If `rec_layer` is None, the first condition is true. Then, it is ensured that `must_exist` is false, otherwise an exception is thrown as expected. If `must_exist` is false, None is returned, indicating that there is no recurrent step information. This behavior is used by `have_rec_step_info()`.

If `rec_layer` is not None, but all contained layers have been optimized out of the loop, my modification ensures that no recurrent step information is returned. This is necessary, because it was never created (see https://github.com/rwth-i6/returnn/blob/master/TFNetworkRecLayer.py#L1580).

If it is not None and contains internal layers, then the step information is returned as expected.

NativeOp.cpp
```diff
@@ -190,6 +190,78 @@ static void tf_cuda_sgemm(
...
+        blas_launch_status =
+            dev_stream
+                ->BlockHostUntilDone()
+                .ok();
+        OP_REQUIRES(context, blas_launch_status, errors::Aborted("BlockHostUntilDone failed!"));
+    }
+#else  // GOOGLE_CUDA
+    context->SetStatus(errors::InvalidArgument("CuBlasGemm needs CUDA."));
+#endif  // GOOGLE_CUDA
+}
```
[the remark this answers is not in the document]
ChristopherBrix: `sgemm_batched` is needed for performance reasons. Usually, only one matrix matrix multiplication is needed at any time. But for the 2D-LSTM, we can compute a complete diagonal at once, so multiple operations can be performed in parallel. 

For CPU support, this could be changed to a sequence of multiple `sgemm` calls.

ChristopherBrix:
I don't know how much work will be required to add CPU support for 2D-LSTMs and when I'll have the time to implement it.

Maybe we can postpone this?

albertz:
Sure, as I said in my comment, let's skip the CPU for now.

albertz:
Btw, also see that the tests all run through (here on GitHub, there is the Travis link to your pull request).

TFNetworkRecLayer.py
ChristopherBrix: The output format depends on whether or not the 2D layer was optimized out of the recurrent layer. Currently, I determine this by checking `self.network.have_rec_step_info()`. 

In the class method `get_out_data_from_opts`, the `network` doesn't appear to be available - how can I perform the check in there?

TFNetworkRecLayer.py
ChristopherBrix: Don't I have to overwrite this method, as explained in the [`LayerBase` class](https://github.com/rwth-i6/returnn/blob/master/TFNetworkLayer.py#L256)?

However, moving it to  `__init__` would have the benefit that I can directly check `have_rec_step_info` and don't have to rely on `train_flag`. For my use case, both are equivalent, but they don't necessary have to be.

albertz: No, it should only be used if not possible via `__init__` (which is very rarely the case).

NativeOp.cpp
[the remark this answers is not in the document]
albertz: I don't exactly understand. What is the matrix in `a_ptrs`, `b_ptr`, `c_ptr`, or better, what are their shapes?

TFNetworkRecLayer.py
albertz: This still need to be resolved.

TFNetworkRecLayer.py
albertz: This still needs to be resolved.

NativeOp.cpp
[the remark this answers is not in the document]
ChristopherBrix: `a_ptrs` (as well as `b_ptrs` and ` c_ptrs`) wraps the input `in_x` (`in_y`, `out`).
This input is a sequence of matrices (for 2D-LSTM, one matrix for each element of the current diagonal). The number of these matrices is given by `batch_size` (which is equal to `batchCount`, I'll make it consistent).
In the `for` loop, they are wrapped using `AsDeviceMemory()` and added to `a_ptrs`.
The shape of the matrices is given by `m`, `n` and `k`: `in_x` has shape `(m, k)`, `in_y` has shape `(k, n)` and `out` has shape `(m, n)`.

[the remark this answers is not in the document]
albertz: So, (m,k), (k,n) are the input shapes, (m,n) is the output shape. This is a pretty standard matrix multiplication, which should be possible via normal `sgemm`. I don't see what `sgemm_batched` can do better/faster than `sgemm`.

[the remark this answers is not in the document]
ChristopherBrix: But `sgemm` only performs one such multiplication at a time, right? We have to perform *multiple, independent* multiplications in parallel.

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+      self.output.placeholder = y
+
+      # Very generic way to collect all created params.
+      # Note that for the TF RNN cells, there is no other way to do this.
+      # Also, see the usage of :func:`LayerBase.cls_layer_scope`, e.g. for initial vars.
+      params = tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope=re.escape(scope_name_prefix))
+      self._add_params(params=params, scope_name_prefix=scope_name_prefix)
+
+  @classmethod
+  def get_out_data_from_opts(cls, sources, **kwargs):
```
albertz: Why do you have only `sources` here, but not other kwargs which you use below, like `network`, etc?

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+    elif self._fwd_weights_initializer:
+      default_var_initializer = self._fwd_weights_initializer
+    else:
+      default_var_initializer = xavier_initializer(seed=self.network.random.randint(2**31))
+    with reuse_name_scope("rec-twod", initializer=default_var_initializer) as scope:
+      assert isinstance(scope, tf.VariableScope)
+      self._rec_scope = scope
+      scope_name_prefix = scope.name + "/"  # e.g. "layer1/rec/"
+      with self.var_creation_scope():
+        print("============!!!!!!!!!!!!========= Cell created")
```
albertz: Better remove those debug msgs.
If you really want them, forward them to the log stream, e.g. `file=log.v4`.
But please remove the `============!!!!!!!!!!!!========= `, and make the message more meaningful.

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+      (not has_cell_attribute and isinstance(parent_layer, RecLayer)):
+      shape = (kwargs["n_out"],)
+      time_dim_axis = None
+      batch_dim_axis = 0
+      size_placeholder = {}
+    else:
+      shape = (None, kwargs["n_out"],)
+      time_dim_axis = 0
+      batch_dim_axis = 1
+      size_placeholder = sources[1].output.size_placeholder.copy()
```
albertz: Do not add workarounds. Esp not such fragile workarounds which rely on some specific random behavior which could maybe change at some future point.
If there is a problem, the problem itself should be fixed, or if there is no good way to get some information, just introduce that. If needed. In this cases, it should not really be needed.
As I said, look just at how other layers do that, which is much simpler.
E.g. `SelfAttentionLayer`, or `RnnCellLayer`. Or `WindowLayer`, or `CumsumLayer`.
I guess the output shape is anyway very simple, just in all cases like `sources[1].output.shape`, or not?
Or you could check for `sources[1].output.time_dim_axis is None` or so.

TFNetworkRecLayer.py
albertz: It also seems as if you assume that `len(sources) == 2`, right? If so, you should add such an `assert`.

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+    """
+    super(TwoDLSTMLayer, self).__init__(**kwargs)
+    import re
+    from TFUtil import is_gpu_available
+    from tensorflow.contrib import rnn as rnn_contrib
+    from tensorflow.python.util import nest
+    if is_gpu_available():
+      from tensorflow.contrib import cudnn_rnn
+    else:
+      assert(False) # currently, there's no CPU support
```
albertz: The proper syntax is `assert False`.
And the comment you added could just go into the assert message, which is more helpful for the user then. I.e. 
```suggestion
      assert False, "currently, there's no CPU support"
```

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+      **unit_opts)
+    return cell
+
+  @classmethod
+  def get_rec_initial_extra_outputs(cls, batch_dim, rec_layer, **kwargs):
+    # state and output are dummy values that aren't used
+    # we have to specifiy them (and they must have 4 dims), because during inference (when multiple iterations
+    # for one 2DLSTM happen, as opposed to the training where everything is computed at once) we pass the intermediate
+    # values from one iteration to the other. 42 is added to show from the value of the variable that it's not
+    # supposed to be used.
```
albertz: The values are used. They are the initial values.
In case they are not needed (training), do not set them.
Again, see other layers as example, e.g. `SelfAttentionLayer`.

TFNetworkRecLayer.py
```diff
@@ -4460,3 +4460,283 @@ def __call__(self, inputs, state, scope=None):
...
+
+  @classmethod
+  def get_rec_initial_extra_outputs_shape_invariants(cls, **kwargs):
+    """
+    :return: optional shapes for the tensors by get_rec_initial_extra_outputs
+    :rtype: dict[str,tf.TensorShape]
+    """
+    return {"state": tf.TensorShape((None, None, None, None)),
+            "output": tf.TensorShape((None, None, None, None)),
+            "iteration": tf.TensorShape((None,))}
```
albertz: See above.

TFNetworkRecLayer.py
ChristopherBrix: Sorry, I added a lot of debug messages while looking for the `size_placeholder` bug. Apparently, I forgot to delete this one afterwards.

ChristopherBrix:
I've fixed the issues you pointed out - is there anything else?
Also, in the meantime I've added two small improvements in my codebase: the ability to pick the maximum instead of the last state on the horizontal axis and to use dropout. Would you like me to add it to this PR, or to create a new one after this one is merged?

albertz:
Remove merge commits. Rebase on the latest master commit.
Also `get_rec_initial_extra_outputs` still looks wrong. If it is outside the rec layer, there should not be a recurrent state.

albertz:
Any new things, once this is merged in, I don't think it needs another PR, unless you are unsure maybe about the change.

albertz:
Hey, what's the state here?

albertz:
Hey, what's the state here? Also, can you rebase? (It conflicts currently to the master.)

ChristopherBrix:
Sorry, I didn't work on this for some time.
I've updated the `get_rec_initial_extra_outputs` method, now it returns an empty dict if there's no time dimension to loop over.

There's just one test failing, and the same tests fails on the master branch as well, so all should be fine.

PR #76, opened 2018-10-15
NCHW support for ConvLayer

squarenabla:


albertz:
Please cleanup the commits. Just merge them together into a single commit.

TFNetworkLayer.py
```diff
@@ -2953,8 +2957,15 @@ def ceildiv(a, b):
...
   @classmethod
   def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilation_rate=1, sources=(),
-                              input_expand_dims=0, input_add_feature_dim=False, input_split_feature_dim=None, **kwargs):
-    shape = [None] * len(filter_size) + [n_out]
+                              input_expand_dims=0, input_add_feature_dim=False, input_split_feature_dim=None,
+                              auto_use_channel_first=False,
+                              **kwargs):
+    data = get_concat_sources_data_template(sources)
+    # The output format is the same as the input.
+    if data.feature_dim_axis == 1:
```
albertz: This is wrong. I think/guess, you actually want to check if the input data is also channel first. Then also check that, and not if `feature_dim_axis == 1` (not the same thing).

albertz: Also, the comment should be extended. Say that you want channel first if the input is also channel first.
But actually, I think the code is simpler if you just remove the code here completely, and leave it like it was originally (`shape = [None] * len(filter_size) + [n_out]`), and then just handle it in the already existing code in the end.

TFNetworkLayer.py
```diff
@@ -2970,14 +2981,15 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
...
     if input_expand_dims == 0 and not input_add_feature_dim and not input_split_feature_dim:
       # Maybe we have a chance to correctly define the output shapes.
-      data = get_concat_sources_data_template(sources)
+      index_shift = data.time_dim_axis_excluding_batch
       for i in range(len(filter_size)):
-        if data.shape[i] is not None:
-          shape[i] = cls.calc_out_dim(
-            in_dim=data.shape[i],
+        if data.shape[i + index_shift] is not None:
+          shape[i + index_shift] = cls.calc_out_dim(
```
albertz: Follow my comment above (just leave `shape = [None] * len(filter_size) + [n_out]`).
Then here you must have `shape[i] = cls.calc_out_dim(...`.

TFNetworkLayer.py
```diff
@@ -2970,14 +2981,15 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
...
+            in_dim=data.shape[i + index_shift],
             filter_size=filter_size[i], stride=strides[i], dilation_rate=dilation_rate[i], padding=padding)
-    feature_dim_axis = NotSpecified
-    if TFUtil.is_gpu_available() and False:  # TODO...
+    feature_dim_axis = data.feature_dim_axis
```
albertz: This is wrong. It should be:
`feature_dim_axis = data.feature_dim_axis_or_unspecified`

TFNetworkLayer.py
```diff
@@ -2970,14 +2981,15 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
...
+    # Swap the dims if the input dim order doesn't fit the flag auto_use_channel_first.
+    if TFUtil.is_gpu_available() and auto_use_channel_first and data.feature_dim_axis != 1:
```
albertz: The check (`data.feature_dim_axis != 1`) is wrong. You actually want to check whether it is channel first or not.
But given my comment initially, anyway, it should be changed to:
`if (TFUtil.is_gpu_available() and auto_use_channel_first) or data.is_batch_feature_major:`

TFNetworkLayer.py
albertz: Actually, with my comment above, just leave it like it was (`feature_dim_axis = NotSpecified`).

TFNetworkLayer.py
```diff
@@ -2961,11 +2961,10 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
                               auto_use_channel_first=False,
                               **kwargs):
     data = get_concat_sources_data_template(sources)
-    # The output format is the same as the input.
-    if data.feature_dim_axis == 1:
-      shape = [n_out] + [None] * len(filter_size)
-    else:
-      shape = [None] * len(filter_size) + [n_out]
+    auto_use_channel_first = auto_use_channel_first or data.is_batch_feature_major
+    # The output format is the same as the input for now.
```
albertz: Why "for now"? Remove that. Maybe write "we maybe change to channel-first below".

TFNetworkLayer.py
```diff
@@ -2961,11 +2961,10 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
...
+    shape = [None] * len(filter_size) + [n_out]
+
```
albertz: Why the extra empty line? Remove that.

TFNetworkLayer.py
```diff
@@ -2981,15 +2980,16 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
     padding = padding.upper()
     if input_expand_dims == 0 and not input_add_feature_dim and not input_split_feature_dim:
       # Maybe we have a chance to correctly define the output shapes.
-      index_shift = data.time_dim_axis_excluding_batch
+      if data.is_batch_feature_major:
+        data = data.copy_with_feature_dim_axis(-1)
```
albertz: Why? Remove that. Use `index_shift` as you did before (for `data.shape`; not for `shape`, as I wrote earlier). There is no need in an additional copy of the input (would be totally wasted resources).

TFNetworkLayer.py
```diff
@@ -2981,15 +2980,16 @@ def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilatio
...
-            in_dim=data.shape[i + index_shift],
+        if data.shape[i] is not None:
+          shape[i] = cls.calc_out_dim(
+            in_dim=data.shape[i],
             filter_size=filter_size[i], stride=strides[i], dilation_rate=dilation_rate[i], padding=padding)
-    feature_dim_axis = data.feature_dim_axis
+    feature_dim_axis = NotSpecified
     # Swap the dims if the input dim order doesn't fit the flag auto_use_channel_first.
-    if TFUtil.is_gpu_available() and auto_use_channel_first and data.feature_dim_axis != 1:
+    if TFUtil.is_gpu_available() and auto_use_channel_first:
```
albertz: See what I wrote before. There also should be `... or data.is_batch_feature_major`. (Please follow all my comments...)

albertz:
Of course, when you think that you are done, please clean up the commits again.
There is also a button where you can request for another review, so that I will get notified.

TFNetworkLayer.py
```diff
@@ -2953,7 +2957,12 @@ def ceildiv(a, b):
 
   @classmethod
   def _get_out_type_from_opts(cls, n_out, filter_size, padding, strides=1, dilation_rate=1, sources=(),
-                              input_expand_dims=0, input_add_feature_dim=False, input_split_feature_dim=None, **kwargs):
+                              input_expand_dims=0, input_add_feature_dim=False, input_split_feature_dim=None,
+                              auto_use_channel_first=False,
+                              **kwargs):
+    data = get_concat_sources_data_template(sources)
+    auto_use_channel_first = auto_use_channel_first or data.is_batch_feature_major
```
albertz: Why this line with `auto_use_channel_first`? I think this is not needed.

TFNetworkLayer.py
```diff
@@ -2953,7 +2957,12 @@ def ceildiv(a, b):
...
+    # The output format is the same as the input. We maybe change to channel-first below.
```
albertz: I think this comment is more confusing than helping. Maybe better just remove it.

squarenabla:
The ConvLayer and the test won't work until the TFUtil.py is updated according to the #79

PR #77, opened 2018-10-15
NCHW support in PoolLayer

squarenabla:
Now the layer converts the input according to the flag auto_use_channel_first

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -3111,13 +3111,16 @@ class PoolLayer(_ConcatInputLayer):
   layer_class = "pool"
   recurrent = True  # we should not shuffle in the time-dimension
 
-  def __init__(self, mode, pool_size, padding="VALID", dilation_rate=1, strides=None, **kwargs):
+  def __init__(self, mode, pool_size, padding="VALID", dilation_rate=1, strides=None,
+               use_channel_first=False,
+               **kwargs):
     """
     :param str mode: "max" or "avg"
     :param tuple[int] pool_size: shape of the window of each reduce
     :param str padding: "valid" or "same"
     :param tuple[int]|int dilation_rate:
     :param tuple[int]|int|None strides: in contrast to tf.nn.pool, the default (if it is None) will be set to pool_size
+    :param bool use_channel_first: if set, will transform input to NCHW format
     """
     assert "n_out" not in kwargs
     assert "out_type" not in kwargs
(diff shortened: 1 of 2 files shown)


TFNetworkLayer.py
```diff
@@ -3002,13 +3002,17 @@ class PoolLayer(_ConcatInputLayer):
...
+               auto_use_channel_first=False,
+               **kwargs):
     """
     :param str mode: "max" or "avg"
     :param tuple[int] pool_size: shape of the window of each reduce
     :param str padding: "valid" or "same"
     :param tuple[int]|int dilation_rate:
     :param tuple[int]|int|None strides: in contrast to tf.nn.pool, the default (if it is None) will be set to pool_size
+    :param bool auto_use_channel_first: if set, will transform input to NCHW format
+    :param bool auto_use_channel_first: convert the input to NCHW or not
```
albertz: why twice?

TFNetworkLayer.py
```diff
@@ -3031,29 +3035,45 @@ def __init__(self, mode, pool_size, padding="VALID", dilation_rate=1, strides=No
...
+                             auto_use_channel_first=False,
+                             **kwargs):
     # y shape is [batch] + spatial_dims + [n_out].
     data = get_concat_sources_data_template(sources, name="%s_output" % name)
-    shape = [None] * len(pool_size) + [data.dim]
+    # The output format is the same as the input.
+    if data.feature_dim_axis == 1:
+      shape = [data.dim] + [None] * len(pool_size)
+    else:
+      shape = [None] * len(pool_size) + [data.dim]
```
albertz: See my comments in #76.

TFNetworkLayer.py
```diff
@@ -3070,18 +3090,25 @@ def get_out_data_from_opts(cls, name, pool_size, strides=None, dilation_rate=1,
...
-          in_dim=data.shape[i],
+      if data.shape[i + index_shift] is not None:
+        shape[i + index_shift] = ConvLayer.calc_out_dim(
+          in_dim=data.shape[i + index_shift],
           filter_size=pool_size[i], stride=strides[i], dilation_rate=dilation_rate[i], padding=padding)
+    feature_dim_axis = data.feature_dim_axis
+    # Swap the dims if the input dim order doesn't fit the flag auto_use_channel_first.
+    if TFUtil.is_gpu_available() and auto_use_channel_first and data.feature_dim_axis != 1:
+      feature_dim_axis = 1
+      shape = shape[-1:] + shape[:-1]
```
albertz: Also here: see my comments in #76.

squarenabla:
Again, like in the #76, the tests won't work without changes in #79. That's why I wanted to do everything in one pull request. Locally I merged all branches and everything works.

albertz:
It's better to have several pull request. Actually, your pull request #79 becomes also a bit too big, that is why we need several iterations now to get it into shape, and that makes it more complicated, also for me to review this again and again (which is much simpler, the shorter each pull request is), and it all takes more time than necessary because of that. Anyway, #79 is almost finished now, probably after the next iteration when you do my requested changes, then we can proceed here.

PR #78, opened 2018-10-15
NCHW support in LinearLayer and SoftmaxLayer

squarenabla:
Now the layer converts the input according to the flag auto_use_channel_first

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -1884,7 +1884,7 @@ class LinearLayer(_ConcatInputLayer):
 
   def __init__(self, activation, with_bias=True, grad_filter=None,
                forward_weights_init="glorot_uniform", bias_init=0.0,
-               out_feature_dim=None,
+               out_feature_dim_axis=None,
                **kwargs):
     """
     :param str|None activation: e.g. "relu", or None
@@ -1893,7 +1893,7 @@ class LinearLayer(_ConcatInputLayer):
     :param str forward_weights_init: see :func:`TFUtil.get_initializer`
     :param str recurrent_weights_init: see :func:`TFUtil.get_initializer`
     :param str|float bias_init: see :func:`TFUtil.get_initializer`
-    :param int|None out_feature_dim: if set, will convert feature dim of the input and output data
+    :param int|None out_feature_dim_axis: if set, will convert feature dim of the input and output data
     """
     super(LinearLayer, self).__init__(**kwargs)
     from TFUtil import get_initializer
@@ -1901,11 +1901,10 @@ class LinearLayer(_ConcatInputLayer):
     self.activation = activation
     self.with_bias = with_bias
 
-    if out_feature_dim is None:
+    if out_feature_dim_axis is None:
       input_data = self.input_data
     else:
-      input_data = self.input_data.copy_with_feature_dim_axis(out_feature_dim)
-      self.output = self.output.copy_with_feature_dim_axis(out_feature_dim)
+      input_data = self.input_data.copy_with_feature_dim_axis(out_feature_dim_axis)
     n_in = input_data.dim
     n_out = self.output.dim
     assert n_in and n_out, "%r and %r" % (input_data, self.output)
@@ -1960,11 +1959,36 @@ class LinearLayer(_ConcatInputLayer):
       self.output_before_activation = OutputWithActivation(x)
     x = self.output_before_activation.y
 
-    if out_feature_dim is None:
+    if out_feature_dim_axis is None:
       assert self.output.batch_dim_axis == self.input_data.batch_dim_axis
       assert self.output.time_dim_axis == self.input_data.time_dim_axis
 
     self.output.placeholder = x
+
+  @classmethod
+  def _get_out_type_from_opts(cls, n_out, sources=(), out_feature_dim_axis=None,
+                              **kwargs):
+    if out_feature_dim_axis is None:
+      return None
+    data = get_concat_sources_data_template(sources)
+    data = data.copy_with_feature_dim_axis(out_feature_dim_axis)
+
+    shape = list(data.shape)
+    shape[out_feature_dim_axis] = n_out
+    time_dim_axis = data.time_dim_axis
+
+    return {
+      "dim": n_out,
+      "shape": tuple(shape),
+      "batch_dim_axis": 0,
+      "time_dim_axis": time_dim_axis,
+      "feature_dim_axis": out_feature_dim_axis,
+      "sparse": False}
+
+  @classmethod
+  def get_out_data_from_opts(cls, **kwargs):
+    out_type = cls._get_out_type_from_opts(**kwargs)
+    return super(LinearLayer, cls).get_out_data_from_opts(out_type=out_type, **kwargs)
 
 
 class SoftmaxLayer(LinearLayer):
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -1974,7 +1974,7 @@ class SoftmaxLayer(LinearLayer):
   layer_class = "softmax"
 
   def __init__(self, activation="softmax", **kwargs):
-    super(SoftmaxLayer, self).__init__(activation=activation, **kwargs)
+    super(SoftmaxLayer, self).__init__(activation=activation, out_feature_dim=-1, **kwargs)
 
 
 class LengthLayer(LayerBase):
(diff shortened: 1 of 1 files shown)


TFNetworkLayer.py
```diff
@@ -1892,14 +1893,19 @@ def __init__(self, activation, with_bias=True, grad_filter=None,
     :param str forward_weights_init: see :func:`TFUtil.get_initializer`
     :param str recurrent_weights_init: see :func:`TFUtil.get_initializer`
     :param str|float bias_init: see :func:`TFUtil.get_initializer`
+    :param int|None out_feature_dim: if set, will convert feature dim of the input and output data
```
albertz: It's not a dim, it's an axis. The parameter name should reflect that. E.g. like `out_feature_dim_axis`.

albertz: Also the documentation ("feature dim axis", not "feature dim").

TFNetworkLayer.py
```diff
@@ -1892,14 +1893,19 @@ def __init__(self, activation, with_bias=True, grad_filter=None,
...
 
     self.activation = activation
     self.with_bias = with_bias
 
-    input_data = self.input_data
+    if out_feature_dim is None:
+      input_data = self.input_data
+    else:
+      input_data = self.input_data.copy_with_feature_dim_axis(out_feature_dim)
+      self.output = self.output.copy_with_feature_dim_axis(out_feature_dim)
```
albertz: The format of `self.output` should not be changed here. This should be done via overwriting `get_out_data_from_opts`.

albertz: Actually, maybe better would be to extend the logic in `LayerBase._base_get_out_data_from_opts`.

albertz: I mean for the case that the input data is channel first.

TFNetworkLayer.py
```diff
@@ -1966,7 +1974,7 @@ class SoftmaxLayer(LinearLayer):
   layer_class = "softmax"
 
   def __init__(self, activation="softmax", **kwargs):
-    super(SoftmaxLayer, self).__init__(activation=activation, **kwargs)
+    super(SoftmaxLayer, self).__init__(activation=activation, out_feature_dim=-1, **kwargs)
```
albertz: I don't understand why that is needed.

squarenabla:
Of course everything works.

TFNetworkLayer.py
squarenabla: Ok

squarenabla: In the base class it's to abstract to decide what the output should look like. So, just coping from the input is ok. 
I'l do the output transformation in get_out_data_from_opts.

TFNetworkLayer.py
squarenabla: It's needed to require output with the feature axis at the end (NWHC format).

TFNetworkLayer.py
albertz: In the base class, it will just copy the input format by default (i.e. output will be batch major if input is batch major; or output will be time major if input is time major). That is exactly what you also want here. So why not just add this additional default behavior?

squarenabla: It may be the case, that the input is in NCHW, but it should be transformed to the NHWC due to the two reasons:
- Channels first are not supported later in the loss function (https://github.com/rwth-i6/returnn/blob/1a1d901b75e7175a999d469891ccda24de9f10ae/TFNetworkLayer.py#L5093). And the labels have channels on the last position.
- In the LinearLayer there's a function dot, which assumes features in the last axis (https://github.com/rwth-i6/returnn/blob/1a1d901b75e7175a999d469891ccda24de9f10ae/TFUtil.py#L2168). I think, it's possible to implement dot(a, b, feature_axis). But it's not needed in my case. In the official ResNet implementation, they transform the data flow to NHWC right before the last linear layer and softmax.

TFNetworkLayer.py
```diff
@@ -1966,7 +1998,11 @@ class SoftmaxLayer(LinearLayer):
   layer_class = "softmax"
 
   def __init__(self, activation="softmax", **kwargs):
-    super(SoftmaxLayer, self).__init__(activation=activation, **kwargs)
+    super(SoftmaxLayer, self).__init__(activation=activation, out_feature_dim_axis=-1, **kwargs)
```
albertz: I still don't understand why that is needed.
Or to say it differently: It's inconsistent. Why should the `SoftmaxLayer` have a different default than the `LinearLayer`? There is no reason for it. So don't do that.

TFNetworkLayer.py
```diff
@@ -1954,10 +1959,37 @@ def __init__(self, activation, with_bias=True, grad_filter=None,
...
+      "shape": tuple(shape),
+      "batch_dim_axis": 0,
+      "time_dim_axis": time_dim_axis,
+      "feature_dim_axis": out_feature_dim_axis,
+      "sparse": False}
+
+  @classmethod
+  def get_out_data_from_opts(cls, **kwargs):
+    out_type = cls._get_out_type_from_opts(**kwargs)
+    return super(LinearLayer, cls).get_out_data_from_opts(out_type=out_type, **kwargs)
```
albertz: As discussed, please just extend `LayerBase._base_get_out_data_from_opts` instead.

TFNetworkLayer.py
```diff
@@ -1892,14 +1893,18 @@ def __init__(self, activation, with_bias=True, grad_filter=None,
     :param str forward_weights_init: see :func:`TFUtil.get_initializer`
     :param str recurrent_weights_init: see :func:`TFUtil.get_initializer`
     :param str|float bias_init: see :func:`TFUtil.get_initializer`
+    :param int|None out_feature_dim_axis: if set, will convert feature dim of the input and output data
```
albertz: This option is not really needed. The option `out_type` already supports to define a custom `feature_dim_axis`.
So please remove it.

TFNetworkLayer.py
```diff
@@ -1892,14 +1893,18 @@ def __init__(self, activation, with_bias=True, grad_filter=None,
...
     from TFUtil import get_initializer
 
     self.activation = activation
     self.with_bias = with_bias
 
-    input_data = self.input_data
+    if out_feature_dim_axis is None:
+      input_data = self.input_data
+    else:
+      input_data = self.input_data.copy_with_feature_dim_axis(out_feature_dim_axis)
```
albertz: If you specify `out_type` for a layer, you can specify the `feature_dim_axis` for the output.
Here, the code should just be like:

    input_data = self.input_data
    if not self.input_data.is_sparse:
        assert input_data.batch_ndim == self.output.batch_ndim
        input_data = input_data.copy_with_feature_dim_axis(self.output_feature_dim_axis_or_unspecified)

(Actually, did you try the code with `dot`? I'm not sure if it really works if it is channel-first. As said, please write a test case. Check that you get the same output when you don't use channel first.)

albertz:
Btw, also check the tests (Travis link here). They all should go through.

TFNetworkLayer.py
albertz: You should not add a workaround in some unrelated layer just because that would not be supported in the loss function. Then just fix the loss function instead.
About the `dot` function: Also see `DotLayer`, which is more generic and has implemented sth like that already.

squarenabla:
There's no need in these modifications anymore

albertz:
In which? In the whole pull request? Why not? What did you change?

squarenabla:
I realized that NCHW support is not needed in the Linear layer. I just added redesigned the functionality of the Pool and Conv layers. If use_channel_first is set they convert the tensor to NCHW, otherwise to NHWC. So, I just have use_channel_first=False in the last PoolLayer.

squarenabla:
I also have an idea, that we need two flags in the Conv and Pool (maybe some other layers which use kernels). The first one tells in which format to perform the operation and the second one - in which format to make an output.

albertz:
Well, the idea is that the layer do not reformat the output but just directly provide the output as they have calculate it internally. There is also no reason for such an option, because in theory, every layer, every loss, anything, should just accept any possible input format, and then maybe reformat the input if they need to. If you find some layer / loss / whatever, which does not work correctly for some input format, please report and/or fix that. (If one would really want some reformat of some output, which should not be needed as explained, you could also introduce just a special `ReformatLayer` or so. Actually there is already `SwapAxesLayer` which does mostly that.)

PR #79, opened 2018-10-15
NCHW support in Data class

squarenabla:


TFUtil.py
```diff
@@ -2688,24 +2702,38 @@ def reversed(x):
...
+    x = check_dim_equal(x, batch_dim_axis, seq_lens, batch_dim_axis, ["batch-dim does not match"])  # batch dim
     # int64? -> https://github.com/tensorflow/tensorflow/issues/6518
-    mask = sequence_mask(seq_lens, maxlen=tf.shape(x)[1])  # shape (batch,time)
+    # Batch and time dims have to be in front of the tensor in order to apply the mask.
+    dyn_axes = [batch_dim_axis, time_dim_axis]
+    perm = [i for i in dyn_axes] + [i for i in range(len(x.shape)) if i not in dyn_axes]
+    batch_dim_axis = 0
+    time_dim_axis = 1
+    x = tf.transpose(x, perm=perm)
+    mask = sequence_mask(seq_lens, maxlen=tf.shape(x)[time_dim_axis])  # shape (batch,time)
```
albertz: You are first doing (maybe) `swapaxes`, and then here an additional `transpose`. That is bad. Just do a single `transpose` instead.
Also, only do it (the `transpose`) if it is needed.

TFUtil.py
```diff
@@ -2688,24 +2702,38 @@ def reversed(x):
...
+  :param int time_dim_axis: index of time_dim in x
+  :param bool time_major: whether time axis is 0 (redundant, kept for compatibility)
+  :return: tensor of shape (time', ...s...s'...) where time' = sum(seq_len) <= batch*time
   :rtype: tf.Tensor
   """
+  # time_major is set(old_variant) => batch_dim_axis and time_dim_axis have to be None.
+  if time_major is not None:
+    assert batch_dim_axis is None and time_dim_axis is None
+    batch_dim_axis = int(time_major)
+    time_dim_axis = int(not time_major)
```
albertz: The code is not clear to read.
Just write it explicitly:
```
if time_major:
  batch_dim_axis, time_dim_axis = 1, 0
else:
  batch_dim_axis, time_dim_axis = 0, 1
```

TFUtil.py
```diff
@@ -949,8 +962,9 @@ def get_placeholder_flattened(self, keep_dims=False):
       dyn_axes = [0]
     assert dyn_axes == [0]
     if keep_dims and orig_num_dyn_axes >= 2:
-      for i in range(orig_num_dyn_axes - 1):
-        x = tf.expand_dims(x, axis=1)
+      for i in self.get_spatial_batch_axes() + [self.batch_dim_axis]:
+        if i != 0:
+          x = tf.expand_dims(x, axis=i)
```
albertz: This is confusing.
Better, initially where you have `dyn_axes = self.get_spatial_batch_axes() + [self.batch_dim_axis]`, do instead:
```
orig_dyn_axes = self.get_spatial_batch_axes() + [self.batch_dim_axis]
dyn_axes = list(orig_dyn_axes)
```
And then, here at this place, do this:
```suggestion
for i in orig_dyn_axes:
  if i not in dyn_axes:
    x = tf.expand_dims(x, axis=i)
```
Also, to have another check that we have the correct number of dims, add this after the loop:
```
x.set_shape([None] * self.batch_ndim)
```
Also, please add some test cases for this, for several of the possible cases.

TFUtil.py
```diff
@@ -2688,24 +2702,40 @@ def reversed(x):
...
+  :return: tensor of shape (time', ...s...s'...) where time' = sum(seq_len) <= batch*time
   :rtype: tf.Tensor
   """
+  # time_major is set(old_variant) => batch_dim_axis and time_dim_axis have to be None
+  if time_major is not None:
+    assert batch_dim_axis is None and time_dim_axis is None
+    if time_major:
+      batch_dim_axis, time_dim_axis = 1, 0
+    else:
+      batch_dim_axis, time_dim_axis = 0, 1
```
albertz: Add a check here (for all cases, not within the `if`):
```
assert batch_dim_axis is not None and time_dim_axis is not None
```

albertz: Also assert `batch_dim_axis != time_dim_axis`.

TFUtil.py
```diff
@@ -2688,24 +2702,40 @@ def reversed(x):
...
   with tf.name_scope("flatten_with_seq_len_mask"):
     seq_lens = check_input_ndim(seq_lens, 1)
-    if time_major:
-      x = swapaxes(x, 0, 1)  # get (batch,time,...s...)
-    x = check_dim_equal(x, 0, seq_lens, 0, ["batch-dim does not match"])  # batch dim
+    # If not (batch,time,...s...), transform.
+    if batch_dim_axis != 0 or time_dim_axis != 1:
+      assert batch_dim_axis is not None and time_dim_axis is not None
```
albertz: This assert is not needed here anymore, when you add it initially.

TFUtil.py
```diff
@@ -2688,24 +2702,40 @@ def reversed(x):
...
+      dyn_axes = [batch_dim_axis, time_dim_axis]
+      perm = [i for i in dyn_axes] + [i for i in range(len(x.shape)) if i not in dyn_axes]
```
albertz: Simpler:
```suggestion
      perm = dyn_axes + [i for i in range(len(x.shape)) if i not in dyn_axes]
```

TFUtil.py
```diff
@@ -937,10 +953,7 @@ def get_placeholder_flattened(self, keep_dims=False):
                   for i in dyn_axes]
       ndim -= 1
     if len(dyn_axes) > 1:
-      assert 0 in dyn_axes, "would need some transpose, not supported at the moment"
-      for i in dyn_axes:
-        if i > 0:
-          assert i - 1 in dyn_axes, "would need some transpose, not supported at the moment"
+      # Transpose x to get dyn axes first.
```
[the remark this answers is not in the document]
albertz: I'm pretty sure now that it will not be correct with `0 not in dyn_axes`, because `flatten_with_seq_len_mask` will always have the new resulting flattened axis in 0.
Also, in your use case, you will anyway have `0 in dyn_axes`. So just leave the check in.

tests/test_TFUtil.py
```diff
@@ -143,6 +143,26 @@ def test_Data_unknown_feature_no_time():
   assert d1.batch_shape == (None, None)
 
 
+def test_Data_copy_with_feature_dim_axis():
+  import numpy as np
+  size_placeholder = tf.constant(np.full((10,), 10), dtype=tf.int32)
+  d = Data(name='test_data', shape=(None, 64, 64), dtype='float32',
```
albertz: Do not use `(None, 64, 64)` but better different numbers, e.g. `(None, 13, 17)`.

tests/test_TFUtil.py
```diff
@@ -143,6 +143,26 @@ def test_Data_unknown_feature_no_time():
...
+           size_placeholder={0: size_placeholder}, batch_dim_axis=0,
+           time_dim_axis=1, feature_dim_axis=3)
+  d_copy = d.copy_with_feature_dim_axis(1)
+  assert d_copy.shape == (64, None, 64)
+  assert d_copy.batch_dim_axis == 0
+  assert d_copy.time_dim_axis == 2
+  assert d_copy.feature_dim_axis == 1
+  assert d_copy.size_placeholder.keys() == [1]
```
albertz: The check is incorrect I think. Should be:
```suggestion
  assert list(d_copy.size_placeholder.keys()) == [1]
```

TFUtil.py
```diff
@@ -393,19 +393,34 @@ def copy_with_feature_dim_axis(self, feature_dim_axis):
...
+        if a < old_feature_axis:
+          a += 1
+        else:
+          a -= 1
+      elif a > feature_dim_axis:
+        if a < old_feature_axis:
+          a += 1
+      else:
+        if a > old_feature_axis:
+          a -= 1
```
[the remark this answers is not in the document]
albertz: With possible cases, I mean for both `old_feature_axis < feature_dim_axis` and the other way around, and that there is some "special" axes (e.g. `time_dim_axis`) before, in between or after (that are 6 cases).

TFUtil.py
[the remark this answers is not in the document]
albertz: Also add a test case for `get_placeholder_flattened`.

tests/test_TFUtil.py
```diff
@@ -143,6 +143,26 @@ def test_Data_unknown_feature_no_time():
...
+  d_copy = d_copy.copy_with_feature_dim_axis(-1)
+  assert d_copy.shape == (None, 64, 64)
+  assert d_copy.batch_dim_axis == 0
+  assert d_copy.time_dim_axis == 1
+  assert d_copy.feature_dim_axis == 3
+  assert d_copy.size_placeholder.keys() == [0]
```
albertz:
```suggestion
  assert list(d_copy.size_placeholder.keys()) == [0]
```

TFUtil.py
```diff
@@ -393,19 +393,37 @@ def copy_with_feature_dim_axis(self, feature_dim_axis):
...
+      elif a > feature_dim_axis:
+        if a < old_feature_axis:
+          a += 1
+      else:
+        if a > old_feature_axis:
+if old_feature_axis <= a < feature_dim_axis:
+  a -= 1
+elif feature_dim_axis <= a < old_feature_axis:
+  a += 1
+      setattr(data, k, a)
```
albertz: I'm not sure what happened here. There is still your old code, but then also my suggested code, but with a wrong indentation. Please fix that.

TFUtil.py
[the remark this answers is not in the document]
squarenabla: Well, I think the code will work if `0 not in dyn_axes` but I didn't test that. And I doubt anybody needs this case. That's why there's  `assert 0 in [self.time_dim_axis, self.batch_dim_axis]` in the `get_placeholder_time_flattened()`.

tests/test_TFUtil.py
```diff
@@ -153,6 +173,44 @@ def test_Data_copy_compatible_to_time_major():
...
+  res = session.run(d.placeholder, feed_dict={placeholder: feed_data})
+  print(res.shape)
+  flat_placeholder = d.get_placeholder_flattened(keep_dims=True)
+  res = session.run(flat_placeholder, feed_dict={placeholder: feed_data})
+  assert res.shape == (7, 17)
+  flat_placeholder = d.get_placeholder_flattened(keep_dims=False)
+  res = session.run(flat_placeholder, feed_dict={placeholder: feed_data})
+  assert res.shape == (7, 17)
+
+
```
albertz: Please, as I said before, also 6 more tests for `copy_with_feature_dim_axis`, always with existing time dim, to cover these cases:

* `time_dim_axis < old_feature_dim_axis < new_feature_dim_axis`
* `old_feature_dim_axis < time_dim_axis < new_feature_dim_axis`
* `old_feature_dim_axis < new_feature_dim_axis <= time_dim_axis`
* `time_dim_axis < new_feature_dim_axis < old_feature_dim_axis`
* `new_feature_dim_axis <= time_dim_axis < old_feature_dim_axis`
* `new_feature_dim_axis < old_feature_dim_axis < time_dim_axis`

TFUtil.py
```diff
@@ -2688,24 +2688,41 @@ def reversed(x):
...
-  :return: tensor of shape (time', ...s...) where time' = sum(seq_len) <= batch*time
+  :param int batch_dim_axis: index of batch_dim in x
+  :param int time_dim_axis: index of time_dim in x
+  :param bool time_major: whether time axis is 0 (redundant, kept for compatibility)
+  :return: tensor of shape (time', ...s...s'...) where time' = sum(seq_len) <= batch*time
   :rtype: tf.Tensor
   """
+  # time_major is set(old_variant) => batch_dim_axis and time_dim_axis have to be None
+  assert (batch_dim_axis is None and time_dim_axis is None) or time_major is None
+  assert (batch_dim_axis is not None and time_dim_axis is not None) or time_major is not None
```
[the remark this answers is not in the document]
albertz: Also remove the comment (it's confusing and clear from the code).

TFUtil.py
```diff
@@ -2688,24 +2688,41 @@ def reversed(x):
...
+  if time_major is not None:
+    if time_major:
+      batch_dim_axis, time_dim_axis = 1, 0
+    else:
+      batch_dim_axis, time_dim_axis = 0, 1
```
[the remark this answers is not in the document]
squarenabla: I don't understand, where to put the assert. Right after the if expression?

tests/test_TFUtil.py
```diff
@@ -153,6 +173,44 @@ def test_Data_copy_compatible_to_time_major():
...
+  res = session.run(flat_placeholder, feed_dict={placeholder: feed_data})
+  print(res.shape)
+  assert res.shape[0] == 7 * 9 * 13
+  assert len(res.shape) == 4
+  flat_placeholder = d.get_placeholder_flattened(keep_dims=False)
+  res = session.run(flat_placeholder, feed_dict={placeholder: feed_data})
+  print(res.shape)
+  assert res.shape[0] == 7 * 9 * 13
+  assert len(res.shape) == 2
+  print("Test2")
```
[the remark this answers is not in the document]
albertz: So? Still, it's a separate test, so make it a separate test.

TFUtil.py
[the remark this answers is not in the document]
albertz: Just that it is executed in all possible cases. Right after the `if` expression. Isn't that clear from the code? You are just verifying, that in all possible cases, `batch_dim_axis is not None and time_dim_axis is not None` (because you just have set it, or you expect it to be set from the input). So from now on, you don't have to care about that. That makes code simpler to read.

PR #81, opened 2018-11-06
MetaDataset: support for Nones in seq_list indicating missing sequences

patrick-wilken:
I hesitated to commit this, as it is somewhat complicated and at least in one case hacky, but now the second person asked me for the code... ;)

This is needed for multi-task setups, where for different tasks different datasets are used. This is implemented here by setting the sequence tag in seq_list to "None" for the unused datasets. In that case a dummy value (0) is provided for the inputs/outputs of a dataset (the hacky part...) instead of real data (which might not be available). One can then condition the computational graph on this value being 0 (or of length 1) to handle the different tasks.

The difficulty is that the Nones have to be removed from the seq_list when handed to the datasets and thus an index mapping has to be created.

### DIFF
--- a/MetaDataset.py
+++ b/MetaDataset.py
@@ -72,9 +72,11 @@ class EpochWiseFilter:
 
 class MetaDataset(CachedDataset2):
   """
-  This wraps around one or multiple datasets and might provide extra information.
-  Every dataset is expected to provide the the same sequences, where the sequence list
-  is given by a file.
+  This wraps around one or multiple datasets to make multiple kinds of inputs/outputs available
+  (e.g. to combine text and audio input features). To match corresponding sequences in the
+  datasets, "seq_list_file" has to be provided. It contains a list of sequence tags for each dataset
+  such that sequences at same indices correspond. A tag may be "None" indicating that this sequence is not
+  present in the given dataset.
   """
 
   def __init__(self,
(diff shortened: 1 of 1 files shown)


patrick-wilken:
I see now that this has significant overlap with CombinedDataset. Let me first check how to make it work with that before accepting this.

albertz:
I'm currently on vacation. I'm back end of November. But maybe just try to
decide what would be the most clean way to do it.


Am Mi., 7. Nov. 2018, 22:05 hat Patrick Wilken <notifications@github.com>
geschrieben:

albertz:
Closed because you pushed it already, or how have you solved it now?


Am Fr., 9. Nov. 2018, 20:31 hat Patrick Wilken <notifications@github.com>
geschrieben:

patrick-wilken:
This is really what CombinedDataset is intended to do, so no need to it in MetaDataset too. I got the CombinedDataset working, I will commit some changes there later today as it was outdated in some places in the code.

albertz:
Ok. Although it might make sense maybe to merge both in functionality. In
the end I would prefer a solution which is easiest for the user to use, and
also most straight forward to understand for the user.


Am Fr., 9. Nov. 2018, 21:09 hat Patrick Wilken <notifications@github.com>
geschrieben:

PR #82, opened 2018-11-09
small changes to make CombinedDataset work again

patrick-wilken:
iteritems() -> items()

num_seqs was unset because of `self._num_seqs = None` in `CachedDataset2.init_seq_order()`.

data type of empty array has to be set to make assert in `CombinedDataset.get_data_dtype()` work.

I thought about partition_epoch for CombinedDataset, but I think this should be done by the sub-datasets, thus the assert.


albertz:
Hey, did you check Travis? It fails for this pull request, for the demos. Is this related?

patrick-wilken:
Hmm, I can't see how this could be related. These changes have no impact on anything outside of CombinedDataset (and it is not used in the demo).

PR #84, opened 2018-11-15
Dataset for Siamese training with triplet loss

orech:
This dataset class implements triplet selection feature for training siamese networks with triplet loss.

### DIFF
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -472,3 +472,250 @@ class NextGenHDFDataset(CachedDataset2):
     :rtype: string
     """
     return name.replace('/', '\\')
+
+
+class SiameseHDFDataset(CachedDataset2):
+  """
+  SiameseHDFDataset class allows to do sequence sampling for weakly-supervised training.
+  It accepts data in the format of NextGenHDFDataset and performs sampling of sequence triplets before each epoch.
+  Triplets are tuples of the format: (anchor seq, random seq with the same label, random seq with a different label)
+  Here we assume that each dataset from the input .hdf has a single label.
+  In the config we can access streams by e.g. ["data:features_0"], ["data:features_1"], ["data:features_2"].
+  Split names depend on stream names in the input data, e.g. "features", "data", "classes", etc.
+  SiameseHDFDataset method _collect_single_seq(self, seq_idx) returns a DatasetSeq with extended dictionary of targets.
+  "data:features_0" key stands for features of anchor sequences from the input data.
+  In NexGenHDFDataset it would correspond to "data:features" or "data".
+  "data:features_1" is a key, which denote a pair of "data:features_0".
+  For each anchor sequence SiameseHDFDataset randomly samples a sequence with the same label.
+  "data:features_2" denotes the third element in a triplet tuple.
+  For each anchor sequence SiameseHDFDataset randomly samples a sequence with a different label.
+  Targets are splitted into different streams as well, e.g. "data:classes_0", "data:classes_1", "data:classes_2".
+
+  SiameseHDFDataset also supports non-uniform sampling and accepts a path to .npz matrix.
+  Rows of this matrix should have probabilities for each of the classes to be sampled.
+  This probability distribution might reflect class similarities.
+
+  This dataset might be useful for metric learning,
+  where we want to learn such representations of input sequences, that those which belong to the same class are close together,
+  while those with different labels should have representations far away from each other.
+  """
+  parsers = { 'feature_sequence'  : FeatureSequenceStreamParser,
+              'sparse'            : SparseStreamParser,
+              'segment_alignment' : SegmentAlignmentStreamParser }
+
+  def __init__(self, input_stream_name, seq_label_stream='words', class_distribution=None, files=None, *args, **kwargs):
+    """
+    :param str input_stream_name: name of a feature stream
+    :param str seq_label_stream: name of a stream with labels
+    :param str class_distribution: path to .npz file of size n x n (n is a number of classes),
+           where each line i contains probs of other classes to be picked in triplets
+           when sampling a pair for element from class i
+    :param files: list of paths to .hdf files
+    :param args: dict[str]
+    :param kwargs: dict[str]
+    """
+    super(SiameseHDFDataset, self).__init__(*args, **kwargs)
+    self.input_stream_name = input_stream_name
+    if class_distribution is not None:
+      self.class_probs = numpy.load(class_distribution)['arr_0']
+    else:
+      self.class_probs = None
+    self.files = []
+    self.h5_files = []
+    self.all_seq_names = []  # all_seq_names[(int)seq_index] = (string) sequence_name
+    self.seq_name_to_idx = {}  # (string) sequence_name -> seq_index (int)
+    self.file_indices = []  # file_indices[(int)seq_index] = file_index => indices of files to which sequences belongs to
+    self.seq_order = []
+    self.all_parsers = collections.defaultdict(list)
+    self.seq_to_target = {}  # (string) sequence_name -> (int) class_index
+    self.target_to_seqs = {}  # (int) class_index -> (string) sequence_names
+    self.curr_epoch_triplets = []
+    self.targets_stream = seq_label_stream
+    if files:
+      for fn in files:
+        self.add_file(fn)
+
+  def add_file(self, path):
+    """
+    register input files and sequences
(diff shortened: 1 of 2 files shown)


SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+
+from CachedDataset2 import CachedDataset2
+from HDFDataset import FeatureSequenceStreamParser, SparseStreamParser, SegmentAlignmentStreamParser
+from Dataset import DatasetSeq
+
+
+
+class SiameseHDFDataset(CachedDataset2):
+  """
+  SiameseHDFDataset class allows to do sequence sampling for weakly-supervised training
```
albertz: Can you make a dot (`.`) at the end of a sentence?

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+import numpy.random as random
+import numpy as np
+
+from CachedDataset2 import CachedDataset2
+from HDFDataset import FeatureSequenceStreamParser, SparseStreamParser, SegmentAlignmentStreamParser
+from Dataset import DatasetSeq
+
+
+
+class SiameseHDFDataset(CachedDataset2):
```
albertz: I think it would be cleaner (for now) to not have this in an own separate file but to move it to the `HDFDataset.py`.

orech:
> I don't exactly understand what the `SiameseHDFDataset` does. There is also no documentation describing that. Can you maybe add that? Esp I also want to know/understand why it is needed, why e.g. `NextGenHDFDataset` is not enough, or maybe `MetaDataset`. And this documentation should also be in the code.

I provided more documentation for the class. I hope it is more clear now, why using `NextGenHDFDataset` is not enough.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+    self.seq_to_target       = {} # (string) sequence_name -> (int) class_index
+    self.target_to_seqs      = {} # (int) class_index -> (string) sequence_names
+    self.curr_epoch_triplets = []
+
+    self.targets_stream = seq_label_stream
+    if files:
+      for fn in files:
+        self.add_file(fn)
+
+
```
albertz: Please follow proper Python conventions. There should just be a single empty line between methods. E.g. use PyCharm and follow the comments you get there.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+    :param kwargs: dict[str]
+    """
+    super(SiameseHDFDataset, self).__init__(*args, **kwargs)
+    self.input_stream_name = input_stream_name
+    if class_distribution is not None:
+      self.class_probs = np.load(class_distribution)['arr_0']
+    else:
+      self.class_probs = None
+
+    self.files           = []
```
albertz: There should just be a single space between `self.files` and `=`.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+    self.h5_files        = []
+    self.all_seq_names   = [] # all_seq_names[(int)seq_index] = (string) sequence_name
```
albertz: There should be two spaces between the end of the command (`[]`) and `#`.
Use PyCharm, it will notify you about standard Python conventions / code formatting standards.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+    self.targets_stream = seq_label_stream
+    if files:
+      for fn in files:
+        self.add_file(fn)
+
+
+  def add_file(self, path):
+    """
+    register input files and sequences
+    :param path: path to single .hdf file
```
albertz: Specify the type of path. I guess it's a string? Use the standard syntax, i.e. `:param str path: ...`.

SiameseDataset.py
albertz: Same for all remaining code...

SiameseDataset.py
albertz: Same for all remaining methods...

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+  while those with different labels should have representations far away from each other
+  """
+  parsers = { 'feature_sequence'  : FeatureSequenceStreamParser,
+              'sparse'            : SparseStreamParser,
+              'segment_alignment' : SegmentAlignmentStreamParser }
+
+  def __init__(self, input_stream_name, seq_label_stream='words', class_distribution=None, files=None, *args, **kwargs):
+    """
+
+    :param str input_stream_name: name of a feature stream
```
albertz: What is an example of this?

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+
+class SiameseHDFDataset(CachedDataset2):
+  """
+  SiameseHDFDataset class allows to do sequence sampling for weakly-supervised training
+  It accepts data in the format of NextGenHDFDataset and performs sampling of sequence triplets before each epoch
+  Here we assume that each dataset from the input .hdf has a single label
+  Sampled triplets are splitted into different streams
+  In the config we can access streams by e.g. ["data:features_0", "data:features_1", "data:features_2"],
```
albertz: So the dataset provides exactly those 3 input data streams, always named `features_0`, `features_1` and `features_2`? Or does it depend somehow on the parameter `input_stream_name` or so?
And then also exactly always 3 output data streams, named `classes_0`, `classes_1` and `classes_2`?
This should be very clear from the documentation.

SiameseDataset.py
albertz: Same for all of the following documentation...

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+from CachedDataset2 import CachedDataset2
+from HDFDataset import FeatureSequenceStreamParser, SparseStreamParser, SegmentAlignmentStreamParser
+from Dataset import DatasetSeq
+
+
+
+class SiameseHDFDataset(CachedDataset2):
+  """
+  SiameseHDFDataset class allows to do sequence sampling for weakly-supervised training
+  It accepts data in the format of NextGenHDFDataset and performs sampling of sequence triplets before each epoch
```
albertz: Describe what that means. I have no idea what sequence triplets are.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+  Here we assume that each dataset from the input .hdf has a single label
+  Sampled triplets are splitted into different streams
+  In the config we can access streams by e.g. ["data:features_0", "data:features_1", "data:features_2"],
+  "data:features_0" stands for features of all sequences from the input data,
```
albertz: So, multiple sequences? Or what is "all sequences"? How many sequences? And they are concatenated?
(All this should be clear from the documentation.)

albertz: Also, why do you call it `features_0`? Why not `features_all` or sth like that? The name should better reflect what data the data stream contains.

SiameseDataset.py
```diff
@@ -0,0 +1,271 @@
...
+  "data:features_1" is a stream of pairs, where for each sequence from "data:features_0" SiameseHDFDataset randomly samples a sequence with the same label
```
albertz: I don't really understand what you mean by "stream of pairs". Can you specify that?
Esp, specify the dtype (I guess int32?), and the shape (is it like (num_sampled_sequences, 2)?).

SiameseDataset.py
albertz: If concatenated, concatenated in time? So you end up with a single sequence of shape (sum_of_time, dimension)?

albertz: If concatenated, how would you be able to split them later? Or you never need that?

tests/test_siamese_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+    dt = h5py.special_dtype(vlen=str)
+    sequence_names_data = dataset.create_dataset("seq_names", shape=(len(seq_names),), dtype=dt)
+    for ind, val in enumerate(seq_names):
+      sequence_names_data[ind] = val
+
+    dataset.close()
+  return ['./dummy.%i.hdf5' % idx for idx in range(1, num_datasets + 1)]
+
+
+def test_triplet_sampling():
```
albertz: Do not add this to a new separate test file. This can be part of `test_HDFDataset.py`.

albertz:
I merged it in now. But look through my earlier comments, some of them still should be addressed (documentation mostly). But you don't need to make new pull requests for this; just push directly (I think you have access, right?).

PR #88, opened 2018-11-30
TFNetworkLayer.py added flag masked_time to BatchNorm layer

squarenabla:
When possible (only standard set of parameters is given) use tf.layers.batch_normalization with fused ops. Works faster.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -818,7 +818,8 @@ class LayerBase(object):
                  use_shift=True, use_std=True, use_sample=0.0, force_sample=False,
                  momentum=0.99, epsilon=1e-3,
                  sample_mean=None, sample_variance=None,
-                 gamma=None, beta=None):
+                 gamma=None, beta=None,
+                 masked_time=True):
     """
     :param Data data:
     :param bool use_shift:
@@ -831,6 +832,7 @@ class LayerBase(object):
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool masked_time: flatten and mask input tensor
     :rtype: tf.Tensor
 
     http://arxiv.org/abs/1502.03167
@@ -840,8 +842,12 @@ class LayerBase(object):
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
     with tf.variable_scope("batch_norm"):
-      x = data.get_placeholder_flattened(keep_dims=True)  # shape (time',...)
-      mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      if masked_time:
+        x = data.get_placeholder_flattened(keep_dims=True)
+        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      else:
+        x = data.placeholder
+        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
       if sample_mean is None:
         with self.var_creation_scope():
           sample_mean = self.add_param(tf.Variable(
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -868,6 +868,63 @@ def test_pool_layer_NCHW():
     print(seq_lens)
 
 
+def test_batch_norm():
+  with make_scope() as session:
+    import numpy as np
+    from tensorflow.python.client import timeline
+    net = TFNetwork(extern_data=ExternData())
+    net.train_flag = True
+    with tf.variable_scope("src_nchw"):
+      src_nhwc = InternalLayer(name="src_nchw", network=net, out_type={"dim": 16,
+                                                                       "shape": (None, 16, 16),
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 1,
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    mean =  tf.constant(np.random.rand(1, 1, 16, 16), name="rand_mean", dtype=tf.float32)
+    variance = tf.constant(np.random.rand(1, 1, 16, 16), name="rand_var", dtype=tf.float32)
+    input_data = np.random.rand(10, 11, 16, 16)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
+
+    with tf.variable_scope("batch_norm_masked_nchw"):
+      batch_norm_1 = BatchNormLayer(name="batch_norm_masked_nchw", network=net, masked_time=True,
+                                    sample_mean=mean, sample_variance=variance,
+                                    sources=[src_nhwc],
+                                    output=BatchNormLayer.get_out_data_from_opts(name="batch_norm_masked_nchw",
+                                                                                 sources=[src_nhwc],
+                                                                                 network=net))
+      batch_norm_1.post_init(layer_desc=None)
+
+    with tf.variable_scope("batch_norm_nonmasked_nchw"):
+      batch_norm_2 = BatchNormLayer(name="batch_norm_nonmasked_nchw", network=net, masked_time=False,
+                                    sample_mean=mean, sample_variance=variance,
+                                    sources=[src_nhwc],
+                                    output=BatchNormLayer.get_out_data_from_opts(name="batch_norm_nonmasked_nchw",
+                                                                                 sources=[src_nhwc],
+                                                                                 network=net))
+      batch_norm_2.post_init(layer_desc=None)
+
+
+    tf.global_variables_initializer().run()
+    out_1, seq_lens_1 = session.run([batch_norm_1.output.placeholder,
+                                 batch_norm_1.output.size_placeholder[0]],
+                                feed_dict={src_nhwc.output.placeholder: input_data,
+                                           src_nhwc.output.size_placeholder[0]: seq_lens}
+                                )
+    out_2, seq_lens_2 = session.run([batch_norm_2.output.placeholder,
+                                 batch_norm_2.output.size_placeholder[0]],
+                                feed_dict={src_nhwc.output.placeholder: input_data,
+                                           src_nhwc.output.size_placeholder[0]: seq_lens}
+                                )
+
+    assert np.array_equal(out_1, out_2)
+    print(np.sum(out_1 - out_2))
+
+
 def test_ResizeLayer_fill_value():
   with make_scope() as session:
     net = TFNetwork(extern_data=ExternData())
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -1026,10 +1026,11 @@ class Data(object):
       x.set_shape([None] * self.batch_ndim)
     return x
 
-  def get_axes(self, exclude_time=False, exclude_batch=False):
+  def get_axes(self, exclude_time=False, exclude_batch=False, exclude_feature=False):
     """
     :param bool exclude_time: will filter out the time-axis
     :param bool exclude_batch: will filter out the batch-axis
+    :param bool exclude_feature: will filter out the feature-axis
     :return: list of axes, like `range(len(self.shape))`, calculated with batch dim.
     :rtype: list[int]
     """
@@ -1038,6 +1039,8 @@ class Data(object):
       axes.pop(axes.index(self.time_dim_axis))
     if exclude_batch and self.batch_dim_axis is not None:
       axes.pop(axes.index(self.batch_dim_axis))
+    if exclude_feature and self.feature_dim_axis is not None:
+      axes.pop(axes.index(self.feature_dim_axis))
     return axes
 
   def get_axes_from_description(self, axes):


TFNetworkLayer.py
```diff
@@ -867,8 +873,12 @@ def batch_norm(self, data,
       if use_std:
         if gamma is None:
           with self.var_creation_scope():
+            if gamma_init == "zeros":
+              gamma_initializer = tf.zeros_initializer()
+            else:
+              gamma_initializer = tf.ones_initializer()
             gamma = self.add_param(tf.Variable(
-              initial_value=tf.ones(data.get_bc_spatial_batch_shape()),
+              initial_value=gamma_initializer(data.get_bc_spatial_batch_shape()),
```
albertz: Remove if not needed.

TFNetworkLayer.py
```diff
@@ -882,6 +892,46 @@ def batch_norm(self, data,
...
+    """
+    x = data.get_placeholder_as_batch_major()
+    axis = data.feature_dim_axis
+    training = self.network.train_flag
+    with self.var_creation_scope():
+      if gamma_init == "zeros":
+        gamma_initializer = tf.zeros_initializer()
+      else:
+        gamma_initializer = tf.ones_initializer()
+      bn = tf.layers.batch_normalization(
```
albertz: Use `tf.nn.fused_batch_norm` instead.

TFNetworkLayer.py
```diff
@@ -882,6 +892,46 @@ def batch_norm(self, data,
...
+    import re
+    scope_name_prefix = tf.get_variable_scope().name + "/"  # e.g. "layer1/"
+    params = tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope=re.escape(scope_name_prefix))
+    for p in params:
+      if not p.name.startswith(scope_name_prefix):
+        continue
+      assert p.name.startswith(scope_name_prefix) and p.name.endswith(":0")
+      self.params[p.name[len(scope_name_prefix):-2]] = p
+    return bn
+
```
albertz: Remove the function `LayerBase.fused_batch_norm`.
Directly use `tf.nn.fused_batch_norm` in the `LayerBase. batch_norm` function. Also the complex parameter handling is not needed when you directly use `tf.nn.fused_batch_norm`.

TFNetworkLayer.py
```diff
@@ -831,6 +832,7 @@ def batch_norm(self, data,
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param str gamma_init: gamma initializer
```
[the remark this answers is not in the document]
squarenabla: I use this in some cases. e.g. trying this variant:
https://github.com/tensorpack/tensorpack/blob/ef334e31c34bab909d2c7469d3905347a0d5137a/examples/ResNet/resnet_model.py#L31

TFNetworkLayer.py
squarenabla: Ok

TFNetworkLayer.py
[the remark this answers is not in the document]
albertz: Hm, ok.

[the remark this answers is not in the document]
albertz: (They don't really provide any justification. I think the effect of this is that for a short while gamma will be very small and that is similar to having a small learning rate, or to initialize some of the weight matrices with very small values. But you could instead just do that, i.e. initialize the weights with smaller values, or use some learning rate warmup.)

TFNetworkLayer.py
```diff
@@ -855,31 +866,44 @@ def batch_norm(self, data,
...
       use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
       mean = (1. - use_sample) * mean + use_sample * sample_mean
       variance = (1. - use_sample) * variance + use_sample * sample_variance
-      bn = (data.placeholder - mean) * tf.rsqrt(variance + epsilon)
       if use_std:
         if gamma is None:
           with self.var_creation_scope():
+            if gamma_init == "zeros":
+              gamma_initializer = tf.zeros_initializer()
+            else:
```
albertz: The `else:` is too generic. You should fail if there is an unexpected `gamma_init`.

TFNetworkLayer.py
```diff
@@ -855,31 +866,44 @@ def batch_norm(self, data,
...
         sample_variance = tf.assign_add(sample_variance, (variance - sample_variance) * momentum)
       # If train or if force_sample, use default use_sample=0.0, otherwise use_sample=1.0.
       use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
       mean = (1. - use_sample) * mean + use_sample * sample_mean
       variance = (1. - use_sample) * variance + use_sample * sample_variance
-      bn = (data.placeholder - mean) * tf.rsqrt(variance + epsilon)
       if use_std:
         if gamma is None:
           with self.var_creation_scope():
+            if gamma_init == "zeros":
```
albertz: This code is kind of replicated (and much more restricted) from similar code which we already have, e.g. `get_initializer`. Better just use just `get_initializer` instead.

TFNetworkLayer.py
albertz: (But see my comment about `get_initializer`, then this is anyway obsolete.)

TFNetworkLayer.py
```diff
@@ -840,12 +844,19 @@ def batch_norm(self, data,
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
     with tf.variable_scope("batch_norm"):
-      x = data.get_placeholder_flattened(keep_dims=True)  # shape (time',...)
-      mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      if masked_time:
+        keep_dims = False if fused else True
+        x = data.get_placeholder_flattened(keep_dims=keep_dims)
+        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=keep_dims)
+        shape = [data.dim] if fused else data.get_bc_spatial_batch_shape()
```
albertz: `shape` looks very much wrong to me. If `fused`, you collapse batch + time together into a single, but then you still have 2 axes (at least), e.g. (batch*time, dim).

albertz: But as shape should match for broadcasting, it should be like `(1, data.dim)` or so. Actually better just use `list(data.get_bc_spatial_batch_shape())` and then remove the time-axis.

TFNetworkLayer.py
```diff
@@ -840,12 +844,19 @@ def batch_norm(self, data,
...
+      else:
+        x = data.placeholder
+        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=False)
+        shape = mean.shape
```
albertz: `mean.shape` is a `TensorShape` but you want to have a bc-able shape here.

TFNetworkLayer.py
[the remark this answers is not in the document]
squarenabla: Ok. Then I'll remove this

TFNetworkLayer.py
albertz: Ah sorry, I think I was wrong about `shape`. But the code is slightly confusing. Make it depend on `keep_dims`, not on `fused`.
(And also consider / keep in mind that at some point, maybe it's actually more clean to have `keep_dims = False` in all case, so maybe we want to change the code later.)

TFNetworkLayer.py
```diff
@@ -863,23 +869,26 @@ def batch_norm(self, data,
       use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
       mean = (1. - use_sample) * mean + use_sample * sample_mean
       variance = (1. - use_sample) * variance + use_sample * sample_variance
-      bn = (data.placeholder - mean) * tf.rsqrt(variance + epsilon)
```
albertz: Leave that in.

TFNetworkLayer.py
```diff
@@ -863,23 +869,26 @@ def batch_norm(self, data,
...
       if use_std:
         if gamma is None:
           with self.var_creation_scope():
             gamma = self.add_param(tf.Variable(
               initial_value=tf.ones(data.get_bc_spatial_batch_shape()),
               name="%s_%s_gamma" % (self.name, data.name),
               trainable=True))
-        bn *= gamma
```
albertz: Leave that in.

TFNetworkLayer.py
```diff
@@ -863,23 +869,26 @@ def batch_norm(self, data,
...
+      else:
+        gamma = 1.0
       if use_shift:
         if beta is None:
           with self.var_creation_scope():
             beta = self.add_param(tf.Variable(
               initial_value=tf.zeros(data.get_bc_spatial_batch_shape()),
               name="%s_%s_beta" % (self.name, data.name),
               trainable=True))
-        bn += beta
```
albertz: Leave that in.

tests/test_TFNetworkLayer.py
```diff
@@ -1078,3 +1126,50 @@ def make_feed_dict(seq_len=10):
     #if len(list(threading.enumerate())) > 1:
     #  print("Warning, more than one thread at exit:")
     #  better_exchook.dump_all_thread_tracebacks()
+  with make_scope() as session:
```
albertz: I think this ended up in the wrong function?
Please sanity check the code/changes you actually committed.

tests/test_TFNetworkLayer.py
```diff
@@ -868,6 +868,54 @@ def test_pool_layer_NCHW():
...
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 2,
+                                                                       "feature_dim_axis": 1,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, 16, None, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {1: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    with tf.variable_scope("batch_norm_masked_nchw"):
+      batch_norm_1 = BatchNormLayer(name="batch_norm_masked_nchw", network=net, masked_time=False,
```
albertz: You write `batch_norm_masked_nchw` but you have `masked_time=False`. This seems wrong...
(Please double check your own code...)

tests/test_TFNetworkLayer.py
```diff
@@ -868,6 +868,54 @@ def test_pool_layer_NCHW():
...
+                                feed_dict={src_nhwc.output.placeholder: np.random.rand(10, 16, 11, 16),
+                                           src_nhwc.output.size_placeholder[1]: np.full(shape=(10,), fill_value=11)}
+                                )
+    print(out.shape)
+    out, seq_lens = session.run([batch_norm_2.output.placeholder,
+                                 batch_norm_2.output.size_placeholder[1]],
+                                feed_dict={src_nhwc.output.placeholder: np.random.rand(10, 16, 11, 16),
+                                           src_nhwc.output.size_placeholder[1]: np.full(shape=(10,), fill_value=11)}
+                                )
+    print(out.shape)
```
albertz: The test is kind of incomplete.
You should also check that the output is correct.
You should also prepare the input in such a way that the output should be the same (of `batch_norm_1` and `batch_norm_2`), and then do an actually assert on that.
Also, the test should cover a more interesting case where the seq lens are not all full, but e.g. some of them are also less than full (e.g. 5 or so). And then prepare the input data in such a way that it's zero for the other values, so that actually `masked_time=True` and `masked_time=False` really gives the same result.

albertz:
Btw, also check the tests... (They fail.)

tests/test_TFNetworkLayer.py
squarenabla: It's not clear how to do this. If the seqs are not full, the result will be different. And filling the data with 0s wouldn't solve this (mean won't change but variance will).

squarenabla: I think, if seq lens are not full, masking has to be applied

squarenabla:
I check the failing test, locally it works.

tests/test_TFNetworkLayer.py
```diff
@@ -868,6 +868,63 @@ def test_pool_layer_NCHW():
...
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    mean =  tf.constant(np.random.rand(1, 1, 16, 16), name="rand_mean", dtype=tf.float32)
+    variance = tf.constant(np.random.rand(1, 1, 16, 16), name="rand_var", dtype=tf.float32)
+    input_data = np.random.rand(10, 11, 16, 16)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: This is not the interesting test case, as I explained. The `seq_lens` should not all be the same.

tests/test_TFNetworkLayer.py
albertz: Yes, but that's the point. Only in that case, the test make sense. And yes, we actually discussed this together, of course the result is different then, but we discussed how to get from one number to the other. Instead of filling with 0s, fill it such that it will become the same.

albertz:
What is the state here?

squarenabla:
Sorry didn't notice last message. I'm working on test with different seq_lens at the moment. I'll finish the pull request soon.

PR #89, opened 2018-11-30
Get sub layer

ArneNx:
Added functionality to allow addressing of sub-layers vie "/" separated path. 
This is useful e.g. if you want to share parameters between two recurrent units.

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -383,18 +383,26 @@ class TFNetwork(object):
     if name in self._constructing_layers:
       raise NetworkConstructionDependencyLoopException(
         layer_name=name, constructing_layers=self._constructing_layers, net_dict=net_dict, network=self)
+    if not get_layer:
+      def get_layer(src_name):
+        return self.construct_layer(net_dict=net_dict, name=src_name)  # set get_layer to wrap construct_layer
     if name not in net_dict:
       if name == "data":
         layer_desc = {"class": "source", "from": []}
       elif name.startswith("data:"):
         layer_desc = {"class": "source", "data_key": name[len("data:"):], "from": []}
+      elif '/' in name:
+        # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
+        # but maybe it's not constructed yet, so try constructing the root layer
+        get_layer(name.split('/')[0])
+        assert name.split('/')[0] in self.layers, ("Root %r layer was not constructed! "
+                                                   "Unable to get layer %r." % (name.split('/')[0], name))
+        # constructing the root layer should have constructed all its children
+        return self.get_layer(name)  # ...so try again now
       else:
         raise LayerNotFound("layer %r not found in %r" % (name, self))
     else:
       layer_desc = net_dict[name]
-    if not get_layer:
-      def get_layer(src_name):
-        return self.construct_layer(net_dict=net_dict, name=src_name)  # set get_layer to wrap construct_layer
     if not add_layer:
       add_layer = self.add_layer
     self.layers_desc[name] = layer_desc
@@ -696,6 +704,12 @@ class TFNetwork(object):
       if not self.parent_net:
         raise LayerNotFound("cannot get layer %r, no parent net for %r" % (layer_name, self))
       return self.parent_net.get_layer(layer_name[len("base:"):])
+    if '/' in layer_name:
+      # this is probably a path to a sub-layer
+      root_layer = self.get_layer(layer_name.split('/')[0])  # get the root-layer (first part of the path)
+      sub_layer = root_layer.get_sub_layer('/'.join(layer_name.split('/')[1:]))  # get the sub-layer from the root-layer
+      if sub_layer:  # get_sub_layer returns None by default (if sub-layer not found)
+        return sub_layer
     if layer_name not in self.layers:
       raise LayerNotFound("layer %r not found in %r" % (layer_name, self))
     return self.layers[layer_name]
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -220,13 +220,15 @@ class RecLayer(_ConcatInputLayer):
   @classmethod
   def transform_config_dict(cls, d, network, get_layer):
     """
+    This method transforms the templates in the config dictionary into references
+    of the layer instances (and creates them in the process).
     :param dict[str] d: will modify inplace
     :param TFNetwork.TFNetwork network:
     :param ((str) -> LayerBase) get_layer: function to get or construct another layer
     """
     if isinstance(d.get("unit"), dict):
       d["n_out"] = d.get("n_out", None)  # disable automatic guessing
-    super(RecLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+    super(RecLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)  # everything except "unit"
     if "initial_state" in d:
       d["initial_state"] = RnnCellLayer.transform_initial_state(
         d["initial_state"], network=network, get_layer=get_layer)
@@ -234,7 +236,7 @@ class RecLayer(_ConcatInputLayer):
       def sub_get_layer(name):
         # Only used to resolve deps to base network.
         if name.startswith("base:"):
-          return get_layer(name[len("base:"):])
+          return get_layer(name[len("base:"):])  # calls get_layer of parent network
       from TFNetwork import TFNetwork, ExternData
       subnet = TFNetwork(parent_net=network, extern_data=network.extern_data)  # dummy subnet
       for sub in d["unit"].values():  # iterate over the layers of the subnet
@@ -725,6 +727,17 @@ class RecLayer(_ConcatInputLayer):
       return layer.is_prev_time_frame
     return False
 
+  def get_sub_layer(self, layer_name):
+    """
+    :param str layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    :rtype: LayerBase|None
+    """
+    if isinstance(self.cell, _SubnetworkRecCell):
+      # try to find layer_name in cell:
+      return self.cell.get_layer_from_outside(layer_name)
+    return None
+
 
 class _SubnetworkRecCell(object):
   """
@@ -1066,6 +1079,22 @@ class _SubnetworkRecCell(object):
         continue
       get_layer(layer_name)
       assert layer_name in self.net.layers
+
+  def get_layer_from_outside(self, layer_name):
+    """
+    :param str layer_name: The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    :rtype: LayerBase|None
+    """
+    if self.output_layers_net and layer_name in self.output_layers_net.layers:
+      return self.output_layers_net.layers[layer_name]
+    elif self.input_layers_net and layer_name in self.input_layers_net.layers:
+      return self.input_layers_net.layers[layer_name]
+    elif self.net and layer_name in self.net.layers:
+      raise Exception(
+        "%r: Cannot get layer %r from outside, because it is only available inside the recurrent loop. \
+         Add 'is_output_layer':True to the layer options." % (self.parent_rec_layer, layer_name))
+    return None
 
   def _get_init_output(self, name):
     """
@@ -2138,7 +2167,7 @@ class _SubnetworkRecCell(object):
       if name.startswith("prev:"):
         return get_prev_layer(name[len("prev:"):])
       if name.startswith("base:"):
-        return self.parent_net.layers[name[len("base:"):]]
+        return self.parent_net.get_layer(name[len("base:"):])
       if name in self.input_layers_moved_out:
         return self.input_layers_net.layers[name]
       if name in self.output_layers_moved_out or name.startswith("data:"):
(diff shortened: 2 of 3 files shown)


albertz:
Hey, can you split this into multiple pull requests? Only keep the commits here which are really about the sub layer.

TFEngine.py
```diff
@@ -954,10 +954,12 @@ def init_network_from_config(self, config=None, net_dict_post_proc=None):
         self_prefix = self.network.get_absolute_name_scope_prefix()  # "" if root, otherwise with "/" at end
         load_if_prefix = opts.get('prefix', '')  # prefix to identify the variables to be restored from the file
         from TFNetwork import CustomCheckpointLoader
+        ignore_missing = opts.get("ignore_missing", False)
         loader = CustomCheckpointLoader(
-          filename=model_filename, saveable_params=self.network.get_trainable_params(),
+          filename=model_filename,
+          saveable_params=self.network.get_params_list() if ignore_missing else self.network.get_trainable_params(),
```
albertz: Why? (But anyway, separate pull request.)

TFNetwork.py
```diff
@@ -684,6 +692,7 @@ def get_layer(self, layer_name):
     :param str layer_name:
     :rtype: LayerBase
     """
+
```
albertz: Why?

TFNetwork.py
```diff
@@ -696,6 +705,12 @@ def get_layer(self, layer_name):
       if not self.parent_net:
         raise LayerNotFound("cannot get layer %r, no parent net for %r" % (layer_name, self))
       return self.parent_net.get_layer(layer_name[len("base:"):])
+    if '/' in layer_name:
```
albertz: This logic should not be in `construct_layer`, but in `get_layer`.

TFNetworkRecLayer.py
```diff
@@ -720,6 +722,15 @@ def is_prev_step_layer(cls, layer):
       return layer.is_prev_time_frame
     return False
 
+  def get_sub_layer(self, layer_name: str):
+    """
+    :param layer_name:  The sub_layer addressed by '/' separated path.
```
albertz: Define the type in the docstring: `:param str layer_name`

TFNetworkRecLayer.py
```diff
@@ -720,6 +722,15 @@ def is_prev_step_layer(cls, layer):
       return layer.is_prev_time_frame
     return False
 
+  def get_sub_layer(self, layer_name: str):
```
albertz: This is Python 3 only. We still want to keep Python 2 support (not sure for how long, but for now).

TFNetworkRecLayer.py
```diff
@@ -720,6 +722,15 @@ def is_prev_step_layer(cls, layer):
...
+    """
+    :param layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
```
albertz: Also define the return type, i.e. like `:rtype: LayerBase|None`.

TFNetworkRecLayer.py
```diff
@@ -1059,6 +1070,19 @@ def get_layer(name):
       get_layer(layer_name)
       assert layer_name in self.net.layers
 
+  def get_sub_layer(self, layer_name: str):
+    """
+    :param layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    """
```
albertz: Same comments as before.

TFNetworkRecLayer.py
```diff
@@ -720,6 +722,15 @@ def is_prev_step_layer(cls, layer):
       return layer.is_prev_time_frame
     return False
 
+  def get_sub_layer(self, layer_name: str):
+    """
+    :param layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    """
+    if '/' in layer_name and layer_name.split('/')[0] == 'unit':
```
albertz: Why this logic? Why only if it is `unit`? I would not do that, but rather for any `layer_name`.
But instead, check `isinstance(self.cell, _SubnetworkRecCell)`, because only in that case you want that.

tests/test_TFNetworkRecLayer.py
```diff
@@ -1429,7 +1429,6 @@ def train(net):
   assert not train_not_optim_out_layer.cell.output_layers_moved_out
   train(train_not_optim_net)
 
-
```
albertz: Why?

albertz:
Also many tests fail (I didn't really looked at the details). Please check.

TFNetwork.py
```diff
@@ -1,6 +1,8 @@
 
 from __future__ import print_function
 
+import pprint
```
albertz: Do `from pprint import pprint` instead. Or better: Don't do that here at all. Do it locally in the method/function where you need it.

TFNetwork.py
```diff
@@ -1,6 +1,8 @@
...
+
```
albertz: No need for this extra empty line.

LmDataset.py
```diff
@@ -84,6 +89,12 @@ def __init__(self,
       self.orth_symbols = orth_symbols
       self.labels["data"] = orth_symbols
       self.seq_gen = None
+    if orth_symbols_map_file and orth_symbols_map_file.endswith('.pkl'):
+      with open(orth_symbols_map_file, 'rb') as f:
+        self.orth_symbols_map = pickle.load(f)
```
albertz: Maybe the `import pickle` locally here.

LmDataset.py
```diff
@@ -13,6 +13,7 @@
 import numpy
 import time
 import re
+import pickle
```
albertz: Maybe move below.

TFNetwork.py
ArneNx: I guess you're talking about line 391 to 396, right? 
Until now, the `get_layer` method is not performing/ calling any construction. Do you really want to change that?

ArneNx:
I think I have addressed all your remarks now. 
The only thing remaining is our discussion about whether to put some of the logic into `construct_layer` (see my response above). Have you taken another look at this? 
And yes, this feature is definitely needed!

TFNetworkRecLayer.py
```diff
@@ -1067,6 +1081,20 @@ def get_layer(name):
       get_layer(layer_name)
       assert layer_name in self.net.layers
 
+  def get_sub_layer(self, layer_name):
```
albertz: Better rename to `get_layer`.

TFNetworkRecLayer.py
```diff
@@ -1067,6 +1081,20 @@ def get_layer(name):
...
+    :param str layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    :rtype: LayerBase|None
+    """
+    if self.output_layers_net and layer_name in self.output_layers_net.layers:
+      return self.output_layers_net.layers[layer_name]
+    elif self.input_layers_net and layer_name in self.input_layers_net.layers:
+      return self.input_layers_net.layers[layer_name]
+    elif self.net and layer_name in self.net.layers:
+      return self.net.layers[layer_name]
```
albertz: I'm not sure if it is a good idea to allow access to `self.net` here as well. Because this is inside the recurrent loop, thus:
* It is without the time dim. This is not consistent when you get the layer from e.g. output_layers_net or input_layers_net. But it should be consistent.
* You also cannot access the `layer.output.placeholder`, because it is inside the recurrent loop. So depending on the use case, this doesn't make sense.

I think I would throw an exception here, like:
```suggestion
      raise Exception("%r: Cannot get layer %r from outside, because it is only available inside the recurrent loop. Add 'is_output_layer':True to the layer options." % (self.parent_rec_layer, layer_name))
```

TFNetworkRecLayer.py
albertz: Or maybe `get_layer_from_outside`.

TFNetwork.py
albertz: I mean the added empty line.

TFNetwork.py
```diff
@@ -388,6 +388,12 @@ def construct_layer(self, net_dict, name, get_layer=None, add_layer=None, check_
         layer_desc = {"class": "source", "from": []}
       elif name.startswith("data:"):
         layer_desc = {"class": "source", "data_key": name[len("data:"):], "from": []}
+      elif '/' in name:
+        # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
+        # but maybe it's not constructed yet, so try constructing the root layer
+        self.construct_layer(net_dict, name.split('/')[0], get_layer, add_layer, check_existing)
+        # constructing the root layer should have constructed all its children
+        return self.get_layer(name)  # ...so try again now
```
albertz: It should not call `self.construct_layer` in all cases, but rather `get_layer` (the param passed to this function).
Thus, move that code `if not get_layer: ...`, which is below, above this, and then call `get_layer` here.

TFNetwork.py
albertz: Maybe this was moved. I think it was in `LayerBase.construct_layer` before. Now it's in `LayerBase.get_layer`, which is what I wanted.

TFNetwork.py
ArneNx: Ah, ok. Fixed that.

TFNetworkRecLayer.py
```diff
@@ -725,6 +727,18 @@ def is_prev_step_layer(cls, layer):
       return layer.is_prev_time_frame
     return False
 
+  def get_sub_layer(self, layer_name):
+    """
+    :param str layer_name:  The sub_layer addressed by '/' separated path.
+    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+    :rtype: LayerBase|None
+    """
+    if '/' in layer_name and isinstance(self.cell, _SubnetworkRecCell):
```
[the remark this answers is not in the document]
albertz: Why the extra `"unit/"`? I think this is confusing, and also more complicated than it needs to be.

TFNetwork.py
```diff
@@ -383,18 +383,24 @@ def construct_layer(self, net_dict, name, get_layer=None, add_layer=None, check_
...
       if name == "data":
         layer_desc = {"class": "source", "from": []}
       elif name.startswith("data:"):
         layer_desc = {"class": "source", "data_key": name[len("data:"):], "from": []}
+      elif '/' in name:
+        # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
+        # but maybe it's not constructed yet, so try constructing the root layer
+        get_layer(name.split('/')[0])
+        # constructing the root layer should have constructed all its children
+        return get_layer(name)  # ...so try again now
```
albertz: Do not call `get_layer(name)` here. This could lead to an infinite recursion (if this is not an existing hierarchical path).
Rather, assert that `name.split('/')[0] in self.layers` now, and then move on right here, so it will go into the code `if '/' in layer_name: ...` below.

TFNetworkRecLayer.py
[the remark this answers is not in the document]
albertz: I would not do that. This is purely artificial. Actually I don't like it too much that it is called `"unit"` currently. Maybe we rename that later. But anyway, this logic is kind of independent (or should be independent) how the parameter of the class is named. So I would just remove that.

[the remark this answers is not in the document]
ArneNx: I can see why this would be simpler (although I still think people will first try to access it through the path they see in the network dictionary).
So I would change this to remain independent of changes in naming of other layers. 
But what would you think of having something like 
```
if 'unit' == layer_name.split('.')[0]: 
  layer_name = layer_name.split('.')[1:] 
```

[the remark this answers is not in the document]
albertz: Hm, I don't think so. I would just remove the `"unit"` part. It does not really add any new functionality or ability. And also keep it simple for now, so no extra cases.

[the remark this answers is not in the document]
ArneNx: Ok... I changed it to expect "base_layer/sub_layer"

TFNetwork.py
ArneNx: You're right. I ignored that case. 
See my newest change.

TFNetwork.py
```diff
@@ -383,18 +383,26 @@ def construct_layer(self, net_dict, name, get_layer=None, add_layer=None, check_
...
     if name not in net_dict:
       if name == "data":
         layer_desc = {"class": "source", "from": []}
       elif name.startswith("data:"):
         layer_desc = {"class": "source", "data_key": name[len("data:"):], "from": []}
+      elif '/' in name:
+        # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
+        # but maybe it's not constructed yet, so try constructing the root layer
+        get_layer(name.split('/')[0])
+        assert name.split('/')[0] in self.layers, "Root %r layer was not constructed! " \
```
albertz: Small detail: Instead of `\` for splitting this into two lines, I prefer to use brackets, like `assert ...., ("Root ... ! "` newline `"Unable to ..." % (...))`.
(Despite personal taste, there is also one small advantage of this: When an exception would occur here, the code line in the stack trace would be the last line of the statement, i.e. in your case the `"Unable to ..." % ..`. When showing that statement alone in a debug stack trace, you might miss that this is part of another statement. E.g. better_exchook looks if the stack frame code line contains any non-matching `)`, and only then it would show also the preceding code line.)

PR #94, opened 2018-12-04
Easier configuration of MetaDataset

patrick-wilken:
1. Get data dims from the sub-datasets instead of explicitly giving it in the config. This is straight forward and I don't know why it hasn't been this way all along 😄 

2. When we combine datasets of the same (or similar) type, it might be straight forward to see which sequences correspond. For example when combining two TranslationDatasets most likely "line-n" of dataset A belongs to "line-n" of dataset B.
For this purpose, I implemented reading the list of sequence tags from one dataset (the one containing the "data" feature) and applying it to all others. In this case a sequence list created by the user is no longer necessary.

### DIFF
--- a/MetaDataset.py
+++ b/MetaDataset.py
@@ -80,7 +80,7 @@ class MetaDataset(CachedDataset2):
   def __init__(self,
                datasets,
                data_map,
-               seq_list_file,
+               seq_list_file=None,
                seq_lens_file=None,
                data_dims=None,
                data_dtypes=None,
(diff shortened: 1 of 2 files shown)


MetaDataset.py
```diff
@@ -127,11 +121,6 @@ def __init__(self,
       self._seq_lens = None
       self._num_timesteps = None
 
-    # Will only init the needed datasets.
```
patrick-wilken: I removed this comment as I never understood it.

MetaDataset.py
```diff
@@ -145,6 +134,38 @@ def __init__(self,
...
     self.data_dtypes = {data_key: _select_dtype(data_key, self.data_dims, data_dtypes) for data_key in self.data_keys}
 
+  def _load_seq_list(self, seq_list_file=None):
+    if seq_list_file:
+      if seq_list_file.endswith(".pkl"):
+        import pickle
+        seq_list = pickle.load(open(seq_list_file, 'rb'))
+      else:
+        seq_list = open(seq_list_file).read().splitlines()
+    else:
```
patrick-wilken: Had to change the order quite a bit, really new is just this else clause.

MetaDataset.py
albertz: In `datasets` you could e.g. define many different datasets (e.g. `dataset1`, `dataset2`, `tmp`), but not all of them will be used. Only those are used which you use by `data_map`. The comment just says, if e.g. `tmp` is not used (via `data_map`), it will also not be initialized.

I would leave the comment in. Maybe make it more clear, if you find it confusing.

MetaDataset.py
```diff
@@ -79,18 +79,19 @@ class MetaDataset(CachedDataset2):
 
   def __init__(self,
                datasets,
-               data_map, data_dims,
-               seq_list_file,
+               data_map,
+               seq_list_file=None,
```
albertz: It should stay compatible with existing configs.
So leave the old support in, do not remove the `data_dims` parameter. Make it optional instead.

There is also one reason to have it explicit:
It's basically a double check. You see more directly from the config what you expect. Like an additional `assert`.

MetaDataset.py
patrick-wilken: Ah, makes sense.

patrick-wilken:
Ok, got it now.

albertz:
Is this ready now? (You can click on "request review again" or sth like that to notify me.)

patrick-wilken:
I can't find such a button 🤔 But yes, it's done.

PR #97, opened 2018-12-09
Keep track of constraints on layers that are moved out of the RecLayer

### LINKED ISSUE #96: Constraints on layers that are moved out of the RecLayer are lost
Because layers that are moved out of the `RecLayer` are moved either to the `input_layers_net` or the `output_layers_net`, they are not checked for possible constraints in [`get_constraints_value()`](https://github.com/rwth-i6/returnn/blob/master/TFNetworkRecLayer.py#L426). Therefore, the constraints are lost.

I've created a PR with a possible fix as well as test cases for all three situations (no constraints, constraints on moved out input/output layer).

ChristopherBrix:
This fixes #96

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -427,7 +427,12 @@ class RecLayer(_ConcatInputLayer):
     v = super(RecLayer, self).get_constraints_value()
     from TFUtil import optional_add
     if isinstance(self.cell, _SubnetworkRecCell):
-      for layer in self.cell.net.layers.values():
+      layers = list(self.cell.net.layers.values())
+      if self.cell.input_layers_net:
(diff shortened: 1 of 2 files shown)


albertz:
Good catch. So I guess you tried and wondered that it did not seem to make a difference? So, with this fix, does it actually help? You tried L2?

ChristopherBrix:
I wanted to try applying the L2 norm to my 2D-LSTM. Because I wasn't sure whether I had to implement some custom logic, I checked what other layers did, and noticed that the `get_constraints_value()` method of the 2D-LSTM layer never got called.

I've scheduled some experiments using L2 norm, but I don't have any results yet.

albertz:
Ok cool. Btw, please finish up the pull request on that... :)

PR #100, opened 2018-12-19
DotLayer: pass on beam size

patrick-wilken:


albertz:
The `assert` fails in some tests (see e.g. [here](https://travis-ci.org/rwth-i6/returnn/jobs/470007260)). Please check what went wrong there.
```
    line: assert a_out.beam_size == b_out.beam_size
    locals:
      a_out = <local> Data(name='enc_ctx_output', shape=(None, 2, 2), beam_size=3)
      a_out.beam_size = <local> 3
      b_out = <local> Data(name='att_query_output', shape=(None, 2, 2), time_dim_axis=None)
      b_out.beam_size = <local> None
```

patrick-wilken:
Ah, in the test we take the dot product of something inside and outside of the recurrent unit. So one has a beam and the other one doesn't. I'm sure somewhere this is handled by broadcasting or so, but I did not understand how and where this is done exactly.
For the updated commit I just assumed that it works and allowed for a mismatch in beam_size if one is None.

albertz:
Actually I'm not totally sure if the broadcast is done correctly. Because as you see from the output, the encoder layer actually has the `beam_size` set/broadcasted, but not the layer inside the loop.
In any case, I'm not sure if the `assert` as you put it now does make sense. Maybe better just remove it. I think this just causes confusion.

albertz:
Hm actually I don't know. I just merge it in for now.

Issue #102, opened 2018-12-20
Factored Neural Machine Translation


patrick-wilken:
I am in the process of implementing (Target-)[Factored Neural Machine Translation](https://arxiv.org/abs/1609.04621). This is what I've got so far: https://github.com/patrick-wilken/returnn-examples/blob/master/factored-nmt

I have problems creating a config that works for both training and search. I have to comment line 126 of https://github.com/patrick-wilken/returnn-examples/blob/master/factored-nmt/config to make it work in training (and uncomment it to make it work in search).

This has to do with the ChoiceLayer in line 124, which provides the true label In training instead of being evaluated. The true label seems to have a time axis, so a shape of `()` is invalid (as it should include the time axis).
In search however, the choice layer is evaluated in each step, so its output has no time dimension. Therefore `"shape": ()` should be correct. But by default it is set to `(None,)`, which causes the following error: https://gist.github.com/patrick-wilken/5eba7a3d209f7a1e470b6082fef2dd6d. Looking at `get_out_data_from_opts` of ChoiceLayer it looks like the time dimension is always included, as we're just getting the shape from data. Is this correct? What would be the proper way to fix it?

patrick-wilken:
Also, it would be nice to discuss, whether parts of the config should become dedicated layers in the end instead of using that many EvalLayers. As you see I'm using an outer product (DotLayer) + reshapes (MergeDimsLayer) to combine the two softmaxes. After that I use div and mod to get back the indices of the different factors. Beam pruning after the two softmaxes is still missing.

albertz:
Setting `shape` explicitly is problematic for the reasons you describe.
But not setting `shape` should work I think in both cases, or maybe there is a bug somewhere. Not setting the shape should then by default take the shape from the first source layer (i.e. `output` here). And this is the right thing, right? I would debug where that goes wrong. Maybe you can create a reduced test case, and then fix it.

About creating dedicated layers: Not sure if that really makes it simpler. Maybe you safe one line in your config, but then imagine someone else reading your config who is not familiar with your very special new layer. He would not understand what the config is doing without also looking at the code of your special layer. That is why I try to only have very generic layers which are more or less clear/straightforward about what they do. If you want to discuss further, maybe independent from the GitHub issue here. I talked with Julian today and we plan to have a meeting together early next year.

patrick-wilken:
I was able to dig deeper into this:
[...]

PR #103, opened 2018-12-20
Support for multiple output layers

patrick-wilken:
This is needed to output the different factors in Factored NMT (also see #102).
I'm not familiar with all the details of search and if multiple output layers can lead to problems in some cases, but in [my case](https://github.com/patrick-wilken/returnn-examples/tree/master/factored-nmt)  it works and I'm pretty sure it does not break the one-output-layer case.

### DIFF
--- a/TFEngine.py
+++ b/TFEngine.py
@@ -1816,11 +1816,11 @@ class Engine(object):
       sys.exit(1)
     return analyzer
 
-  def search(self, dataset, do_eval=True, output_layer_name="output", output_file=None, output_file_format="txt"):
+  def search(self, dataset, do_eval=True, output_layer_names="output", output_file=None, output_file_format="txt"):
     """
     :param Dataset.Dataset dataset:
     :param bool do_eval: calculate errors. can only be done if we have the reference target
-    :param str output_layer_name:
+    :param str|list[str] output_layer_names:
     :param str output_file:
     :param str output_file_format: "txt" or "py"
     """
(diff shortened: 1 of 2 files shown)


TFEngine.py
```diff
@@ -1857,86 +1857,118 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
+                  dataset.serialize_data(key=target_keys[t], data=outputs[t][out_idx + b]),
+                  file=log.v4)
+
         if out_cache is not None:
           corpus_seq_idx = dataset.get_corpus_seq_idx(seq_idx[i])
           assert corpus_seq_idx not in out_cache
           seq_idx_to_tag[corpus_seq_idx] = seq_tag[i]
           if out_beam_size is None:
-            out_cache[corpus_seq_idx] = dataset.serialize_data(key=target_key, data=output[out_idx])
+            out_cache[corpus_seq_idx] = '\t'.join(dataset.serialize_data(key=target_keys[t],
```
patrick-wilken: I simply used tabs to separate the outputs. Maybe different files would be better?

albertz: Yes, no tabs.
But I think separate output files might make it more complicated than necessary. Why not just a dict? (But only for the case `isinstance(output_layer_names, list)`.)

TFEngine.py
```diff
@@ -1950,7 +1991,13 @@ def extra_fetches_callback(seq_idx, seq_tag, output, targets=None, beam_scores=N
       assert len(out_cache) - 1 in out_cache
       if output_file_format == "txt":
         for i in range(len(out_cache)):
-          output_file.write("%s\n" % out_cache[i])
+          if isinstance(out_cache[i], dict):
+            # manually construct dict string to guarantee deterministic order
+            out_string = "{" + ', '.join("%r: %s" % (output_layer_names[t], out_cache[i][output_layer_names[t]]) \
+                for t in range(num_serializable_targets)) + "}\n"
```
albertz: There is `Util.betterRepr` for this purpose.
Also then you don't need the comment (as it becomes clear when you use `Util.betterRepr`).

albertz: However, I think it makes more sense to just don't allow txt format output for this case anyway. That makes it just more complicated to parse it later. As said: Don't make it more complicated than necessary.

albertz: Note that you should use `Util.betterRepr` for the py format case, though.

TFEngine.py
```diff
@@ -1857,86 +1857,127 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
+            if out_cache is not None:
+              if out_beam_sizes[t] is None:
+                  out_data = dataset.serialize_data(key=target_keys[t], data=outputs[t][out_idx])
+              else:
+                assert beam_scores[t] is not None
+                out_data = [
+                    (beam_scores[t][i][b], dataset.serialize_data(key=target_keys[t], data=outputs[t][out_idx + b]))
+                    for b in range(out_beam_sizes[t])]
+
+              if num_serializable_targets == 1:
```
albertz: The check should be `isinstance(output_layer_names, dict)`, and not `num_serializable_targets == 1`.

TFEngine.py
```diff
@@ -1857,86 +1857,127 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
-                "  hyp %i, score %f:" % (b, beam_scores[i][b]),
-                dataset.serialize_data(key=target_key, data=output[out_idx + b]),
-                file=log.v4)
         if out_cache is not None:
           corpus_seq_idx = dataset.get_corpus_seq_idx(seq_idx[i])
           assert corpus_seq_idx not in out_cache
           seq_idx_to_tag[corpus_seq_idx] = seq_tag[i]
-          if out_beam_size is None:
-            out_cache[corpus_seq_idx] = dataset.serialize_data(key=target_key, data=output[out_idx])
+          if num_serializable_targets > 1:
```
albertz: The check should be `isinstance(output_layer_names, dict)`, and not `num_serializable_targets == 1`.

TFEngine.py
```diff
@@ -1857,86 +1857,127 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
+        output_layer_beam_scores.append(output_layer.get_search_choices().beam_scores)
+      out_beam_sizes.append(output_layer.output.beam_size)
+      target_keys.append(output_layer.target or self.network.extern_data.default_target)
 
     out_cache = None
     seq_idx_to_tag = {}
     if output_file:
       assert output_file_format in {"txt", "py"}
-      assert dataset.can_serialize_data(target_key)
+      num_serializable_targets = sum(dataset.can_serialize_data(target_key) for target_key in target_keys)
```
albertz: Why so complicated? Just assert that all can be serialized.

TFEngine.py
```diff
@@ -1857,86 +1857,127 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
-      print("Given output %r is after decision (no beam)." % output_layer, file=log.v1)
-    else:
-      print("Given output %r has beam size %i." % (output_layer, out_beam_size), file=log.v1)
-      output_layer_beam_scores = output_layer.get_search_choices().beam_scores
-    target_key = output_layer.target or self.network.extern_data.default_target
+    if not isinstance(output_layer_names, list):
+      output_layer_names = [output_layer_names]
+    num_targets = len(output_layer_names)
+
+    output_layers, out_beam_sizes, output_layer_beam_scores, target_keys = [], [], [], []
```
albertz: There should be a comment about the size of the lists.

TFEngine.py
```diff
@@ -1857,86 +1857,127 @@ def search(self, dataset, do_eval=True, output_layer_name="output", output_file=
...
     if output_file:
       assert output_file_format in {"txt", "py"}
-      assert dataset.can_serialize_data(target_key)
+      num_serializable_targets = sum(dataset.can_serialize_data(target_key) for target_key in target_keys)
+      assert num_serializable_targets > 0
       assert not os.path.exists(output_file)
       print("Will write outputs to: %s" % output_file, file=log.v2)
       output_file = open(output_file, "w")
-      out_cache = {}  # corpus-seq-idx -> str|list[(float,str)]
+      out_cache = {}  # corpus-seq-idx -> str|list[(float,str)]|dict[str -> str|list[(float,str)]] depending on number of output layers and beam sizes
```
albertz: It should not depend on number of output layers.

albertz: Also not on beam sizes. It actually depends if the output is after a choice (i.e. single best) or before (the whole beam).

TFEngine.py
patrick-wilken: Before, it was possible to do search with a non-serializable target when no output file was given. That just prints the output indices to the console (if verbosity >=4). I wanted to keep that behaviour even if other targets are written to the output file. But I agree, it is not super useful, so I can remove it.

albertz: But where is the problem? That would still be the case (this is inside the `if output_file:` branch). Just do `assert all([dataset.can_serialize_data(target_key) for target_key in target_keys])`.

patrick-wilken: Yes, I will. I just mean, then it is not possible to have a console output for a non-serializable feature A while writing feature B to file.

TFEngine.py
patrick-wilken: list you mean? But I convert output_layer_names to list if it is a string, should I change that?

albertz: Sorry, yes, I meant list. You can also do `want_output_list = isinstance(output_layer_names, list)` in the beginning. Or do not convert `output_layer_names`. Whatever makes the code simpler.

TFEngine.py
patrick-wilken: Oh, forgot that one. Do you mean just a comment saying all lists will be of length num_targets? Or do you mean the data type?

albertz: The data types would probably be clearer (it's not totally obvious in all the cases), but also the length. So maybe make each assign an individual line also.

patrick-wilken:
Ok, I think I have addressed all the requests now.

PR #104, opened 2019-01-03
RecLayer template construction: allow loop exception if get_layer safe

patrick-wilken:
This fixes the invalid shape issue in #102. However, I'm not 100% sure if this is what we want...

The goal of this change is that we can infer the output data shape of layers in the recurrent subnetwork that depend on the "output" layer (usually a target embedding for label feedback). What happens at the moment is, that construction starts at output, recursively we go backwards through the recurrent cell until we end at output again and cannot construct it, because a loop exception is thrown. The first layer that is constructed is then the one that directly depends on the output, which (in most cases, I guess) can only be constructed uninitialized, because the input ("output") is not constructed yet. Therefore the shape is set to the default (None,). For target embeddings, more or less by chance, this works out, but for other layers it leads to shape mismatches.

By not throwing a loop exception for the case of get_layer.safe, we allow constructing the output layer when visiting it again, while at the same time avoiding an infinite loop. 

I am not sure if get_layer.once has to be included. What is it actually doing?

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -380,7 +380,7 @@ class TFNetwork(object):
         return self.get_layer(name)
       except LayerNotFound:
         pass  # ok, we will try to construct it then
-    if name in self._constructing_layers:
+    if name in self._constructing_layers and not (get_layer and hasattr(get_layer, "safe") and get_layer.safe):
       raise NetworkConstructionDependencyLoopException(
         layer_name=name, constructing_layers=self._constructing_layers, net_dict=net_dict, network=self)
     if name not in net_dict:


patrick-wilken:
Ok, after failing checks I realize it's bad to expect get_layer to have an attribute "safe". Even testing for it would be ugly... But what about the concept? 😄

albertz:
I wonder that not more test cases have failed. :)

I'm also not quite sure I understand your idea. If it does not throw the exception, it will end up in a loop, or not? And how would it also know the output format now?

Have you tried my suggestion I made in the issue (#102)?

In principle, you have to come up with a solution which would not end up in such a loop, i.e. which can determine the output format without fully recursing.

patrick-wilken:
Just to prove the point, the tests work with the ugly version (the single one that fails worked before and locally, it's just numerics, I assume).

No, there is no infinite loop, because I suppress the exception in exactly the case where the recursion will stop anyway (safe==True). The initial output format (from which others are then derived) is now taken from the output layer, which is a choice layer, which gets it directly from the target (and this is the only place, where we can the information from...). I think it makes sense to expect from the output layer that it knows the output format. Certainly better than it is now, where the shape of the target embedding is effectively hard coded 😊

albertz:
Well, yes, with the additional `hasattr`.
With infinite loop, I mean an infinite recursion. Because that is what is being checked.

albertz:
I don't remember, what was the state here? Do we still need something here? Or is this obsolete now?

patrick-wilken:
Yes, I was going to comment, that this is a too specific solution to a problem I don't have anymore. Maybe we can improve template generation in the future, but this can be closed for now.

PR #106, opened 2019-01-06
a minor bug related to TFUtil was fixed.

sephiroce:
Command: ${returnn-experiments}/2018-asr-attention/librispeech/full-setup-attention/22_train.sh
I've got this log message.

" default_name type (<class 'list'>) is not a string type. You likely meant to pass this into the values kwarg."

I simply modified "returnn/TFUtil.py".

### DIFF
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -4773,7 +4773,7 @@ def smoothing_cross_entropy(logits,
   :return: Tensor of the same shape as `labels` and of the same dtype as `logits`.
   :rtype: tf.Tensor
   """
-  with tf.name_scope("smoothing_cross_entropy", [logits, labels]):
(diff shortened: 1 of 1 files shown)


sephiroce:
I installed TF by building source.
The exact version I have used is the git commit below.

```
> git log
commit 8e4ca33c8b5861526bcdc43eb1af6d73e1670e92 (HEAD -> master, origin/master, origin/HEAD)
Author: A. Unique TensorFlower <gardener@tensorflow.org>
Date:   Fri Jan 4 04:10:23 2019 -0800

    Correctly handle S32 and U32 in the AR-CRS combiner
    
    Some of the transforms we are doing are only valid on floating point types
    so we have to condition them on the element type of the operations.
    
    PiperOrigin-RevId: 227836047

...
```

PR #109, opened 2019-01-15
SubnetworkRecCell: make all extern data available

patrick-wilken:
This addresses 
`# TODO check subnet, which extern data keys are used...`
in _SubnetworkRecCell.get_output().

Now, the placeholders for all extern data used in the subnetwork are set. This makes it possible to use all extern data in the layers inside the recurrent unit.

I changed the "x", "y" notation, as we now just loop over the used data keys without distinguishing between the true "y" and additional data.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -933,8 +933,7 @@ class _SubnetworkRecCell(object):
       pprint(self.layer_data_templates)
       raise
 
-  def _construct(self, prev_outputs, prev_extra, i,
-                 data=None, classes=None,
+  def _construct(self, prev_outputs, prev_extra, i, data=None,
                  inputs_moved_out_tas=None, needed_outputs=("output",)):
     """
     This is called from within the `tf.while_loop` of the :class:`RecLayer`,
@@ -943,8 +942,9 @@ class _SubnetworkRecCell(object):
     :param dict[str,tf.Tensor] prev_outputs: outputs of the layers from the previous step
     :param dict[str,dict[str,tf.Tensor]] prev_extra: extra output / hidden states of the previous step for layers
     :param tf.Tensor i: loop counter. scalar, int32, current step (time)
-    :param tf.Tensor|None data: optional source data, shape e.g. (batch,dim)
-    :param tf.Tensor|None classes: optional target classes, shape e.g. (batch,) if it is sparse
+    :param dict[str,tf.Tensor] data: All data needed from outside of the loop. Possible keys are 'source'
+        (for the input of the recurrent layer) and the keys in parent_net.extern_data, which notably include
+        the target of the recurrent layer, usually called 'classes'.
     :param dict[str,tf.TensorArray]|None inputs_moved_out_tas:
     :param set[str] needed_outputs: layers where we need outputs
     """
@@ -952,21 +952,15 @@ class _SubnetworkRecCell(object):
     from TFNetworkLayer import InternalLayer, ExtendWithBeamLayer
     from TFUtil import tile_transposed
     needed_beam_size = self.layer_data_templates["output"].output.beam_size
-    if data is not None:
+    for key in data:
       if needed_beam_size:
-        assert not self.parent_rec_layer.input_data.beam_size
-        data = tile_transposed(
-          data,
-          axis=self.net.extern_data.data["source"].batch_dim_axis,
+        if key == "source":
+          assert not self.parent_rec_layer.input_data.beam_size
+        data[key] = tile_transposed(
+          data[key],
+          axis=self.net.extern_data.data[key].batch_dim_axis,
           multiples=needed_beam_size)
-      self.net.extern_data.data["source"].placeholder = data
-    if classes is not None:
-      if needed_beam_size:
-        classes = tile_transposed(
-          classes,
-          axis=self.net.extern_data.data[self.parent_rec_layer.target].batch_dim_axis,
-          multiples=needed_beam_size)
-      self.net.extern_data.data[self.parent_rec_layer.target].placeholder = classes
+      self.net.extern_data.data[key].placeholder = data[key]
     for data_key, data in self.net.extern_data.data.items():
       if data_key not in self.net.used_data_keys:
         continue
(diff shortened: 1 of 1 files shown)


TFNetworkRecLayer.py
```diff
@@ -1595,7 +1596,7 @@ def maybe_transform(layer):
           for k in self._initial_extra_outputs:
             args.update({"%s.extra.%s" % (k, k2): v for (k2, v) in self.net.layers[k].rec_vars_outputs.items()})
             args.update({"prev:%s.extra.%s" % (k, k2): v for (k2, v) in prev_extra[k].items()})
-          res = (identity_with_debug_log(out=self._debug_out, x=res[0], args=args),) + res[1:]
+          res = (identity_with_debug_log(out=self._debug_out, source=res[0], args=args),) + res[1:]
```
albertz: identity_with_debug_log has the arg name `x`, not `source`.

TFNetworkRecLayer.py
```diff
@@ -943,30 +942,25 @@ def _construct(self, prev_outputs, prev_extra, i,
     :param dict[str,tf.Tensor] prev_outputs: outputs of the layers from the previous step
     :param dict[str,dict[str,tf.Tensor]] prev_extra: extra output / hidden states of the previous step for layers
     :param tf.Tensor i: loop counter. scalar, int32, current step (time)
-    :param tf.Tensor|None data: optional source data, shape e.g. (batch,dim)
-    :param tf.Tensor|None classes: optional target classes, shape e.g. (batch,) if it is sparse
+    :param dict[str,tf.Tensor|None] data: All data needed from outside of the loop. Usually includes
+        'source' data (shape e.g. (batch,dim)) and target classes (shape e.g. (batch,) if it is sparse).
```
albertz: Target is `'target'`? Or `'classes'`? Be explicit.

albertz: Why do you allow `None` as value?

patrick-wilken:
Yes, this is supposed to fix something :) I need it for Factored NMT, where I have several targets (factor1, factor2 and also a combination of them).

TFNetworkRecLayer.py
patrick-wilken: Automatic renaming... :D

TFNetworkRecLayer.py
patrick-wilken: None was also allowed before, it means e.g. that the recurrent unit has no external input data. But I guess you mean I should not use None but check if the key is in the dict instead, right? 
For other data than "source" the key is the one used in `parent_net.extern_data` (see line 763), but I can add that in the description.

albertz: `None` was there because `data` has to have some value because it exists always (the parameter). But if it is a dict, then `data["data"]` does not have to exists. I don't understand what the point is to have `data["data"] is None`.

The new description is also not really clear. `classes` is supposed to be a key? Then write `'classes'` to make that clear.

TFNetworkRecLayer.py
```diff
@@ -1244,31 +1242,35 @@ def get_output(self, rec_layer):
...
                  " The length of the sources (", tf.reduce_max(input_seq_len),
-                 ") differ from the length of the target (", y_max_len, ")."])]):
-              y_max_len = tf.identity(y_max_len)
-        y_ta = tf.TensorArray(
-          name="y_ta",
-          dtype=y_data.dtype,
-          element_shape=tf.TensorShape(y_data.copy_template_excluding_time_dim().batch_shape),
-          size=y_max_len,
+                 ") differ from the length of the target (", data_max_len, ")."])]):
+              data_max_len = tf.identity(data_max_len)
```
albertz: What is the point of recreating `data_max_len` for every key (in `used_keys`)?

TFNetworkRecLayer.py
```diff
@@ -1244,31 +1242,35 @@ def get_output(self, rec_layer):
...
+          dtype=data.dtype,
+          element_shape=tf.TensorShape(data.copy_template_excluding_time_dim().batch_shape),
+          size=data_max_len,
           infer_shape=True)
-        y_ta = y_ta.unstack(y, name="y_ta_unstack")
+        data_ta = data_ta.unstack(data_placeholder, name="{}_ta_unstack".format(key))
+        data_tensor_arrays[key] = data_ta
         if max_seq_len is None:
-          max_seq_len = y_max_len
+          max_seq_len = data_max_len
```
albertz: This behavior will depend on the order in `used_data_keys`, which can be non-deterministic. This code should not be non-deterministic.

TFNetworkRecLayer.py
```diff
@@ -1525,12 +1527,12 @@ def body(i, net_vars, acc_tas, seq_len_info=None):
           for (k, v) in zip(sorted(self._initial_extra_outputs), prev_extra_flat)}
         with tf.name_scope("prev_extra"):
           prev_extra = identity_op_nested(prev_extra)
+        data = {key: ta.read(i, name="{}_ta_read".format(key)) if ta else None for key, ta in data_tensor_arrays.items()}
```
albertz: Why is it possible that `ta is None`?

TFNetworkRecLayer.py
```diff
@@ -943,30 +942,25 @@ def _construct(self, prev_outputs, prev_extra, i,
     :param dict[str,tf.Tensor] prev_outputs: outputs of the layers from the previous step
     :param dict[str,dict[str,tf.Tensor]] prev_extra: extra output / hidden states of the previous step for layers
     :param tf.Tensor i: loop counter. scalar, int32, current step (time)
-    :param tf.Tensor|None data: optional source data, shape e.g. (batch,dim)
-    :param tf.Tensor|None classes: optional target classes, shape e.g. (batch,) if it is sparse
+    :param dict[str,tf.Tensor] data: All data needed from outside of the loop. Possible keys are 'source'
+        (for the input of the recurrent layer) and the keys in parent_net.extern_data, which notably may include
+        the target of the recurrent layer.
```
patrick-wilken: This is the updated description, is it clear now?

patrick-wilken:
Ok, I no longer use None values.
I also assert that all extern data have the same number of time steps. Because of this, the order of used_keys makes no difference.

TFNetworkRecLayer.py
albertz: I would mention `'classes'` as an example also.

TFNetworkRecLayer.py
```diff
@@ -943,30 +942,25 @@ def _construct(self, prev_outputs, prev_extra, i,
     :param dict[str,tf.Tensor] prev_outputs: outputs of the layers from the previous step
     :param dict[str,dict[str,tf.Tensor]] prev_extra: extra output / hidden states of the previous step for layers
     :param tf.Tensor i: loop counter. scalar, int32, current step (time)
-    :param tf.Tensor|None data: optional source data, shape e.g. (batch,dim)
-    :param tf.Tensor|None classes: optional target classes, shape e.g. (batch,) if it is sparse
+    :param dict[str,tf.Tensor] data: All data needed from outside of the loop. Possible keys are 'source'
+        (for the input of the recurrent layer) and the keys in parent_net.extern_data, which notably include
+        the target of the recurrent layer, usually called 'classes'.
```
patrick-wilken: Like this?

TFNetworkRecLayer.py
```diff
@@ -1244,31 +1242,46 @@ def get_output(self, rec_layer):
...
+          data_len = tf.shape(data_placeholder)[0]
+          if common_data_len is None:
+            # Check for first key if input length matches data length
+            if input_seq_len is not None:
+              with tf.control_dependencies([tf.assert_equal(tf.reduce_max(input_seq_len), data_len,
+                  ["RecLayer %r with sources %r:" % (rec_layer.name, rec_layer.sources),
+                   " The length of the sources (", tf.reduce_max(input_seq_len),
+                   ") differ from the length of the target ", key, "(", data_len, ")."])]):
+                data_len = tf.identity(data_len)
+            common_data_len = data_len
```
albertz: The order of the loop is non-deterministic, so this behavior is (slightly) non-deterministic. This is not good. It should not be non-deterministic.

patrick-wilken: So just loop over sorted(used_keys)?
Or do you mean we should know from which key we initially set common_data_len? It doesn't matter, because checking equality is the whole point, but nevertheless I could make rec_layer.target the first key.

TFNetworkRecLayer.py
```diff
@@ -1244,31 +1242,46 @@ def get_output(self, rec_layer):
...
-        y_ta = tf.TensorArray(
-          name="y_ta",
-          dtype=y_data.dtype,
-          element_shape=tf.TensorShape(y_data.copy_template_excluding_time_dim().batch_shape),
-          size=y_max_len,
+      common_data_len = None  # used to check whether all extern data have same length
+      used_keys = self.net.used_data_keys.copy()
+      if rec_layer.target:
+        used_keys.add(rec_layer.target)  # we always need the target of the recurrent layer
+      for key in sorted(used_keys):
```
patrick-wilken: Ok, done. sorted should be enough, as the order really doesn't matter.

patrick-wilken:
I noticed that 1bd275a is in conflict with this commit, as we can now have multiple features on the time axis of the recurrent layer. When checking for same time axes, tags are compared, which contain the feature name. Thus we do not create a tensor array for some features, although we should.
What would be a good way to solve this? Make the dataset provide the information, which features lie on the same time axis? This would be quite some overhead.

albertz:
To clarify: Which data keys does your dataset provide? And some of them would share the same time dimension? I think this really should be specified by the dataset. However, maybe we can get away without doing that. There are maybe some cases maybe where we can automatically infer/assume that some data keys share the same time dim. But to be more concrete about this, can you maybe create a small test case which currently fails now? Create that as a pull request, and then we can discuss over there how to solve this.

PR #112, opened 2019-01-22
CombinedDataset: redesign to support all seq ordering options

patrick-wilken:
The current implementation of CombinedDataset lacks the ability to jointly sort all sequences using seq orders that depend on the length of the sequences ('sorted' and 'laplace'). The concrete use case where we need this is the dynamic combination of bilingual corpora (instead of concatenating the text files) for training multi-lingual MT systems.

I changed the calculation of the sequence order such that it uses the "standard" LayerBase.get_seq_order_for_epoch() and applies it to the set of all sequences. However, I kept the old behaviour as a special case, which is still the default.

At the same time I adopted reading data dims from data as in 5851c19 and did some renaming (if you insist, I can still make those separate commits). 

I also added a repeat_epoch option to Dataset, such that one can adjust the relative sizes of the datasets that are combined.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -55,7 +55,7 @@ class Dataset(object):
 
   def __init__(self, name=None,
                window=1, context_window=None, chunking=None,
-               seq_ordering='default', partition_epoch=None,
+               seq_ordering='default', partition_epoch=None, repeat_epoch=None,
                shuffle_frames_of_nseqs=0, min_chunk_size=0,
                estimated_num_seqs=None,):
     """
@@ -67,6 +67,9 @@ class Dataset(object):
     :param str seq_ordering: "batching"-option in config. e.g. "default", "sorted" or "random".
       See self.get_seq_order_for_epoch() for more details.
     :param int|None partition_epoch:
+    :param int|None repeat_epoch: Repeat the sequences in an epoch this many times. Useful to scale the dataset
+      relative to other datasets, e.g. when used in CombinedDataset. Not allowed to be used in combination with
+      partition_epoch.
     :param int shuffle_frames_of_nseqs: shuffles the frames. not always supported
     :param None|int estimated_num_seqs: for progress reporting in case the real num_seqs is unknown
     """
@@ -77,6 +80,10 @@ class Dataset(object):
     self.window = window
     self.seq_ordering = seq_ordering  # "default", "sorted" or "random". See self.get_seq_order_for_epoch().
     self.partition_epoch = partition_epoch or 1
+    self.repeat_epoch = repeat_epoch or 1
+    # There is probably no use case for combining the two, so avoid potential misconfiguration.
+    assert self.partition_epoch == 1 or self.repeat_epoch == 1, \
+        "Combining partition_epoch and repeat_epoch is prohibited."
     self.timestamps = None
     self.labels = {}; """ :type: dict[str,list[str]] """
     self.weights = {}
@@ -254,6 +261,7 @@ class Dataset(object):
     :rtype: list[int]
     """
     partition_epoch = self.partition_epoch or 1
+    repeat_epoch = self.repeat_epoch or 1
     if not epoch:
       epoch = 1
     full_epoch = epoch
@@ -314,6 +322,8 @@ class Dataset(object):
       assert len(partitions) == partition_epoch + 1
       seq_index = seq_index[partitions[current_partition]:partitions[current_partition + 1]]
       assert len(seq_index) == partition_sizes[current_partition]
+    if repeat_epoch > 1:
+      seq_index = seq_index * repeat_epoch
     return seq_index
 
   def init_seq_order(self, epoch=None, seq_list=None):
--- a/Dataset.py
+++ b/Dataset.py
@@ -315,17 +315,31 @@ class Dataset(object):
     else:
       assert False, "invalid batching specified: " + self.seq_ordering
     if partition_epoch > 1:
-      current_partition = ((epoch or 1) - 1) % partition_epoch
-      seqs_per_epoch = num_seqs // partition_epoch
-      partition_sizes = ([seqs_per_epoch + 1] * (num_seqs % partition_epoch) +
-                         [seqs_per_epoch] * (partition_epoch - num_seqs % partition_epoch))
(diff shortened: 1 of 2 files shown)


Dataset.py
```diff
@@ -267,6 +267,8 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
     seq_index = list(range(num_seqs)); """ :type: list[int]. the real seq idx after sorting """
     if self.seq_ordering == 'default':
       pass  # Keep order as-is.
+    elif self.seq_ordering == 'reverse':
+      seq_index = list(reversed(seq_index))
```
patrick-wilken: Just because this is possible with the current CombinedDataset. I don't actually need this ;)

Dataset.py
```diff
@@ -265,7 +265,7 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
       full_epoch = (epoch - 1) // partition_epoch + 1
     assert num_seqs > 0
     seq_index = list(range(num_seqs)); """ :type: list[int]. the real seq idx after sorting """
-    if self.seq_ordering == 'default':
+    if self.seq_ordering in ['default', 'in-order']:
```
patrick-wilken: To be able to still do 'in-order' if default is something else.

MetaDataset.py
```diff
@@ -491,55 +491,57 @@ class CombinedDataset(CachedDataset2):
...
     """
     assert window == 1  # not implemented
     super(CombinedDataset, self).__init__(**kwargs)
     assert self.shuffle_frames_of_nseqs == 0  # not implemented. anyway only for non-recurrent nets
-    assert self.partition_epoch == 1, "partition_epoch not supported for CombinedDataset, apply it to the sub-datasets."
 
     self.rnd = Random(self.epoch)
-    self.dataset_keys = set(datasets.keys()); ":type: set[str]"
-    self.dataset_idxs = dict(enumerate(sorted(self.dataset_keys)))  # idx -> dataset-key
+    self.dataset_keys = set([m[0] for m in data_map.keys()]); ":type: set[str]"
```
patrick-wilken: To only load needed datasets. Copied from MetaDataset. This was wrong before.

MetaDataset.py
```diff
@@ -550,75 +552,147 @@ def __init__(self,
...
+    return dataset.get_seq_length(dataset_seq_idx)["data"]
+
   def _expand_dataset_sec_idxs(self, num_values):
     """
     :param num_values: int Add num_values entries to the dataset-segment-idx mapping table
     :return:
     """
     for i in range(num_values):
-      if self.seq_ordering in ("default", "random"):  # default is random. this is different from base class!
+      if self.seq_ordering == "default":
```
patrick-wilken: "default" and "random" are now different things. "random" shuffles all sequences, while "default" keeps the order of the sub-datasets and just randomly picks the dataset.

patrick-wilken: Maybe we should give it another explicit name, like "random_dataset".

Dataset.py
```diff
@@ -305,15 +311,23 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
...
-      assert len(partitions) == partition_epoch + 1
-      seq_index = seq_index[partitions[current_partition]:partitions[current_partition + 1]]
-      assert len(seq_index) == partition_sizes[current_partition]
+      seq_index = self._apply_partition_epoch(seq_index, partition_epoch, num_seqs, epoch)
+    if repeat_epoch > 1:
+      seq_index = seq_index * repeat_epoch
+    return seq_index
+
+  @classmethod
+  def _apply_partition_epoch(cls, seq_index, partition_epoch, num_seqs, epoch):
```
albertz: The parameters (esp their type) should be documented.

Dataset.py
albertz: I don't have a big problem with this.
However, instead of commenting this here, maybe you should write the comment into the source code.
Also, I wonder: Would it make sense that some datasets have another default? Wouldn't that be confusing?

Dataset.py
```diff
@@ -67,6 +67,8 @@ def __init__(self, name=None,
     :param str seq_ordering: "batching"-option in config. e.g. "default", "sorted" or "random".
       See self.get_seq_order_for_epoch() for more details.
     :param int|None partition_epoch:
+    :param int|None repeat_epoch: Repeat the sequences in an epoch this many times. Useful to scale the dataset
+      relative to other datasets, e.g. when used in CombinedDataset.
```
albertz: How does it work together with `partition_epoch`? Would it repeat that partition? (Extend the documentation to cover this.) Or I guess it doesn't really make much sense to use both options together.
Another thing: Maybe allowing some fraction here, some float, also would make sense? But if you don't need that now, and it complicates other code, then leave it away for now.

MetaDataset.py
albertz: Yes. "default" should behave as it behaves in all other datasets. It would be confusing otherwise, I think.
This case, "random_dataset", do you really need that? Do you think this is useful? I'm not sure. Don't add this if you don't need it now.

Dataset.py
patrick-wilken: It is applied after partition_epoch, so yes, you would have a part of the epoch repeated. There is not really a use case for this, I agree. Should I assert that only one is used?

Yes, I even thought, maybe you should be able to set the relative sizes of the datasets in the CombinedDataset. But this requires more code than just using partition and repeat epoch in the sub-datasets. I guess I will wait and see if this is needed in practice.

MetaDataset.py
patrick-wilken: Ok, I just tried to keep the old behaviour. So then I make 'in-order' the default, i.e. to go through all sequences of dataset1, then dataset2 and so on.
I didn't add 'random_dataset', this is the only case that was there before ;) And it is useful, for example if one dataset has much longer sequence lengths than the other (i.e. text vs. audio). Then a joint 'laplace' sorting for training makes no sense, but you can still sort each dataset and then pick datasets randomly.

Dataset.py
patrick-wilken: This isn't needed if I make 'in-order' the default, so I will remove this.

patrick-wilken:
The updated commits should address all the points.

PR #113, opened 2019-01-24
ChoiceLayer: multiple targets (factors)

patrick-wilken:
This is needed for [Factored NMT](https://arxiv.org/abs/1609.04621).
ChoiceLayer now accepts two sources and two corresponding targets (e.g. target factors) and does a joint decision for both. Criterion is the sum of the two incoming scores in log space. Beam pruning is applied to both sources to reduce the number of possible combinations.

As the layer has several outputs (labels for both targets), I used sub-layers to refer to them. For this I also needed sub-layer templates.
An example config can be found here: https://github.com/patrick-wilken/returnn-examples/blob/master/factored-nmt/config

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -387,6 +387,7 @@ class TFNetwork(object):
       def get_layer(src_name):
         return self.construct_layer(net_dict=net_dict, name=src_name)  # set get_layer to wrap construct_layer
     if name not in net_dict:
+      layer_desc = None
       if name == "data":
         layer_desc = {"class": "source", "from": []}
       elif name.startswith("data:"):
@@ -394,12 +395,11 @@ class TFNetwork(object):
       elif '/' in name:
         # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
         # but maybe it's not constructed yet, so try constructing the root layer
-        get_layer(name.split('/')[0])
-        assert name.split('/')[0] in self.layers, ("Root %r layer was not constructed! "
-                                                   "Unable to get layer %r." % (name.split('/')[0], name))
-        # constructing the root layer should have constructed all its children
-        return self.get_layer(name)  # ...so try again now
-      else:
+        root_layer = get_layer(name.split('/')[0])
+        sub_layer = root_layer.get_sub_layer('/'.join(name.split('/')[1:]))  # get the sub-layer from the root-layer
+        if sub_layer:
+          return sub_layer
+      if not layer_desc:
         raise LayerNotFound("layer %r not found in %r" % (name, self))
     else:
       layer_desc = net_dict[name]
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -479,9 +479,24 @@ class LayerBase(object):
   def get_sub_layer(self, layer_name):
     """
     The default behavior for any layer is to return None.
-    :param str layer_name:  The sub_layer addressed by '/' separated path.
-    :return: The sub_layer addressed in layer_name or None if no sub_layer exists
+
+    :param str layer_name: name of the sub_layer (right part of '/' separated path)
+    :return: the sub_layer addressed in layer_name or None if no sub_layer exists
     :rtype: LayerBase|None
+    """
+    return None
+
+  @classmethod
+  def get_sub_layer_out_data_from_opts(cls, layer_name, parent_layer_kwargs):
+    """
+    Called by _TemplateLayer.get_sub_layer(). Gets a Data template for the sub-layer with name 'layer_name'.
+    Also returns the network the sub-layer is in and the class type of the sub-layer. There is no good default
+    behaviour here, as this heavily depends on how the current layer uses sub-layers.
+
+    :param str layer_name: name of the sub_layer (right part of '/' separated path)
+    :param dict[str] parent_layer_kwargs: kwargs for the parent layer (as kwargs in cls.get_out_data_from_opts())
+    :return: Data template, network and the class type of the sub-layer
+    :rtype: (Data, TFNetwork, type)|None
     """
     return None
 
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -366,16 +366,23 @@ class LayerBase(object):
       d["reuse_params"] = ReuseParams.from_config_dict(d["reuse_params"], network=network, get_layer=get_layer)
     if d.get("loss", None) and "target" not in d:
       d["target"] = network.extern_data.default_target
+    targets = None
     if d.get("target"):
+      targets = d["target"]
+      # we might have multiple targets, e.g. in choice layer, so convert to list
+      if isinstance(targets, str):
+          targets = [targets]
       if network.eval_flag:
-        # Not resolving this in the dict, but call get_layer to make it available.
-        assert isinstance(d["target"], str)
-        if d["target"].startswith("layer:"):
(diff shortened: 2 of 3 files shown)


TFNetwork.py
```diff
@@ -387,19 +387,19 @@ def construct_layer(self, net_dict, name, get_layer=None, add_layer=None, check_
...
       elif name.startswith("data:"):
         layer_desc = {"class": "source", "data_key": name[len("data:"):], "from": []}
       elif '/' in name:
         # it may be a hierarchical path to a sub-layer, which should have been found by get_layer()
         # but maybe it's not constructed yet, so try constructing the root layer
-        get_layer(name.split('/')[0])
-        assert name.split('/')[0] in self.layers, ("Root %r layer was not constructed! "
-                                                   "Unable to get layer %r." % (name.split('/')[0], name))
-        # constructing the root layer should have constructed all its children
-        return self.get_layer(name)  # ...so try again now
```
patrick-wilken: self.get_layer() does not work during template construction.

TFNetworkRecLayer.py
```diff
@@ -862,6 +862,16 @@ def __call__(lself, name):
...
+        if '/' in name:
+          # this is probably a path to a sub-layer
+          root_name = name.split('/')[0]
+          root_layer = lself.__call__(root_name)  # get the root-layer (first part of the path)
+          sub_layer = root_layer.get_sub_layer('/'.join(name.split('/')[1:]))  # get the sub-layer from the root-layer
+          if sub_layer:  # get_sub_layer returns None by default (if sub-layer not found)
+            # add to templates so we will collect output in self.get_output if this is an output layer
+            self.layer_data_templates[name] = sub_layer
+            sub_layer.dependencies.add(root_layer)
+            return sub_layer
```
albertz: The logic for sub layers inside the rec layer template construction, do you really need that? If not, I would vote for removing this, to keep the code simpler for now. But at least it should be a separate pull request.

TFNetworkRecLayer.py
```diff
@@ -2260,6 +2272,25 @@ def init(self, output, layer_class, template_type="template", **kwargs):
       self.search_choices = SearchChoices(owner=self, beam_size=self._get_search_choices_beam_size())
     self.collocate_with = kwargs.get("collocate_with", None) or []
 
+  def get_sub_layer(self, layer_name):
+    """
+    Creates a template sub-layer using self.layer_class_type.get_sub_layer_out_data_from_opts().
+    """
```
albertz: Documentation is missing (layer_name).

TFNetworkRecLayer.py
```diff
@@ -3118,24 +3268,58 @@ def transform_config_dict(cls, d, network, get_layer):
...
-    return out
+      out_data.beam_size = beam_size
+
+    return out_data
+
+  def get_sub_layer(self, layer_name):
+    """
+    Used to get outputs in case of multiple targets. For all targets we create a sublayer that can be referred to
+    by "self.name + '/out_' + index" (e.g. output/out_0). These sublayers can then be used as input to other layers,
+    e.g. "output_0": {"class": "copy", "from": ["output/out_0"].
```
albertz: Documentation missing (layer_name).

TFNetworkRecLayer.py
albertz: Ah I see, this is needed for the ChoiceLayer which provides the different sub layers.

TFNetworkRecLayer.py
```diff
@@ -3106,10 +3168,98 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+  def _prune_and_combine_sources(self, sources, beam_sizes, batch_dim):
+    """
+    Applies beam bruning to the sources and then calculates all possible sums of scores.
+    Returns the scores, the (static) number of targets after pruning and a list of
+    labels corresponding to the top scores.
+    :param list[LayerBase] sources: input layers providing the scores
+    :param list[int] beam_sizes: beam sizes used for pruning of the individual sources
+    :param tf.Tensor|int batch_dim: dim of batch axis (batch size * incoming beam)
+    :rtype: (tf.Tensor, int, list[tf.Tensor])
+    """
```
albertz: Documentation missing (`:return: ...`). Just reformat your text "Returns the scores ...".

albertz: Btw, not sure about this, but I think readthedocs/Sphinx has problems when there is no empty line between your description and the first parameter (`:param ...`). Maybe check that and if needed, add an empty line in between.

TFNetworkRecLayer.py
```diff
@@ -3106,10 +3168,98 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+  def _get_combined_labels(self, pruned_labels, beam_sizes, combined_ids):
+    """
+    Gets output labels by converting 'combined_ids' (corresponding to the flattend shape created in
+    self._prune_and_combine_sources()) back to separate ids and then using those as indices to the labels
+    that survived pruning.
+    :param list[tf.Tensor] pruned_labels: labels before pruning, see self._prune_and_combine_sources()
+    :param list[int] beam_sizes: beam sizes used for pruning of the individual sources
+    :param tf.Tensor combined_ids: indices to the flattened scores, see self._prune_and_combine_sources()
+    :rtype: list[tf.Tensor]
+    """
```
albertz: `:return:` missing.

TFNetworkLayer.py
```diff
@@ -485,6 +492,16 @@ def get_sub_layer(self, layer_name):
     """
     return None
 
+  @classmethod
+  def get_sub_layer_out_data_from_opts(cls, layer_name, parent_layer_kwargs):
+    """
+    Called by _TemplateLayer.get_sub_layer(). Gets a Data template for the sub-layer with name 'name'. Also returns
+    the network the sub-layer is in and the class type of the sub-layer. There is no good default behaviour here,
+    as this heavily depends on how the current layer uses sub-layers.
+    :rtype: (Data, TFNetwork, type)|None
```
albertz: Documentation (`:return:`, layer_name, parent_layer_kwargs) missing.

TFNetworkRecLayer.py
```diff
@@ -3149,24 +3268,58 @@ def transform_config_dict(cls, d, network, get_layer):
...
     """
-    if not network.search_flag and not d.get("scheduled_sampling"):
-      # In the dependency graph, we don't want it.
-      # This can enable some optimizations in the RecLayer.
-      # We do it here because we should know about the deps early in the template creation in RecLayer.
-      d["from"] = []
+    if isinstance(d["target"], str):
+      d["target"] = [d["target"]]
+    if not network.search_flag:
+      network.used_data_keys.update(d["target"])
```
patrick-wilken: This is a bit hacky. We need to find out during template creation which targets are used in the sub-net. I'm not sure where it is best to set this. We could also just set mark_data_key_as_used=True in get_out_data_from_opts(). Because there is no place in the code where we call get_out_data_from_opts() and then end up not using the target, I think. Or maybe make mark_data_key_as_used a parameter of get_out_data_from_opts()...

patrick-wilken:
I added all the missing documentation.

TFNetworkLayer.py
albertz: `layer_name` type is not specified...

patrick-wilken: Oops, fixed.

PR #118, opened 2019-01-27
Respect trainable=False for RecLayer subnetwork layers

### LINKED ISSUE #117: Trainable=False flag is ignored for sublayers of RecLayer
Currently, setting `trainable=False` for layers that are part of a `RecLayer` has no effect. I guess this is a bug, so I've created failing tests as well as a fix. See my PR.

ChristopherBrix:
This PR:
 - allows to set `trainable=False` to layers that are part of a `RecLayer` 
 - adds test cases to ensure that both trainable and untrainable layers are handled correctly, whether they were optimized out of the loop or not

This fixes #117

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -798,8 +798,6 @@ class TFNetwork(object):
     for layer_name in sorted(self._selected_train_layers):
       layer = self.layers[layer_name]
       assert isinstance(layer, LayerBase)
-      if not layer.trainable:
-        continue
       for param_name, param in sorted(layer.params.items()):
         assert isinstance(param, tf.Variable)
         if param in trainable_vars_col:
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -3422,6 +3422,201 @@ def test_GenericAttentionLayer_extra_spatial_multi_head():
   assert list(layer.output.size_placeholder.values())[0] is layer.weights.output.size_placeholder[0]
 
 
+def test_untrainable_sublayers():
+  with make_scope() as session:
+    config = Config()
+    n_in, n_out = 2, 3
+    net_dict = {
+      "source_embed": {"class": "linear", "activation": None, "with_bias": False, "n_out": 6},
+
+      "lstm0_fw": {"class": "rec", "unit": "standardlstm", "unit_opts": {"use_peepholes": True, "forget_bias": 0.0},
+                   "initial_state": "var", "n_out": 10, "direction": 1, "from": ["source_embed"]},
+      "lstm0_bw": {"class": "rec", "unit": "standardlstm", "unit_opts": {"use_peepholes": True, "forget_bias": 0.0},
+                   "initial_state": "var", "n_out": 10, "direction": -1, "from": ["source_embed"]},
+
+      "lstm1_fw": {"class": "rec", "unit": "standardlstm", "unit_opts": {"use_peepholes": True, "forget_bias": 0.0},
+                   "initial_state": "var", "n_out": 10, "direction": 1, "from": ["lstm0_fw", "lstm0_bw"]},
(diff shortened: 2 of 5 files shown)


TFNetwork.py
```diff
@@ -779,16 +779,33 @@ def get_trainable_params(self):
...
-      assert isinstance(layer, LayerBase)
-      if not layer.trainable:
-        continue
-      for param_name, param in sorted(layer.params.items()):
-        assert isinstance(param, tf.Variable)
-        if param in trainable_vars_col:
-          l.append(param)
-          trainable_vars_col.remove(param)
+
+    def get_trainable(layer, l, trainable_vars_col):
```
albertz: Parameters should be documented.

TFNetwork.py
```diff
@@ -779,16 +779,33 @@ def get_trainable_params(self):
...
+      if not isinstance(layer, LayerBase):
+        return l, trainable_vars_col
+      from TFNetworkRecLayer import RecLayer, _SubnetworkRecCell
+      if isinstance(layer, RecLayer) and isinstance(layer.cell, _SubnetworkRecCell) and layer.trainable:
```
albertz: There should be no layer specific code here in TFNetwork.

TFNetwork.py
```diff
@@ -779,16 +779,33 @@ def get_trainable_params(self):
...
+        for param_name, param in layer.params.items():
+          assert isinstance(param, tf.Variable)
+          if param in trainable_vars_col:
+            l.append(param)
+            trainable_vars_col.remove(param)
+      return l, trainable_vars_col
+
+    layers = [self.layers[layer_name] for layer_name in sorted(self._selected_train_layers)]
+    for layer in layers:
+      l, trainable_vars_col = get_trainable(layer, l, trainable_vars_col)
```
albertz: This makes it way too complicated.
Why not check sth like `trainable = param in param.graph.get_collection_ref(tf.GraphKeys.TRAINABLE_VARIABLES)`? Or extend `LayerBase.add_param`?

ChristopherBrix: We cannot rely on `tf.GraphKeys.TRAINABLE_VARIABLES`, because even untrainable variables are added to this group. Eg. in the [LinearLayer](https://github.com/rwth-i6/returnn/blob/master/TFNetworkLayer.py#L2027), the matrix `W` is created without passing the `trainable` flag. Therefore, TF adds the `TRAINABLE_VARIABLES` graph key whether or not the layer should be trained.

I think this is the reason why the whole `get_trainable_params()` method is needed at all - we cannot just select all `TRAINABLE_VARIABLES`.
Maybe this is the main issue that should be fixed? It would require to update the parameter definition of every layer, but should remove the need for this method and the filtering. I do not know whether there will be any side effects of setting this flag.

albertz: Of course we can do that. We could just change the `LayerBase.add_param` function.

ChristopherBrix:
I haven't forgotten about this issue, I'll try an alternative (simpler) way to solve it after 3rd of March, once I've more time for it.

tests/test_TFNetworkRecLayer.py
```diff
@@ -2473,6 +2473,200 @@ def get_net_dict():
...
+    assert(weight_input_layer_moved_out in set(network.get_trainable_params()))
+
+    weight_output_layer_moved_out = network.layers["output"].params["readout_in/W"]
+    assert(weight_output_layer_moved_out in set(network.get_trainable_params()))
+
+    weight_internal = network.layers["output"].params["prev_s_transformed/W"]
+    assert(weight_internal in set(network.get_trainable_params()))
+
+
+
```
albertz: There are too many empty lines after the function.

albertz:
There are conflicts now. Can you rebase?

ChristopherBrix:
Done

PR #122, opened 2019-01-29
adding flag to tools/compile_native_op.py to better control linker params

pavelgolik:
When we use TF with Intel MKL (built with `--config=mkl`) and build native ops, OpMaker retrieves both OpenBLAS and MKL from the run-time environment, such that libopenblas has a higher priority.

With `--skip_openblas` we can drop the OpenBLAS libraries from ld_flags such that MKL's version is picked at run-time.

### DIFF
--- a/TFNativeOp.py
+++ b/TFNativeOp.py
@@ -56,7 +56,8 @@ class OpMaker(object):
   op_cache = {}  # cache_key -> op
 
   def __init__(self, description, compiler_opts=None,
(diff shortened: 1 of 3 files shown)


pavelgolik:
ok, will re-structure as suggested

albertz:
You don't have to close this. You can just update the pull request.
(Btw, was the revert commit intended?)

pavelgolik:
no, I screwed up, that's why I'm redoing the PR

PR #124, opened 2019-01-29
adding flag to tools/compile_native_op.py to better control linker param

pavelgolik:
When we use TF with Intel MKL (built with `--config=mkl`) and build native ops, OpMaker retrieves both OpenBLAS and MKL from the run-time environment, such that libopenblas has a higher priority.

With `--blas_lib` we can specify either
 * `/path/to/libmyblas.so` -> link with `-L/path/to -lmyblas`, or
 * `libmyblas.so` -> perform search and add `-L... -l...` for library that contains `libmyblas.so` in file name

### DIFF
--- a/TFNativeOp.py
+++ b/TFNativeOp.py
@@ -56,7 +56,8 @@ class OpMaker(object):
   op_cache = {}  # cache_key -> op
 
   def __init__(self, description, compiler_opts=None,
-               search_for_runtime_blas=True, search_for_numpy_blas=True, search_for_system_blas=True):
+               search_for_runtime_blas=True, search_for_numpy_blas=True, search_for_system_blas=True,
+               blas_lib=None):
     """
     :param OpDescription description:
     :param dict[str]|None compiler_opts: passed on to OpCodeCompiler as kwargs
@@ -68,6 +69,7 @@ class OpMaker(object):
     self.search_for_runtime_blas = search_for_runtime_blas
     self.search_for_numpy_blas = search_for_numpy_blas
     self.search_for_system_blas = search_for_system_blas
+    self.blas_lib = blas_lib
 
   @classmethod
   def _cls_init(cls):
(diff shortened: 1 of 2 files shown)


TFNativeOp.py
```diff
@@ -402,14 +404,23 @@ def _make_mod(self):
...
+    if not have_blas_lib and self.search_for_runtime_blas:
       import Util
       libs = Util.find_sgemm_libs_from_runtime()
       if libs:
         numpy_libs = [fn for fn in libs if "/numpy/.libs/" in fn]
         if numpy_libs:
           # Prefer Numpy; move to front.
           libs = numpy_libs + [fn for fn in libs if fn not in numpy_libs]
+        if self.blas_lib is not None:
+          libs = filter(lambda l: self.blas_lib in l, libs)
```
albertz: Instead of `filter`, I would use list comprehensions, which makes it easier to read.
But also, I would then add an `assert` that there is actually such a lib.
I.e.:
```suggestion
          libs = [l for l in libs if self.blas_lib in l]
```

albertz: Ah, wait, I see, you might later add more to `libs`. So ignore the `assert`.

TFNativeOp.py
```diff
@@ -425,6 +436,8 @@ def _make_mod(self):
         from glob import glob
         for f in glob("%s/.libs/*.so" % numpy_dir):
           f = os.path.basename(f)
+          if self.blas_lib is not None and self.blas_lib not in f:
+            continue
           if f.startswith("lib"):
             f = f[3:]
           if f.endswith(".so"):
```
albertz: The `assert` I was mentioning should be added at some point in the end.
Maybe actually that filtering should be done also there, so you don't need to have it before.

TFNativeOp.py
albertz: See also below. Do the filtering just in the end, and then also add the `assert libs` there.

albertz:
Can you also add the assert?

pavelgolik:
Not sure I understand what to assert - it's fine to fail multiple times, get a warning and compile anyway

albertz:
I meant that if you set `blas_lib`, and it is not a valid filename, and also it is nowhere found (not in the runtime, or anywhere else), that would be strange, and should lead to an exception, because likely `blas_lib` is incorrectly specified.
But maybe not so important now. I will just merge your changes for now.

albertz:
Btw, also, it would be cleaner to cleanup the commit history, and rebase everything into a single commit. But anyway, will commit like that now.

PR #125, opened 2019-01-29
Loss out per seq

jan-rosendahl-rwth:
added options
++eval_output_file_per_seq FILE_NAME  
++loss_name LOSS_NAME 
++output_per_seq_format "seq_tags;seq_len;score;error;pos_score;pos_error"

### DIFF
--- a/TFEngine.py
+++ b/TFEngine.py
@@ -1344,54 +1344,100 @@ class Engine(object):
         return False
     return True
 
-  def eval_model(self, output_file=None, output_per_seq_file=None):
+  def eval_model(self, output_file=None, output_per_seq_file=None, loss_name=None, output_per_seq_format=("seq_len", "score", "pos_score")):
     """
     Eval the current model on the eval datasets (dev + eval, whatever is set).
     See also :func:`self.search` for performing beam search.
 
     :param str|None output_file: if given, will save the results to this file (total err/score for each dataset)
     :param str|None output_per_seq_file: if given, will save the err/score for each sequence
+    :param str|None loss_name: specifies the loss which will be written to output_file
+    :param list[str]|tuple[str]|None output_per_seq_format:
+      which properies of `loss_name` should be written to `output_per_seq_file`.
+      allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+
     :return: nothing
     """
     extra_fetches = None
-    results_per_seq = {}  # seq_tag -> dict[str,float]
+
     if output_per_seq_file:
-      extra_fetches = {}
+      allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+
+      assert isinstance(output_per_seq_format, (tuple, list))
+      assert set(output_per_seq_format) - allowed_outputs == set(), (
+        "Only %r are allowed in function eval_model as output_per_seq_format, but got: %r " % (allowed_outputs, output_per_seq_format))
+
+      # always fetch seq_tag to map loss values to the corresponding line
+      extra_fetches = {"seq_tags": self.network.get_seq_tags()}
+
       from TFUtil import identity
       losses_dict, _, _ = self.network.get_losses_initialized(reduce_func=identity, with_total=False)
-      for loss_name, loss_holder in losses_dict.items():
-        loss_holder.loss.layer = None  # enforce reinit
-        loss_holder.loss.use_flatten_frames = False  # we need that such that we get (B*T,...) unreduced values
-        loss_value = loss_holder.get_normalized_loss_value_per_seq()
-        error_value = loss_holder.get_normalized_error_value_per_seq()
-        if loss_value is not None:
-          extra_fetches["loss:score:%s" % loss_name] = loss_value
-        if error_value is not None:
-          extra_fetches["loss:error:%s" % loss_name] = error_value
-        if loss_holder.loss.output.have_time_axis():
-          seq_lens = loss_holder.loss.output.get_sequence_lengths()
-          extra_fetches["loss:seq_lens:%s" % loss_name] = seq_lens
-        extra_fetches["seq_tags"] = self.network.get_seq_tags()
-      assert extra_fetches
+      assert loss_name in losses_dict, (
+        "Unknown loss defined. Got %r. Possible losses are %r" % (loss_name, losses_dict.keys()))
 
+      loss_holder = losses_dict[loss_name]
+      # enforce reinit, otherwise the new value of 'loss_holder.loss.use_flatten_frames' will be ignored
+      loss_holder.loss.layer = None
+      loss_holder.loss.use_flatten_frames = False  # we need that such that we get (B*T,...) unreduced values
+
+      # we need sequence lengths for positional fetches
+      has_positional_fetch = ("pos_score" in output_per_seq_format) or ("pos_error" in output_per_seq_format)
+
+      if "seq_len" in output_per_seq_format or has_positional_fetch:
+        extra_fetches["seq_len"] = loss_holder.loss.output.get_sequence_lengths()
+      if "score" in output_per_seq_format:
+        extra_fetches["score"] = loss_holder.get_normalized_loss_value_per_seq()
+      if "error" in output_per_seq_format:
+        extra_fetches["error"] = loss_holder.get_normalized_error_value_per_seq()
+      if "pos_score" in output_per_seq_format:
+        extra_fetches["pos_score"] = loss_holder.get_normalized_loss_value_per_seq(per_pos=True)
+      if "pos_error" in output_per_seq_format:
+        extra_fetches["pos_error"] = loss_holder.get_normalized_error_value_per_seq(per_pos=True)
+
+    results_per_seq = {}  # seq_tag -> dict[str,float]. Results of fetches will be written in this dict
+
+    # function to save the return values of each callback to the dict `results_per_seq`
     def extra_fetches_callback(seq_tags, **extra_fetches_out):
       """
       :param list[str] seq_tags:
       :param dict[str,numpy.ndarray] extra_fetches_out: see extra_fetches
       """
+
       for name, value in extra_fetches_out.items():
-        assert name.startswith("loss:")
-        name = name[len("loss:"):]
+        assert name in allowed_outputs
         assert isinstance(value, numpy.ndarray)
-        assert value.shape == (len(seq_tags),)
-        for i, seq_tag in enumerate(seq_tags):
-          results_per_seq.setdefault(seq_tag, {})[name] = value[i]
+
+        # in case of positional values, we have to handle a 2-dim ndarray
+        if name[:4] == "pos_":
+          assert 'seq_len' in extra_fetches_out
+          seq_lens = extra_fetches_out['seq_len']
+          shorted_scores = [ps[:l] for ps,l in zip(value, seq_lens)]
+          for i, seq_tag in enumerate(seq_tags):
+            results_per_seq.setdefault(seq_tag, {'seq_tags': seq_tag})[name] = shorted_scores[i]
+        else:
+          for i, seq_tag in enumerate(seq_tags):
+            results_per_seq.setdefault(seq_tag, {'seq_tags': seq_tag})[name] = value[i]
+
 
     # It's constructed lazily and it will set used_data_keys, so make sure that we have it now.
     self.network.maybe_construct_objective()
     results = {}
     eval_dump_str = []
     train = self._maybe_prepare_train_in_eval()
+
+    if output_per_seq_file:
+      assert len(self.get_eval_datasets()) == 1, (
+        "output per sequence is only supported for one dataset (dev or eval), provided datasets are %r" % list(self.get_eval_datasets().keys()))
+      # try to sort dataset to minimize zero-padding
+      dataset = list(self.get_eval_datasets().values())[0]
+      if dataset.have_corpus_seq_idx():
+        # We can sort it. Sort it in reverse to make sure that we have enough memory right at the beginning.
+        print("Dataset have_corpus_seq_idx == True, i.e. it will be sorted for optimal performance.", file=log.v3)
+        dataset.seq_ordering = "sorted_reverse"
+      else:
+        print("Dataset have_corpus_seq_idx == False, i.e. it will not be sorted for optimal performance.", file=log.v3)
+        dataset.seq_ordering = "default"  # enforce order as-is, so that the order in the written file corresponds
+      dataset.init_seq_order(epoch=self.epoch)
 
     for dataset_name, dataset in self.get_eval_datasets().items():
       if dataset_name not in self.dataset_batches or not dataset.batch_set_generator_cache_whole_epoch():
@@ -1427,8 +1473,25 @@ class Engine(object):
     if output_per_seq_file:
       print('Write eval results per seq to %r' % output_per_seq_file, file=log.v3)
       from Util import betterRepr
+
+      # write fetches to file
       with open(output_per_seq_file, 'w') as f:
-        f.write(betterRepr(results_per_seq) + '\n')
+        def create_output_string(seq_tag):
+          """
+          Creates a colon (;) separated output string in the order given by `requested_outputs`
+          :param str seq_tag:
+          """
+          value_list = [results_per_seq[seq_tag][req_out] for req_out in output_per_seq_format]
+          value_list = [' '.join(map(str, v)) if isinstance(v, numpy.ndarray) else str(v) for v in value_list]
+          assert all([all([c not in "\n;" for c in v]) for v in value_list])
+          return ';'.join(value_list)
+
+        corpus_seq_idx_to_idx = {dataset.get_corpus_seq_idx(idx): idx for idx in range(len(results_per_seq))}
+        results_per_seq_sorted = [dataset.get_tag(corpus_seq_idx_to_idx[corp_idx]) for corp_idx in range(len(results_per_seq))]
+        results_per_seq = [create_output_string(seq_tag) for seq_tag in results_per_seq_sorted]
+
+        for res in results_per_seq:
+          f.write(str(res) + '\n')
 
   def check_last_epoch(self):
     if self.start_epoch == 1:
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -1411,31 +1411,45 @@ class LossHolder:
     self._prepare()
     return self._norm_factor
 
-  def _normalized_loss_value_per_seq(self, value):
+  def _normalized_value_per_seq(self, value, per_pos=False):
     """
-    :param tf.Tensor|None loss:
-    :return: (batch,) or None if loss is None
+    :param tf.Tensor|None value: (batch*time,) or (time*batch,)
+    :param bool per_pos: one value per time position
+    :return: if per_pos return (batch,time) else (batch,) or None if loss is None
     :rtype: tf.Tensor|None
     """
     if value is None:
       return None
-    return self.loss.reduce_to_batch(value, normalize=True)
 
-  def get_normalized_loss_value_per_seq(self):
+    if per_pos:
+      value = tf.reshape(value, tf.shape(self.loss.output.placeholder)[:2])  # (batch,time) or (time,batch)
+
+      # We want output of the form (B,T)
+      if self.loss.output.time_dim_axis == 0:
+        from TFUtil import swapaxes
+        value = swapaxes(value, 0, 1)  # resulting in (B,T,...)
+
+      return value
+    else:
+      return self.loss.reduce_to_batch(value, normalize=True)
+
+  def get_normalized_loss_value_per_seq(self, per_pos=False):
     """
-    :return: (batch,) or None if loss is None
+    :param bool per_pos: one value per time position
+    :return: if per_pos return (batch,time) else (batch,) or None if loss is None
     :rtype: tf.Tensor|None
     """
     self._prepare()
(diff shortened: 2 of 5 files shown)


TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
 
-  def eval_model(self, output_file=None, output_per_seq_file=None):
+  def eval_model(self, output_file=None, output_per_seq_file=None, loss_name=None, output_per_seq_format="seq_len;score;pos_score"):
     """
     Eval the current model on the eval datasets (dev + eval, whatever is set).
     See also :func:`self.search` for performing beam search.
 
     :param str|None output_file: if given, will save the results to this file (total err/score for each dataset)
     :param str|None output_per_seq_file: if given, will save the err/score for each sequence
+    :param [str]|None loss_name=None: Needed if output_file is given.
```
albertz: This is not the proper documentation syntax. `[str]` is not allowed (and also not clear what this should mean). `=None` is also not allowed, and also redundant.
Also, the description itself (after `: `) is not really clear to me. I don't really know what this is, and why it's needed.

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+    :param str|None output_per_seq_format="seq_len;score;pos_score":
+                                            which properies of `loss_name` should be written to `output_per_seq_file`.
+                                            allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+                                            requested outputs must be separated by a colon ";", e.g. "seq_len;score;error"
```
albertz: I would remove some indent spaces here. All the contiguous comments starting in the next line can be just 2 spaces indented from the first line. Like this:
```
:param str foo: comment.
  comment continues here.
  and here some more.
```

albertz: Also, I really would not introduce custom syntax. This can become ugly. Just require a `list` or `tuple`.

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+
     extra_fetches = None
-    results_per_seq = {}  # seq_tag -> dict[str,float]
+    requested_outputs=[]
+
     if output_per_seq_file:
-      extra_fetches = {}
+      allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+
+      requested_outputs = output_per_seq_format.split(';')
```
albertz: As said, do not use `split` here, and also no need for the extra `output_per_seq_format` variable. (Actually, maybe that is a better param name?)
Instead, just do:
```suggestion
      assert isinstance(output_per_seq_format, (tuple, list))
```

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
-        if loss_holder.loss.output.have_time_axis():
-          seq_lens = loss_holder.loss.output.get_sequence_lengths()
-          extra_fetches["loss:seq_lens:%s" % loss_name] = seq_lens
-        extra_fetches["seq_tags"] = self.network.get_seq_tags()
-      assert extra_fetches
-
+      assert loss_name in losses_dict, \
+        "Unknown loss defined. Got \"%s\". Possible losses are %s" % (loss_name, losses_dict.keys())
+
+      #for loss_name, loss_holder in losses_dict.items():
```
albertz: Do not add commented out code.
Also, why is it commented out?

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+      loss_holder = losses_dict[loss_name]
```
albertz: Why do you want to return only this loss?

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+      loss_holder.loss.layer = None                  # enforce reinit, otherwise the new value of 'loss_holder.loss.use_flatten_frames' will be ignored
```
albertz: The number of spaces between some Python statement and the comment should be exactly 2.
Also, this line looks too long. In such cases, I would put the comment just in the line before the Python statement.

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+      loss_holder.loss.use_flatten_frames = False    # we need that such that we get (B*T,...) unreduced values
```
albertz: Same.

TFEngine.py
```diff
@@ -1344,48 +1345,84 @@ def _do_save(self):
...
+      if "pos_error" in requested_outputs:
+        extra_fetches["pos_error"] = loss_holder.get_normalized_error_value_per_seq(per_pos=True)
+
+    results_per_seq = {}  # seq_tag -> dict[str,float]. Results of fetches will be written in this dict
+
+    # function to save the return values of each callback to the dict `results_per_seq`
     def extra_fetches_callback(seq_tags, **extra_fetches_out):
       """
-      :param list[str] seq_tags:
+      :param list[str] seq_tags
```
albertz: This syntax is wrong. Why did you remove the `:`?
(Btw, curious, does PyCharm not print a warning/hint about this? E.g. like that `seq_tags` is not documented? But what you should notice in any case is that now PyCharm will not know the type of `seq_tags` anymore, which is something which you should always check, and always try to avoid if possible.)

TFEngine.py
```diff
@@ -1427,8 +1464,24 @@ def extra_fetches_callback(seq_tags, **extra_fetches_out):
...
       with open(output_per_seq_file, 'w') as f:
-        f.write(betterRepr(results_per_seq) + '\n')
+        def create_output_string(st):
+          """
+          Creates a colon (;) separated output string in the order given by `requested_outputs`
+          :param str st : a seq_tag
+          """
+          value_list = [results_per_seq[st][req_out] for req_out in requested_outputs]
+          value_list = [' '.join(map(str, v)) if isinstance(v, numpy.ndarray) else str(v) for v in value_list]
+          return ';'.join(value_list)
```
albertz: Do not do sth like this. No custom format/syntax.

TFEngine.py
```diff
@@ -1427,8 +1464,24 @@ def extra_fetches_callback(seq_tags, **extra_fetches_out):
...
       print('Write eval results per seq to %r' % output_per_seq_file, file=log.v3)
       from Util import betterRepr
+
+      # write fetches to file
       with open(output_per_seq_file, 'w') as f:
-        f.write(betterRepr(results_per_seq) + '\n')
+        def create_output_string(st):
+          """
+          Creates a colon (;) separated output string in the order given by `requested_outputs`
+          :param str st : a seq_tag
```
albertz: There is an extra space between `str st` and `:`. Why? Make it consistent. (I'm even not sure if this is allowed.)

albertz: Also, instead of this simple description ("a seq_tag"), why don't you just call the variable like that? Then the comment would not be needed, and the remaining code becomes also easier to read.

albertz: (But anyway, the whole function should be removed, as said below.)

TFNetwork.py
```diff
@@ -1447,11 +1461,11 @@ def _tf_summary(self):
       return  # skip summaries. the root net should also do this
     name = self.get_tf_name()
     if self._loss_value is not None:
-      tf.summary.scalar("loss_%s" % name, self._loss_value * self._norm_factor)
+      tf.summary.tensor_summary("loss_%s" % name, self._loss_value * self._norm_factor)
       if self._network.get_config().bool("calculate_exp_loss", False):
         tf.summary.scalar("exp_loss_%s" % name, tf.exp(self._loss_value * self._norm_factor))
     if self._error_value is not None:
-      tf.summary.scalar("error_%s" % name, self._error_value * self._norm_factor)
+       tf.summary.tensor_summary("error_%s" % name, self._error_value * self._norm_factor)
```
albertz: Why the extra space?

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
+        from TFUtil import swapaxes
+        value = swapaxes(value, 0, 1)  # resulting in (B,T,...)
+
+      return value
+    else:
+      return self.loss.reduce_to_batch(value, normalize=True)
+
+  def get_normalized_loss_value_per_seq(self, per_pos=False):
     """
+    :param bool per_pos: if true return tf.Tensor(B,T) else tf.Tensor(B)
```
albertz: Keep the documentation/comments somewhat consistent. Just below, I write just the tuple (`(batch,)`), so write it in the same way.

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
     :return: (batch,) or None if loss is None
-    :rtype: tf.Tensor|None
+    :rtype: tf.Tensor(B,T)|tf.Tensor(B)|None
```
albertz: This is not a valid type. The type is `tf.Tensor`.
All extra information about what is being returned should be written in the `:return:` description (just the line above).

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
     self._prepare()
     return self._norm_factor
 
-  def _normalized_loss_value_per_seq(self, value):
+  def _normalized_value_per_seq(self, value, per_pos=False):
     """
-    :param tf.Tensor|None loss:
+    :param tf.Tensor|None value: (batch*time,) or (time*batch,)
+    :param bool per_pos: if true return tf.Tensor(B,T) else tf.Tensor(B)
```
albertz: Keep the comments consistent. Write `... (batch,time) else (batch,)`.
Or actually this should rather be written for `:return:`.

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
     :return: (batch,) or None if loss is None
```
albertz: `:return:` is incorrect now in the documentation. Fix that.

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
     :rtype: tf.Tensor|None
     """
     if value is None:
       return None
-    return self.loss.reduce_to_batch(value, normalize=True)
 
-  def get_normalized_loss_value_per_seq(self):
+    if per_pos:
+      value = tf.reshape(value, tf.shape(self.loss.output.get_sequence_mask()))  # (batch,time) or (time,batch)
```
albertz: This is a very inefficient way to get the shape. Why do you need to create a sequence mask for that? Instead, do sth like:
```suggestion
      value = tf.reshape(value, tf.shape(self.loss.output.placeholder)[:2])  # (batch,time) or (time,batch)
```

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
+        value = swapaxes(value, 0, 1)  # resulting in (B,T,...)
+
+      return value
+    else:
+      return self.loss.reduce_to_batch(value, normalize=True)
+
+  def get_normalized_loss_value_per_seq(self, per_pos=False):
     """
+    :param bool per_pos: if true return tf.Tensor(B,T) else tf.Tensor(B)
     :return: (batch,) or None if loss is None
```
albertz: `:return:` must be fixed now.

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
+    :rtype: tf.Tensor(B,T)|tf.Tensor(B)|None
     """
     self._prepare()
-    return self._normalized_loss_value_per_seq(self._loss_value)
+    return self._normalized_value_per_seq(self._loss_value, per_pos=per_pos)
 
-  def get_normalized_error_value_per_seq(self):
+  def get_normalized_error_value_per_seq(self, per_pos=False):
     """
+    :param bool per_pos: if true return tf.Tensor(B,T) else tf.Tensor(B)
```
albertz: Again, keep consistent.

TFNetwork.py
```diff
@@ -1411,31 +1411,45 @@ def get_norm_factor(self):
...
     :return: (batch,) or None if error is None
```
albertz: `:return:` needs to be fixed now.

TFNetwork.py
```diff
@@ -1447,11 +1461,11 @@ def _tf_summary(self):
       return  # skip summaries. the root net should also do this
     name = self.get_tf_name()
     if self._loss_value is not None:
-      tf.summary.scalar("loss_%s" % name, self._loss_value * self._norm_factor)
+      tf.summary.tensor_summary("loss_%s" % name, self._loss_value * self._norm_factor)
```
albertz: Why this change?

albertz: Ah, because it is not necessarily a scalar now?
Add a comment (extra line, above this) about that.

TFEngine.py
```diff
@@ -15,6 +15,7 @@
 import os
 import sys
 import time
+import re
```
albertz: Can be removed, see my comments below.

TFEngine.py
albertz: Actually, what was the problem with the format before? It should be much simpler, and better, in many ways:
* The code here is simpler.
* The parsing code will be simpler.
* It's bullet proof, this cannot break.

TFEngine.py
jan-rosendahl-rwth: because otherwise the format of the output file is very unclear (e.g. order of losses) as we already have a variable amount of output values depending on "output_per_seq_format" (where all positional values can have arbitrary length). For further parsing usually a csv file with one line per input sentence is used (especially for huge datasets e.g. in data filtering).

Also their can be many losses but we typically are only interested in a single one.

TFNetwork.py
jan-rosendahl-rwth: we adapted this from:
https://github.com/rwth-i6/returnn/blob/b3117f972bf0c6c4c607c4165034f7f1e257e87b/TFNetworkLayer.py#L5422

we changed both lines now

TFEngine.py
jan-rosendahl-rwth: what comment are you referring to?

TFNetwork.py
jan-rosendahl-rwth: we actually are not sure, it did crash before. (even without positional loss)

albertz: Do not change code when you are not sure. Understand what the problem is. But I guess it's what I wrote.

TFEngine.py
albertz: No, that's not true. Reading a dict should be way faster. And of course much easier.
Also, you do not have the problem that this likely will break at some point.
The space overhead should be really low. And if this is a problem, maybe you can just use on-the-fly gzip or so.

TFEngine.py
```diff
@@ -1344,48 +1345,81 @@ def _do_save(self):
...
-        if error_value is not None:
-          extra_fetches["loss:error:%s" % loss_name] = error_value
-        if loss_holder.loss.output.have_time_axis():
-          seq_lens = loss_holder.loss.output.get_sequence_lengths()
-          extra_fetches["loss:seq_lens:%s" % loss_name] = seq_lens
-        extra_fetches["seq_tags"] = self.network.get_seq_tags()
-      assert extra_fetches
-
+      assert loss_name in losses_dict, \
+        "Unknown loss defined. Got \"%s\". Possible losses are %s" % (loss_name, losses_dict.keys())
```
albertz: Use `%r` instead of `\"%s\"` and also instead of the second `%s`.
Also, better use brackets instead of `\` to break the newline, as in `assert loss_name in losses_dict, (\n "Unknown loss ...")`, as this makes it easier to identify in what line the assertion happened, when the code is parsed, for tools like better_exchook. Which is some useful information here, as the user would basically see also the variable names, which can be helpful.

TFEngine.py
albertz: Where you use it. `re.sub`.

TFEngine.py
```diff
@@ -1344,48 +1345,81 @@ def _do_save(self):
...
     extra_fetches = None
-    results_per_seq = {}  # seq_tag -> dict[str,float]
+
     if output_per_seq_file:
-      extra_fetches = {}
+      allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+
+      assert isinstance(output_per_seq_format, (tuple, list))
+      assert set(output_per_seq_format) - allowed_outputs == set(), \
+        "Only %s are allowed in function eval_model as output_per_seq_format, but got: %s " % (allowed_outputs, output_per_seq_format)
```
albertz: Similar as in next comment, better use brackets instead of `\` for new line.

TFEngine.py
albertz: Ok. But always check you commits also, line by line, if your change is really meaningful. Some change like this should be seen this way.

TFNetwork.py
albertz: Also, when you understood what the problem is, add a comment into the code, why you use `tensor_summary` here, and not `scalar`.

TFEngine.py
```diff
@@ -1344,48 +1345,81 @@ def _do_save(self):
...
-  def eval_model(self, output_file=None, output_per_seq_file=None):
+  def eval_model(self, output_file=None, output_per_seq_file=None, loss_name=None, output_per_seq_format=["seq_len", "score", "pos_score"]):
     """
     Eval the current model on the eval datasets (dev + eval, whatever is set).
     See also :func:`self.search` for performing beam search.
 
     :param str|None output_file: if given, will save the results to this file (total err/score for each dataset)
     :param str|None output_per_seq_file: if given, will save the err/score for each sequence
+    :param str|None loss_name: specifies the loss which will be written to output_file
+    :param list[str]|tuple[str]|None output_per_seq_format:
```
albertz: Your default in the config is not a list or tuple, but just a `str`. Maybe you also want to allow that here.

TFEngine.py
```diff
@@ -1344,48 +1345,81 @@ def _do_save(self):
...
     """
+
     extra_fetches = None
-    results_per_seq = {}  # seq_tag -> dict[str,float]
+
     if output_per_seq_file:
-      extra_fetches = {}
+      allowed_outputs = {"seq_tags", "seq_len", "score", "error", "pos_score", "pos_error"}
+
+      assert isinstance(output_per_seq_format, (tuple, list))
```
albertz: Maybe you also want to allow a single `str` here.
```suggestion
      assert isinstance(output_per_seq_format, (str, tuple, list))
      if isinstance(output_per_seq_format, str):
          output_per_seq_format = [output_per_seq_format]
```

rnn.py
```diff
@@ -420,8 +420,10 @@ def executeMainTask():
     config.set('load_epoch', engine.epoch)
     print("Evaluate epoch", engine.epoch, file=log.v4)
     engine.eval_model(
-      output_file=config.value("eval_output_file", ""),
-      output_per_seq_file=config.value("eval_output_file_per_seq", ""))
+      output_file=config.value("eval_output_file", None),
+      output_per_seq_file=config.value("eval_output_file_per_seq", None),
+      loss_name=config.value("loss_name", None),
+      output_per_seq_format=config.value("output_per_seq_format", "score"),)
```
albertz: Why the extra `,` in the end before the `)`?
Also, you write that you only allow a list or tuple. But see my comment above. Maybe str is also fine.

TFNetworkLayer.py
```diff
@@ -5419,7 +5419,7 @@ def reduce_to_batch(self, loss, normalize):
       # We expect to get (batch*time) or (time*batch) in the first dimension of the loss and the output.
       loss = self._reduce_to_batch_time_with_mask(loss, normalize=normalize)  # (batch*time,) or (time*batch,)
       loss.set_shape((None,))
-      loss = tf.reshape(loss, tf.shape(self.output.get_sequence_mask()))  # (batch,time) or (time,batch)
+      loss = tf.reshape(loss, tf.shape(self.loss.output.placeholder)[:2])  # (batch,time) or (time,batch)
```
albertz: Why `self.loss.output` and not `self.output`?

TFEngine.py
jan-rosendahl-rwth: changed the default value in the config. List or tuple is required now

TFNetworkLayer.py
jan-rosendahl-rwth: that was your suggestion :)
Is fixed now

TFEngine.py
```diff
@@ -1344,48 +1344,80 @@ def _do_save(self):
         return False
     return True
 
-  def eval_model(self, output_file=None, output_per_seq_file=None):
+  def eval_model(self, output_file=None, output_per_seq_file=None, loss_name=None, output_per_seq_format=["seq_len", "score", "pos_score"]):
```
albertz: Small other problem (and I'm quite sure PyCharm warns about this): You should not use mutable default values (such as a list). Make this a tuple instead.

TFEngine.py
```diff
@@ -1427,8 +1473,25 @@ def extra_fetches_callback(seq_tags, **extra_fetches_out):
...
+          """
+          Creates a colon (;) separated output string in the order given by `requested_outputs`
+          :param str seq_tag:
+          """
+          value_list = [results_per_seq[seq_tag][req_out] for req_out in output_per_seq_format]
+          value_list = [' '.join(map(str, v)) if isinstance(v, numpy.ndarray) else str(v) for v in value_list]
+          assert all([all([c not in "\n;" for c in v]) for v in value_list])
+          return ';'.join(value_list)
+
+        corpus_seq_idx_to_idx = {dataset.get_corpus_seq_idx(idx):idx for idx in range(len(results_per_seq))}
```
albertz: There is a space missing before `idx`. (I think PyCharm should hint about this?)

PR #128, opened 2019-02-01
added TripletLoss

orech:
Triplet loss implements `loss = max(margin + d(x_a, x_s) - d(x_a, x_d), 0.0)`. It can be used  in siamese/triplet network. Distances between instances of the same class `d(x_a, x_s)` are pushed to be lower than distances between instances of different classes ` d(x_a, x_d)` by at least margin value.


TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+  Triplet loss is used for metric learning in a siamese/triplet network.
+  It should be used as a part of CopyLayer with 3 inputs corresponding to x_a, x_s and x_d in a loss.
+  Here we assume that x_a are anchor samples, x_s are samples where at each position i in a minibatch
+  x_ai and x_si belong to the same class, while pairs x_ai and x_di belong to different classes.
+  In this implementation the number of training examples is increased
+  by extracting all possible same/different pairs within a minibatch.
+  """
+  class_name = "minibatch_triplet_loss"
+
+  def __init__(self, margin=0.6, multi_view_training=False, **kwargs):
```
albertz: `margin=0.6` seems very arbitrary.
I would rather enforce that this must be set (i.e. no default value), and you can maybe say in the documentation what a good default value would be.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+    super(TripletLoss, self).__init__(**kwargs)
+    """
+    :param margin: how much the distance between instances of the same class should be smaller then distances between instances of different classes.
```
albertz: You do not define the type. You should write `:param float margin: ...`.
Also make sure that the line does not exceeds 120 chars. If you need more, then add a new line (but indent the next line with 2 spaces, not less and also not more).

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+    :param multi_view_training: True if we have a pair of inputs (x_a, x_s, x_d) extracted from two different data representations (i.e. acoustic and orthographic)
```
albertz: Same as before.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+    self.multi_view = multi_view_training
+
+  def init(self, output, output_with_activation=None, target=None, **kwargs):
+    """
+    :param Data output: generated output
+    :param OutputWithActivation|None output_with_activation:
+    :param Data target: reference target from dataset
+    """
+    super(TripletLoss, self).init(output=output, output_with_activation=output_with_activation, target=target, **kwargs)
+    self.loss_norm_factor = 1.0 / tf.cast(tf.shape(3.0 * self.output_flat)[0], tf.float32)
```
albertz: This does not quite make sense. Why do you multiply output * 3, and then take the shape from it?
Also, what is it that you really want here? If you need actually the sequence lengths, or number of frames, then do that, but don't use output flat in that case.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+        embeds = tf.concat(values=[out_0, out_1, out_2], axis=0)
+        a = targets[:, 0]
+        p = targets[:, 1]
+        d = targets[:, 2]
+        labels = tf.concat(values=[a, p, d], axis=0) 
+        loss_out = self._triplet_loss(embeds, labels)
+
+    return loss_out
+
+  def _triplet_loss(self, embeds, labels):
```
albertz: The parameters are not documented.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+  """
+  Triplet loss: loss = max(margin + d(x_a, x_s) - d(x_a, x_d), 0.0)
+  Triplet loss is used for metric learning in a siamese/triplet network.
+  It should be used as a part of CopyLayer with 3 inputs corresponding to x_a, x_s and x_d in a loss.
+  Here we assume that x_a are anchor samples, x_s are samples where at each position i in a minibatch
+  x_ai and x_si belong to the same class, while pairs x_ai and x_di belong to different classes.
+  In this implementation the number of training examples is increased
+  by extracting all possible same/different pairs within a minibatch.
+  """
+  class_name = "minibatch_triplet_loss"
```
albertz: Why is it called like this? Why not just `"triplet_loss"`? Also, it's inconsistent to the class name.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+    :param Data target: reference target from dataset
+    """
+    super(TripletLoss, self).init(output=output, output_with_activation=output_with_activation, target=target, **kwargs)
+    self.loss_norm_factor = 1.0 / tf.cast(tf.shape(3.0 * self.output_flat)[0], tf.float32)
+
+  def get_value(self):
+    if self.multi_view:
+      with tf.name_scope("multi_view_loss"):
+        out = self.output_flat
+        sources = tf.split(out, num_or_size_splits=6, axis=1)
```
albertz: It looks a lot like you have certain assumptions about the dimensions and number of classes. You should check that (add an `assert`).

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+        out_5 = sources[5]
+        embeds_1 = tf.concat(values=[out_0, out_4, out_5], axis=0)
+        embeds_2 = tf.concat(values=[out_1, out_4, out_2], axis=0)
+        a = targets[:, 0]
+        p = targets[:, 1]
+        d = targets[:, 2]
+        labels = tf.concat(values=[a, p, d], axis=0)
+        loss_out = self._triplet_loss(embeds_1, labels) + self._triplet_loss(embeds_2, labels)
+    else:
+      with tf.name_scope("acoustic_loss"):
```
albertz: This name space sounds very specific for acoustics. But the loss can be used for non-acoustics stuff as well, or not? So make the name more generic then.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+        out = self.output_flat
+        sources = tf.split(out, num_or_size_splits=3, axis=1)
```
albertz: Same as before, you should check the dimensions.

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,112 @@ def full_softmax_fn():
...
+      diff_distances = tf.gather_nd(dist, diff_indices)
+      diff_max = tf.maximum(self.margin - diff_distances, 0.0)
+      diff_loss = 0.5 * tf.reduce_sum(diff_max, 0)
+
+    return same_loss + diff_loss
+
+  def _check_init(self):
+    """
+    Does some checks on self.target and self.output, e.g. if the dense shapes matches.
+    Since self.target is not used in triplet_loss, these checks are not relevant
```
albertz: This is wrong. You actually use it (target_flat).

TFNetworkLayer.py
orech: Fixed that. A good value might be different in each particular case

TFNetworkLayer.py
orech: It is a specific implementation of triplet loss, which considers all possible same/different pairs within a minibatch. The standard implementation would be different. But I only committed this implementation, so I changed the name.

TFNetworkLayer.py
orech: Added `assert` in _check_init().

TFNetworkLayer.py
orech: You are right, this loss can be used for non-acoustic data

TFNetworkLayer.py
```diff
@@ -6595,6 +6595,122 @@ def full_softmax_fn():
...
+        diff_targets = targets[:, 2]
+        labels = tf.concat(values=[anchor_targets, pair_targets, diff_targets], axis=0) 
+        loss_out = self._triplet_loss(embeds, labels)
+
+    return loss_out
+
+  def _triplet_loss(self, embeds, labels):
+    """
+    param tf.Tensor embeds: all embeddings concatenated in batch dim
+    param tf.Tensor labels: all output labels concatenated in batch dim
```
albertz: The documentation misses the shape and length information about the tensors.

TFNetworkLayer.py
albertz: An `assert` also should be added here.

orech: Ok, I added additional checks.

albertz: I would also check that axis 1 is the right axis...

albertz:
I'll merge this now.
Note that I think that all of this should have been implemented very differently. Also not sure if this code will survive in the long run. (@curufinwe)

* The dataset should not be specific to this loss. There is no real need for that. You could just use any dataset, and do any of your custom logic inside the network.
* The loss itself is also too specific. I doubt that much people will use it in the future. And there was also no real need for it, as this also could have been done directly in the config as part of the network.

PR #129, opened 2019-02-06
dataset description and ConvLayer assert

JackTemaki:
It would be really helpful for beginners, if during debugging the axis information would not be dropped for some cases, as this can be very confusing without looking directly into the code.

I also ran in some issue with ConvLayer, an assert seems more helpful than:
`TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'`

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -3226,6 +3226,7 @@ class ConvLayer(_ConcatInputLayer):
     padding = padding.upper()
     if input_expand_dims == 0 and not input_add_feature_dim and not input_split_feature_dim:
       # Maybe we have a chance to correctly define the output shapes.
+      assert data.time_dim_axis is not None, "time_dim_axis was not specified, consider setting dim options of ConvLayer"
       index_shift = data.time_dim_axis_excluding_batch
       for i in range(len(filter_size)):
         if data.shape[i + index_shift] is not None:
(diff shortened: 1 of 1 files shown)


albertz:
There are some reasons why I did it the way it is currently:

* I wanted it to be not too verbose. Some debugging output already is quite long, and this hurts readability. So it drops some very often used default values. Of course, yes, you are right, if you don't know the default values, the output might be misleading/confusing, so it hurts readability for newcomers. But you have to ask yourself: Which group is more important, the newcomers, or people who work with RETURNN all the time?

* `get_description` is used in `__repr__`. The general rule is, that `eval(repr(obj)) == obj` should always be true. This is not the case anymore with your changes, as for the `feature_dim_axis`, it makes a difference if `_feature_dim_axis is NotSpecified` or not. If you specify it explicitly (i.e. not `NotSpecified`), it is a fixed value which will behave differently during some transformations. If it is not specified explicitly, it will always be implicit, which is in some cases a different behavior, and the behavior you would prefer. (Actually, the whole logic, when it is a different behavior and when not, is by itself a bit non-straightforward, but let's keep that as a separate issue.)

I'm actually planning since a while now to do some cleanup there, and simplify it again. And also allow for some better representation, sth like `Data(batch_shape=("B", "T", "D"))` or so. But this will require many changes throughout many parts of RETURNN, and I did not have time yet to do that. Also, this could potentially break some corner cases. Also, I still have not fully decided how the best representation would look like. E.g. there can be multiple spatial/time axes. I already have some support to distinguish them (see e.g. `DimensionTag`).

So, I think for now, I would not merge in your `Data.get_description` changes. Maybe we can think later about how to make this better in a more clean way.

albertz:
About the `ConvLayer` change: It would rather add this:

    assert data.time_dim_axis is not None

This makes the code easier to read, and also the exception output on stdout will contain more information, such as `data` (which is really what you want to see during this exception).

PR #133, opened 2019-02-09
added check for non deterministic ops

tbscode:


TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
   signal = tf.concat([tf.sin(scaled_time), tf.cos(scaled_time)], axis=1)
   signal = tf.pad(signal, [[0, 0], [0, num_channels % 2]])  # (length, channels)
   return signal
+
+def check_contains_only_det_ops(print_trace=False):
+  """
+  :param bool print_trace: prints a trace of which ops are checked
+  :return:  are all ops used on the available devices deterministic?
```
albertz: Make only one space between `:return:` and the following text, not two. (In general, try to be consistent in style with all the remaining code.)
Also better write `:return: whether all ops from the current graph are deterministic`.
Also missing is the return type, i.e. `:rtype: bool`.
Also, better rename the function. `check` does not explain good what the function does. E.g. I expected that it does some `assert` on that or so. If you do not do a check, but you return whether it has some property or not, the function should be called sth like `is_current_graph_deterministic`.

TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+  """
+  # List of non deterministic (NDT) ops (depending on version and device)
+  non_deterministic_ops = {
+    "Mean": ["1.5.0", ["GPU"]]
+    # "Reshape": ["1.13.0", ["CPU", "GPU"]]  # Not NDT just here for test purposes
+  }
+
+  # Getting all available device types
+  device_types = [type.device_type for type in get_tf_list_local_devices()]
```
albertz: Replace `type` (bad variable name for the device, also shadows builtin function) by `dev`. (Btw, do you use PyCharm? It will warn you about things like this.)

TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+
+  # Getting all used ops by name and filtering duplicates
+  ops_set = {op.type for op in tf.get_default_graph().get_operations()}
+
+  deterministic = True
+  for op in ops_set:
+    if print_trace: print("Testing: {}".format(op))
```
albertz: Make a newline after the `if` statement. (PyCharm will also notify about such things.) (This is PEP8 style I think. Or at least common Python convention.)

TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
   return signal
+
+def check_contains_only_det_ops(print_trace=False):
+  """
+  :param bool print_trace: prints a trace of which ops are checked
+  :return:  are all ops used on the available devices deterministic?
+  """
+  # List of non deterministic (NDT) ops (depending on version and device)
+  non_deterministic_ops = {
+    "Mean": ["1.5.0", ["GPU"]]
```
albertz: In general: Do not use a list, but a tuple for things like this.
But in this case (and also almost in general, in most cases), I think you even should introduce some helper class, which makes it more descriptive what the individual fields mean (e.g. min_version, dev_type, or so).
But then, maybe also this makes the whole code too complicated. Maybe just do not introduce such dict at all here, and directly do the check in the `for op in op_set` loop below. I.e. just do a big `if op == "Mean": ... elif ...` etc.

TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+
+  # Getting all used ops by name and filtering duplicates
+  ops_set = {op.type for op in tf.get_default_graph().get_operations()}
+
+  deterministic = True
+  for op in ops_set:
+    if print_trace: print("Testing: {}".format(op))
+    if op in non_deterministic_ops:
+      # Converting Version to tuple
+      pre_version = tuple([int(x) for x in non_deterministic_ops[op][0].split('.') if x.isdigit()])
```
albertz: This is difficult to read/understand from a first glance. But see my comment above. You can make that simpler.

TFUtil.py
```diff
@@ -6667,3 +6667,32 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+      # Do version check | and see if device for which that op is NDT is used
+      if (pre_version >= tf_version_tuple()) and (not set(non_deterministic_ops[op][1]).isdisjoint(device_types)):
+        print("WARING: using {}-op, that is non deterministic in tf.__version__ = {}".format(op, tf.__version__))
+        if deterministic: deterministic = False
```
albertz: Again, newline after `if` statement.

TFUpdater.py
```diff
@@ -105,6 +105,11 @@ def __init__(self, config, network, initial_learning_rate=1.):
     self.optimizer_vars = []  # type: list[tf.Variable]
     self.optimizer_init_vars_op = None  # type: tf.Operation
 
+    # After graph was build: look if it only uses deterministic ops
+    if self.config.is_true('deterministic_train'):
+      if not check_contains_only_det_ops():
+        print("WARNING: Some op used is not deterministic!")
```
albertz: The `print` should use `file=log.v1` also.
But better, maybe the function should return which ops (`list[tf.Operation]`) are not deterministic, and then you should print that list here.

TFUtil.py
albertz: Ok, also see the comment in TFUpdater. Instead of just returning a `bool`, we should maybe better return the list of non-deterministic ops. I.e. the function should be called sth like `get_non_deterministic_ops_from_graph`. And then `rtype: list[tf.Operation]` (not just `list[str]` of the op types, it would be good to know where they are exactly, i.e. `tf.Operation`).

albertz:
Can you rebase the commits into a single commit? Otherwise ok for now.

TFUtil.py
```diff
@@ -6667,3 +6667,25 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+  :return: list of all non deterministic ops names (depending on device and tf version) used in current graph
+  :rtype: list[tf.Operation]
+  """
+  device_types = {device.device_type for device in get_tf_list_local_devices()}
+  non_det_ops = []
+  tf_version = tf_version_tuple()
+  for op in tf.get_default_graph().get_operations():
+    if op.type == "Mean" and tf_version <= (1, 5, 0) and "GPU" in device_types:
+      non_det_ops.append(op)
+    # elif ... more non det ops to be added
```
albertz: Btw, I would also collect all the ops (or op type) which are not yet known (i.e. do an `else:` case in the end and collect all others), and then print a warning about unknown ops (or maybe just op types, would probably be too much otherwise for now).

PR #134, opened 2019-02-10
hdf-dump, fix bug: converting dataset with 'data' as target key

kazuki-irie:
Set "classes" to be the target key after conversion, when the
original dataset uses 'data' as the target key (which is reserved
to be the input key for the HDFDataset).

Add 'labels["delayed"]' in LmDataset.

These fix bug for converting LmDataset to HDFDataset.

### DIFF
--- a/tools/hdf_dump.py
+++ b/tools/hdf_dump.py
@@ -63,11 +63,10 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
     default_data_target_key = data_target_keys[0]
   print("Using target data key:", default_data_target_key)
 
-  if default_data_target_key == "data":
-    valid_hdf_target_key = "classes"  # Replace "data" which is reserved for input key in HDFDataset.
-    assert valid_hdf_target_key not in data_target_keys
-  else:
-    valid_hdf_target_key = default_data_target_key
+  hdf_data_key_map = {key: key for key in data_keys if key != default_data_input_key}
+  if "data" in hdf_data_key_map:
+    hdf_data_key_map["data"] = "classes"  # Replace "data" which is reserved for input key in HDFDataset.
+    assert "classes" not in hdf_data_key_map
 
   # We need to do one run through the dataset to collect some stats like total len.
   print("Collect stats, iterate through all data...", file=log.v3)
@@ -136,14 +135,10 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
     if data_key == default_data_input_key:
       hdf_dataset.create_dataset(
         'inputs', shape=shapes[data_key], dtype=dataset.get_data_dtype(data_key))
-    elif data_key == default_data_target_key:
-      hdf_dataset['targets/data'].create_dataset(
-        valid_hdf_target_key, shape=shapes[data_key], dtype=dataset.get_data_dtype(data_key))
-      hdf_dataset['targets/size'].attrs[valid_hdf_target_key] = dataset.num_outputs[data_key]
     else:
       hdf_dataset['targets/data'].create_dataset(
-        data_key, shape=shapes[data_key], dtype=dataset.get_data_dtype(data_key))
-      hdf_dataset['targets/size'].attrs[data_key] = dataset.num_outputs[data_key]
+        hdf_data_key_map[data_key], shape=shapes[data_key], dtype=dataset.get_data_dtype(data_key))
+      hdf_dataset['targets/size'].attrs[hdf_data_key_map[data_key]] = dataset.num_outputs[data_key]
     if data_key in dataset.labels:
       labels = dataset.labels[data_key]
       if PY3:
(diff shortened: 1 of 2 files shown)


tools/hdf_dump.py
```diff
@@ -145,8 +154,14 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
...
-      for i, label in enumerate(labels):
+      if data_key == default_data_target_key:
+        hdf_dataset['targets/labels'].create_dataset(valid_hdf_target_key,
+                                                     (len(labels),), dtype="S%i" % (max_label_len + 1))
+        for i, label in enumerate(labels):
+          hdf_dataset['targets/labels'][valid_hdf_target_key][i] =\
+            numpy.array(label, dtype="S%i" % (max_label_len + 1))
+      else:
+        hdf_dataset['targets/labels'].create_dataset(data_key, (len(labels),), dtype="S%i" % (max_label_len + 1))
         hdf_dataset['targets/labels'][data_key][i] = numpy.array(label, dtype="S%i" % (max_label_len + 1))
```
albertz: This code line seems strange now. `i` is not defined here.

tools/hdf_dump.py
```diff
@@ -145,8 +154,14 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
     print("Labels for %s:" % data_key, labels[:3], "...", file=log.v5)
     max_label_len = max(map(len, labels))
     if data_key != default_data_input_key:
-      hdf_dataset['targets/labels'].create_dataset(data_key, (len(labels),), dtype="S%i" % (max_label_len + 1))
-      for i, label in enumerate(labels):
+      if data_key == default_data_target_key:
```
albertz: Instead of having this `if` branch here, I think the code would be simpler if you write this:
```
hdf_data_key = hdf_data_key_map[data_key]
```
We need to introduce `hdf_data_key_map` instead of `valid_hdf_target_key`.

tools/hdf_dump.py
```diff
@@ -161,6 +176,8 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
     for data_key in data_keys:
       if data_key == default_data_input_key:
         hdf_data = hdf_dataset['inputs']
+      elif data_key == default_data_target_key:
+        hdf_data = hdf_dataset['targets/data'][valid_hdf_target_key]
```
albertz: This can be removed.

tools/hdf_dump.py
```diff
@@ -161,6 +176,8 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
...
       else:
         hdf_data = hdf_dataset['targets/data'][data_key]
```
albertz: Instead of `data_key` here, you would use `hdf_data_key_map[data_key]`.

kazuki-irie:
Adding labels["delayed"] in LmDataset is actually not relevant to the bug fix here. We could revert this.

tools/hdf_dump.py
```diff
@@ -63,6 +63,12 @@ def hdf_dump_from_dataset(dataset, hdf_dataset, parser_args):
     default_data_target_key = data_target_keys[0]
   print("Using target data key:", default_data_target_key)
 
+  if default_data_target_key == "data":
+    valid_hdf_target_key = "classes"  # Replace "data" which is reserved for input key in HDFDataset.
+    assert valid_hdf_target_key not in data_target_keys
+  else:
+    valid_hdf_target_key = default_data_target_key
```
albertz: You need to introduce `hdf_data_key_map` here. It should be sth like:
```
hdf_data_key_map = {key: key for key in data_keys if key != default_data_input_key}
```
And then you can handle `"data"`. However, there should be extra checks for `"classes"`. E.g. like:
```
if "data" in hdf_data_key_map:
  # Replace "data" which is reserved for input key in HDFDataset.
  hdf_data_key_map["data"] = "classes"
  assert "classes" not in hdf_data_key_map
```

tools/hdf_dump.py
albertz: (Btw, have you used PyCharm? I'm pretty sure it would warn about this.)

albertz:
> Adding labels["delayed"] in LmDataset is actually not relevant to the bug fix here. We could revert this.

It's not relevant, but it also does not hurt, and it's correct, so you can just leave that in.

tools/hdf_dump.py
kazuki-irie: Oh how could I miss this! Thanks!

tools/hdf_dump.py
kazuki-irie: Will change it.

tools/hdf_dump.py
kazuki-irie: OK.

tools/hdf_dump.py
kazuki-irie: Done.

tools/hdf_dump.py
kazuki-irie: Done.

kazuki-irie:
Pushed the requested changes. Please take another look, @albertz. Thank you!

PR #135, opened 2019-02-11
SliceLayer, NCHW support, bug fixes

squarenabla:


albertz:
Some of the tests fail. Please look into that.

TFNetworkLayer.py
```diff
@@ -1933,7 +1933,9 @@ def get_out_data_from_opts(
       out_type["shape"] = list(out_type["shape"])
       if out_type["shape"][axis_wo_batch] is not None:
         out_type["shape"][axis_wo_batch] = len(range(out_type["shape"][axis_wo_batch])[dim_slice])
-      if axis_wo_batch == len(out_type["shape"]) - 1 and not out_type["sparse"]:
+      if "feature_dim_axis" not in out_type:
+        out_type["feature_dim_axis"] = len(out_type["shape"])
```
albertz: This is wrong in general. Also, `Data.__init__` should have all that logic already. Also, it's better to default to `NotSpecified`, and only overwrite it if you specifically want to set it.

albertz: But you are right that the old code was also wrong for cases where feature dim axis is not the default.

TFNetworkLayer.py
```diff
@@ -1933,7 +1933,9 @@ def get_out_data_from_opts(
...
+      if axis_wo_batch == out_type["feature_dim_axis"] - 1 and not out_type["sparse"]:
```
albertz: The calculation is not always correct (only if batch axis is before feature dim axis).

TFNetworkLayer.py
```diff
@@ -1933,7 +1933,9 @@ def get_out_data_from_opts(
...
         assert out_type["shape"][axis_wo_batch]
         out_type["dim"] = out_type["shape"][axis_wo_batch]
```
albertz: Maybe this can be removed, as I think `Data.__init__` might automatically infer it correctly. (Check that, and maybe consider to extend `Data.__init__` if not.

squarenabla:
It's not ready yet

TFNetworkLayer.py
```diff
@@ -1928,12 +1928,15 @@ def get_out_data_from_opts(
     out_type = input_data.get_kwargs()
     out_type["name"] = "%s_output" % name
     axis_wo_batch = input_data.get_batch_axis_excluding_batch(axis)
+    feature_dim_wo_batch = input_data.feature_dim_axis_excluding_batch
+    if feature_dim_wo_batch is None:
+      feature_dim_wo_batch = input_data.get_batch_axis_excluding_batch(len(out_type["shape"]) - 1)
```
albertz: This is wrong. If it is `feature_dim_axis is None`, then for this `Data` instance, it is esp marked that it does not have a feature axis. Maybe you confuse this to `NotSpecified` (but this is only for `feature_dim_axis_or_not_specified`).
You cannot just assume that the last axis might be a feature axis. (Btw, `out_type["shape"]` is without batch, so using `get_batch_axis_excluding_batch` is also wrong here.)

TFNetworkLayer.py
albertz: I mean `out_type["dim"] = out_type["shape"][axis_wo_batch]`.

PR #137, opened 2019-02-12
Loss, NCHW support fixed

squarenabla:
If labels has channels first format, transform to channels last

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -5676,6 +5676,13 @@ class Loss(object):
             target = target.copy_extend_with_beam(output.beam_size)
         else:
           assert not target.beam_size
+      if output.feature_dim_axis is not None and output.feature_dim_axis != output.batch_ndim - 1:
+        if output_with_activation:
+          from TFUtil import move_axis
+          output_with_activation = OutputWithActivation(x=move_axis(output_with_activation.x,
+                                                                    output.feature_dim_axis, -1),
+                                                        act_func=output_with_activation.act_func)
+        output = output.copy_with_feature_dim_axis(-1)
       self.output = output
       self.output_with_activation = output_with_activation
       self.target = target
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -2767,6 +2767,8 @@ def swapaxes(x, axis1, axis2):
                 for i in range(ndim)]
       else:
         perm = list(range(ndim))
+        axis1 = axis1 % len(perm)
+        axis2 = axis2 % len(perm)
         perm[axis1] = axis2
         perm[axis2] = axis1
     else:
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -970,6 +970,47 @@ def test_ReduceLayer_NCHW():
     assert reduce2.output.batch_dim_axis is None
 
 
+def test_Loss_NCHW():
+  with make_scope() as session:
+    import numpy as np
+    net = TFNetwork(extern_data=ExternData())
+    with tf.variable_scope("src_nchw"):
+      src_nchw = InternalLayer(name="src_nchw", network=net, out_type={"dim": 16,
+                                                                       "shape": (16, None),
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 2,
+                                                                       "feature_dim_axis": 1,
+                                                                       "sparse": False
+                                                                       })
+      src_nchw.output.placeholder = tf.placeholder(shape=(None, 16, None), dtype=tf.float32)
+      src_nchw.output.size_placeholder = {1: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    with tf.variable_scope("activation"):
+      activation = ActivationLayer(name="activation", activation="softmax", network=net, sources=[src_nchw])
+
+    target_placeholder = tf.placeholder(shape=(None, None, 16), dtype=tf.float32)
+    target_size_placeholder = tf.placeholder(shape=(None,), dtype=tf.int32)
+    target_data = Data(name="target", shape=(None, 16), placeholder=target_placeholder,
+                       size_placeholder={0: target_size_placeholder},
+                       time_dim_axis=1, feature_dim_axis=2)
+
+    with tf.variable_scope("loss"):
+      loss = CrossEntropyLoss(base_network=net)
+      loss.init(output=activation.output, output_with_activation=activation.output_before_activation,
+                target=target_data, layer=activation)
+
+    random_input = np.random.rand(10, 16, 32)
+    loss_out, out_flat = session.run([loss.get_value(), loss.output_before_softmax_flat],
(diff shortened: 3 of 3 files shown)


TFNetworkLayer.py
```diff
@@ -5468,14 +5468,18 @@ def init_by_layer(self, layer, layer_output_template=None):
...
     :param tf.Tensor x: (B,T,...) or (T,B,...)
     :param tf.Tensor seq_lens: (B,)
     :param bool time_major:
+    :param bool batch_feature_major:
     :return: (B*T|T*B|B',...)
     :rtype: tf.Tensor
     """
+    if batch_feature_major:
+      from TFUtil import swapaxes
+      x = swapaxes(x, 1, 2)
```
albertz: This is wrong in general. You probably mean `-1` instead of `2`.
But also, you are actually swapping it already below, so the whole change here in `_flatten_or_merge` should not be needed?

TFNetworkLayer.py
```diff
@@ -5516,9 +5520,12 @@ def init(self, output, output_with_activation=None, target=None, layer=None):
       if self.output.have_time_axis():
         self.output_seq_lens = output.get_sequence_lengths()
         time_and_batch_dims = (self.output.time_dim_axis, self.output.batch_dim_axis)
-        assert time_and_batch_dims in [(0, 1), (1, 0)], "output time-batch-dim unexpected: %s" % self.output
+        assert time_and_batch_dims in [(0, 1), (1, 0), (2, 0)], "output time-batch-dim unexpected: %s" % self.output
+        if self.output.time_dim_axis == 2:
+          self.output = self.output.copy_with_feature_dim_axis(-1)
```
albertz: If you anyway do this, then why don't you do it right in the beginning, **before** `time_and_batch_dims = (...)`? Then you don't need to change any other code, and it's much simpler. So please do that.

albertz: Also, instead of checking `time_dim_axis == 2`, check for `is_batch_feature_major`.

TFNetworkLayer.py
squarenabla: You mean `x = swapaxes(x, 0, 1)`, which goes below? But it's swapping between time_dim and batch_dim.

albertz: No. I mean what I wrote, I.e. I mean `-1` instead of `2`, i.e. `swapaxes(x, 1, -1)` (which is equivalent to `swapaxes(x, 1, 2)` if `ndim == 3`.

TFNetworkLayer.py
```diff
@@ -5515,10 +5519,13 @@ def init(self, output, output_with_activation=None, target=None, layer=None):
...
       if self.output.have_time_axis():
         self.output_seq_lens = output.get_sequence_lengths()
+        if self.output.is_batch_feature_major:
+          self.output = self.output.copy_with_feature_dim_axis(-1)
         time_and_batch_dims = (self.output.time_dim_axis, self.output.batch_dim_axis)
         assert time_and_batch_dims in [(0, 1), (1, 0)], "output time-batch-dim unexpected: %s" % self.output
         if output_with_activation and output_with_activation.act_func is tf.nn.softmax:
-          self.output_before_softmax_flat = flatten_or_merge(output_with_activation.x, self.output_seq_lens, time_major=output.is_time_major)
+          self.output_before_softmax_flat = flatten_or_merge(output_with_activation.x, self.output_seq_lens,
+                                                             time_major=output.is_time_major, batch_feature_major=output.is_batch_feature_major)
```
albertz: You don't need that.
After you do `self.output.copy_with_feature_dim_axis(-1)` above, you will not have `is_batch_feature_major` anymore. So remove this.

TFNetworkLayer.py
```diff
@@ -5468,14 +5468,18 @@ def init_by_layer(self, layer, layer_output_template=None):
       target=layer._get_target_value(),
       layer=layer)
 
-  def _flatten_or_merge(self, x, seq_lens, time_major):
+  def _flatten_or_merge(self, x, seq_lens, time_major, batch_feature_major=False):
```
albertz: This change here is not needed. See my other comment.

TFNetworkLayer.py
squarenabla: 'self.output.copy_with_feature_dim_axis(-1)' will return the right format. But  'output_with_activation.x' will be still in the old format. So it has to be converted as well.

TFNetworkLayer.py
```diff
@@ -5468,14 +5468,18 @@ def init_by_layer(self, layer, layer_output_template=None):
...
     :param tf.Tensor x: (B,T,...) or (T,B,...)
     :param tf.Tensor seq_lens: (B,)
     :param bool time_major:
+    :param bool batch_feature_major:
     :return: (B*T|T*B|B',...)
     :rtype: tf.Tensor
     """
+    if batch_feature_major:
+      from TFUtil import swapaxes
+      x = swapaxes(x, 1, -1)
```
albertz: I think you should use `move_axis` instead here.

TFNetworkLayer.py
albertz: Yes, but it should not be done only here. Also, `output_with_activation` should be consistent to `output`, always. Otherwise later you do not know anymore, because you just know what `self.output` is.
So, just do it where you do `copy_with_feature_dim_axis`. That makes the code much simpler (you don't need to change `_flatten_or_merge` and care about other problems in other code later).

TFNetworkLayer.py
```diff
@@ -5515,10 +5519,13 @@ def init(self, output, output_with_activation=None, target=None, layer=None):
       self.loss_norm_factor = 1.0
       if self.output.have_time_axis():
         self.output_seq_lens = output.get_sequence_lengths()
+        if self.output.is_batch_feature_major:
+          self.output = self.output.copy_with_feature_dim_axis(-1)
```
albertz: `output_with_activation` should also be changed like this here.

albertz: Esp `self.output_with_activation`.

TFNetworkLayer.py
```diff
@@ -5677,6 +5677,12 @@ def init(self, output, output_with_activation=None, target=None, layer=None):
         else:
           assert not target.beam_size
       self.output = output
+      if self.output.is_batch_feature_major:
+        self.output = self.output.copy_with_feature_dim_axis(-1)
+        if output_with_activation:
+          from TFUtil import move_axis
+          output_with_activation = OutputWithActivation(x=move_axis(output_with_activation.x, 1, -1),
+                                                        act_func=output_with_activation.act_func)
```
albertz: This is slightly inconsistent (you operate on `self.output`, i.e. the attribute, but on `output_with_activation`, i.e. not the attribute but the local var). Why don't you make it consistent? I.e. just do it in both cases on the local var (`output` and `output_with_activation`), before the assignment `self.output = output`.

Also, instead of the check `output.is_batch_feature_major`, you should check `output.feature_dim_axis is not None and output.feature_dim_axis != output.batch_ndim - 1`.

PR #138, opened 2019-02-14
Add option to avoid loading train_data

kazuki-irie:
Does this make sense @albertz, @jan-rosendahl-rwth ?
We should have an option to avoid loading the training data, for example for eval tasks.
@albertz, please suggest if you have a better solution or naming. Thanks!


albertz:
I don't understand why this is needed. Could you not just set `train = None`?

kazuki-irie:
Ah, yes. Solved :) Thanks!

PR #140, opened 2019-02-22
LearningRateControl, alternative relative error

kazuki-irie:
Add 'newbob_relative_error_old_minus_new' option.
If 'newbob_relative_error_old_minus_new' is set to True, compute
relative error as (old - new) / old.
Otherwise (by default), (new - old) / new is used as before.

I would like to have an option to compute the relative change in this
direction and use positive valued lower-bound threshold (isn't this more natural?).
What do you think @albertz ?
This behavior can not be reproduced by the default code, even if you
try to fit the threshold.

Also, in line 176, 
```relativeError = (newError - oldError) / abs(newError)```
I think it is not correct to have "abs" in the denominator in the general case.

### DIFF
--- a/LearningRateControl.py
+++ b/LearningRateControl.py
@@ -40,6 +40,7 @@ class LearningRateControl(object):
                          or config.value('learning_rate_control_error_measure', None),
       "relativeErrorAlsoRelativeToLearningRate": config.bool('learning_rate_control_relative_error_relative_lr', False),
       "minNumEpochsPerNewLearningRate": config.int("learning_rate_control_min_num_epochs_per_new_lr", 0),
+      "relativeErrorDivByOld": config.bool('newbob_relative_error_div_by_old', False),
       "filename": config.value('learning_rate_file', None),
     }
 
@@ -56,12 +57,14 @@ class LearningRateControl(object):
                errorMeasureKey=None,
                relativeErrorAlsoRelativeToLearningRate=False,
                minNumEpochsPerNewLearningRate=0,
+               relativeErrorDivByOld=False,
                filename=None):
     """
     :param float defaultLearningRate: default learning rate. usually for epoch 1
     :param list[float] | dict[int,float] defaultLearningRates: learning rates
     :param str|list[str]|None errorMeasureKey: for getEpochErrorValue() the selector for EpochData.error which is a dict
     :param int minNumEpochsPerNewLearningRate: if the lr was recently updated, use it for at least N epochs
+    :param bool relativeErrorDivByOld: if True, compute relative error as (new - old) / old.
     :param str filename: load from and save to file
     """
     self.epochData = {}  # type: dict[int,LearningRateControl.EpochData]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by albertz** [comment]
This relative error is used in `NewbobRelative` and `NewbobMultiEpoch`. 
The relative error (calculated in `calcRelativeError`) is currently more or less like this:

    relativeError = (newError - oldError) / abs(newError)

The error is e.g. the loss value (cross entropy or so) or frame error rate, i.e. the lower the better.
E.g.:

1.  Let's assume the training improved, e.g. like `oldError = 0.5` and `newError = 0.45`. Then `relativeError = -0.05 / x < 0` (let's ignore the denominator for now).
2. Let's assume the training improved, but not so much, e.g. like `oldError = 0.5` and `newError = 0.49`. Then `relativeError = -0.01 / x < 0`.
3. Let's assume the training did not change the error. Then `relativeError = 0`.
4. Let's assume the error got worse, e.g. like `oldError = 0.5` and `newError = 0.52`. Then `relativeError = 0.02 / x > 0`.

E.g. in `NewbobRelative` we have:

```
  def calcLearningRateForEpoch(self, epoch):
    ...
    relativeError = self.calcRelativeError(last2Epoch, lastEpoch)
    if relativeError is None:
      return learningRate
    if relativeError > self.relativeErrorThreshold:
      learningRate *= self.learningRateDecayFactor
    return learningRate
```

So, if `relativeError > self.relativeErrorThreshold`, then `learningRate *= self.learningRateDecayFactor`.

Let's still ignore the denominator and just assume `x=1`.
E.g.:
1. If you set `relativeErrorThreshold = 0`, then the learning rate will be decayed for case 4 above, but not in the other cases.
2. If you set `relativeErrorThreshold = -1e-10`, then the learning rate will be decayed for case 3 and 4.
3. If you set `relativeErrorThreshold = -0.015`, then the learning rate will be decayed for case 2, 3 and 4.

I think this covers all the cases you are interested in, or not?

With your change, you have it wrong, I think. E.g. if the training improved, you get `relativeError > 0`, and if the training got worse, you get `relativeError < 0`. Then you do the decay if `relativeError > relativeErrorThreshold`, i.e. for the case when it improved.

About the denominator:

1. Yes, you are right, maybe `oldError` would make more sense there. But not sure if this is so important. If you think this is relevant, we maybe can make an option for this.
2. The `abs` is important, though. Otherwise, e.g. if you have some loss which is negative, the logic would be wrong, as this would flip around the sign. And if it is always positive, the `abs` does not matter.
3. Also consider cases where the loss can be both positive and negative (we have that for some of the sequence criteria). And then maybe in some epochs you might also get values very close to 0.

About case 2 and 3 here, I'm anyway not sure if the relative error makes much sense.

kazuki-irie:
You are right. I had somehow `relativeError > self.relativeErrorThreshold` in the wrong direction.
Thanks a lot for your detailed explanation!

As you say, the only thing which makes sense in fact is the division by `oldError`. Can we still add this option? (I still find this more natural). Any suggestion for the option name?

kazuki-irie:
I have pushed the change which only contains the division by `oldError`.
Please take another look. Thank you.

albertz:
Can you cleanup the commits? (Rebase, merge them into a single commit.)

kazuki-irie:
> Rebase into a single commit.

Done.

PR #142, opened 2019-02-25
TFNetworkLayer.py added flag masked_time to BatchNorm layer

squarenabla:
When seq_lens are equal, possible faster computation with masked_time=False.
Some modifications in TFUtil.py, which fix bugs and extend the functionality.
See #88 for previous conversations.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -882,7 +882,8 @@ class LayerBase(object):
                  use_shift=True, use_std=True, use_sample=0.0, force_sample=False,
                  momentum=0.99, epsilon=1e-3,
                  sample_mean=None, sample_variance=None,
-                 gamma=None, beta=None):
+                 gamma=None, beta=None,
+                 masked_time=True):
     """
     :param Data data:
     :param bool use_shift:
@@ -895,6 +896,7 @@ class LayerBase(object):
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool masked_time: flatten and mask input tensor
     :rtype: tf.Tensor
 
     http://arxiv.org/abs/1502.03167
@@ -904,8 +906,12 @@ class LayerBase(object):
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
     with tf.variable_scope("batch_norm"):
-      x = data.get_placeholder_flattened(keep_dims=True)  # shape (time',...)
-      mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      if masked_time:
+        x = data.get_placeholder_flattened(keep_dims=True)
+        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      else:
+        x = data.placeholder
+        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
       if sample_mean is None:
         with self.var_creation_scope():
           sample_mean = self.add_param(tf.Variable(
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -1041,10 +1041,11 @@ class Data(object):
       x.set_shape([None] * self.batch_ndim)
     return x
 
-  def get_axes(self, exclude_time=False, exclude_batch=False):
+  def get_axes(self, exclude_time=False, exclude_batch=False, exclude_feature=False):
     """
     :param bool exclude_time: will filter out the time-axis
     :param bool exclude_batch: will filter out the batch-axis
+    :param bool exclude_feature: will filter out the feature-axis
     :return: list of axes, like `range(len(self.shape))`, calculated with batch dim.
     :rtype: list[int]
     """
@@ -1053,6 +1054,8 @@ class Data(object):
       axes.pop(axes.index(self.time_dim_axis))
     if exclude_batch and self.batch_dim_axis is not None:
       axes.pop(axes.index(self.batch_dim_axis))
+    if exclude_feature and self.feature_dim_axis is not None:
+      axes.pop(axes.index(self.feature_dim_axis))
     return axes
 
   def get_axes_from_description(self, axes):
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -1,6 +1,6 @@
 
 # start: nosetests $this_file --nologcapture
-
+from __future__ import division
 import logging
 logging.getLogger('tensorflow').disabled = True
 import tensorflow as tf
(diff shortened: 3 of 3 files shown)


TFNetworkLayer.py
```diff
@@ -895,6 +897,8 @@ def batch_norm(self, data,
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool masked_time: flatten and mask input tensor
+    :param tf.DType dtype: data type, different from default when testing
```
albertz: `dtype` can also be of type `str`, right? Or probably that is even the usual case which you often have. You should write that, i.e. `:param tf.DType|str dtype:`.

But anyway, I don't quite understand why you need that? It should not have an effect. Also, it's not clear from the documentation and also from the parameter name what this is really used for. Please remove this if this is not really needed.

TFNetworkLayer.py
```diff
@@ -904,8 +908,12 @@ def batch_norm(self, data,
...
     """
     with tf.variable_scope("batch_norm"):
-      x = data.get_placeholder_flattened(keep_dims=True)  # shape (time',...)
-      mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      if masked_time:
+        x = data.get_placeholder_flattened(keep_dims=True)
+        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+      else:
+        x = data.placeholder
+        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
```
albertz: This is not consistent, and thus wrong.
E.g. consider you have `data` with (batch) shape (B, T, H, D).
In the `masked_time` case, you calculate mean/var over (B, T), i.e. the mean/var shape is (1, 1, H, D).
In the other case, you calculate mean/var over (B,T,H), i.e. the mean/var shape is (1,1,1,D).

albertz: The question is also, what do you really want to have here? If you need that new behavior (mean/var over (B,T,H)), then that should be yet another option.
Btw, in both cases (`masked_time` or not), you can use the same `axes` argument, because it should be the same.

TFNetworkLayer.py
```diff
@@ -924,7 +932,7 @@ def batch_norm(self, data,
             trainable=False))
         sample_variance = tf.assign_add(sample_variance, (variance - sample_variance) * momentum)
       # If train or if force_sample, use default use_sample=0.0, otherwise use_sample=1.0.
-      use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
+      use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), dtype) * (use_sample - 1.0)
```
albertz: Why?

TFUtil.py
```diff
@@ -1028,10 +1028,11 @@ def get_placeholder_flattened(self, keep_dims=False):
       ndim -= 1
     if len(dyn_axes) > 1:
       shape = tf.shape(x)
+      shape_as_list = x._shape_as_list()
```
albertz: You are calling a hidden method. You should not do that. Also, why do you need that?

TFUtil.py
```diff
@@ -1028,10 +1028,11 @@ def get_placeholder_flattened(self, keep_dims=False):
...
       x = tf.reshape(
         x,
         [tf.reduce_prod([shape[i] for i in dyn_axes])] +
-        [shape[i] for i in range(ndim) if i not in dyn_axes])
+        [shape_as_list[i] for i in range(ndim) if i not in dyn_axes])
```
albertz: Why? The old could was just the same.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,134 @@ def test_batch_norm_vars():
...
+    net.train_flag = True
+    with tf.variable_scope("src_nhwc"):
+      src_nhwc = InternalLayer(name="src_nhwc", network=net, out_type={"dim": 16,
+                                                                       "shape": (None, 16, 16),
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 1,
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float64)
```
albertz: The `dtype` is wrong. It should be `tf.float32`.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,134 @@ def test_batch_norm_vars():
...
+    assert np.array_equal(np_bn2.astype('f'), out_2.astype('f'))
+    # Manually calculating batch_norm with different seq_lens, having:
+    # Mean_1 = n2 / n1 * Mean_2
+    # Var_1 = n2 / n1 * (Var_2 + Mean_2 ^ 2 (1 - n2 / n1))
+    # bn_1 = (x - Mean_1) * 1 / sqrt(Var_1)
+    # Substituting Mean_1 and Var_1:
+    np_bn1 = (input_data - n2 / n1 * data_mean) * \
+             (1.0 / np.sqrt(n2 / n1 * (data_var + data_mean ** 2 * (1 - n2 / n1))))
+    # Check with tf output.
+    assert np.array_equal(np_bn1.astype('f'), out_1.astype('f'))
```
albertz: I think you rather should use sth like `np.testing.assert_almost_equal` or so.

TFNetworkLayer.py
squarenabla: For some reason numpy and tf calculate mean differently (diff is about e-4 or e-5). The calculations are more accurate with float64 type. So, I need that to do tests and assert on equal results.

albertz: No, you do not need that. Do not assert on equal results anyway. You should use `numpy.testing.assert_almost_equal`.

TFNetworkLayer.py
albertz: Also, I wonder, why did you not catch this error in your test case? That means that your test case is likely broken.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,134 @@ def test_batch_norm_vars():
...
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 1,
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    mean =  tf.constant(np.random.rand(1, 1, 16, 16), name="rand_mean", dtype=tf.float32)
+    variance = tf.constant(np.random.rand(1, 1, 16, 16), name="rand_var", dtype=tf.float32)
```
albertz: You should use `rnd = numpy.random.RandomState(42)` or so initially, such that you get deterministic behavior.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,134 @@ def test_batch_norm_vars():
...
+                                                                       "shape": (None, 16, 16),
+                                                                       "batch_dim_axis": 0,
+                                                                       "time_dim_axis": 1,
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float64)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int64)}
+    mean = tf.constant(np.random.rand(1, 1, 1, 16), name="rand_mean", dtype=tf.float64)
+    variance = tf.constant(np.random.rand(1, 1, 1, 16), name="rand_var", dtype=tf.float64)
```
albertz: This is wrong. It should be of shape (1, 1, 16, 16).

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,134 @@ def test_batch_norm_vars():
...
+    input_data = np.random.rand(10, 11, 16, 16)
+    input_data[2, 5:, :, :] = 0
+    data_mean = np.mean(input_data, axis=(0, 1, 2), keepdims=True, dtype=np.float64)
+    data_var = np.var(input_data, axis=(0, 1, 2), keepdims=True, dtype=np.float64)
```
albertz: This is wrong. The axis to reduce are only (0,1). But see my comment above also on that you have that inconsistent with `masked_time`.

TFNetworkLayer.py
albertz: No, you do not need that. Use `numpy.testing.assert_almost_equal`.

tests/test_TFNetworkLayer.py
squarenabla: No, I don't use batch_norm in 2 dims. Just in the last dim. So, the statistics is gathered over 0, 1 and 2 axes. I'll make another pull request with multiple axes support. There're a few issues, which have to be tested separately. E.g. saving the bn weights in the NCHW and restoring in the NHWC mode (I told you about that problem).

TFNetworkLayer.py
squarenabla: > In the `masked_time` case, you calculate mean/var over (B, T), i.e. the mean/var shape is (1, 1, H, D).
No, the mean/var shape is (1, 1, 1, D), which is a default behavior. `data.get_placeholder_flattened(keep_dims=True)` returns (BxTxH, D) tensor.

TFUtil.py
squarenabla: Technically it was the same, but `[shape[i] for i in range(ndim) if i not in dyn_axes]` consists of tensors and returns None shape for static dims. However it's possible to specify the static dims by gettign the values from `x.get_shape().as_list()`.

TFNetworkLayer.py
albertz: Oh yes, you are right.

TFUtil.py
albertz: No, `shape[i]` is never `None`. But `shape_as_list[i]` could be `None`, so this would be wrong.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,131 @@ def test_batch_norm_vars():
...
+                                                                       "time_dim_axis": 1,
+                                                                       "feature_dim_axis": 3,
+                                                                       "sparse": False
+                                                                       })
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    rnd = np.random.RandomState(42)
+    mean =  tf.constant(rnd.rand(1, 1, 16, 16), name="rand_mean", dtype=tf.float32)
+    variance = tf.constant(rnd.rand(1, 1, 16, 16), name="rand_var", dtype=tf.float32)
```
albertz: Sorry, I was wrong before. This should be the same shape as it is in the `batch_norm` function.

TFUtil.py
squarenabla: `shape[i]` is not `None` but it's a tensor. After `tf.reshape` with a tensor, the resulting dim becomes None even it was defined before.

albertz: Yes, why does that matter? This is the argument for `tf.reshape`, and it must not be `None`.

tests/test_TFNetworkLayer.py
squarenabla: Yes, didn't notice. The shapes here should also be `(1, 1, 1, 16)`.

TFUtil.py
squarenabla: The result of `tf.reshape` with Tensor as an argument gives `None`. Even for `dim`.

squarenabla:
For some reason, the tests fail with python2.7. However everything is ok with python3.6. I'll fix the problem.

tests/test_TFNetworkLayer.py
```diff
@@ -146,6 +146,131 @@ def test_batch_norm_vars():
...
+    # Manually calculating batch_norm and compare to the tf output
+    np_bn2 = (input_data - data_mean) * (1.0 / np.sqrt(data_var))
+    npt.assert_array_almost_equal(np_bn2, out_2, decimal=5)
+    # Manually calculating batch_norm with different seq_lens, having:
+    # Mean_1 = n2 / n1 * Mean_2
+    # Var_1 = n2 / n1 * (Var_2 + Mean_2 ^ 2 (1 - n2 / n1))
+    # bn_1 = (x - Mean_1) * 1 / sqrt(Var_1)
+    # Substituting Mean_1 and Var_1:
+    np_bn1 = (input_data - n2 / n1 * data_mean) * \
+             (1.0 / np.sqrt(n2 / n1 * (data_var + data_mean ** 2 * (1 - n2 / n1))))
```
albertz: `n2`/`n1` is an `int`, so this behaves different in Pyhton 2/3.
You maybe should add `from __future__ import division` in the beginning of the file, then they are always the same.

TFUtil.py
albertz: This is not the result of `tf.reshape`, this is the argument to `tf.reshape`. For that argument, you must not give `None`.

albertz: Maybe you misunderstand me. To make it clear:
Sth like this is allowed:

    tf.reshape(x, (1, 2, 3, some_tensor, 4, 5))

Sth like this is **not** allowed (i.e. `None` is not allowed):

    tf.reshape(x, (1, 2, None, some_tensor, 4, 5))

squarenabla: I know, that this is not a result of `tf.reshape`. My point is: why should we pass a tensor as an argument (smth like `tf.reshape(x, (1, 2, 3, some_tensor, 4, 5))`), when we could explicitly pass a number there (`tf.reshape(x, (1, 2, 3, 34, 4, 5))`))?

albertz: `shape_as_list[i]` can be `None`. You must not pass `None`.

squarenabla: Yes, it can be. But `[shape_as_list[i] for i in range(ndim) if i not in dyn_axes]` doesn't have `None`. because these are such elements, which are `not in dyn_axes`.

albertz: Do not rely on that. That does not always have to be the case. `shape_as_list[i]` could be `None`. E.g. the feature dim can be None, and is not part of `dyn_axes` here. But there could be other cases. Even if this works now, for some of the cases, this might break in the future. Don't write code like that.

PR #149, opened 2019-03-05
Recurrent Audio Feature Prediction with Stop Token

JackTemaki:
This would be my suggestion for enabling speech synthesis with RETURNN, open for discussion on how to solve the issues with the stop token and searching without a beam/choice layer.

### DIFF
--- a/SprintDataset.py
+++ b/SprintDataset.py
@@ -129,6 +129,7 @@ class SprintDatasetBase(Dataset):
     if self.orth_vocab:
       self.num_outputs["orth_classes"] = (self.orth_vocab.num_labels, 1)
     self.num_outputs["orth"] = (256, 1)
+    self.num_outputs['sequence_end'] = (1,2)
     self._base_init()
     # At this point, we are ready for data. In case we don't use the Sprint PythonSegmentOrdering
     # (SprintInterface.getSegmentList()), we must call this at least once.
@@ -330,6 +331,11 @@ class SprintDatasetBase(Dataset):
       assert isinstance(orth, (str, unicode))
       assert "orth_classes" not in targets
       targets["orth_classes"] = numpy.array(self.orth_vocab.get_seq(orth), dtype="int32")
+
+    #build sequence end array
+    sequence_end = numpy.zeros((T,1), dtype="float32")
+    sequence_end[-1][0] = 1
+    targets["sequence_end"] = sequence_end
 
     # Maybe convert some targets.
     if self.target_maps:
--- a/SprintDataset.py
+++ b/SprintDataset.py
@@ -293,7 +293,6 @@ class SprintDatasetBase(Dataset):
     :returns the sorted seq index
     :rtype: int
     """
-
     # is in format (feature,time)
     assert self.num_inputs == features.shape[0]
     T = features.shape[1]
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -1294,8 +1294,13 @@ class _SubnetworkRecCell(object):
 
       common_data_len = None  # used to check whether all extern data have same length
       used_keys = self.net.used_data_keys.copy()
+
       if rec_layer.target:
         used_keys.add(rec_layer.target)  # we always need the target of the recurrent layer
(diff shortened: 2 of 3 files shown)


albertz:
Can you remove the SprintDataset change? We don't need that. This can be done via the network. And it also should be generic to any dataset.

albertz:
Btw, also always cleanup commits (e.g. 4883cea).

TFNetworkRecLayer.py
```diff
@@ -1435,7 +1435,7 @@ def get_choice_source_batches():
         end_template = self.layer_data_templates["end"]
         needed_outputs.add("end")
         assert tf.as_dtype(end_template.output.dtype) is tf.bool
-        assert end_template.output.batch_shape == (None,)  # (batch*beam,)
+        assert end_template.output.batch_shape == (None,) or end_template.output.batch_shape == (None,1)  # (batch*beam,)
```
albertz: Why do you need that?
Also, there should be a space after `,` (please use PyCharm, it will tell you PEP8 / Python code conventions).

tests/test_TFEngine.py
```diff
@@ -536,6 +536,70 @@ def test_engine_rec_subnet_count():
...
+  engine.search(dataset=dataset)
+  print("error keys:")
+  pprint(engine.network.losses_dict)
+  assert engine.network.total_objective is not None
+
+  engine.finalize()
+
+def test_engine_end_layer():
+  check_engine_end_layer()
+
```
albertz: There should be two empty lines between every function (please use PyCharm).

tests/test_TFEngine.py
```diff
@@ -536,6 +536,70 @@ def test_engine_rec_subnet_count():
   engine.finalize()
 
 
+def check_engine_end_layer(extra_rec_kwargs=None):
```
albertz: Why do you need that as an extra function? There is only a single place where you call it.

TFNetworkRecLayer.py
```diff
@@ -1639,17 +1644,29 @@ def maybe_transform(layer):
...
-            end_flag = tf.logical_or(end_flag, self.net.layers["end"].output.placeholder)  # (batch * beam,)
-          with tf.name_scope("dyn_seq_len"):
-            dyn_seq_len = select_src_beams(dyn_seq_len, src_beams=choices.src_beams)
-            dyn_seq_len += tf.where(
-              end_flag,
-              constant_with_shape(0, shape=tf.shape(end_flag)),
-              constant_with_shape(1, shape=tf.shape(end_flag)))  # (batch * beam,)
-            seq_len_info = (end_flag, dyn_seq_len)
+          if choices:
+            #assert choices, "no search choices in layer %r" % self.net.layers["end"]
```
albertz: There should never be commented out code.

SprintDataset.py
```diff
@@ -331,6 +331,11 @@ def addNewData(self, features, targets=None, segmentName=None):
       assert "orth_classes" not in targets
       targets["orth_classes"] = numpy.array(self.orth_vocab.get_seq(orth), dtype="int32")
 
+    #build sequence end array
+    sequence_end = numpy.zeros((T,1), dtype="float32")
+    sequence_end[-1][0] = 1
+    targets["sequence_end"] = sequence_end
```
albertz: In TF `EvalLayer`, sth like:
```
tf.equal(source(0, as_data=True).get_sequence_length(), tf.range(....))
```

albertz: `tf.range` should be of shape (batch=1, time).
`...get_sequence_length()` is of shape (batch,)
Maybe needs `tf.expand_dims`.
Then you get the result of shape (batch,time).

albertz:
There are still a lot of issues with the code. Both syntax, Python/PEP conventions (please check PyCharm or so), but also logical. What is the state here? Do you still need that? E.g. as already discussed, you don't need any change in the dataset.

albertz:
Also, btw, the tests are failing. Please check why that is and fix that.

JackTemaki:
Sorry, i forgot to close this request, it is completely outdated

PR #150, opened 2019-03-05
RNNCellLayer, add get_rec_initial_state_inner method

kazuki-irie:
get_rec_initial_state_inner class method replaces the previous 'make'
method in get_rec_initial_state.
The new method is more flexible and can be used outside RNNCellLayer
(for example in SelfAttentionLayer).

Include a small change in get_state_by_key, which allows flexible output
shape.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -2656,7 +2656,7 @@ class RnnCellLayer(_ConcatInputLayer):
     return self._hidden_state
 
   @classmethod
-  def get_state_by_key(cls, state, key):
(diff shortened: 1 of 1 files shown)


albertz:
Some of the tests fail (ignore TFNativeOp, but all others are relevant). Please check why that is. You can also run them locally, e.g. `python3 tests/test_TFEngine test_rec_subnet_train_t3d`.

albertz:
This should be a single commit.

kazuki-irie:
> This should be a single commit.

Rebased.

PR #151, opened 2019-03-05
SelfAttentionLayer, add initial_state

kazuki-irie:
Adding this primarily to be able to use `SelfAttentionLayer` in step-by-step evaluation graph in C++.

Not fully sure if I have covered other use scenarios related to `initial_state` (e.g., I'm not using `initial_state` in `__init__`): Is there anything else I should have thought about?

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -2660,6 +2660,7 @@ class RnnCellLayer(_ConcatInputLayer):
     """
     :param tf.Tensor|tuple[tf.Tensor]|namedtuple state:
     :param int|str|None key:
+    :param tuple[int|None] shape: Shape of the state.
     :rtype: tf.Tensor
     """
     from tensorflow.python.util import nest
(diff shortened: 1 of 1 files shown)


TFNetworkRecLayer.py
```diff
@@ -4134,12 +4135,16 @@ def get_rec_initial_extra_outputs(cls, batch_dim, rec_layer, num_heads, total_ke
...
       # Assume inside RecLayer.
       # Before, we used a tf.TensorArray.
       # However, that has higher memory consumptions than just using a tensor and concatenating to it.
-      total_value_dim = n_out
       # (batch,heads,time,kv-dim//heads)
-      kv_left = tf.zeros((batch_dim, num_heads, 0, (total_key_dim + total_value_dim) // num_heads))
+      kv_left = RnnCellLayer.get_rec_initial_state_inner(rec_layer=rec_layer, state_key="kv_left",
+                                                         name=kwargs['name'],
+                                                         initial_state=kwargs['initial_state'],
+                                                         initial_shape=(batch_dim, num_heads, 0, kv_dim // num_heads),
```
albertz: `initial_state` is an optional argument, or not? So it would be `kwargs.get('initial_state', None)`.
But better, why do you access `kwargs` at all here? Just make it an explicit arg of the method `get_rec_initial_extra_outputs`, with the same default as in `__init__`.
The same also for `name`.

TFNetworkRecLayer.py
```diff
@@ -2660,6 +2660,7 @@ def get_state_by_key(cls, state, key, shape=None):
     """
     :param tf.Tensor|tuple[tf.Tensor]|namedtuple state:
     :param int|str|None key:
+    :param tuple shape: Shape of the state.
```
albertz: You should specify also what the type is, e.g. `tuple[int]`.

albertz: Or `tuple[int|None]`. Or `tuple[int|tf.Tensor]`. Not sure which one it is here.

TFNetworkRecLayer.py
kazuki-irie: Changed accordingly. Thanks!

TFNetworkRecLayer.py
kazuki-irie: Done. Specified as `tuple[int|None]`

kazuki-irie:
Pushed the requested changes (rebased to one commit).
Thanks a lot!

PR #155, opened 2019-03-19
Added .gz support for segment files in MetaDataset, Docs update

JackTemaki:
The documentation for the dataset combination was taken from Patrick Wilkens pdf, transformed into markdown.

### DIFF
--- a/MetaDataset.py
+++ b/MetaDataset.py
@@ -195,6 +195,9 @@ class MetaDataset(CachedDataset2):
       if seq_list_file.endswith(".pkl"):
         import pickle
         seq_list = pickle.load(open(seq_list_file, 'rb'))
+      elif seq_list_file.endswith(".gz"):
+        import gzip
+        seq_list = gzip.open(seq_list_file, "rt").read().splitlines()
       else:
         seq_list = open(seq_list_file).read().splitlines()
     else:
--- a/docs/dataset_reference/combination.rst
+++ b/docs/dataset_reference/combination.rst
@@ -1,0 +1,92 @@
+.. _dataset_combination:
+
+===================
+Dataset Combination
+===================
+
+
+Meta Dataset
+------------
+
+The MetaDataset (API: :ref:`meta_dataset`) is to be used in the case of **Multimodality**.
+Here, the datasets are expected to describe different features of the **same training sequences**.
+These features will all be available to the network at the same time.
+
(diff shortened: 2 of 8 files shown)


MetaDataset.py
```diff
@@ -195,6 +195,10 @@ def _load_seq_list(self, seq_list_file=None):
       if seq_list_file.endswith(".pkl"):
         import pickle
         seq_list = pickle.load(open(seq_list_file, 'rb'))
+      elif seq_list_file.endswith(".gz"):
+        import gzip
+        # needs strip, as gzip returns sequences with \n at end
```
albertz: `readlines` always returns with `"\n"`, also for normal files.
You could do `read().splitlines()` instead, which will not have that `"\n"`. Not sure which is faster. Anyway, the comment is slightly confusing.

albertz:
Also, I would have split the doc update into a separate commit.

JackTemaki:
The API documentation is difficult to browse, I think the aim of the docs should be to provide sufficient information on how to build an experiment without having to look into the code directly.

albertz:
I thought that you already started in building up a separate more structured index which just references the classes, and then includes their docstrings?
Why not do the same for the datasets?

patrick-wilken:
@JackTemaki Thanks for converting! As you skipped the definitions of "Multimodality", "Multi-Task Learning" and "Combination of Corpora" maybe it looks weird now that those words are bold and upper-case, but that's a minor thing.

@albertz One purpose of this doc is to clarify when to use which dataset. This information belongs at some higher level than the dataset itself, I guess. But I agree that ideally the description of the parameters should only go into the doc string. Is there some good way to format this? I haven't worked with the documentation yet...

albertz:
I would say that an overview over available datasets should be separate (but maybe just the docstring of the module?). But some documentation about some specific class, and also its use cases and usage examples, belong to the docstring of the class itself.

In all cases, i.e. docstring, and separate documentation, it uses reStructuredText.

JackTemaki:
If we could move the dataset files into a folder hierarchy, it would be possible to define module docstrings on a hierarchical level, and then we can just import those from the documentation .rst. I would suggest moving all datasets files into a "datasets" folder, and then defining module-docstrings for the folder __init__.py and in the respective dataset files.

JackTemaki:
@patrick-wilken Sorry that i forgot the introduction, i will add it later

PR #160, opened 2019-03-25
Added HDF support for cases without a second input (target)

ArneNx:
Some small modifications to `HDFDataset` and `dump_from_dataset` to make it work with data that does not have a target key. 
So this works e.g. if you use `HDFDataset` on a `LMDataset` with delayed sequence disabled. 
I need this for masked language modeling and others might find this useful as well.

### DIFF
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -69,7 +69,7 @@ class HDFDataset(CachedDataset):
       self.labels = {
         k: [self._decode(item) for item in fin["targets/labels"][k][...].tolist()]
         for k in fin['targets/labels']}
-    if not self.labels:
+    if not self.labels and 'labels' in fin:
       labels = [item.split('\0')[0] for item in fin["labels"][...].tolist()]; """ :type: list[str] """
       self.labels = {'classes': labels}
       assert len(self.labels['classes']) == len(labels), "expected " + str(len(self.labels['classes'])) + " got " + str(len(labels))
@@ -951,9 +951,13 @@ class HDFDatasetWriter:
     print("Data keys:", data_keys, file=log.v3)
     if "orth" in data_keys:  # special workaround for now, not handled
       data_keys.remove("orth")
-    data_target_keys = [key for key in dataset.get_target_list() if key in data_keys]
+
+    if len(data_keys) > 1:
+      data_target_keys = [key for key in dataset.get_target_list() if key in data_keys]
+    else:
+      data_target_keys = []
     data_input_keys = [key for key in data_keys if key not in data_target_keys]
-    assert len(data_input_keys) > 0 and len(data_target_keys) > 0
+
     if len(data_input_keys) > 1:
       if "data" in data_input_keys:
         default_data_input_key = "data"
@@ -962,14 +966,17 @@ class HDFDatasetWriter:
     else:
       default_data_input_key = data_input_keys[0]
     print("Using input data key:", default_data_input_key)
-    if len(data_target_keys) > 1:
-      if "classes" in data_target_keys:
-        default_data_target_key = "classes"
+    if len(data_target_keys) > 0:
+      if len(data_target_keys) > 1:
+        if "classes" in data_target_keys:
+          default_data_target_key = "classes"
+        else:
+          raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
       else:
-        raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
+        default_data_target_key = data_target_keys[0]
+      print("Using target data key:", default_data_target_key)
     else:
-      default_data_target_key = data_target_keys[0]
-    print("Using target data key:", default_data_target_key)
+      default_data_target_key = default_data_input_key
 
     hdf_data_key_map = {key: key for key in data_keys if key != default_data_input_key}
     if "data" in hdf_data_key_map:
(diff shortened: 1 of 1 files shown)


HDFDataset.py
```diff
@@ -965,9 +965,13 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
     print("Data keys:", data_keys, file=log.v3)
     if "orth" in data_keys:  # special workaround for now, not handled
       data_keys.remove("orth")
-    data_target_keys = [key for key in dataset.get_target_list() if key in data_keys]
+
+    if len(data_keys) > 1:
```
albertz: Why is this needed?
Also, isn't this wrong?

HDFDataset.py
```diff
@@ -1039,15 +1047,18 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
         progress_bar_with_time(float(i) / num_seqs)
 
     print("Set seq len info...", file=log.v3)
-    hdf_dataset.create_dataset(attr_seqLengths, shape=(num_seqs, 2), dtype="int32")
+    hdf_dataset.create_dataset(attr_seqLengths, shape=(num_seqs, 1+len(data_target_keys)), dtype="int32")
```
albertz: Please write PEP conform code (spacing etc; use PyCharm or any proper editor).

HDFDataset.py
```diff
@@ -1039,15 +1047,18 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
-      if targets_len is None:
-        targets_len = data_len
-      hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
+      if len(data_target_keys) > 0:
+        targets_len = seq_len[default_data_target_key]
+        for data_key in data_target_keys:
+          assert seq_len[data_key] == targets_len, "different lengths in multi-target not supported"
+        if targets_len is None:
+          targets_len = data_len
+        hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
```
albertz: This code is wrong now. E.g. if `len(data_target_keys) == 2`, then you must set 3 lengths here.

HDFDataset.py
```diff
@@ -1039,15 +1047,18 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
-      targets_len = seq_len[default_data_target_key]
-      for data_key in data_target_keys:
-        assert seq_len[data_key] == targets_len, "different lengths in multi-target not supported"
-      if targets_len is None:
-        targets_len = data_len
-      hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
+      if len(data_target_keys) > 0:
+        targets_len = seq_len[default_data_target_key]
+        for data_key in data_target_keys:
+          assert seq_len[data_key] == targets_len, "different lengths in multi-target not supported"
```
albertz: This is actually not needed anymore with your change of the `attr_seqLengths` shape above.

HDFDataset.py
```diff
@@ -1039,15 +1047,18 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
+        if targets_len is None:
+          targets_len = data_len
+        hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
+      else:
+        hdf_dataset[attr_seqLengths][i] = [data_len]
```
albertz: Do you actually need the `else` case? If you fix the code, probably it should directly work in all cases.

HDFDataset.py
ArneNx: I don't see why this should be wrong.
If you only have one data_key, you want it to go into data_input_keys, isn't that right?

HDFDataset.py
ArneNx: That's interesting. I did not intend to do anything here except enclosing it into an if block. I think something got messed up in a merge. I'll change it back.

HDFDataset.py
albertz: Ah, sorry, you are right.

HDFDataset.py
```diff
@@ -976,14 +980,18 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
+        else:
+          raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
       else:
-        raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
+        default_data_target_key = data_target_keys[0]
+      print("Using target data key:", default_data_target_key)
     else:
-      default_data_target_key = data_target_keys[0]
-    print("Using target data key:", default_data_target_key)
+      default_data_target_key = default_data_input_key
```
albertz: If you do that, maybe you also should add it to `data_target_keys`? Not sure if this is relevant.

HDFDataset.py
albertz: It's wrong because you changed the seqLength shape (which before was always 2 items).

ArneNx: which I thought was the intended behaviour, but nvm. it should be back to the original version now.

albertz: Actually I think the shape `(num_seqs, 1 + len(data_target_keys))` is more correct. When I wrote that, I thought that the HDFDataset always expected shape[-1] in {1,2}, but only later I noticed that it is also fine for multiple targets.

HDFDataset.py
ArneNx: Which change exactly are you talking about? Is this still relevant in the updated commit?

albertz: The check that all the seq lens are the same. This was needed for the old shape of seq lengths.

albertz: Ping?

ArneNx: I'm sorry, I totally forgot about this. 
I don't have any time to do this properly right, but hopefully after I handed in my thesis.

albertz: Why is that `assert` still needed? It's actually wrong, isn't it?

HDFDataset.py
```diff
@@ -962,14 +966,17 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
+    if len(data_target_keys) > 0:
+      if len(data_target_keys) > 1:
+        if "classes" in data_target_keys:
+          default_data_target_key = "classes"
+        else:
+          raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
       else:
-        raise Exception("not sure which target data key to use from %r" % (data_target_keys,))
+        default_data_target_key = data_target_keys[0]
+      print("Using target data key:", default_data_target_key)
```
albertz: Why is the `print` only here now?
Btw, do we even need `default_data_target_key` now with the change?

albertz:
There are conflicts now. Also, did you see my comments? Are you planning to fix this? Otherwise I will close this for now.

albertz:
Closing this now. Please reopen if you plan to update this.

PR #161, opened 2019-03-27
added scheduled sampling for regression-type outputs

JackTemaki:


TFNetworkRecLayer.py
```diff
@@ -3268,20 +3268,29 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+          samples = tf.where(
+            tf.less(tf.random_uniform(tf.shape(samples)), self.scheduled_sampling.get("gold_mixin_prob")),
+            gold_targets, samples)
+      else:
         gold_targets = self._get_target_value().get_placeholder_as_batch_major()  # (batch,), int32
-        samples = tf.where(
-          tf.less(tf.random_uniform(tf.shape(samples)), self.scheduled_sampling.get("gold_mixin_prob")),
-          gold_targets, samples)
+        choice = tf.distributions.Bernoulli(probs=self.scheduled_sampling.get("gold_mixin_prob"),dtype=tf.bool).sample(sample_shape=tf.shape(scores_in)[0])
+        samples = tf.where(choice, gold_targets, scores_in)
```
albertz: This code violates PEP8. (Spacing.)
Btw, isn't it possible to share the same code with the code above, for `samples`?
Also, this code should only run if `self.scheduled_sampling.get("gold_mixin_prob")` is actually set (like in the code above).

TFNetworkRecLayer.py
```diff
@@ -3268,20 +3268,29 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+
       self.output = Data(
         name="%s_sampled_output" % self.name,
         batch_dim_axis=0,
         shape=self.output.shape,
-        sparse=True,
+        sparse=not input_type == "regression",
```
albertz: This is against Python convention. It should be `input_type != "regression"`. (Btw, I think PyCharm warns about this, or not?)

TFNetworkRecLayer.py
JackTemaki: The only line that could be shared is the gold_targets = ... line, I think there is no convenient way to share it

albertz: Also the `tf.where`, or not?

TFNetworkRecLayer.py
```diff
@@ -3268,20 +3268,32 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+        samples = tf.to_int32(tf.reshape(samples, [-1]))  # (batch,), int32
+        if self.scheduled_sampling.get("gold_mixin_prob"):
+          gold_targets = self._get_target_value().get_placeholder_as_batch_major()  # (batch,), int32
+          samples = tf.where(
+            tf.less(tf.random_uniform(tf.shape(samples)), self.scheduled_sampling.get("gold_mixin_prob")),
+            gold_targets, samples)
+      else:
+        if self.scheduled_sampling.get("gold_mixin_prob"):
+          gold_targets = self._get_target_value().get_placeholder_as_batch_major()  # (batch,), int32
+          choice = tf.distributions.Bernoulli(probs=self.scheduled_sampling.get("gold_mixin_prob"), dtype=tf.bool).sample(sample_shape=tf.shape(scores_in)[0])
```
albertz: `choice` here is the same as `tf.less(tf.random_uniform(...` above.

TFNetworkRecLayer.py
```diff
@@ -3268,20 +3268,32 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+          samples = tf.where(choice, gold_targets, scores_in)
```
albertz: This `tf.where` is the same as the `tf.where` above.

JackTemaki:
Is this now ready to merge?

albertz:
Did you see my latest comments? I think this can be simplified.

JackTemaki:
I agree that the code can be shortened, but then it is much less readable

``` python
      if input_type is not "regression":
        samples = tf.multinomial(scores_in, num_samples=1, seed=get_random_seed())  # (batch, num_samples), int64
        samples = tf.to_int32(tf.reshape(samples, [-1]))  # (batch,), int32
      if self.scheduled_sampling.get("gold_mixin_prob"):
        gold_targets = self._get_target_value().get_placeholder_as_batch_major()  # (batch,), int32
        if input_type is not "regression":
          choice = tf.less(tf.random_uniform(tf.shape(samples)), self.scheduled_sampling.get("gold_mixin_prob"))
        else:
          choice = tf.distributions.Bernoulli(probs=self.scheduled_sampling.get("gold_mixin_prob"), dtype=tf.bool).sample(sample_shape=tf.shape(scores_in)[0])
        samples = tf.where(choice, gold_targets, scores_in)
      elif input_type is not "regression":
        samples = scores_in
```

I would even go as far as to replace the variable samples, because in the case of regression the variable does not represent samples.

And when you only exclude the gold_targets, you will get a maybe not initialized warning, or adding `gold_targets = None`

JackTemaki:
Also, even if we unify choice, then we would have to split at shape

albertz:
I don't understand. `choice` should be the same in both cases. It is the probability that you take either the ground truth or the prediction. I find it quite confusing that the code looks as if you do sth different there in both cases. Also the shape for `choice` should be the same in both cases.

Btw, instead of `if ... is not ...: ... else:`, it is often easier to read the code if the positive (not negated) case comes first. Also, `is not` is wrong. That checks whether it is the same object. It should be `==` or `!=`. Sth like:

    if input_type == "regression":
        samples = scores_in  # (batch,dim), float32
    else:
        samples = tf.multinomial(scores_in, num_samples=1, seed=get_random_seed())  # (batch, num_samples), int64
        samples = tf.to_int32(tf.reshape(samples, [-1]))  # (batch,), int32

And then:

    if self.scheduled_sampling.get("gold_mixin_prob"):
        choice = tf.less(tf.random_uniform(tf.shape(scores_in)[:-1]), self.scheduled_sampling.get("gold_mixin_prob"))
        samples = tf.where(choice, gold_targets, samples)

JackTemaki:
I tried to rewrite it completely, how about this? Also, TF uses random_uniform for the Bernoulli internally, so there is no benefit in using it.

TFNetworkRecLayer.py
```diff
@@ -3346,23 +3346,33 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
         raise Exception("%r: invalid input type %r" % (self, input_type))
-      samples = tf.multinomial(scores_in, num_samples=1, seed=get_random_seed())  # (batch, num_samples), int64
-      samples = tf.to_int32(tf.reshape(samples, [-1]))  # (batch,), int32
-      if self.scheduled_sampling.get("gold_mixin_prob"):
+
+      if input_type == "regression":
+        feedback_output = scores_in
+      else:
+        # sample from scores
+        samples = tf.multinomial(scores_in, num_samples=1, seed=get_random_seed())  # (batch, num_samples), int64
```
albertz: Call it also `feedback_output` here.

TFNetworkRecLayer.py
```diff
@@ -3346,23 +3346,33 @@ def __init__(self, beam_size, input_type="prob", explicit_search_source=None, le
...
+        feedback_output = tf.to_int32(tf.reshape(samples, [-1]))  # (batch,), int32
+
+      gold_mixing_prob = self.scheduled_sampling.get("gold_mixin_prob", False)
+      if gold_mixing_prob:
         gold_targets = self._get_target_value().get_placeholder_as_batch_major()  # (batch,), int32
-        samples = tf.where(
-          tf.less(tf.random_uniform(tf.shape(samples)), self.scheduled_sampling.get("gold_mixin_prob")),
-          gold_targets, samples)
+        choice_shape = tf.shape(feedback_output) if input_type is not "regression" else (tf.shape(feedback_output)[0],)
```
albertz: `is not` is wrong. Also, why not just `tf.shape(scores_in)[:-1]`. It's simpler and shorter. Also I would not introduce the local var for it but just have it in `choice = ...` directly.

JackTemaki:
but wouldn't it be (tf.shape(feedback_output)[0],), because tf.where only accepts a single dimension when the choice does not match the inputs exactly?

albertz:
The current version of `tf.where`, yes. But I still hope that they fix that. Anyway, you will get an exception in that case, and maybe with later TF versions, it's fine.
However, with your code, if there are other dimensions, it would also fail with some exception (shape mismatch or so), and fail even with later TF versions.

JackTemaki:
But why would it fail in my case, when the choice is only made across the batch? The variable is stored as batch major, so it will never fail as long as the dimensions of the target and prediction are the same. Or do I misunderstand something there? I do not see the case where my code would fail for arbitrary shapes right now.

albertz:
If e.g. the shape of scores_in is (batch,something,feature|dim), and something is maybe e.g. time, or something else, I would have thought that you want to have a separate sampling for each dimension in the other axis.
But maybe you are right, maybe the better default is to just have it separate over the batch axis and no other axis. So then it would be `tf.shape(scores_in)[:1]` (or `samples` instead of `scores_in`, does not matter).

albertz:
I might have waited before the checks finished, but I think it's ok. :)

PR #165, opened 2019-04-04
add all layer targets to used_data_keys in SubnetworkRecCell

JackTemaki:
Used keys is currently lacking any target that was specified in layers of the subnetwork. The subnetwork copies all targets from the parent network, and adds the defined target, but no additional targets.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -437,6 +437,13 @@ class LayerBase(object):
           # Not resolving this in the dict as target, but call get_layer to make it available.
           if target.startswith("layer:"):
             target_layers[target] = get_layer(target[len("layer:"):])
+          else:
+            # Note: This is a workaround for cases where we need to know about used data keys before the layer
+            # itself is constructed (e.g. in _SubnetworkRecCell.get_output).
+            # A nicer solution would be to not modify this here,
+            # but instead lazily handle it in TFNetwork.get_extern_data,
+            # such that we do not need to know in advance which data keys we need.
+            network.used_data_keys.add(target)
     if "n_out" not in d and targets and network.eval_flag:
       # Must be done here now because loss might be set to None later.
       target = targets[0]  # guess using first target
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -3612,13 +3612,11 @@ class ChoiceLayer(LayerBase):
     """
     if isinstance(d["target"], str):
       d["target"] = [d["target"]]
-    if not network.search_flag:
-      network.used_data_keys.update(d["target"])
-      if not d.get("scheduled_sampling"):
-        # In the dependency graph, we don't want it.
-        # This can enable some optimizations in the RecLayer.
-        # We do it here because we should know about the deps early in the template creation in RecLayer.
-        d["from"] = []
+    if not network.search_flag and not d.get("scheduled_sampling"):
+      # In the dependency graph, we don't want it.
+      # This can enable some optimizations in the RecLayer.
+      # We do it here because we should know about the deps early in the template creation in RecLayer.
+      d["from"] = []
     if d.get("explicit_search_source"):
       d["explicit_search_source"] = get_layer(d["explicit_search_source"]) if network.search_flag else None
     super(ChoiceLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)


patrick-wilken:
For which specific layer do you need this?
I think marking targets as used should be the job of the layers, because only they know which targets they actually use for given search flags, etc.
But I addressed the same issue here: https://github.com/rwth-i6/returnn/pull/113#discussion_r250703451 The problem is, that inside the SubnetworkRecCell we need the used targets before creating the actual layers. So I added the target in transform_config_dict, because there we already have the necessary information to decide whether the target is going to be used. But this also doesn't feel optimal...

albertz:
Yes, I think @patrick-wilken is right. This (#113) is already merged. And `network` there is the subnetwork, right? So it means it already should be in `self.net.used_data_keys`? So then why is this code here needed at all? Or if it is not in `self.net.used_data_keys`, where did it end up? Why does this not work yet?

JackTemaki:
The layer I use in the recurrent unit is:
'stop_token': {'class': 'linear', 'activation': None, 'n_out': 1, 'loss': 'bin_ce', 'loss_scale': 0.2, 'target': 'sequence_end', 'from': ['decoder_2', 'att0']}

'sequence_end' is in my case part of the dataset.

The error message I get is that the variable is not part of my feed dict, so not available as input. 
When I run 'print(used_keys)', only 'data' is inside, so it is definitely missing.

albertz:
I think you should debug why it is not in `used_keys`. See the linked code by @patrick-wilken. That should add it, I think.

patrick-wilken:
Well, this was only for the choice layer. If the target of the linear layer is never used in a choice layer than it won't be added currently.

albertz:
Ah, right. I think this should not be in the `ChoiceLayer`. I think we can just put it into `LayerBase.transform_config_dict`, where we handle `"target"`, or not?

JackTemaki:
something like this?

``` python
   if d.get("target", None):
      targets = d["target"]
      # we might have multiple targets, e.g. in choice layer, so convert to list
      if isinstance(targets, str):
        targets = [targets]
      if network.eval_flag:
        # _target_layers is a small workaround for further code which might not have access to the right get_layer.
        d["_target_layers"] = target_layers
        for target in targets:
          assert isinstance(target, str)
--------->network.used_data_keys.add(target)
          # Not resolving this in the dict as target, but call get_layer to make it available.
          if target.startswith("layer:"):
            target_layers[target] = get_layer(target[len("layer:"):])
```

albertz:
Yea, probably only if not startswith "layer:". Also not sure if this potentially could break sth, but let's see if the tests run through.

albertz:
Also, the corresponding code in `ChoiceLayer` can be removed then.

JackTemaki:
before or after the `if network.eval`?

JackTemaki:
maybe like this? It is running in my setup (training)

TFNetworkRecLayer.py
```diff
@@ -1408,6 +1408,7 @@ def get_output(self, rec_layer):
       used_keys = self.net.used_data_keys.copy()
       if rec_layer.target:
         used_keys.add(rec_layer.target)  # we always need the target of the recurrent layer
+
```
albertz: Maybe remove this new line.

TFNetworkLayer.py
```diff
@@ -437,6 +437,8 @@ def transform_config_dict(cls, d, network, get_layer):
           # Not resolving this in the dict as target, but call get_layer to make it available.
           if target.startswith("layer:"):
             target_layers[target] = get_layer(target[len("layer:"):])
+          else:
+            network.used_data_keys.add(target)
```
albertz: Maybe add a comment, why we do this.

albertz:
Yes. See my comment. And also let's wait for the test results.

patrick-wilken:
The question is how smart we want to be. For layers not inside a subnetwork the data key is only marked used if accessed via TFNetwork.get_extern_data(), which can depend on arbitrary things inside the layer. Setting it in the LayerBase is much less precise, but obviously works.

JackTemaki:
I directly walked into the next issue. It is working fine for the external data, but not if I replace it by a layer. If I set 'target': 'layer:base:stop_token_data', I get a layer not found error although it searches in the network root.

albertz:
Yes, this is a workaround. And this should be clearly written in the comment.

A nicer solution maybe is this:
We overwrite `get_extern_data` of the subnetwork somehow. And there we lazily get it from the parent network. That also means to lazily create the corresponding `tf.TensorArray` (and make sure that it will be outside the loop, which you can do via `tf.control_dependencies(None)`).
Not sure how much changes this involves, and if I forget sth now.
So maybe the workaround is fine for now.

albertz:
> I directly walked into the next issue. It is working fine for the external data, but not if I replace it by a layer. If I set 'target': 'layer:base:stop_token_data', I get a layer not found error although it searches in the network root.

I think this is a separate issue, so let's discuss this in a separate issue. (But I wonder about it. There is `if target.startswith("layer:"): ... get_layer(target[len("layer:"):])` in the same function, which should handle this? Or maybe not...)

TFNetworkLayer.py
albertz: The comment also should explain why this is a workaround, and what probably a better solution would be.

JackTemaki: can you provide a text? because I do not fully understand neither the problem nor the possible fix

albertz: Make sure that these do not exceed 120 chars.

> Note: This is a workaround for cases where we need to know about used data keys before the layer itself is constructed
> (e.g. in `_SubnetworkRecCell.get_output`).
> A nicer solution would be to not modify this here,
> but instead lazily handle it in `TFNetwork.get_extern_data`,
> such that we do not need to know in advance which data keys we need.

albertz:
Hm, this currently breaks some code of search. I wonder why none of the test cases cover this. Maybe I have some more special case (I use search during training).
I'm investigating now.

albertz:
Ok, I have a test case. The problem is, `ChoiceLayer` should not mark this as a used target if it is doing search. I still wonder why this does not break more tests, and the search in general.

albertz:
Ok, this was not so difficult. But the code feels even more hacky/ugly now. I think this should be fixed in the way I suggested in the comment. Maybe I do that when I have some time for it.

PR #167, opened 2019-04-08
SprintControl.py: convert segment name from bytes to string

pavelgolik:
When building RASR with Python 3, the segment name is encoded as bytes, so we need to decode it into string. It is not guaranteed, but in 100% of corpus files I've worked with, the segment name is pure ascii.

This PR should allow to keep using RASR binaries built with Python 2 as well as to switch to Python 3.

### DIFF
--- a/SprintControl.py
+++ b/SprintControl.py
@@ -443,7 +443,7 @@ class PythonControl:
     with self.cond:
       self.control_thread__have_new_seg = True
       self.control_thread__have_new_error_signal = False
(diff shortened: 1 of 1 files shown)


albertz:
Why not just use utf8? Should work probably with all the ASCII files, but I guess also with everything else what you will ever have (as that should be utf8, or not?).

albertz:
Also, can you clean up the commits (merge into a single commit)?

PR #168, opened 2019-04-09
determinism test

tbscode:


albertz:
When you make a pull request, can you clean up the commits? (In this case, merge them together into a single commit.)

tests/test_TF_determinismus.py
```diff
@@ -0,0 +1,66 @@
+# tests if the graph values of two different training runs on GPU are equal
+# make sure GPU is available
+# in some cases the test will give a false negative
```
albertz: Make this a docstring instead.

tests/test_TF_determinismus.py
```diff
@@ -0,0 +1,66 @@
...
+
+def load_data():
+  from rnn import load_data
+  dev_data, _ = load_data(config, 0, 'dev', chunking="200:200", seq_ordering="sorted", shuffle_frames_of_nseqs=0)
+  eval_data, _ = load_data(config, 0, 'eval', chunking="200:200", seq_ordering="sorted", shuffle_frames_of_nseqs=0)
+  train_data, _ = load_data(config, 0, 'train')
+  return dev_data, eval_data, train_data
+
+
+def test_determinism_of_vanillalstm():
```
albertz: Add a docstring for each test, and document your results on the test. Sth like this:
"With GPU (or CPU), TF x.y, this runs deterministic."

albertz:
As I understood you, in all configurations you tried (which ones exactly? please document in the code), this is deterministic now, right?
But this means there was no other change in RETURNN needed anymore? Last time I thought that you said there is still sth non-deterministic?
If this is all deterministic now: Go on to some other example. Maybe native LSTM (v1 or v2), same demo. And after that, some of the attention examples.
When you catch something non-deterministic: Try to find out exactly what it is, extend `get_non_deterministic_ops_from_graph`, and extend the option `deterministic_train` with alternative code which makes it deterministic.

albertz:
Can you rebase on the latest master?

albertz:
Also, can you mark my comments as resolved when you addressed them?

tests/test_TF_determinismus.py
```diff
@@ -0,0 +1,95 @@
...
+
+  def train_engine_fetch_vars(engine):
+    data_provider = engine._get_new_data_provider(dataset=engine.train_data, batches=engine.train_batches)
+    feed_dict, _ = data_provider.get_feed_dict(single_threaded=True)
+    trainer = Runner(engine=engine, dataset=engine.train_data, batches=engine.train_batches, train=True)
+    feed_dict, _ = data_provider.get_feed_dict(single_threaded=True)
+    trainer.run(report_prefix="One Run")
+    return [x.eval(session=engine.tf_session)
+            for x in (tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope='fw0') +
+                      tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope='output'))]
```
albertz: This should be more generic. You should get all variables.
(You can also use `engine.network` for that.)

tests/test_TF_determinism.py
```diff
@@ -0,0 +1,93 @@
...
+    data_provider = engine._get_new_data_provider(dataset=engine.train_data, batches=engine.train_batches)
+    feed_dict, _ = data_provider.get_feed_dict(single_threaded=True)
+    trainer = Runner(engine=engine, dataset=engine.train_data, batches=engine.train_batches, train=True)
+    feed_dict, _ = data_provider.get_feed_dict(single_threaded=True)
+    trainer.run(report_prefix="One Run")
+
+    # all variables of all layers
+    return [[x.eval(session=engine.tf_session)
+             for x in (tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES, scope=layer_name))]
+            for layer_name in engine.network.layers]
```
albertz: This is slightly wrong. It should be `scope=LayerBase.cls_get_tf_scope_name(layer_name)`.

But also, this is too complicated.
Why don't you just use `network.get_params_list()` or so?
Or maybe `network.get_param_values_dict()`.

albertz:
Sorry for late reply. I'm merging this in now. I will write you a mail with next steps.

albertz:
Btw, I squashed it myself now, as this all was basically one single change. But next time, it would be good if you can cleanup the commits yourself (merge multiple commits into a single one, if they basically change the same thing).

PR #169, opened 2019-04-10
added zoneout lstm cell from tacotron-2

JackTemaki:


TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+
+class ZoneoutLSTMCell(BaseRNNCell):
+  '''Wrapper for tf LSTM to create Zoneout LSTM Cell
+
+  inspired by:
+  https://github.com/teganmaharaj/zoneout/blob/master/zoneout_tensorflow.py
+
+  Published by one of 'https://arxiv.org/pdf/1606.01305.pdf' paper writers.
+
+  Many thanks to @Ondal90 for pointing this out. You sir are a hero!
```
albertz: This comment is from you, or from Tacotron?
Better adapt this, and add a reference to the related Tacotron code.
Also, use `"""` instead of `'''` (keep the code consistent; also, please check PyCharm warnings).

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+  '''
+
+  def __init__(self, num_units, zoneout_factor_cell=0., zoneout_factor_output=0., state_is_tuple=True, name=None, **kwargs):
+    '''Initializer with possibility to set different zoneout values for cell/hidden states.
+    '''
```
albertz: Use `"""`. Keep the style consistent.
Also, please document all arguments, including specifying their types (see other code for example).

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+    if kwargs.get('is_training', None) is None:
+      from TFNetwork import TFNetwork
+      self.is_training = TFNetwork.get_current_network().train_flag
+    else:
+      self.is_training = kwargs['is_training']
+
+    self.state_is_tuple = state_is_tuple
+
+  @property
+  def state_size(self):
```
albertz: This function is not documented.

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+    return self._cell.state_size
+
+  @property
+  def output_size(self):
```
albertz: This function is not documented.

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+    return self._cell.output_size
+
+  def __call__(self, inputs, state, scope=None):
+    '''Runs vanilla LSTM Cell and applies zoneout.
+    '''
```
albertz: `"""`. Also, document types (also return types).

TFNetworkRecLayer.py
albertz: (Also should document return type, i.e. `:rtype:`.)

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+    self._cell = tf.nn.rnn_cell.LSTMCell(num_units, state_is_tuple=state_is_tuple, name=name)
+    self._zoneout_cell = zoneout_factor_cell
+    self._zoneout_outputs = zoneout_factor_output
+    if kwargs.get('is_training', None) is None:
+      from TFNetwork import TFNetwork
+      self.is_training = TFNetwork.get_current_network().train_flag
+    else:
+      self.is_training = kwargs['is_training']
+
+    self.state_is_tuple = state_is_tuple
```
albertz: Do we really need that option?

TFNetworkRecLayer.py
```diff
@@ -5955,3 +5955,72 @@ def _get_output_native_rec_op(self, cell):
...
+
+    if self.state_is_tuple:
+      (prev_c, prev_h) = state
+      (new_c, new_h) = new_state
+    else:
+      num_proj = self._cell._num_units if self._cell._num_proj is None else self._cell._num_proj
+      prev_c = tf.slice(state, [0, 0], [-1, self._cell._num_units])
+      prev_h = tf.slice(state, [0, self._cell._num_units], [-1, num_proj])
+      new_c = tf.slice(new_state, [0, 0], [-1, self._cell._num_units])
+      new_h = tf.slice(new_state, [0, self._cell._num_units], [-1, num_proj])
```
albertz: Is there any case where `not state_is_tuple` is better? If not, I think we should just remove that.

TFNetworkRecLayer.py
```diff
@@ -6002,3 +6002,73 @@ def _get_output_native_rec_op(self, cell):
       y = y[0]
 
     return y
+
+
+class ZoneoutLSTMCell(BaseRNNCell):
+  """
+  Wrapper for tf LSTM to create Zoneout LSTM Cell.
+  This code is an adapted version of Rayhane Mamas version of Tacotron-2
```
albertz: There should be a link here to the original code (Tacotron 2), and also to the paper.
(The formating is RST.)

TFNetworkRecLayer.py
```diff
@@ -6002,3 +6002,73 @@ def _get_output_native_rec_op(self, cell):
...
+  """
+
+  def __init__(self, num_units, zoneout_factor_cell=0., zoneout_factor_output=0., name=None, **kwargs):
+    """
+    Initializer with possibility to set different zoneout values for cell/hidden states.
+    :param int num_units: 
+    :param float zoneout_factor_cell: 
+    :param float zoneout_factor_output: 
+    :param string name: 
+    :param dict kwargs:
```
albertz: `string` is not a valid type. Please check PEP (or PyCharm).
Also, the type of `name` is wrong. It looks like it can also be `None`. Then you should write `str|None`.
Also, `kwargs` will not be documented this way. Better just remove it.
And I think you need one empty line between the comment and the first param.

TFNetworkRecLayer.py
```diff
@@ -6002,3 +6002,73 @@ def _get_output_native_rec_op(self, cell):
...
+    :rtype: int 
+    """
+    return self._cell.output_size
+
+  def __call__(self, inputs, state, scope=None):
+    """
+    Apply ZoneoutLSTM on input with given state 
+    :param tf.Tensor inputs: input tensor to the cell
+    :param tf.nn.rnn_cell.LSTMStateTuple state: previous state of the LSTM
+    :param scope:
```
albertz: What is the type of `scope`? This is missing from the documentation.

albertz: Also, better add a newline between the comment and the first param.

TFNetworkRecLayer.py
```diff
@@ -6002,3 +6002,73 @@ def _get_output_native_rec_op(self, cell):
...
+    return self._cell.output_size
+
+  def __call__(self, inputs, state, scope=None):
+    """
+    Apply ZoneoutLSTM on input with given state 
+    :param tf.Tensor inputs: input tensor to the cell
+    :param tf.nn.rnn_cell.LSTMStateTuple state: previous state of the LSTM
+    :param scope: 
+    :return: tuple of output and LSTMStateTuple 
+    :rtype: tuple[tf.Tensor, tf.nn.rnn_cell.LSTMStateTuple]
```
albertz: The syntax for `rtype` is wrong here.
(Btw, just use PyCharm and check that it works correct.)
It should be `(tf.Tensor, tf.nn.rnn_cell.LSTMStateTuple)`.

PR #170, opened 2019-04-11
batch_norm, new flags, faster implementation, save/load bug fixed

squarenabla:
For conv nets it's beneficial to use both channel and frequency dims to gather the statistics in batch normalization. Added `use_frequency` flag. 
Also conv nets are trained faster in NCHW format, however when eval on cpu only NHWC format is available. => train in NCHW and recog in NHWC. But an error occurs: shape mismatch when saving in NCHW after training and loading in NHWC for recog. => perform all bn operations in one format (NHWC) . In case of input is in NCHW transpose to NHWC, perform operations, transpose back to NCHW.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -11,7 +11,7 @@ import contextlib
 import typing
 import TFUtil
 from Util import unicode, NotSpecified, CollectionReadCheckCovered
-from TFUtil import Data, OutputWithActivation, CustomUpdate, dimshuffle, swapaxes
+from TFUtil import Data, OutputWithActivation, CustomUpdate, CustomUpdateExpAverage, dimshuffle, swapaxes
 from Log import log
 
 
@@ -1081,13 +1081,16 @@ class LayerBase(object):
     return c
 
   def batch_norm(self, data,
+                 axes='f',
                  use_shift=True, use_std=True, use_sample=0.0, force_sample=False,
                  momentum=0.99, epsilon=1e-3,
                  sample_mean=None, sample_variance=None,
                  gamma=None, beta=None,
+                 fused=False,
                  masked_time=True):
     """
     :param Data data:
+    :param int|str|list[int|str] axes: axis/axes to normalize
     :param bool use_shift:
     :param bool use_std:
     :param float use_sample: defaults to 0.0 which is used in training
@@ -1098,6 +1101,7 @@ class LayerBase(object):
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool fused: use fused version of batch normalisation
     :param bool masked_time: flatten and mask input tensor
     :rtype: tf.Tensor
 
@@ -1107,51 +1111,102 @@ class LayerBase(object):
       tf.nn.batch_normalization()
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
+    from operator import mul
+    from functools import reduce
+
     with tf.variable_scope("batch_norm"):
-      if masked_time:
-        x = data.get_placeholder_flattened(keep_dims=True)
-        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
-      else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+      # Creating variables
+      fused = fused and not force_sample
+      # Perform batch_norm operations in NHWC format in order to
+      # keep consistency when saving/loading the variables.
+      # Not relevant when fused or masked_time.
+      input_data = data.copy() if fused or masked_time else data.copy_as_batch_spatial_major()
+      ax = input_data.get_axes_from_description(axes)
+      opts = dict(zip(ax, [-1] * len(ax)))
+      variables_shape = input_data.get_bc_shape(opts)
+      variables_shape_flat = [reduce(mul, variables_shape, 1)]
+      # fused bn possible => create flat variables
+      shape = variables_shape_flat if fused else variables_shape
+
       if sample_mean is None:
         with self.var_creation_scope():
           sample_mean = self.add_param(tf.Variable(
-            initial_value=tf.zeros(data.get_bc_spatial_batch_shape()),
+            initial_value=tf.zeros(shape),
             name="%s_%s_mean" % (self.name, data.name),
             trainable=False))
-        # Use exponential moving average of batch mean.
-        # Note: We could also use cumulative moving average. Our Theano implementation does that for inference.
-        sample_mean = tf.assign_add(sample_mean, (mean - sample_mean) * momentum)
+      else:
+        sample_mean = tf.reshape(sample_mean, shape)
       if sample_variance is None:
-        # Note: Our Theano implementation does not use a moving average for this.
         with self.var_creation_scope():
           sample_variance = self.add_param(tf.Variable(
-            initial_value=tf.ones(data.get_bc_spatial_batch_shape()),
+            initial_value=tf.ones(shape),
             name="%s_%s_variance" % (self.name, data.name),
             trainable=False))
+      else:
+        sample_variance = tf.reshape(sample_variance, shape)
+      if use_std:
+        if gamma is None:
+          with self.var_creation_scope():
+            gamma = self.add_param(tf.Variable(
+              initial_value=tf.ones(shape),
+              name="%s_%s_gamma" % (self.name, data.name),
+              trainable=True))
+        else:
+          gamma = tf.reshape(gamma, shape)
+      if use_shift:
+        if beta is None:
+          with self.var_creation_scope():
+            beta = self.add_param(tf.Variable(
+              initial_value=tf.zeros(shape),
+              name="%s_%s_beta" % (self.name, data.name),
+              trainable=True))
+        else:
+          beta = tf.reshape(beta, shape)
+
+      if fused:
+        # It's possible to use fused bn operation
+        return self._fused_batch_norm(
+          data=data, axes=axes,
+          use_shift=use_shift, use_std=use_std,
+          momentum=momentum, epsilon=epsilon,
+          sample_mean=sample_mean, sample_variance=sample_variance,
+          gamma=gamma, beta=beta,
+          masked_time=masked_time)
+      if masked_time:
+        # Default case
+        x = input_data.get_placeholder_flattened(keep_dims=True)
+        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+        mean = tf.reshape(mean, shape=variables_shape)
+        variance = tf.reshape(variance, shape=variables_shape)
+      else:
+        axes = input_data.get_axes_from_description(axes)
+        assert input_data.feature_dim_axis in axes
+        assert input_data.batch_dim_axis not in axes  # Not supported
+        assert input_data.time_dim_axis not in axes  # Not supported
+        x = input_data.placeholder
+        axes_norm = axes
+        axes_stat = list(set(input_data.get_axes()) - set(axes_norm))
+        mean, variance = tf.nn.moments(x, axes=axes_stat, keep_dims=True)
+      # Use exponential moving average of batch mean.
+      # Note: We could also use cumulative moving average. Our Theano implementation does that for inference.
+      if isinstance(sample_mean, tf.Variable):
+        sample_mean = tf.assign_add(sample_mean, (mean - sample_mean) * momentum)
+      if isinstance(sample_variance, tf.Variable):
         sample_variance = tf.assign_add(sample_variance, (variance - sample_variance) * momentum)
       # If train or if force_sample, use default use_sample=0.0, otherwise use_sample=1.0.
       use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
       mean = (1. - use_sample) * mean + use_sample * sample_mean
       variance = (1. - use_sample) * variance + use_sample * sample_variance
-      bn = (data.placeholder - mean) * tf.rsqrt(variance + epsilon)
+      bn = (input_data.placeholder - mean) * tf.rsqrt(variance + epsilon)
       if use_std:
-        if gamma is None:
-          with self.var_creation_scope():
-            gamma = self.add_param(tf.Variable(
-              initial_value=tf.ones(data.get_bc_spatial_batch_shape()),
-              name="%s_%s_gamma" % (self.name, data.name),
-              trainable=True))
         bn *= gamma
       if use_shift:
-        if beta is None:
-          with self.var_creation_scope():
-            beta = self.add_param(tf.Variable(
-              initial_value=tf.zeros(data.get_bc_spatial_batch_shape()),
-              name="%s_%s_beta" % (self.name, data.name),
-              trainable=True))
         bn += beta
+      if data.is_batch_feature_major and not masked_time:
+        # Transpose back to NCHW
+        perm = data.get_axes()
+        perm = perm[:1] + perm[-1:] + perm[1:-1]
+        bn = tf.transpose(bn, perm=perm)
       return bn
 
   def get_hidden_state(self):
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -1209,6 +1209,120 @@ class LayerBase(object):
         bn = tf.transpose(bn, perm=perm)
       return bn
 
+  def _fused_batch_norm(self, data,
+                        axes,
+                        use_shift, use_std,
+                        momentum, epsilon,
+                        sample_mean, sample_variance,
+                        gamma, beta,
+                        masked_time):
+    """
+    :param Data data:
+    :param int|str|list[int|str] axes: axis/axes to normalize
+    :param bool use_shift:
+    :param bool use_std:
+    :param float momentum: for the running average of sample_mean and sample_std
+    :param float epsilon:
+    :param tf.Variable sample_mean:
+    :param tf.Variable sample_variance:
+    :param tf.Variable gamma:
+    :param tf.Variable beta:
+    :param bool masked_time: flatten and mask input tensor
+    :rtype: tf.Tensor
+
+    See tf.nn.fused_batch_norm()
+    """
+    assert sample_mean is not None
+    assert sample_variance is not None
+    assert not use_std or gamma is not None
+    assert not use_shift or beta is not None
+
+    # with tf.variable_scope("fused_batch_norm"):
+    axes = data.get_axes_from_description(axes)
+    assert data.feature_dim_axis in axes
+    assert data.batch_dim_axis not in axes  # Not supported
+    assert data.time_dim_axis not in axes  # Not supported
+    if masked_time:
+      assert len(axes) == 1 and axes[0] == data.feature_dim_axis
+      x = data.get_placeholder_flattened(keep_dims=True)
+    else:
+      # Put feature_dim_axis at the end s.t. the merged axis is a feature.
+      axes.remove(data.feature_dim_axis)
+      axes.append(data.feature_dim_axis)
+      x = data.get_placeholder_merge_axes(axes=axes, keep_dims=True)
+    data_format = "NCHW" if data.is_batch_feature_major else "NHWC"
+
+    def _fused_batch_norm_training():
+      return tf.nn.fused_batch_norm(
+        x,
+        gamma,
+        beta,
+        epsilon=epsilon,
+        data_format=data_format)
+
+    def _fused_batch_norm_inference():
+      return tf.nn.fused_batch_norm(
+        x,
+        gamma,
+        beta,
+        mean=sample_mean,
+        variance=sample_variance,
+        epsilon=epsilon,
+        is_training=False,
+        data_format=data_format)
+
+    output, mean, variance = tf.cond(
+      tf.cast(self.network.train_flag, tf.bool),
+      true_fn=_fused_batch_norm_training,
+      false_fn=_fused_batch_norm_inference)
+    # Remove Bessel's correction to be consistent with non-fused batch norm.
+    sample_size = tf.cast(
+      tf.size(x) / tf.size(variance), variance.dtype)
+    factor = (sample_size - tf.cast(1.0, variance.dtype)) / sample_size
+    variance *= factor
+
+    # Controlling updates in training and inference modes via momentum
+    mean_momentum = tf.cast(self.network.train_flag, tf.float32) * momentum
+    variance_momentum = tf.cast(self.network.train_flag, tf.float32) * momentum
+    # Add custom updates for moving_mean and moving_variance
+    mean_update = CustomUpdateExpAverage(average=mean, alpha=mean_momentum)
+    variance_update = CustomUpdateExpAverage(average=variance, alpha=variance_momentum)
+    if isinstance(sample_mean, tf.Variable):
+      mean_update.set_on_var(sample_mean)
+    if isinstance(sample_variance, tf.Variable):
+      variance_update.set_on_var(sample_variance)
+
+    if len(axes) > 1:
+      # Output has a different shape => reshape.
+      old_shape = tf.shape(data.placeholder)
+      out_shape = tf.shape(output)
+      expand_axes = axes
+      new_shape = []
+      major_axis = expand_axes[-1]
+      for i in data.get_axes():
+        if i == major_axis:
+          for j in expand_axes:
+            new_shape.append(old_shape[j])
+          continue
+        if i not in expand_axes:
+          new_shape.append(out_shape[i])
+      output = tf.reshape(output, shape=new_shape)
+      # Transform the axes to the original position
+      perm = data.get_axes()
+      for i in axes:
+        perm.remove(i)
+      perm = perm[:axes[-1]] + axes + perm[axes[-1]:]
+      perm = dict(zip(perm, range(len(perm))))
+      perm = [perm[i] for i in data.get_axes()]
+      output = tf.transpose(output, perm=perm)
+    elif masked_time:
+      # Output has a different shape => reshape.
+      # Much simpler reshape than in the previous case.
+      new_shape = tf.shape(data.placeholder)
+      output = tf.reshape(output, shape=new_shape)
+    output.set_shape(data.placeholder.shape)
+    return output
+
   def get_hidden_state(self):
     """
     If this is a recurrent layer, this would return the hidden state.
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -1736,8 +1736,12 @@ class Data(object):
       dyn_axes = [(i if (i < removed_axis) else (i - 1))
                   for i in dyn_axes]
       ndim -= 1
+    shape = tf.shape(x)
+    dyn_axes.sort()
+    # Transpose remaining dyn_axes, s.t. they are together before merge
+    perm = dyn_axes + [i for i in range(self.ndim) if i not in dyn_axes]
+    x = tf.transpose(x, perm=perm)
     if len(dyn_axes) > 1:
-      shape = tf.shape(x)
       x = tf.reshape(
         x,
         [tf.reduce_prod([shape[i] for i in dyn_axes])] +
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -1669,6 +1669,47 @@ class Data(object):
     return flatten_with_seq_len_mask(self.placeholder, seq_lens, batch_dim_axis=self.batch_dim_axis,
                                      time_dim_axis=self.time_dim_axis)
 
+  def get_placeholder_merge_axes(self, axes, keep_dims=False):
+    """
+    :param list[int|str] axes: axes to merge
+    :param bool keep_dims: if set, it will add broadcast dimensions after the flattening behind the first axis
+    :rtype: tf.Tensor
+    :return: placeholder where all axes specified in 'axes' are flattened into a single axis.
+      The position of the resulting axis will be axes[-1].
+    """
+    assert self.placeholder is not None
+    x = self.placeholder
+    axes_merge = self.get_axes_from_description(axes)
+    assert self.time_dim_axis not in axes_merge and self.batch_dim_axis not in axes_merge  # Not supported
+
+    shape = tf.shape(x)
+    new_shape = [(1 if i in axes_merge else shape[i])
+                 for i in self.get_axes()]
+    new_shape[axes_merge[-1]] = tf.reduce_prod([shape[i] for i in axes_merge])
+    # To be consistent transpose the placeholder s.t. axes to merge are together before the merge
+    perm = self.get_axes()
+    major_axis = axes_merge[-1]
+    for i in axes_merge:
+      if i == major_axis:
+        continue
+      perm.remove(i)
+      perm.insert(perm.index(major_axis), i)
+    new_shape = [new_shape[i] for i in perm]
+    x = tf.transpose(x, perm=perm)
+    x = tf.reshape(x, shape=list(new_shape))
+    # Squeeze only axes from axes_merge except major_axis:
+    squeeze_axes = [perm.index(i) for i in axes_merge if i != major_axis]
+    x = tf.squeeze(x, axis=squeeze_axes)
+
+    if keep_dims:
+      axes_merge.sort()
+      for i in axes_merge:
+        if i == major_axis:
+          continue
+        x = tf.expand_dims(x, axis=i)
+    x.set_shape([None] * self.batch_ndim)
+    return x
+
   def get_placeholder_flattened(self, keep_dims=False):
     """
     :param bool keep_dims: if set, it will add broadcast dimensions after the flattening behind the first axis
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -212,19 +212,19 @@ def test_batch_norm_vars():
     assert_equal(layer.params[bn_prefix + "variance"].get_shape().as_list(), [1, 1, n_out])
 
 
-def test_batch_norm():
+def test_batch_norm_masked_vs_nonmasked_equal_seq_len():
   with make_scope() as session:
     import numpy as np
     net = TFNetwork(extern_data=ExternData())
     net.train_flag = True
-    with tf.variable_scope("src_nchw"):
-      src_nhwc = InternalLayer(name="src_nchw", network=net, out_type={"dim": 16,
-                                                                       "shape": (None, 16, 16),
-                                                                       "batch_dim_axis": 0,
-                                                                       "time_dim_axis": 1,
-                                                                       "feature_dim_axis": 3,
-                                                                       "sparse": False
-                                                                       })
+    with tf.variable_scope("src_nhwc"):
+      src_nhwc = InternalLayer(name="src_nhwc", network=net, out_type={
+        "dim": 16,
+        "shape": (None, 16, 16),
+        "batch_dim_axis": 0,
+        "time_dim_axis": 1,
+        "feature_dim_axis": 3,
+        "sparse": False})
       src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
       src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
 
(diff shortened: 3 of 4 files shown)


TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
...
+        default_shape = data.get_bc_shape(opts)
+        variables_shape = [np.prod(default_shape)]
+        mean, variance = tf.nn.moments(x, axes=axes, keep_dims=True)
+        mean = tf.reshape(mean, shape=variables_shape)
+        variance = tf.reshape(variance, shape=variables_shape)
       if sample_mean is None:
         with self.var_creation_scope():
           sample_mean = self.add_param(tf.Variable(
-            initial_value=tf.zeros(data.get_bc_spatial_batch_shape()),
+            initial_value=tf.random_normal(variables_shape),
```
albertz: Why? Should be 0.

TFNetworkLayer.py
```diff
@@ -1035,6 +1036,7 @@ def batch_norm(self, data,
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool use_frequency: whether to use frequency dim to collect statistics
```
albertz: what do you mean by "frequency dim"?

TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
     with tf.variable_scope("batch_norm"):
+      import numpy as np
```
albertz: Don't import it as `np`.

TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
...
       if masked_time:
+        assert not use_frequency
+        # Default case
         x = data.get_placeholder_flattened(keep_dims=True)
         mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+        variables_shape = data.get_bc_spatial_batch_shape()
+        squeeze_variables = False
```
albertz: Don't make the code more complicated. Why do you have `squeeze_variables = False` in the one case and `squeeze_variables = True` in the other case? Make it consistent.

TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
...
         x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+        axes = data.get_axes(exclude_feature=True)
+        if use_frequency:
+          opts = {data.feature_dim_axis: -1, data.get_axis_from_description("s:1"): -1}
+          axes.remove(data.get_axis_from_description('s:1'))
+        else:
+          opts = {}
+        default_shape = data.get_bc_shape(opts)
+        variables_shape = [np.prod(default_shape)]
```
albertz: Do not make a flat version. Instead just keep those which are not broadcasted (basically just `data.batch_shape[a] for a in range(data.batch_ndim) if a not in axes`, or not?).

TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
...
+        squeeze_variables = True
         x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+        axes = data.get_axes(exclude_feature=True)
+        if use_frequency:
+          opts = {data.feature_dim_axis: -1, data.get_axis_from_description("s:1"): -1}
+          axes.remove(data.get_axis_from_description('s:1'))
+        else:
+          opts = {}
+        default_shape = data.get_bc_shape(opts)
```
albertz: Don't call this `default_shape`. Why `default`? Call this `data_bc_shape` or so.

TFNetworkLayer.py
albertz: (However, that also introduces some incompatibility to older models. Make sure that this is catched and handled correctly.)

TFNetworkLayer.py
```diff
@@ -1045,16 +1047,34 @@ def batch_norm(self, data,
...
       else:
+        # Variables have to be squeezed in order to save/load the variables in
+        # different data formatrs (NCHW/NHWC) => set squeeze_variables
+        squeeze_variables = True
         x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+        axes = data.get_axes(exclude_feature=True)
+        if use_frequency:
+          opts = {data.feature_dim_axis: -1, data.get_axis_from_description("s:1"): -1}
+          axes.remove(data.get_axis_from_description('s:1'))
```
albertz: Why `'s:1'`?

TFNetworkLayer.py
```diff
@@ -1064,30 +1084,39 @@ def batch_norm(self, data,
...
             name="%s_%s_variance" % (self.name, data.name),
             trainable=False))
         sample_variance = tf.assign_add(sample_variance, (variance - sample_variance) * momentum)
       # If train or if force_sample, use default use_sample=0.0, otherwise use_sample=1.0.
       use_sample = 1.0 + tf.cast(tf.logical_or(self.network.train_flag, force_sample), tf.float32) * (use_sample - 1.0)
       mean = (1. - use_sample) * mean + use_sample * sample_mean
       variance = (1. - use_sample) * variance + use_sample * sample_variance
+      if squeeze_variables:
+        # Double check, that it's not a masked_time case
+        assert not masked_time
```
albertz: Why is this check needed? This is actually confusing here. This should work no matter if `masked_time` or not `masked_time`. And also don't add redundant comments (which say exactly the same as the code).

TFNetworkLayer.py
albertz: (But you anyway have that incompatibility already with this change for the `not masked_time` case.)

albertz:
Actually:

> Also conv nets are trained faster in NCHW format, however when eval on cpu only NHWC format is available. => train in NCHW and recog in NHWC. But an error occurs: shape mismatch when saving in NCHW after training and loading in NHWC for recog. => squeeze variables to 1dim tensors to be format independent.

This is wrong! Yes, you don't get an error by this hack, but you are actually doing it wrong then, because the axes are swapped, and then the order of the dimensions your 1D tensor are wrong.

TFNetworkLayer.py
albertz: Also, this hides some serious bugs which can happen here, which would silently go through but behave incorrect. Never ever introduce such behavior which will make it hard to find this bug in the future.

albertz: (Also see my comment below.)

TFNetworkLayer.py
squarenabla: Oh, sorry. Didn't delete after debugging

squarenabla:
> Actually:
> 
> > Also conv nets are trained faster in NCHW format, however when eval on cpu only NHWC format is available. => train in NCHW and recog in NHWC. But an error occurs: shape mismatch when saving in NCHW after training and loading in NHWC for recog. => squeeze variables to 1dim tensors to be format independent.
> 
> This is wrong! Yes, you don't get an error by this hack, but you are actually doing it wrong then, because the axes are swapped, and then the order of the dimensions your 1D tensor are wrong.

I check that in the tests. I save the variables in the NCHW format and load in the NHWC. And assert on the equal output.

albertz:
Then your test is wrong. As I explained, it will be different.

squarenabla:
> Then your test is wrong. As I explained, it will be different.

I see what you mean. I'll think, why the test asserts are passing. And modify the batch_norm and tests s.t. to be sure that all variables are exactly the same but transposed (nchw/nhwc). 

> Do not make a flat version. Instead just keep those which are not broadcasted (basically just `data.batch_shape[a] for a in range(data.batch_ndim) if a not in axes`, or not?).
I don't understand what you mean here.
The matter is that all bn variables are not  broadcasted from checkpoints saved in different data format. Another solution is to transform the input to nhwc and do all operations in nhwc format; then transform back. I think it'll be more compatible to the old versions.

albertz:
> > Do not make a flat version.
> 
> I don't understand what you mean here.
> The matter is that all bn variables are not broadcasted from checkpoints saved in different data format. Another solution is to transform the input to nhwc and do all operations in nhwc format; then transform back. I think it'll be more compatible to the old versions.

Not sure what you don't understand. Just don't make it flat. I don't say that you should add the broadcast axes.

TFNetworkLayer.py
squarenabla: It's a frequency dim. s:0 is time, s:1 is freq

TFNetworkLayer.py
squarenabla: frequency bin dimension, it's "s:1"

albertz: No, "s:1" is "s:1". Maybe you have some case where you call this frequency dim for some reason. But it is just "s:1". I don't understand why you think that "s:1" is in general always called a "frequency dim".

TFNetworkLayer.py
albertz: No. "s:0" can be anything, "s:1" can be anything.

TFNetworkLayer.py
```diff
@@ -1044,16 +1046,33 @@ def batch_norm(self, data,
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
     with tf.variable_scope("batch_norm"):
+      input_data = data.copy()
```
albertz: Why do you rename it here? And why the copy?

TFNetworkLayer.py
```diff
@@ -1044,16 +1046,33 @@ def batch_norm(self, data,
...
         mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
+        variables_shape = input_data.get_bc_spatial_batch_shape()
       else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+        # Perform batch_norm operations in NHWC format in order to
+        # keep consistency when saving/loading the variables.
+        if data.is_batch_feature_major:
+          # Transform the input to NHWC
+          input_data = data.copy_as_batch_spatial_major()
```
albertz: I thought you want to have it efficient? This would make it quite inefficient.

squarenabla: What would be your suggestions?

TFNetworkLayer.py
squarenabla: Then how should I get the frequency bin dim?

albertz: Well, you could call the parameter `use_spatial1` or so, which is already much better than `use_frequency`.

But this is still not good. This is too specific. What if you want to have `use_spatial2` or so later? You should come up with a generic solution. See e.g. `dropout_noise_shape` as an example. Or the general way to specify axes, `Data.get_axes_from_description`.

TFNetworkLayer.py
albertz: It was efficient before? (But anyway, maybe also depends what the fused variant can handle.)

squarenabla: I came up to 3 solutions:
1.  Perform all operations in one format (NHWC), and  transform the input to NHWC and transform the output back to the initial format (+ 2 transpose operations)
2.  Keep all variables in one format (NHWC) and transform them according to the input format (+ num_vars transpose operations, which is 4 by default)
3.  Keep all variables flat and reshape according to the input format (+num_vars reshape operations)
3d variant isn't good as it's hard to keep consistency with the old implementation
Now you say, that the 1st isn't efficient. But what solution would you suggest as efficient?

albertz: Any wrong/incorrect solution is not a solution, so let's not discuss that (e.g. your flat variables).

It is not inefficient because I suddenly say so. You should see that for yourself. You also should be able to tell yourself which other solution is probably more efficient. Actually, the whole argument e.g. of using the fused op, or convolution on NCHW was for efficiency, so I wonder why you now think that efficiency is not important anymore?

Transforming the variables on-the-fly is also some overhead which you don't need. Although that is maybe not so relevant (much much faster than transforming the data here). Keep it in the format which is faster for training on GPU, then this does not give you any overhead for the training case.

I thought about detecting whether the parameters were saved in the other format and the automatically convert them (similarly as we automatically convert params in other cases, like CuDNN LSTM, or NativeLSTM/BlockLSTM, etc). But the on-the-fly transformation is probably simpler.

Maybe check first what the fused op really supports here, and just do it accordingly.

TFNetworkLayer.py
squarenabla: The axes order might be changed in `input_data` later. But I want to keep the `data` unchanged.

squarenabla:
Some tests fail, I'll fix soon

squarenabla:
bn
Speed/memory benchmark for different flags

tests/test_TFUtil.py
```diff
@@ -2583,6 +2610,20 @@ def test_softmax_cross_entropy_over_size_gradient():
       last_loss = loss
 
 
+def test_flatten_with_seq_len_mask():
+  import numpy as np
+  import numpy.testing as npt
+  batch_dim_axis = 0
+  time_dim_axis = 2
+  x = tf.constant(np.random.rand(10, 16, 11, 16))
+  seq_lens = tf.constant([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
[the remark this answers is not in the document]
squarenabla: I agree, but this function tests whether the axes are merged correctly

TFUtil.py
```diff
@@ -1834,6 +1880,23 @@ def update_var(self, var):
     raise NotImplementedError
 
 
+class CustomUpdateMomentum(CustomUpdate):
```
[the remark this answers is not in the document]
squarenabla: Oh, it is. I'll replace it with `CustomUpdateExpAverage`

TFUtil.py
```diff
@@ -1260,6 +1260,48 @@ def get_placeholder_time_flattened(self):
...
+    """
+    import numpy as np
+
+    assert self.placeholder is not None
+    x = self.placeholder
+    axes_merge = self.get_axes_from_description(axes)
+    assert self.time_dim_axis not in axes_merge and self.batch_dim_axis not in axes_merge  # Not supported
+
+    shape = tf.shape(x)
+    new_shape = [(tf.Dimension(1) if i in axes_merge else shape[i])
```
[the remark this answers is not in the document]
squarenabla: To make all elements in `new_shape` be the same type. But of course this can be just `1`.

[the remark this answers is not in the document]
albertz: But `tf.Dimension` is not the same type as `tf.Tensor` (`shape[i]` is of type `tf.Tensor`)?

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,238 @@ def batch_norm(self, data,
...
+    :param tf.Tensor beta:
+    :param bool masked_time: flatten and mask input tensor
+    :rtype: tf.Tensor
+
+    See tf.nn.fused_batch_norm()
+    """
+    from operator import mul
+    from functools import reduce
+    from tensorflow.python.ops import array_ops
+    from tensorflow.python.ops import math_ops
```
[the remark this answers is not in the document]
squarenabla: I need this below. Where should I make imports? At the beginning of the file?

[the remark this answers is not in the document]
albertz: No, you don't need it. Please read a TensorFlow tutorial to get familiar with TF. You can write `tf.size`, `tf.cast`, etc., like you usually do.

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,238 @@ def batch_norm(self, data,
...
               trainable=True))
         bn += beta
+      if data.is_batch_feature_major and not masked_time:
+        # Transpose back to NCHW
+        perm = data.get_axes()
+        perm = perm[:1] + perm[-1:] + perm[1:-1]
+        bn = tf.transpose(bn, perm=perm)
       return bn
 
+  def fused_batch_norm(self, data,
```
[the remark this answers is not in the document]
squarenabla: Apart from the variables creation the code is completely different. Merging everything into one function will make the code very bulky and unreadable.

TFNetworkLayer.py
```diff
@@ -1034,6 +1037,7 @@ def batch_norm(self, data,
     :param tf.Tensor sample_variance:
     :param tf.Tensor gamma:
     :param tf.Tensor beta:
+    :param bool fused: use fused version of batch normalisation
```
albertz: Why do you have this as an option? Or rather, why is the default `False`? I thought that you made sure via the tests that it does not matter at all in behavior? In that case, why can't you always have this `True`?

TFNetworkLayer.py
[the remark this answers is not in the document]
albertz: I'm not sure about that. Why? They are both doing the same thing. Only that some of the calculations are fused together.
In any case, the variable creation should definitely also share the same code (i.e. created in `batch_norm` and just passed over to this function).
Also, this function should be private.

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,238 @@ def batch_norm(self, data,
...
+        bn = tf.transpose(bn, perm=perm)
       return bn
 
+  def fused_batch_norm(self, data,
+                       axes='f',
+                       use_shift=True, use_std=True,
+                       momentum=0.99, epsilon=1e-3,
+                       sample_mean=None, sample_variance=None,
+                       gamma=None, beta=None,
+                       masked_time=True):
```
albertz: Please remove the default values. They are all already given by `batch_norm`, and you should never call the function directly.

TFNetworkLayer.py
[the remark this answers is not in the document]
squarenabla: fused batch norm can use only 1D variables, so variables created for the regular version have different shape than the fused version.

TFNetworkLayer.py
squarenabla: It's false to make the old version default.

albertz: I thought they are equivalent? If they are equivalent, why is there any reason to not use the fused version if possible? So shouldn't it be the default then (if it is possible to use)? (Maybe use `None`, or `NotSpecified` as the default value, and then in the code automatically set it to `True` if it is possible to use.)
If they are not equivalent, then this is wrong and should be fixed. It should be equivalent.

TFNetworkLayer.py
albertz: Ping?

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,237 @@ def batch_norm(self, data,
...
+        sample_mean = tf.reshape(sample_mean, shape=variables_shape)
       if sample_variance is None:
         # Note: Our Theano implementation does not use a moving average for this.
         with self.var_creation_scope():
           sample_variance = self.add_param(tf.Variable(
-            initial_value=tf.ones(data.get_bc_spatial_batch_shape()),
+            initial_value=tf.ones(variables_shape),
             name="%s_%s_variance" % (self.name, data.name),
-            trainable=False))
+            trainable=True))
```
albertz: Why the change to `trainable=True`? This is wrong, or not? It is not a variable which you train. The update happens because you have `tf.assign_add` below.

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,237 @@ def batch_norm(self, data,
...
+        tf.size(x) / tf.size(variance), variance.dtype)
+      factor = (sample_size - tf.cast(1.0, variance.dtype)) / sample_size
+      variance *= factor
+
+      # Controlling updates in training and inference modes via momentum
+      momentum_mean = tf.cast(self.network.train_flag, tf.float32) * momentum
+      momentum_variance = tf.cast(self.network.train_flag, tf.float32) * momentum
+      # Add custom updates for moving_mean and moving_variance
+      mean_update = CustomUpdateExpAverage(average=mean, alpha=momentum_mean)
+      variance_update = CustomUpdateExpAverage(average=variance, alpha=momentum_variance)
```
albertz: Why do you have the updating here different than in the other function? (Here `CustomUpdateExpAverage`, in the other it is `tf.assign_add` directly.) It should be the same.

albertz: Also, please don't duplicate the code. Use the same code in both cases. Simplify this. I think you should not have this in an extra function.

TFNetworkLayer.py
```diff
@@ -1043,53 +1047,237 @@ def batch_norm(self, data,
...
+              initial_value=tf.ones(variables_shape_flat),
+              name="%s_%s_gamma" % (self.name, data.name),
+              trainable=True))
+      if use_shift:
+        if beta is None:
+          with self.var_creation_scope():
+            beta = self.add_param(tf.Variable(
+              initial_value=tf.zeros(variables_shape_flat),
+              name="%s_%s_beta" % (self.name, data.name),
+              trainable=True))
```
albertz: Please don't duplicate the code.

tests/test_TFUtil.py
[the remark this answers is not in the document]
albertz: But you don't test set if all seq lens are the same.

TFNetworkLayer.py
squarenabla: If it's `False`,  these variables are not "seen" by the updater even if the custom update is set.

TFNetworkLayer.py
squarenabla: I see your point, I'll do this way

albertz:
What is the state here?

albertz:
Ping? What is the state here?

albertz:
Hello?

squarenabla:
Sorry, didn't notice the messages. I'll return back to the pull request in a few days

tests/test_TFUtil.py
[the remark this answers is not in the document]
squarenabla: Added a test if seq_lens are not equal

TFNetworkLayer.py
squarenabla: Fixed

TFNetworkLayer.py
squarenabla: > I thought they are equivalent? If they are equivalent, why is there any reason to not use the fused version if possible? So shouldn't it be the default then (if it is possible to use)? (Maybe use `None`, or `NotSpecified` as the default value, and then in the code automatically set it to `True` if it is possible to use.)
> If they are not equivalent, then this is wrong and should be fixed. It should be equivalent.

I think it's better keep the default as `False`. This way, when loading old configs, the nonfused version is used. Otherwise, old checkpoints can't be loaded into fused variables.

TFNetworkLayer.py
squarenabla: `tf.assign_add` works in the old batch_norm because the variables are than added to the computatin graph via `use_sample`:
https://github.com/rwth-i6/returnn/blob/ddc40b9ac02433b6fc16a8f71344a27b726024dd/TFNetworkLayer.py#L1136

In case of fused version, all computations are done inside `tf.nn.fused_batch_norm`. And later updates can't be added to the graph. So, it's done via `CustomUpdateExpAverage`.

squarenabla:
@albertz, could you have a look at the changes?

TFNetworkLayer.py
albertz: But then this is a bug. Old checkpoints should be able to be loaded, also when you have `fused=True` here. This is what I explained earlier. And also you should have a test case for this (as I also explained earlier).

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
-      if masked_time:
-        x = data.get_placeholder_flattened(keep_dims=True)
-        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
-      else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+      # Creating variables
+      fused = fused and not force_sample
+      # Perform batch_norm operations in NHWC format in order to
+      # keep consistency when saving/loading the variables.
```
albertz: That does not make sense. The way how you load/store variables can be independent from how you operate.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+
     with tf.variable_scope("batch_norm"):
-      if masked_time:
-        x = data.get_placeholder_flattened(keep_dims=True)
-        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
-      else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+      # Creating variables
+      fused = fused and not force_sample
```
albertz: Why?
Also, don't write code like this. Writing it like this is much easier to read and understand:
```
if force_sample:
  fused = False
```

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+
+    output, mean, variance = tf.cond(
+      tf.cast(self.network.train_flag, tf.bool),
+      true_fn=_fused_batch_norm_training,
+      false_fn=_fused_batch_norm_inference)
+    # Remove Bessel's correction to be consistent with non-fused batch norm.
+    sample_size = tf.cast(
+      tf.size(x) / tf.size(variance), variance.dtype)
+    factor = (sample_size - tf.cast(1.0, variance.dtype)) / sample_size
+    variance *= factor
```
albertz: So the fused variant is actually different than what we have?
That is good to know.
But instead of just undoing what they do, maybe what they do actually makes sense? You should not just undo it. If that is actually the better way to do it (using Bessel's correction), then we also should have that (have an option for it).

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+
+    # Controlling updates in training and inference modes via momentum
+    mean_momentum = tf.cast(self.network.train_flag, tf.float32) * momentum
+    variance_momentum = tf.cast(self.network.train_flag, tf.float32) * momentum
```
albertz: Don't write code like that. Use `TFUtil.cond`.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+        bn = tf.transpose(bn, perm=perm)
       return bn
 
+  def _fused_batch_norm(self, data,
+                        axes,
+                        use_shift, use_std,
+                        momentum, epsilon,
+                        sample_mean, sample_variance,
+                        gamma, beta,
+                        masked_time):
```
albertz: As I explained already, don't make it a separate function. You have lot of duplicated code here.

TFUtil.py
```diff
@@ -1695,8 +1736,12 @@ def get_placeholder_flattened(self, keep_dims=False):
       dyn_axes = [(i if (i < removed_axis) else (i - 1))
                   for i in dyn_axes]
       ndim -= 1
+    shape = tf.shape(x)
+    dyn_axes.sort()
+    # Transpose remaining dyn_axes, s.t. they are together before merge
+    perm = dyn_axes + [i for i in range(self.ndim) if i not in dyn_axes]
+    x = tf.transpose(x, perm=perm)
```
albertz: Why this change in `get_placeholder_flattened`? Please make a separate pull request, and show a test case first, where the old behavior was wrong (and why).

tests/test_TFUtil.py
```diff
@@ -385,6 +385,33 @@ def test_4D_Data_get_placeholder_flattened():
   assert res.shape[0] == 7 * 9 * 13
   assert len(res.shape) == 2
 
+
+def test_4D_nhwc_Data_get_placeholder_flattened():
+  import numpy as np
+  import numpy.testing as npt
```
albertz: Don't do that. Why? You don't need that. Just write directly `numpy` and `numpy.testing`.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+    assert data.feature_dim_axis in axes
+    assert data.batch_dim_axis not in axes  # Not supported
+    assert data.time_dim_axis not in axes  # Not supported
+    if masked_time:
+      assert len(axes) == 1 and axes[0] == data.feature_dim_axis
+      x = data.get_placeholder_flattened(keep_dims=True)
+    else:
+      # Put feature_dim_axis at the end s.t. the merged axis is a feature.
+      axes.remove(data.feature_dim_axis)
+      axes.append(data.feature_dim_axis)
```
albertz: Why does the order in `axes` matter?

TFUtil.py
```diff
@@ -1669,6 +1669,47 @@ def get_placeholder_time_flattened(self):
     return flatten_with_seq_len_mask(self.placeholder, seq_lens, batch_dim_axis=self.batch_dim_axis,
                                      time_dim_axis=self.time_dim_axis)
 
+  def get_placeholder_merge_axes(self, axes, keep_dims=False):
+    """
+    :param list[int|str] axes: axes to merge
+    :param bool keep_dims: if set, it will add broadcast dimensions after the flattening behind the first axis
```
albertz: This is inconsistent with what you write later (`axes[-1]`).

TFUtil.py
```diff
@@ -1669,6 +1669,47 @@ def get_placeholder_time_flattened(self):
...
+    :rtype: tf.Tensor
+    :return: placeholder where all axes specified in 'axes' are flattened into a single axis.
+      The position of the resulting axis will be axes[-1].
+    """
+    assert self.placeholder is not None
+    x = self.placeholder
+    axes_merge = self.get_axes_from_description(axes)
+    assert self.time_dim_axis not in axes_merge and self.batch_dim_axis not in axes_merge  # Not supported
+
+    shape = tf.shape(x)
```
albertz: Use `get_shape` instead of `tf.shape`.

TFUtil.py
```diff
@@ -1669,6 +1669,47 @@ def get_placeholder_time_flattened(self):
...
+    new_shape = [(1 if i in axes_merge else shape[i])
+                 for i in self.get_axes()]
+    new_shape[axes_merge[-1]] = tf.reduce_prod([shape[i] for i in axes_merge])
```
albertz: Use `optional_mul` instead of `tf.reduce_prod`.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+    :rtype: tf.Tensor
+
+    See tf.nn.fused_batch_norm()
+    """
+    assert sample_mean is not None
+    assert sample_variance is not None
+    assert not use_std or gamma is not None
+    assert not use_shift or beta is not None
+
+    # with tf.variable_scope("fused_batch_norm"):
```
albertz: Don't ever add commented out code just like that.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+    axes = data.get_axes_from_description(axes)
```
albertz: You add this code to allow to specify different (and maybe multiple) axes, but do you actually use it? It makes the remaining code somewhat more complicated.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
-        mean, variance = tf.nn.moments(x, axes=[0], keep_dims=True)
-      else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+      # Creating variables
+      fused = fused and not force_sample
+      # Perform batch_norm operations in NHWC format in order to
+      # keep consistency when saving/loading the variables.
+      # Not relevant when fused or masked_time.
+      input_data = data.copy() if fused or masked_time else data.copy_as_batch_spatial_major()
```
albertz: This is not clear. Why do you make it batch-spatial major if not fused and not masked time?

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+      ax = input_data.get_axes_from_description(axes)
+      opts = dict(zip(ax, [-1] * len(ax)))
```
albertz: Don't write code like that. Write directly a dict literal (`{...}`).

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+      variables_shape = input_data.get_bc_shape(opts)
```
albertz: This does not make sense. Why do you use this function `get_bc_shape`? You defined `opts` in a way that this totally does not make sense to use this function at all. Do you want to get a broadcast (bc) shape? But then your `opts` don't make sense, as you will not get any broadcast shape. Do you just want to get the shape? Then why not directly get the shape?

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
       tf.nn.batch_normalization()
       https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/batch_norm.py
     """
+    from operator import mul
+    from functools import reduce
```
albertz: Not needed.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
-      else:
-        x = data.placeholder
-        mean, variance = tf.nn.moments(x, axes=data.get_axes(exclude_feature=True), keep_dims=True)
+      # Creating variables
+      fused = fused and not force_sample
+      # Perform batch_norm operations in NHWC format in order to
+      # keep consistency when saving/loading the variables.
+      # Not relevant when fused or masked_time.
+      input_data = data.copy() if fused or masked_time else data.copy_as_batch_spatial_major()
+      ax = input_data.get_axes_from_description(axes)
```
albertz: Do you really use the support of multiple axes? Do you think it is important?

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+      opts = dict(zip(ax, [-1] * len(ax)))
+      variables_shape = input_data.get_bc_shape(opts)
+      variables_shape_flat = [reduce(mul, variables_shape, 1)]
+      # fused bn possible => create flat variables
+      shape = variables_shape_flat if fused else variables_shape
```
albertz: Your variable names are very bad. Please fix that. `shape` is always also the variable shape. `variable_shape` is not (always) the variable shape. This is confusing.

TFNetworkLayer.py
```diff
@@ -1107,53 +1111,218 @@ def batch_norm(self, data,
...
+      # Creating variables
+      fused = fused and not force_sample
+      # Perform batch_norm operations in NHWC format in order to
+      # keep consistency when saving/loading the variables.
+      # Not relevant when fused or masked_time.
+      input_data = data.copy() if fused or masked_time else data.copy_as_batch_spatial_major()
+      ax = input_data.get_axes_from_description(axes)
+      opts = dict(zip(ax, [-1] * len(ax)))
+      variables_shape = input_data.get_bc_shape(opts)
+      variables_shape_flat = [reduce(mul, variables_shape, 1)]
```
albertz: Do not use `reduce` and `mul` here, but rather e.g. `optional_mul`.

TFNetworkLayer.py
albertz: Also, this makes it inefficient, or not? And is not really needed?

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+  gpu = tf.test.is_gpu_available(
+    cuda_only=False,
+    min_cuda_compute_capability=None
+  )
+  if not gpu:
+    print("The CPU implementation of FusedBatchNorm only supports NHWC tensor format for now.")
+    print("Quiting the test")
+    return
+  import numpy as np
+  import numpy.testing as npt
```
albertz: Don't do that (as explained earlier).

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+  save_path = "/tmp/model.ckpt"
```
albertz: Don't hardcode such a path. Use `tempfile`. See other test code as example.

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+        # Evaluate vars after update
+        bn_vars[k] = v.eval()
+      return out, bn_vars
+
+  rnd = np.random.RandomState(42)
+  input_data = rnd.rand(10, 11, 16, 16)
+  input_data_nchw = np.transpose(input_data, axes=(0, 3, 1, 2))
+  # Make sure that the data is equal
+  assert np.array_equal(input_data[:, :, :, 2], input_data_nchw[:, 2, :, :])
+  seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: As I wrote before, you should not use this case where all seq lens are the same. You are hiding many bugs by that.

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+        feed_dict={
+          src_nhwc.output.placeholder: data, src_nhwc.output.size_placeholder[0]: seq_lens,net.train_flag: False})
+      return np.array(out)
+
+  rnd = np.random.RandomState(42)
+  input_data = rnd.rand(10, 11, 16, 16)
+  input_data_nchw = np.transpose(input_data, axes=(0, 3, 1, 2))
+  # Make sure that the input data is equal
+  assert np.array_equal(input_data[:, :, :, 2], input_data_nchw[:, 2, :, :])
+  seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: Don't use these seq lens.

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+          "feature_dim_axis": 3,
+          "sparse": False})
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    rnd = np.random.RandomState(42)
+    mean =  tf.constant(rnd.rand(16), name="rand_mean", dtype=tf.float32)
+    variance = tf.constant(rnd.rand(16), name="rand_var", dtype=tf.float32)
+    input_data = rnd.rand(10, 11, 16, 16)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: Don't use that.

tests/test_TFNetworkLayer.py
```diff
@@ -331,10 +362,440 @@ def test_batch_norm_unequal_seq_len():
...
+          "batch_dim_axis": 0,
+          "time_dim_axis": 1,
+          "feature_dim_axis": 3,
+          "sparse": False})
+      src_nhwc.output.placeholder = tf.placeholder(shape=(None, None, 16, 16), dtype=tf.float32)
+      src_nhwc.output.size_placeholder = {0: tf.placeholder(shape=(None,), dtype=tf.int32)}
+
+    rnd = np.random.RandomState(42)
+    input_data = rnd.rand(10, 11, 16, 16)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: Don't use that.

TFNetworkLayer.py
```diff
@@ -11,7 +11,7 @@
 import typing
 import TFUtil
 from Util import unicode, NotSpecified, CollectionReadCheckCovered
-from TFUtil import Data, OutputWithActivation, CustomUpdate, dimshuffle, swapaxes
+from TFUtil import Data, OutputWithActivation, CustomUpdate, CustomUpdateExpAverage, dimshuffle, swapaxes
```
albertz: Don't globally import that. Import it locally where you use it.

albertz:
There are also code conflicts now. Please rebase to the latest version.

Also, please clean up the commits. E.g. a commit message like "batch_norm, new flags added" is not good. Also, e.g. first you add the `fused` flag, but only in a later commit you actually implement it? That does not make sense.

To make the discussion and progress more productive here, it would also help to move unrelated changes to separate pull requests.

tests/test_TFUtil.py
```diff
@@ -2783,6 +2783,34 @@ def body(t, _x):
...
+  x = tf.constant(np.random.rand(10, 16, 11, 16))
+  seq_lens = tf.constant([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
+
+  x_flat = flatten_with_seq_len_mask(x, seq_lens, batch_dim_axis, time_dim_axis)
+
+  x, x_flat = session.run([x, x_flat])
+  npt.assert_array_almost_equal(x[0, :, 1, :], x_flat[1])
+
+
+def test_flatten_with_unequal_seq_len_mask():
```
albertz: The function should always start with `test_`, followed by the function name you are testing, i.e. `flatten_with_seq_len_mask`, and then followed by `_whatever_specific_case_you_test`.

tests/test_TFUtil.py
```diff
@@ -2783,6 +2783,34 @@ def body(t, _x):
   print("magic (totally arbitrary) res:", session.run(x))
 
 
+def test_flatten_with_seq_len_mask():
+  import numpy as np
+  import numpy.testing as npt
+  batch_dim_axis = 0
+  time_dim_axis = 2
+  x = tf.constant(np.random.rand(10, 16, 11, 16))
+  seq_lens = tf.constant([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
```
albertz: That does not make any sense. The whole test is senseless with such seq lens here.

tests/test_TFUtil.py
```diff
@@ -2783,6 +2783,34 @@ def body(t, _x):
...
+  npt.assert_array_almost_equal(x[0, :, 1, :], x_flat[1])
+
+
+def test_flatten_with_unequal_seq_len_mask():
+  import numpy as np
+  import numpy.testing as npt
+  batch_dim_axis = 0
+  time_dim_axis = 2
+  x = tf.constant(np.random.rand(10, 16, 11, 16))
+  seq_lens = tf.constant([11, 11, 5, 11, 11, 11, 11, 11, 11, 11])
```
albertz: The whole test does not make sense with these seq lens.

albertz:
Btw, @mennetob , I think you are one of the main users of batch norm, right? Maybe it make sense if you also keep an eye on this. If we are not careful here, we could either have incompatible changes to earlier versions (so models cannot be imported; or worse, they will be imported without error, but not in the correct way), or also the behavior could maybe change slightly (which should not be the case, but I think it's easy to get this wrong).
Maybe also @doetsch @doetsch-apptek . Or @pavelgolik .

albertz:
This has been implemented now via #888, #898, #897.

Issue #166, opened 2019-04-05
Specify different layer name for parameter loading


ChristopherBrix:
I need to load parameters from a trained model, but modify the config a bit. Specifically, I want to switch the embedding matrix for both source and target. 
Is there a flag which can be used to specify which saved parameters should be used for the initialization? Specifically, I need the following:

`output/rec/target_embed/W` should to be initialized using `source_embed/W` and `source_embed/W` should to be initialized using `output/rec/target_embed/W`

If this is not yet possible, I could write a PR for this feature.

albertz:
There is `preload_from_files`, which should make this possible. See the related code.

albertz:
I don't remember the exact syntax. But maybe sth like that:
```
preload_from_files =  {
'01_init': {
  'init_for_train': ...,
  'filename': ...,
  'prefix': ...
},
'02_swap': {
  'init_for_train': ...,
  'filename': ...,
  'prefix': ...
}
}
```

ChristopherBrix:
If I understand the [code](https://github.com/ChristopherBrix/returnn/blob/master/TFEngine.py#L939) and the [tests](https://github.com/ChristopherBrix/returnn/blob/master/tests/test_TFEngine.py#L1535) correctly, `preload_from_files` only allows me to specify a prefix which is removed from the new parameter name to match the old one. However, this does not allow arbitrary renaming.

I could use it to load the layer `output/rec/target_embed` using `source_embed` by naming it `x_source_embed` and choosing the prefix `output/rec/x_`.
To initialize my new layer `source_embed` using `output/rec/target_embed`, I could probably just name it `output/rec/target_embed`. Not sure whether this is valid in RETURNN, but it's definitely ugly and prone to cause confusion later on.

Did I miss something?

albertz:
Maybe you are right. We can also extend it.


ChristopherBrix <notifications@github.com> schrieb am Fr., 5. Apr. 2019,
19:17:

albertz:
Yet another way might be: Use `forward_weights_init` and `bias_init`. This configures the param init (by default this is random or 0). This can be anything, any custom function, which e.g. loads the right param.

ChristopherBrix:
But `forward_weights_init` is only used during the (random) initialization, right? In my case I want to load a saved checkpoint (with some modifications in the network layout).

[...]

PR #173, opened 2019-04-27
Added option to rename layers using preload_from_files

ChristopherBrix:
This PR allows to rename a layer using the preload_from_files option. The correct syntax is:

    "preload_from_files": {
      'train_base': {
        'filename': model_filename,
        'prefix': 'swap_',
        'init_for_train': True,
        'layer_mapping': {
          'l1': 'l2',
          'l2': 'l1'
        }
      }
    },

if the layer `swap_l1` should be initialized using the trained weights of layer `l2`, and `swap_l2` should be initialized using layer `l1`.
If you don't specify the filename, the current checkpoint will be used. This is useful during inference, when one config file may be used for the evaluation of multiple checkpoints.

This should solve issue #166

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -1947,7 +1947,7 @@ class CustomCheckpointLoader:
   """
 
   def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="", ignore_missing=False,
-               network=None):
+               network=None, layer_mapping=None):
     """
     :param str filename: filepattern for NewCheckpointReader
     :param list[tf.Variable|tensorflow.python.training.saver.BaseSaverBuilder.SaveableObject] saveable_params:
@@ -1964,14 +1964,23 @@ class CustomCheckpointLoader:
     self.params_prefix = params_prefix
     self.load_if_prefix = load_if_prefix
     self.saveable_params = []
+    self.param_mapping = {}
+    layer_mapping = layer_mapping or {}
     for param in saveable_params:
       custom_post_init = getattr(param, "custom_post_init", None)
       if custom_post_init:
         print("Not loading pre-initialized variables %s" % param, file=log.v2)
         continue
-      if load_if_prefix and self._get_param_name(param, assert_load_if_prefix_match=False) is None:
+      param_name = self._get_param_name(param, assert_load_if_prefix_match=False)
+      if load_if_prefix and param_name is None:
         continue
       self.saveable_params.append(param)
+      param_layer_name = "/".join(param_name.split("/")[:-1])
+      if param_layer_name in layer_mapping:
+        self.param_mapping[param_name] = layer_mapping[param_layer_name] + "/" + param_name.split("/")[-1]
+      else:
+        self.param_mapping[param_name] = param_name
+
     assert self.saveable_params, "no saveable vars"
     self.reader = tf.train.NewCheckpointReader(filename)
     self.net_vars = [v for v in self.saveable_params if isinstance(v, tf.Variable)]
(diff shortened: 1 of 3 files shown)


albertz:
I'm not sure if this is not too complicated now. Can't this be simplified in some way?

TFNetwork.py
```diff
@@ -1947,7 +1947,7 @@ class CustomCheckpointLoader:
   """
 
   def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="", ignore_missing=False,
-               network=None):
+               network=None, layer_mapping=None):
```
albertz: Documentation is missing.

TFNetwork.py
```diff
@@ -1964,14 +1964,23 @@ def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="
...
-      if load_if_prefix and self._get_param_name(param, assert_load_if_prefix_match=False) is None:
+      param_name = self._get_param_name(param, assert_load_if_prefix_match=False)
+      if load_if_prefix and param_name is None:
         continue
       self.saveable_params.append(param)
+      param_layer_name = "/".join(param_name.split("/")[:-1])
+      if param_layer_name in layer_mapping:
+        self.param_mapping[param_name] = layer_mapping[param_layer_name] + "/" + param_name.split("/")[-1]
+      else:
+        self.param_mapping[param_name] = param_name
```
albertz: This code is too complicated and too specific and also prone to errors. It should be more generic.

TFNetwork.py
```diff
@@ -1964,14 +1964,23 @@ def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="
     self.params_prefix = params_prefix
     self.load_if_prefix = load_if_prefix
     self.saveable_params = []
+    self.param_mapping = {}
```
albertz: This also makes it more complicated.

albertz: Also, it's not documented. What is the type of this?

TFNetwork.py
```diff
@@ -1981,9 +1990,10 @@ def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="
     # All variables of the model to be loaded:
     self.var_net_names = set([self._get_param_name(v) for v in self.saveable_params])
     # Model variables missing in the checkpoint:
-    self.missing_var_names = [v for v in sorted(self.var_net_names) if v not in self.var_ckpt_names]
+    self.missing_var_names = [v for v in sorted(self.var_net_names) if self.param_mapping[v] not in self.var_ckpt_names]
     # Checkpoint variables which are not used in this model:
-    self.obsolete_var_names = [v for v in sorted(self.var_ckpt_names) if v not in self.var_net_names]
+    self.obsolete_var_names = [v for v in sorted(self.var_ckpt_names) if v not in self.var_net_names and
+                               v not in self.param_mapping.values()]
```
albertz: This is also too complicated now.

tests/test_TFEngine.py
```diff
@@ -66,7 +66,7 @@ def _get_tmp_file(suffix):
   return fn
 
 
-session = tf.InteractiveSession()
+tf_interactive_session = tf.InteractiveSession()
```
albertz: Why this rename?

ChristopherBrix: `session` is used within most tests, too (`with make_scope() as session`). This hides the interactive session. This rename fixes this issue.

TFNetwork.py
ChristopherBrix: Line 1980 could be replaced with `self.param_mapping[param_name] = param_name.replace(param_layer_name, layer_mapping[param_layer_name])`. This should make clearer that we're modifying the name of the layer.
Do you agree?

TFNetwork.py
ChristopherBrix: It's a dictionary `new layer name => old layer name`. I'll add the documentation. 
I'm not sure how we can avoid the additional logic. We could omit this attribute, but then we have to rely on the user defined `layer_mapping`. This has two downsides:
 - the user would have to specify the mapping for all parameters of a layer, not just once
 - layers that are not to be renamed are not part of `layer_mapping`. This is the case e.g. if we just want to perform preloading without renaming. 
Using `self.param_mapping`, we can fix this, by checking whether the parameter if part of any renamed layer, and mapping it to itself otherwise.

TFNetwork.py
albertz: No, what I mean is esp `param_layer_name = "/".join(param_name.split("/")[:-1])`. This is too arbitrary. This is also wrong in some cases. Also, this is too complicated.
I am not sure if I would introduce this layer mapping at all. Maybe param mapping, which is simpler.

albertz:
I'm really not sure about this. I find that the code is already quite complicated and this adds somewhat more to the complexity. And this is maybe for a feature which only you, only right now, maybe only once, need. Instead, you could maybe write a checkpoint convert script. Or use some other workaround (we discussed some in the other issue).

tests/test_TFEngine.py
albertz: Why is this a problem? When you introduce a new `session` in some of the tests, you don't need access to the interactive session.

albertz:
Another idea: In `reuse_params`, maybe you could just specify the scope name.

TFNetwork.py
albertz: If we have to make a tradeoff between easier code and making sth easier for the user, but for some very rarely used feature, I would very much vote for the easier code.

ChristopherBrix:
You may be right - as long as I'm the only person needing this feature, I can keep the code in my private fork. We can revisit this in case anyone else has a need for it, too.

PR #178, opened 2019-05-22
Support max pooling while keeping the batch dimension

ChristopherBrix:
With this PR, it becomes possible to perform a max-/min pooling operation only across the time-dimension, while keeping the batch dimension untouched.

For `tf.reduce_max` and `tf.reduce_min`, we can replace all masked values with the minimal/maximal possible value. This way, they won't be selected during max/min operation.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -3938,7 +3938,7 @@ class ReduceLayer(_ConcatInputLayer):
     x_ = x.placeholder
     # Check if we should ignore some frames, e.g. via masking.
     if use_time_mask:
-      if f is tf.reduce_sum:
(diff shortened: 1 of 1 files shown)


albertz:
I merged it in. Note though that there is a small (probably minor) problem: E.g. for mode="min", if your input has only values of `dtype.max` (i.e. not outside of the sequence), then this value will be the output of this layer. Which is correct. The problem is the gradient: The error signal to this min value will get forwarded normally to the `argmin` entry of the input. If there are multiple min values, it will be distributed across them (see `_MinOrMaxGrad` in TF). And this factor will be slightly wrong now. But as I said, probably this is really minor and does not matter.

PR #179, opened 2019-05-22
Mark time dimension and store sequence length

ChristopherBrix:
If the `ExpandDimsLayer` is used to add a time dimension, the time axis index is now stored as the new `output.time_dim_axis` and an entry in the `output.size_placeholder` dictionary is added.

This is needed, because some layers (at least the `HDFDumpLayer`) expect the time dimension to be variable and to be listed in the `output.size_placeholder`. If `HDFDumpLayer` is the only layer with this behavior, it could also be a bug over there.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -3345,12 +3345,14 @@ class ExpandDimsLayer(_ConcatInputLayer):
     data = self.input_data
     if isinstance(axis, int):
       data = data.copy_as_batch_major()
-    axis = self._get_axis(data=data, axis=axis)
+    axis_index = self._get_axis(data=data, axis=axis)
     from TFUtil import expand_dims_unbroadcast
-    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis, dim=dim)
+    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis_index, dim=dim)
     self.output.size_placeholder = {
-      (i if (data.get_batch_axis(i) < axis) else i + 1): v
+      (i if (data.get_batch_axis(i) < axis_index) else i + 1): v
       for (i, v) in data.size_placeholder.items()}
+    if axis.lower() in ["spatial", "time", "t"] and self.output.time_dim_axis is None:
+      self.output.time_dim_axis = axis_index
 
   @classmethod
   def _get_axis(cls, data, axis):


ArneNx:
I also implemented the same thing locally, but thought I was the only person needing this. 
Also I don't see any reason why this shouldn't be the default behavior when a spatial axis is added.

TFNetworkLayer.py
```diff
@@ -3345,12 +3345,19 @@ def __init__(self, axis, dim=1, **kwargs):
...
-    axis = self._get_axis(data=data, axis=axis)
+    axis_index = self._get_axis(data=data, axis=axis)
     from TFUtil import expand_dims_unbroadcast
-    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis, dim=dim)
+    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis_index, dim=dim)
     self.output.size_placeholder = {
-      (i if (data.get_batch_axis(i) < axis) else i + 1): v
+      (i if (data.get_batch_axis(i) < axis_index) else i + 1): v
       for (i, v) in data.size_placeholder.items()}
+    if axis in ["spatial", "time", "T"]:
```
albertz: Better. `axis.lower()`, and check for `"t"`.

TFNetworkLayer.py
```diff
@@ -3345,12 +3345,19 @@ def __init__(self, axis, dim=1, **kwargs):
...
+      self.output.time_dim_axis = axis_index
```
albertz: It should not be overridden if it was already set before.

TFNetworkLayer.py
```diff
@@ -3345,12 +3345,19 @@ def __init__(self, axis, dim=1, **kwargs):
...
+
+      # The sequence length of the time axis is know to be 1 at this point. But subsequent layers (eg. HDFDumpLayer)
+      # may expect the sequence length to be variable
```
albertz: No, this should not be needed. This should be fixed in `HDFDumpLayer`, or other layers which cannot handle that correctly.

TFNetworkLayer.py
ChristopherBrix: RETURNN does not support multiple time dimensions, right? So we could assert that it is not yet set.

albertz:
About adding a dynamic seq len: No, in principle, this should not be needed. And this is also not a good idea. E.g. maybe you have some other layer which can only work correctly if the spatial axis is static. And introducing this variable length (which is actually static) will make it hard to check for that. I'm also quite sure that there are such layers, or such cases already, although I don't have some specific in mind. Although, at least you make a lot of other layers also slower. For example `ReduceLayer` will do some extra masking if it reduces on some axis which has variable length.

TFNetworkLayer.py
albertz: Well, it supports, kind of. `time_dim_axis` can only be a single axis. But that does not matter very much in many cases. It can have multiple spatial axes, and it also can have multiple axes with variable length. And when you remove the axis which is currently `time_dim_axis` (e.g. by `ReduceLayer` or so), and there is another spatial axis, it usually will switch `time_dim_axis` over to the next one.

ChristopherBrix:
Ok, then I will fix this issue in the `HDFDumpLayer`.

TFNetworkLayer.py
```diff
@@ -3345,12 +3345,14 @@ def __init__(self, axis, dim=1, **kwargs):
...
+    axis_index = self._get_axis(data=data, axis=axis)
     from TFUtil import expand_dims_unbroadcast
-    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis, dim=dim)
+    self.output.placeholder = expand_dims_unbroadcast(data.placeholder, axis=axis_index, dim=dim)
     self.output.size_placeholder = {
-      (i if (data.get_batch_axis(i) < axis) else i + 1): v
+      (i if (data.get_batch_axis(i) < axis_index) else i + 1): v
       for (i, v) in data.size_placeholder.items()}
+    if axis.lower() in ["spatial", "time", "t"] and self.output.time_dim_axis is None:
+        self.output.time_dim_axis = axis_index
```
albertz: The indentation here is too much.

albertz: Actually the logic to set `time_dim_axis` should better also be in `get_out_data_from_opts`, and not in `__init__`.

ChristopherBrix: Ok, I'll change that tomorrow.

ChristopherBrix:
`get_out_data_from_opts` [already marks the time dimension axis](https://github.com/rwth-i6/returnn/blob/master/TFNetworkLayer.py#L3402). So this PR is not needed.

What's left to do is to fix the `HDFDumpLayer`, so it doesn't expect the time dimension to be specified in the `size_placeholder`. I'll work on this next week, and create a new PR.

PR #181, opened 2019-05-23
removed unnecessary constraints on PaddingLayer

JackTemaki:
I am not sure why the code was written like this before, but i checked all tensorflow versions from 1.8 to 1.13, and it has always been that constant_value is assumed to be 0 by default, and it is only used when the mode is "CONSTANT", but also accepts every value given.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -2886,7 +2886,7 @@ class PadLayer(_ConcatInputLayer):
   """
   layer_class = "pad"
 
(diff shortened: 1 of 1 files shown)


albertz:
Note that some people used 1.4.1 for quite a long time, or maybe still (also myself; it was the last TF where the official binary runs on all CPUs). And maybe even older. Although I would say there is no reason for anyone not to upgrade to 1.4.1. Anyway, merged it in now. Let's see if anyone complains with problems.

PR #184, opened 2019-06-05
add support for character based vocabulary

JackTemaki:


albertz:
Why do we need that? We already have the class `CharacterTargets`. What is the difference?

JackTemaki:
> Why do we need that? We already have the class `CharacterTargets`. What is the difference?

The CharacterTargets is a different Class, and i want to control the Vocabulary from the Config. By specifying an additional Parameter in Vocabulary, i can do the following in my Sprint dataset without touching other code:
```
train = {'class': 'ExternSprintDataset',
  ...
  'orth_vocab': {'vocab_file': "orth_vocab.pkl", 'unknown_label': None, 'character_based': True},
  ...
}
```
I think having it this way is more flexible

JackTemaki:
And yes, if you run on english you do not need the vocabulary, as you can use the 'orth' of the Sprint dataset directly, but this only works for english, not for other languages with non-ascii symbols.

albertz:
No, we should not duplicate functionality. In `SprintDatasetBase`, we just call `Vocabulary(**orth_vocab)` for historical reasons. We should instead just call `Vocabulary.create_vocab(**orth_vocab)`, and it should work already.

JackTemaki:
Ok, for me it was not clear that this functionality exists, then i will change the `SprintDatasetBase` accordingly, and add some comments

PR #187, opened 2019-06-07
Bugfix for HDFDumpLayer: Support for sequences of fixed length

ChristopherBrix:
As previously mention in PR #179, the HDFDumpLayer expected the time dimension to be a sequence of variable length, ie. `size_placeholder` had to be set.
This PR makes the input requirements (a sequence, not a single state; only one input, not multiple) explicit and allows the sequence length to be fixed.

If the dump layer is supposed to be able to handle an input that is only a single state and not a whole sequence, this is yet another issue. For now, a simple workaround is to first pass the individual state through the `expand_dims` layer, and then dump it afterwards. This approach is used in my test case.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -6271,6 +6271,8 @@ class HDFDumpLayer(LayerBase):
     :param bool dump_whole_batches: dumps the whole batch as a single sequence into the HDF
     """
     super(HDFDumpLayer, self).__init__(**kwargs)
+    assert len(self.sources) == 1
+    assert self.sources[0].output.have_time_axis()
     self.output = self.sources[0].output.copy("%s_output" % self.name)
     data = self.output.copy_as_batch_major()  # need batch-major for SimpleHDFWriter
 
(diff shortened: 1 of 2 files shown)


TFNetworkLayer.py
```diff
@@ -6189,6 +6192,12 @@ def py_write(data_np, tags, *sizes):
         n_batch = data_np.shape[0]
         assert len(sizes) == len(data.size_placeholder)
         seq_lens = {i: size for (i, size) in zip(sorted(data.size_placeholder.keys()), sizes)}
+        # there may be axes with a fixed length other than the batch and feature axes
+        # these have the indices 0, ..., (ndim-1), as the batch dimension is skipped
+        for dim in range(ndim-1):
```
albertz: Please use PyCharm and check for Python formatting warnings.

ChristopherBrix: PyCharm didn't warn me about any issues - I've capitalized the beginning of the sentences and added periods. Is that what you're referring to? (Sorry for the late reply, I mistakenly never submitted my comment)

albertz: `ndim-1`. There should be spaces before and after the operand `-`.

ChristopherBrix:
Can this be merged?

PR #192, opened 2019-06-26
made loss_norm_factor a tf.Tensor

tran-khoa:
it appears tensorflow needs a tensor here


albertz:
Why do you think so?
(I guess you try to fetch it as-is, but have not defined any other loss or so, in training, which is a case which likely will not happen; anyway, then for the fetch, you should use `tf.convert_to_tensor` in there or so; this should be done here.)

albertz:
Ping?

albertz:
Hello?

albertz:
Closing for now. Please respond if you there is a problem.

PR #193, opened 2019-06-26
enable end layer usage without a choice layer

JackTemaki:
Currently there was no way to use the end layer for stopping decoding without having a choice layer. The test case is modelling a simple sigmoid based end layer, which is similar to the network needed for TTS.


albertz:
Note that this could also have been realized by just making a ChoiceLayer based on the sigmoid output, and then the end_layer checking on that.
But I guess we can also merge this in.
(Waiting for tests now.)

PR #194, opened 2019-06-27
Implemented relative positional encoding

tran-khoa:
Relative positional encoding (Shaw et al., 2018) propose adding a distance term in self-attention instead of using sinusoids after embedding. This PR introduces a relative positional encoding layer that should take the same input as the self-attention layer.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -4493,9 +4493,9 @@ class SelfAttentionLayer(_ConcatInputLayer):
       q_t = tf.transpose(q, [2, 0, 1, 3])  # [num_queries|1,batch,heads,key_dim]
       q_t_r = tf.reshape(
         q_t, [tf.shape(q_t)[0], batch_dim * num_heads, total_key_dim // num_heads])  # [num_queries|1,batch*heads,k-dim]
-      with tf.control_dependencies(tf.assert_equal(
+      with tf.control_dependencies([tf.assert_equal(
             message="check_shape_of_key_shift:",
-            x=tf.shape(k_), y=[tf.shape(q_t)[0], tf.shape(energy)[-1], total_key_dim // num_heads])):
+            x=tf.shape(k_), y=[tf.shape(q_t)[0], tf.shape(energy)[-1], total_key_dim // num_heads])]):
         energy_ = tf.matmul(q_t_r, k_, transpose_b=True)  # [num_queries|1,batch*heads,num_keys]
       energy_ = tf.reshape(
         energy_, [tf.shape(q_t)[0], batch_dim, num_heads, tf.shape(energy)[-1]])  # [num_queries|1,batch,heads,num_keys]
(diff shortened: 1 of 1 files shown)


tran-khoa:
Example usage:
```
d[output + '_rel_pos'] = {"class": "relative_positional_encoding", 
                          "from": [output + '_self_att_laynorm'],
                          "n_out": self.EncKeyTotalDim // self.AttNumHeads,
                          "forward_weights_init": self.ff_init}
d[output + '_self_att_att'] = {"class": "self_attention", 
                               "num_heads": self.AttNumHeads,
                               "total_key_dim": self.EncKeyTotalDim,
                               "n_out": self.EncValueTotalDim, "from": [output + '_self_att_laynorm'],
                               "attention_left_only": False, "attention_dropout": self.attention_dropout,
                               "forward_weights_init": self.ff_init, "key_shift": output + '_rel_pos'}
```

TFNetworkRecLayer.py
```diff
@@ -6116,3 +6116,60 @@ def __call__(self, inputs, state, scope=None):
     new_state = tf.nn.rnn_cell.LSTMStateTuple(c, h)
 
     return output, new_state
+
```
albertz: There must be two empty lines before a new class. (For all code, always, ever: Keep it consistent to the existing code.) (Please use PyCharm, and check for the warnings. PyCharm should have notified about that.)

TFNetworkRecLayer.py
```diff
@@ -6116,3 +6116,60 @@ def __call__(self, inputs, state, scope=None):
...
+  Parts of the code are adapted from Tensor2Tensor (https://github.com/tensorflow/tensor2tensor).
+  """
+  layer_class = "relative_positional_encoding"
+  recurrent = True
+
+  def __init__(self, n_out, forward_weights_init="glorot_uniform", clipping=16, fixed=False, **kwargs):
+    """
+    :param n_out: Feature dimension of encoding.
+    :param clipping: After which distance to fallback to the last encoding
+    :param fixed: Uses sinusoid positional encoding instead of learned parameters
```
albertz: The argument types are not specified (`int n_out`, `int clipping`, `bool fixed`). Always specify them if possible.
Also, you don't specify all arguments (`forward_weights_init` is missing).

TFNetworkRecLayer.py
```diff
@@ -6116,3 +6116,60 @@ def __call__(self, inputs, state, scope=None):
...
+    distance_mat_clipped = tf.clip_by_value(distance_mat, -clipping, clipping)
+    # Shift values to be >= 0. Each integer still uniquely identifies a relative
+    # position difference.
+    position_info_indices = distance_mat_clipped + clipping
+
+    encoding = tf.gather(encoding_matrix, position_info_indices)
+
+    self.output.placeholder = encoding
+    self.output.shape = (None, None, n_out)
+    self.output.batch_dim_axis = None
```
albertz: You must not change `self.output.shape` (or any others, except `output.placeholder and `output.size_placeholder`) here in `__init__`. Rather, you must define the class method `get_out_data_from_opts`, which defines what the output type/shape (`Data`) is.

tran-khoa:
Alright, I have changed the code accordingly.

albertz:
Please clean up the commits (I think this probably should be just a single commit).

tran-khoa:
Done.

Issue #195, opened 2019-07-05
LearningRateControl should be able to use combined losses


ChristopherBrix:
I just realized that the learning rate control logic depends on the name of the output layer, if multiple outputs are defined.

If there is only one output defined, [`dev_score` and `dev_error` are set to the respective values of the output layer](https://github.com/rwth-i6/returnn/blob/master/LearningRateControl.py#L260).
For multiple outputs, [they are stored like `dev_score_LAYER_NAME`](https://github.com/rwth-i6/returnn/blob/master/LearningRateControl.py#L262).

Later, these values are needed to determine whether the LR needs to be updated. For one output, [this is easy, because the key value (`dev_score`) is known](https://github.com/rwth-i6/returnn/blob/master/LearningRateControl.py#L296).
For multiple outputs, where no output layer is called `output`, [the first (alphabetically sorted) `dev_` entry is used](https://github.com/rwth-i6/returnn/blob/master/LearningRateControl.py#L306).

This is already not intuitive, I would have expected that for two outputs, the two losses are eg. added.
But in my case, the `decision` layer happens to be the first layer found. But its value is 0, so later,  [the computation of the relative error leads to NaN](https://github.com/rwth-i6/returnn/blob/master/LearningRateControl.py#L235), and my learning rate is never reduced.

Before I write the PR for this, I would like to clarify whether there is a reason to choose just one of the output layers for the LR scheduling. Should the default be to add all losses, and do we need a flag to ensure backwards compatibility?

JackTemaki:
Do I understand it correct, that the return value (key) is arbitrary (alphabetic) when there is no layer called 'output'?

ChristopherBrix:
If you only have one output layer, then the name should not matter. 
But if you have multiple output layers, then yes, the alphabetically first output layer name is chosen for the learning rate scheduling.

ChristopherBrix:
Oh, but you probably have a `decision` layer, right? And for all of my experiments, this layer logs an error of 0. This never really bothered me, I just ignored it. But if you have the same setup, then you automatically have at least 2 output layers, and now the name of your "real" output layer does matter. It should be "output", or some alphabetically smaller name than "decision". 
I will test this later. It should be visible in the logs, if your learning rate never decreases.

JackTemaki:
This is my learning rate file with multiple outputs (no decision though):

```
200: EpochData(learningRate=6.461081889226677e-05, error={
'dev_error_decoder/output': 0.7288903655845147,
[...]

PR #196, opened 2019-07-05
Meta Loss to combine losses for learning rate scheduling

ChristopherBrix:
This introduces a `MetaLoss`. It can be configured by the `meta_loss` setting in the config file. It may be any mathematical operation, using all defined losses, eg: `5 * decoder_src_output_prob_src + decoder_trg_output_prob_trg` (here, I have two recursive layers, `decoder_src` and `decoder_trg`, each containing `output_prob_src` and `output_prob_trg`, respectively.

This is meant to be used to control the learning rate scheduling if there are multiple outputs given, see #195.
The logic to actually choose the meta loss still needs to be implemented.

### DIFF
--- a/LearningRateControl.py
+++ b/LearningRateControl.py
@@ -236,6 +236,8 @@ class LearningRateControl(object):
       relative_error = (new_error - old_error) / abs(old_error)
     else:
       relative_error = (new_error - old_error) / abs(new_error)
+    import math
+    assert not math.isnan(relative_error)
     if self.relative_error_also_relative_to_learning_rate:
       learning_rate = self.get_most_recent_learning_rate(new_epoch, exclude_current=False)
       # If the learning rate is lower than the initial learning rate,
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -730,6 +730,11 @@ class TFNetwork(object):
             else:
               total_constraints += constraints
 
+    if self._config and self._config.has('meta_losses'):
+      for name, formula in self._config.typed_value('meta_losses', []):
+        assert name not in losses_dict
+        losses_dict[name] = MetaLossHolder(losses_dict, formula)
+
     return losses_dict, total_loss, total_constraints
 
   def _construct_objective(self):
@@ -1965,6 +1970,57 @@ class LossHolder:
       loss=self.loss, reduce_func=reduce_func,
       loss_value=loss_value, error_value=error_value,
       norm_factor=self._norm_factor, only_on_eval=self._only_on_eval)
+
+
+class MetaLossHolder:
+  """
+  Combines multiple losses in a customizable way.
+  It can be eg. used to compute the average over multiple losses which may then control the learning rate.
+  Depending on the way the losses are combined, the result may not be normalized.
+  This class supports only a subset of the functions in LossHolder. It is not supposed to be used eg. to compute
+  the objective function.
+  """
+  def __init__(self, losses_dict, eval_string):
+    """
+    :param dict[str,TFNetworkLayer.Loss] losses_dict: Mapping from loss names to losses.
+    :param string eval_string: Formula to combine the relevant losses. Will be eval'ed.
+    """
+    self.eval_string = eval_string
+
+    import re
+    term_pattern = re.compile('([a-zA-Z0-9_-]+)')
+    terms = re.findall(term_pattern, self.eval_string)
+    self.used_losses = {k.replace("/", "_"): v for k,v in losses_dict.items() if k.replace("/", "_") in terms}
+
+  def get_only_on_eval(self):
+    """
+    :return: only_on_eval flag. assumes that it is set for all combined losses
+    :rtype: bool
+    """
+    return any([used_loss.get_only_on_eval() for used_loss in self.used_losses.values()])
+
+  def get_loss_value_for_fetch(self):
+    """
+    :return: loss value for fetch. scalar.
+    :rtype: tf.Tensor|None
+    """
+    d = {key: loss.get_loss_value_for_fetch() * loss.get_norm_factor() for key, loss in self.used_losses.items()}
+    return eval(self.eval_string, d)
+
+  def get_error_value(self):
+    """
+    :return: error value for fetch. scalar
+    :rtype: tf.Tensor|None
+    """
+    d = {key: loss.get_error_value() * loss.get_norm_factor() for key, loss in self.used_losses.items()}
+    return eval(self.eval_string, d)
+
+  def get_norm_factor(self):
+    """
+    :return: always 1, because we cannot normalize arbitrary combined losses. scalar
+    :rtype: tf.Tensor
+    """
+    return tf.constant(1, dtype=tf.float32)
 
 
 class NetworkConstructionDependencyLoopException(Exception):
--- a/docs/configuration_reference/optimizer_settings.rst
+++ b/docs/configuration_reference/optimizer_settings.rst
@@ -61,7 +61,10 @@ optimizer_epsilon
 
 user_learning_rate_control_always
 
-
+meta_losses
+    A list of (name, formula) tuples that define arbitrary meta losses.
+    Every meta loss may combine any number of other losses. The formula is a string, which will be eval'ed.
+    This can be used in combination with learning_rate_control_error_measure to make the LR scheduling dependant on multiple losses.
 
 
 


LearningRateControl.py
```diff
@@ -236,6 +236,8 @@ def calc_relative_error(self, old_epoch, new_epoch):
       relative_error = (new_error - old_error) / abs(old_error)
     else:
       relative_error = (new_error - old_error) / abs(new_error)
+    import math
+    assert not math.isnan(relative_error)
```
albertz: This can happen in certain situations (some losses sometimes can result in inf, and that can be ok in some situations), and I think it should not error here.

TFNetwork.py
```diff
@@ -1967,6 +1972,57 @@ def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
       norm_factor=self._norm_factor, only_on_eval=self._only_on_eval)
 
 
+class MetaLossHolder:
```
albertz: It is not a good idea to introduce a class which is sort-of-like `LossHolder` but not quite. It should properly derive from the base class. You can throw an exceptions in any functions which are not supported or so.

TFNetwork.py
```diff
@@ -1967,6 +1972,57 @@ def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
...
+    """
+    :param dict[str,TFNetworkLayer.Loss] losses_dict: Mapping from loss names to losses.
+    :param string eval_string: Formula to combine the relevant losses. Will be eval'ed.
+    """
+    self.eval_string = eval_string
+
+    import re
+    term_pattern = re.compile('([a-zA-Z0-9_-]+)')
+    terms = re.findall(term_pattern, self.eval_string)
+    self.used_losses = {k.replace("/", "_"): v for k,v in losses_dict.items() if k.replace("/", "_") in terms}
```
albertz: A space is missing between `k,v`. (Does PyCharm not warn about this?)
Also, isn't the line too long?

TFNetwork.py
```diff
@@ -1967,6 +1972,57 @@ def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
...
+  """
+  Combines multiple losses in a customizable way.
+  It can be eg. used to compute the average over multiple losses which may then control the learning rate.
+  Depending on the way the losses are combined, the result may not be normalized.
+  This class supports only a subset of the functions in LossHolder. It is not supposed to be used eg. to compute
+  the objective function.
+  """
+  def __init__(self, losses_dict, eval_string):
+    """
+    :param dict[str,TFNetworkLayer.Loss] losses_dict: Mapping from loss names to losses.
```
albertz: That's the wrong type, or not? Isn't it `dict[str,LossHolder]`? (Does PyCharm not warn about this?)

albertz:
I would not call it meta loss. There is already some other concept in RETURNN which is called meta loss (see class `MetaLosses`). This is used e.g. to get a loss for other losses or for other gradients (e.g. if you want to to regularization on the gradients, or also implement sth like synthetic gradients, or so). This is also only used during training, because it assumes that you calculate the gradient for the other losses (the non-meta losses), and it wants access to that.

I'm also not sure if it is a good idea to merge your requirements into that class/concept. These meta losses are only added by the updater, later when the other losses and their gradients are already computed.

But then, your thing should be called differently. E.g. sth like `CombinedLosses` or so.

Note that the whole thing is actually not really needed, as you could already do (almost) the same via `EvalLayer`. Do you really think there is an advantage of adding this?

TFNetwork.py
```diff
@@ -1967,6 +1972,57 @@ def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
...
+    :param string eval_string: Formula to combine the relevant losses. Will be eval'ed.
```
albertz: `string` is the wrong type. It is called `str`. (Please always check in PyCharm that the types are correctly inferred.)

TFNetwork.py
```diff
@@ -1967,6 +1972,57 @@ def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
...
+    """
+    self.eval_string = eval_string
+
+    import re
+    term_pattern = re.compile('([a-zA-Z0-9_-]+)')
```
albertz: Why the last `-`?

TFNetwork.py
albertz: Also, I don't understand, how can `/` be in `k`?

LearningRateControl.py
albertz: (Also, this is anyway not really related to your pull request.)

TFNetwork.py
ChristopherBrix: If `-` is the last character of the regex, it is treated as a regular character, not a "from-to sign". And `-` could be a part of a valid layer name, right? So we need to include it here.

LearningRateControl.py
ChristopherBrix: Ok, I didn't know that this can happen in valid situations. Do you think a warning should be printed?

I added this here, because such a warning would have made me aware of the failed LR scheduling much sooner (as written in the issue, the loss couldn't be found, so the error became NaN, disabling the LR scheduling without any notice). But I can remove it/create a separate PR for this.

ChristopherBrix:
> Note that the whole thing is actually not really needed, as you could already do (almost) the same via `EvalLayer`. Do you really think there is an advantage of adding this?

How would you implement this in an eval layer? In my setup, I'm training one network to generate two distinct sequences. Each has a loss, but I would like the LR to depend on both.

LearningRateControl.py
albertz: Normally, the training should stop if there is a nan/inf. You should get the message "Model seems broken, got inf or nan final score", and then it stops (unless you set `stop_on_nonfinite_train_score`).
You explicitly have to set `stop_on_nonfinite_train_score = False` in the config to disable this. But if you do you, you actually also say that it inf/nan scores are valid.
Anyway, this is handled elsewhere, and should be handled elsewhere in any case, not by the LR scheduling. If you don't like any of the behavior, create a new issue and let's discuss over there, but remove it from here to keep us focused here.

albertz:
Well, you can in general just calculate anything what you want in an EvalLayer. I.e. sth like this:

```
"my_custom_loss": {
"class": "eval",
"eval": ...,  # whatever
"from": ...,  # whatever you need to calculate it
"loss": "as_is",
"loss_scale": 0.0,  # so that the optimizer will not use this
"out_type": {...}  # depends if you define it frame-wise or sequence-wise...
}
```

LearningRateControl.py
ChristopherBrix: Interesting. I'll check why it didn't abort in my case. But I agree, I'll remove it here.

TFNetwork.py
albertz: No, I think it can not be part of a layer name (or should not).
Anyway, this is not a good idea, as a string like "a-b" becomes ambiguous (is that loss a - loss b or the layer with the name "a-b")

albertz:
What is the state here?

ChristopherBrix:
I understand that I can use the output of the `eval` layer as my loss, but I don't know how to compute the correct loss in there.

As far as I know, I can give a list of layer names in the `from` parameter, and get the corresponding `Data` objects. But these don't contain the losses. So I would have to compute the cross entropy loss myself (duplicating the code), to combine multiple losses?

I have a layer
```
'output_prob': { 'class': 'softmax',
  'dropout': 0.3,
  'from': ['readout'],
  'loss': 'ce',
  'loss_opts': {'label_smoothing': 0.1},
  'target': 'classes', 'is_output_layer': True},
```
inside my `RecLayer`, and 
```
"output_prob_y" : {"class": "get_rec_accumulated", "from": ["output"], "sub_layer": "readout"},
```
outside.
But this output is after `softmax` is applied, so I can no longer apply label smoothing. I would need the `output_before_softmax`, but I don't know where to access it.

albertz:
In general, you can just calculate anything in Returnn. If not, this is a separate issue, and you can open an issue (or pull request) for that.
Also in general, via `EvalLayer`, you can almost calculate anything. Surely it is not a problem to calculate a cross-entropy loss in there. It would look sth like `"eval": "tf.nn.sparse_softmax_cross_entropy_with_logits(...)"`. Or maybe `smoothing_cross_entropy`. Or whatever you like.
You maybe could even call `layer.get_losses` directly (I'm not sure if I maybe forget sth, why this is a bad idea or so).
Label smoothing is not sth you apply before or after softmax. Label smoothing just changes the hard 0-1 target labels to soft targets.
(Btw, off-topic, don't use the `get_rec_accumulated` layer anymore. There is no need for it. You can directly access any sub-layer of a rec-layer via e.g. `"output/readout"` for your example.)

ChristopherBrix:
Ok, thanks. I'll try it that way. Then I guess this PR is not needed.

PR #198, opened 2019-07-16
Efficient HDF creation tool for TranslationDatasets

patrick-wilken:
When using huge MT corpora (100M + lines) hdf_dump.py becomes unusable because of memory usage and run time.

This script is much faster (e.g. in my tests factor >30 on 7M lines) and uses only a fraction of the memory (e.g. 0.6 vs. 9GB on 7M sentences). The output HDF file contains exactly the same data.

The issue with hdf_dump.py is that it accesses the file for every sequence, which is slow. Also, it uses TranslationDataset, which loads the whole corpus into memory.

hdf_dump could be improved by buffering as in this script. But TranlsationDataset as implemented now cannot be used for that large amounts of data.

I realize a generic solution to create HDF files is more desirable, but I think this special solution is worth sharing.

### DIFF
--- a/tools/hdf_dump_translation_dataset.py
+++ b/tools/hdf_dump_translation_dataset.py
@@ -1,0 +1,321 @@
+#!/usr/bin/env python3
+
+import sys
+import argparse
+import gzip
+import pickle
+
+import numpy
+import h5py
+from operator import itemgetter
+from itertools import islice, zip_longest
+
+UNKNOWN_LABEL = "<UNK>"
+POSTFIX = " </S>"
+
+
+class HDFTranslationDatasetCreator(object):
+  """
+  Creates the same HDF file as hdf_dump.py called on a TranslationDataset, but does so much faster
+  and using much less memory.
+
+  Input parameters are basically the files in a TranslationDataset folder. Additionally, the number of lines has to
+  be given.
+  """
+
+  def __init__(self, hdf_file, source_file, target_file, source_vocabulary, target_vocabulary, number_of_lines,
+               compression_method=None, line_buffer_size=100000, data_buffer_size=5000000):
+    """
+    :param str hdf_file: filename for the hdf file being created
+    :param str source_file: filename of the source text file
+    :param str target_file: filename of the target text file
+    :param str source_vocabulary: filename of the source vocabulary (in pickle format)
+    :param str target_vocabulary: filename of the target vocabulary (in pickle format)
+    :param int number_of_lines: line count in source_file and target_file
+    :param str compression_method: Optional compression method as supported by h5py.File.create_dataset().
+        Applied to the main data only ('inputs' and 'target/data/classes').
+    :param int line_buffer_size: number of corpus lines to read and process at once
+    :param int data_buffer_size: space to reserve inside the hdf file at once, in numbers of integers
+    """
+
+    self.hdf_file_name = hdf_file
+    self.hdf_file = None
+    self.compression_method = compression_method
+
+    self.source_file_handle = self._open_file(source_file)
+    self.target_file_handle = self._open_file(target_file)
+
+    source_vocabulary = self._read_vocabulary(source_vocabulary)
+    target_vocabulary = self._read_vocabulary(target_vocabulary)
+
+    self._vocabularies = {"source": source_vocabulary, "target": target_vocabulary}
+    self._vocabulary_sizes = {"source": len(source_vocabulary), "target": len(target_vocabulary)}
+    self._unknown_ids = {"source": source_vocabulary[UNKNOWN_LABEL], "target": target_vocabulary[UNKNOWN_LABEL]}
+
+    self.number_of_lines = number_of_lines
+    self.line_buffer_size = line_buffer_size
+    self.data_buffer_size = data_buffer_size
+
+    self._number_of_processed_lines = 0
+    self._write_offsets = {"source": 0, "target": 0}
+
+  def create(self):
+    """
+    Main function writing the HDF file.
+    """
+    self._init_hdf_file()
+
+    print("Setting attributes...", file=sys.stderr)
+    sys.stderr.flush()
+    self._write_attributes()
+    print("Done.", file=sys.stderr)
+
+    print("Writing labels (vocabulary)...", file=sys.stderr)
+    sys.stderr.flush()
+    self._write_labels()
+    print("Done.", file=sys.stderr)
+
+    print("Writing source, target, sequence lengths and tags for all lines...", file=sys.stderr)
+    sys.stderr.flush()
+    end_of_file = False
+    while not end_of_file:
+      end_of_file = self._write_data()
+      if not end_of_file:
+        print("> Processed {} lines.".format(self._number_of_processed_lines), file=sys.stderr)
+        sys.stderr.flush()
+    print("Done.", file=sys.stderr)
+    sys.stderr.flush()
+
+    self.hdf_file.close()
+
+  def _init_hdf_file(self):
+    """
+    Sets up the HDF file and initializes the datasets that will be filled.
+    """
+    self.hdf_file = h5py.File(self.hdf_file_name, "w")
+
+    self.hdf_file.create_group("targets/data")
+    self.hdf_file.create_group("targets/size")
+    self.hdf_file.create_group("targets/labels")
+
+    self.hdf_file.create_dataset("seqLengths", (self.number_of_lines, 2), dtype="int32")
+    max_tag_length = len("line-") + len(str(self.number_of_lines))
+    self._tag_dtype = "S{}".format(max_tag_length)
+    self.hdf_file.create_dataset("seqTags", (self.number_of_lines,), dtype=self._tag_dtype)
+
+    self.hdf_file.create_dataset("inputs", (self.data_buffer_size,), maxshape=(None,),
+                                 dtype="int32", compression=self.compression_method)
+    self.hdf_file["targets/data"].create_dataset("classes", (self.data_buffer_size,), maxshape=(None,),
+                                                 dtype="int32", compression=self.compression_method)
+
+  def _write_attributes(self):
+    """
+    Writes several attributes to the HDF file.
+    """
+    self.hdf_file["targets/size"].attrs["classes"] = (self._vocabulary_sizes["target"], 1)
+
+    # Those should be deprecated, but include nevertheless to exactly reproduce hdf_dump.
+    self.hdf_file.attrs["inputPattSize"] = self._vocabulary_sizes["source"]
+    self.hdf_file.attrs["numLabels"] = self._vocabulary_sizes["target"]
+
+  def _write_labels(self):
+    """
+    Writes the labels (i.e. target vocabulary) to the HDF file.
+    """
+    sorted_vocabulary_tuples = sorted(self._vocabularies["target"].items(), key=itemgetter(1))
+    labels = [word.encode("utf8") for (word, _) in sorted_vocabulary_tuples]
+
+    assert len(labels) == self._vocabulary_sizes["target"], "Word ids were not unique."
+
+    max_label_length = max([len(label) for label in labels])
+    dtype = "S{}".format(max_label_length + 1)
+
+    labels = [numpy.array(label, dtype=dtype, ndmin=1) for label in labels]
+    labels = numpy.concatenate(labels)
+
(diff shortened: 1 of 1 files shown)


tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,253 @@
...
+import h5py
+from operator import itemgetter
+from itertools import islice, zip_longest
+
+UNKOWN_LABEL = "<UNK>"
+POSTFIX = " </S>"
+
+
+class HDFTranslationDatasetCreator(object):
+  def __init__(self, hdf_file, source_file, target_file, source_vocabulary, target_vocabulary, number_of_lines,
```
albertz: The arguments (including types) are not documented. (Please enable the warning about missing argument documentation in PyCharm to get notified about this.)

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,253 @@
...
+    line = line.strip() + POSTFIX
+    words = line.split()
+
+    indices = [self._vocabularies[side].get(word, self._unknown_ids[side]) for word in words]
+
+    indices_numpy = numpy.array(indices, dtype=numpy.int32)
+    return indices_numpy
+
+  @staticmethod
+  def _open_file(file_name):
```
albertz: Not documented. Probably type `str`?

patrick-wilken:
I added documentation.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+
+      source_indices = self._line_to_indices(source_line, "source")
+      target_indices = self._line_to_indices(target_line, "target")
+
+      source_chunk.append(source_indices)
+      target_chunk.append(target_indices)
+
+    return source_chunk, target_chunk
+
+  def _write_lengths(self, source_lengths, target_lengths):
```
albertz: Arguments not documented here.

albertz: Also, if these are numpy arrays (I don't know...), you should document the dtype and shape.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+
+  def _write_tags(self, chunk_size):
+    offset = self._number_of_processed_lines
+
+    tags = [numpy.array("line-" + str(offset + i), dtype=self._tag_dtype, ndmin=1) for i in range(chunk_size)]
+    tags = numpy.concatenate(tags)
+
+    self.hdf_file["seqTags"][offset: offset + chunk_size] = tags
+
+  def _write_data_indices(self, chunk, side):
```
albertz: Arguments.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+    # Make sure the number of lines given by the user was correct.
+    # Otherwise lengths and labels would have trailing zeros.
+    assert self.number_of_lines == self._number_of_processed_lines, "Fewer lines ({}) in the corpus files " \
+        "than specified ({}).".format(self._number_of_processed_lines, self.number_of_lines)
+
+    # Trim datasets to actually occupied length, i.e. remove unused reserved space.
+    self.hdf_file["inputs"].resize((self._write_offsets["source"],))
+    self.hdf_file["targets/data/classes"].resize((self._write_offsets["target"],))
+
+  def _line_to_indices(self, line, side):
```
albertz: Arguments.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+  @staticmethod
+  def _open_file(file_name):
+    """
+    :param str file_name: filename of a plain text file, possibly zipped
+    :return: file handle
+    :rtype: TextIOWrapper|GzipFile
+    """
+    if file_name.endswith(".gz"):
+      open_function = gzip.open
+      mode = "rt"
```
albertz: Why not directly `return gzip.open(file_name, "rt")`? And then `else:` and `return open(file_name, "r")`. Would be simpler and shorter and easier to read.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+      mode = "r"
+
+    file_handle = open_function(file_name, mode)
+    return file_handle
+
+  @staticmethod
+  def _read_pickle_file(file_name):
+    """
+    :param str file_name: filename of a pickle file
+    :return: contents of the pickle file
```
albertz: What return type do you expect here?

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,280 @@
...
+
+    indices_numpy = numpy.array(indices, dtype=numpy.int32)
+    return indices_numpy
+
+  @staticmethod
+  def _open_file(file_name):
+    """
+    :param str file_name: filename of a plain text file, possibly zipped
+    :return: file handle
+    :rtype: TextIOWrapper|GzipFile
```
albertz: I think it should be `:rtype: io.TextIOWrapper|gzip.GzipFile`.

tools/hdf_dump_translation_dataset.py
patrick-wilken: The function is meant to be generic. But when loading vocabularies dict[str, int]. Should I add this?

albertz: Yes, and also rename the function to `_read_vocab` (you don't need a generic pickle loading function).

tools/hdf_dump_translation_dataset.py
patrick-wilken: Ah yes, Pycharm imported this for me automatically, didn't notice it.

patrick-wilken:
Added more documentation and changed file read functions.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,325 @@
...
+    self.number_of_lines = number_of_lines
+    self.line_buffer_size = line_buffer_size
+    self.data_buffer_size = data_buffer_size
+
+    self._number_of_processed_lines = 0
+    self._write_offsets = {"source": 0, "target": 0}
+
+  def __del__(self):
+    self.source_file_handle.close()
+    self.target_file_handle.close()
```
albertz: Normally you never need (nor should) define `__del__`. Also in this case, I don't see the point. If the GC catches the object, it will also catch these handles, and then also close them. Do you really need that? Depending on the Python interpreter, it anyway has no effect.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,325 @@
...
+    self._number_of_processed_lines += chunk_size
+
+    return False
+
+  def _get_chunk(self):
+    """
+    Reads in the next chunk of lines from the corpus files.
+
+    :return: word indices for all lines in the current chunk, for both source and target
+    :rtype: tuple[list[numpy.ndarray[int]]]
```
albertz: Better be more explicit about the type, i.e. `:rtype: list[numpy.ndarray], list[numpy.ndarray]`. I'm also not sure if `numpy.ndarray[int]` is valid. Did you check? I don't think it is valid.

albertz: Also, the shape should be documented (in `:return:`) (just like the dtype).

patrick-wilken:
Oops... Wait a second :D

patrick-wilken:
Reopened.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,322 @@
...
+    self._number_of_processed_lines += chunk_size
+
+    return False
+
+  def _get_chunk(self):
+    """
+    Reads in the next chunk of lines from the corpus files.
+
+    :return: Word indices for all lines in the current chunk, for both source and target.
+        Indices for one line are stored as a numpy array of dtype int and shape (None,).
```
patrick-wilken: Is there a standard syntax for defining the array attributes?

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,322 @@
...
+    self.hdf_file["targets/data/classes"].resize((self._write_offsets["target"],))
+
+  def _line_to_indices(self, line, side):
+    """
+    Converts a line of text to an array of word indices.
+
+    :param str line: input line
+    :param str side: "source" or "target"
+    :return: word indices
+    :rtype: numpy.ndarray[int]
```
patrick-wilken: Ah, here too...

tools/hdf_dump_translation_dataset.py
albertz: I think the PEP defines this partly (or at least suggests ways).
PyCharm has some documentation about that (which is maybe a bit extended from PEP).
`:rtype:` is just about the Python type (what you would get when you call `type(obj)`). The Python type is `numpy.ndarray` here (I guess). You can also use the `typing` module, and some PEP/PyCharm extended syntax (e.g. `dict[int,str]` works here; but not in other situations).

albertz: It would be `:rtype: numpy.ndarray` here. All other info in `:return:`.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,322 @@
...
+    tags = [numpy.array("line-" + str(offset + i), dtype=self._tag_dtype, ndmin=1) for i in range(chunk_size)]
+    tags = numpy.concatenate(tags)
+
+    self.hdf_file["seqTags"][offset: offset + chunk_size] = tags
+
+  def _write_data_indices(self, chunk, side):
+    """
+    Writes the main data (word indices for the source or target corpus) to the HDF file.
+
+    :param list[numpy.ndarray[int] chunk: word indices for all lines in the current chunk
```
albertz: Missing `]`, and invalid type.

tools/hdf_dump_translation_dataset.py
patrick-wilken: Not sure I understand what to change now. The type() of the returned object would just be tuple...
I meant to ask whether the way I described the dtype and shape is ok.

albertz: Ah, sorry, misunderstood.
But you don't need to say that it is a numpy array, because that can already be seen in `:rtype:`. Keep it short. That makes it easier to read and comprehend (I think). I would write it maybe just like:
```
:return: word indices for source and target, dtype int, shape [Lines]
:rtype: list[numpy.ndarray], list[numpy.ndarray]
```
You already mention that it is for the lines in the sentence just before, same for chunk. Don't repeat, it just blows it up unnecessarily and makes it harder to read (in my opinion).

patrick-wilken: Ok, done, I think.

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,321 @@
...
+    tags = [numpy.array("line-" + str(offset + i), dtype=self._tag_dtype, ndmin=1) for i in range(chunk_size)]
+    tags = numpy.concatenate(tags)
+
+    self.hdf_file["seqTags"][offset: offset + chunk_size] = tags
+
+  def _write_data_indices(self, chunk, side):
+    """
+    Writes the main data (word indices for the source or target corpus) to the HDF file.
+
+    :param list[numpy.ndarray] chunk: word indices for all lines in the current chunk
```
albertz: Also misses shape and dtype...

tools/hdf_dump_translation_dataset.py
```diff
@@ -0,0 +1,321 @@
...
+    indices = [self._vocabularies[side].get(word, self._unknown_ids[side]) for word in words]
+
+    indices_numpy = numpy.array(indices, dtype=numpy.int32)
+    return indices_numpy
+
+  @staticmethod
+  def _open_file(file_name):
+    """
+    :param str file_name: filename of a plain text file, possibly zipped
+    :return: file handle
```
albertz: Btw, this `:return:` info is kind of redundant, or at least doesn't add any more information over the `:rtype:`, which says the same, but more explicitly. In such cases, I would just leave it away...

albertz:
Merged in now. Feel free to optimize / cleanup further (without pull request) as you like (not so important).

PR #199, opened 2019-07-16
CachedDataset: save index inversion

patrick-wilken:
dict[int, int] self._seq_index_inv gets huge and creation is slow when used on large amounts of sequences (100M). So just create it if we really need it. For the case of disabled cache, reinitialization is cheaper, so never map indices here.

### DIFF
--- a/CachedDataset.py
+++ b/CachedDataset.py
@@ -85,7 +85,9 @@ class CachedDataset(Dataset):
       # Give some hint to the user in case he is wondering why the cache is reloading.
       print("Reinitialize dataset seq order for epoch %i." % epoch, file=log.v4)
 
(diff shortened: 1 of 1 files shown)


albertz:
Maybe @doetsch @doetsch-apptek wants to take a look. Looking just at the code change (without too much remembering the details/internals of CachedDataset/HDFDataset), it looks ok. Let's see if all tests are passing, and I guess then we can (rebase and) merge it in.

patrick-wilken:
True. I've updated the commit. This should be fine, right?

albertz:
Ok, let's wait for the test cases, and I guess this can be (rebased and) merged then.

PR #200, opened 2019-07-18
Reduced memory consumption of HDFDataset

patrick-wilken:
We had trouble using the HDFDataset on large amounts of sequences as experiments needed too much RAM. (The most extreme was >60 GB on 500M sequences.)

This is because the dataset stores many data structures of O(total number of sequences): sequence lengths, sequence tags, file index for each sequence, and start index for each sequence within the file. Note, that we don't use the cache, so the memory is consumed by this "meta data" only.

With the provided changes I managed to reduce that substantially. I realize that some of the changes might be controversial in terms of design decisions and speed performance, so let's discuss.

...by the way, you could even go one step further and remove the last O(total number of sequences) object `self.file_seq_start` by storing this information in the HDF file instead of the sequence lengths. Then, theoretically, the number of sequences would be unlimited (while sacrificing more speed).

### DIFF
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -36,7 +36,6 @@ class HDFDataset(CachedDataset):
     self.h5_files = []  # type: list[h5py.File]
     self.file_start = [0]
     self.file_seq_start = []; """ :type: list[numpy.ndarray] """
-    self.file_index = []; """ :type: list[int] """
     self.data_dtype = {}; ":type: dict[str,str]"
     self.data_sparse = {}; ":type: dict[str,bool]"
     if files:
@@ -58,7 +57,6 @@ class HDFDataset(CachedDataset):
     """
     Setups data:
       self.seq_lengths
-      self.file_index
       self.file_start
       self.file_seq_start
     Use load_seqs() to load the actual data.
@@ -105,7 +103,6 @@ class HDFDataset(CachedDataset):
     self.file_seq_start.append(seq_start)
     nseqs = len(seq_start) - 1
     self._num_seqs += nseqs
-    self.file_index.extend([len(self.files) - 1] * nseqs)
     self.file_start.append(self.file_start[-1] + nseqs)
     self._num_timesteps += numpy.sum(seq_lengths[:, 0])
     if self._num_codesteps is None:
@@ -185,7 +182,7 @@ class HDFDataset(CachedDataset):
     for idc in selection:
       if self.sample(idc):
         ids = self._seq_index[idc]
-        file_info[self.file_index[ids]].append((idc,ids))
+        file_info[self._get_file_index(ids)].append((idc,ids))
       else:
         self.preload_set.add(idc)
     for i in range(len(self.files)):
@@ -220,7 +217,7 @@ class HDFDataset(CachedDataset):
 
     # Otherwise, directly read it from file now.
     real_seq_idx = self._seq_index[seq_idx]
-    file_idx = self.file_index[real_seq_idx]
+    file_idx = self._get_file_index(real_seq_idx)
     fin = self.h5_files[file_idx]
 
     real_file_seq_idx = real_seq_idx - self.file_start[file_idx]
@@ -276,6 +273,11 @@ class HDFDataset(CachedDataset):
                       "sequences: %i" % self.num_seqs,
                       "frames: %i" % self.get_num_timesteps()])
 
+  def _get_file_index(self, real_seq_idx):
+    file_index = 0
+    while file_index < len(self.file_start) - 1 and real_seq_idx >= self.file_start[file_index + 1]:
+      file_index += 1
+    return file_index
 
 # ------------------------------------------------------------------------------
 
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -56,7 +56,6 @@ class HDFDataset(CachedDataset):
   def add_file(self, filename):
     """
     Setups data:
-      self.seq_lengths
       self.file_start
       self.file_seq_start
     Use load_seqs() to load the actual data.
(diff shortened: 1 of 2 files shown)


albertz:
Again, @doetsch @doetsch-apptek should probably also take a look here.

HDFDataset.py
```diff
@@ -25,20 +25,23 @@
 
 class HDFDataset(CachedDataset):
 
-  def __init__(self, files=None, use_cache_manager=False, **kwargs):
+  def __init__(self, files=None, use_cache_manager=False, tag_function=None, **kwargs):
     """
     :param None|list[str] files:
     :param bool use_cache_manager: uses :func:`Util.cf` for files
+    :param function tag_function: Function that takes a sequence index (int) and returns the corresponding tag (str).
```
albertz: `function` is an incorrect type. The type would be `(int)->str`.

albertz:
I just skimmed over it. Can you maybe summarize how you save the memory now? And also, are there any drawbacks with these changes? If not, I don't see a problem in merging this in (if all tests are passing).

patrick-wilken:
Well, self._seq_lengths and self.file_index don't exist anymore. Also self._tags (strings!) doesn't, if you use the tag function. This frees up a lot of memory.

I calculate the sequence length now by subtracting subsequent start indices of the sequences. This means a few more operations per length request, but that should not matter too much.

The file index can simply be calculated by comparing the index of the first sequence of each file. E.g. if we have files with 5, 10 and 5 sequences in it, then sequence 4 will be in the first file, sequence 16 in the second and so on. Again, this requires a few more basic operations each time we select a file.

The tag function is meant to be used if the tag has a trivial form like "line-<idx>". (Then it would be `lambda n: "line-" + n`). If you don't use it, a list ["line-0", "line-1", ..., "line-10000000"] is stored, which is a huge overhead.

I didn't rigorously compare the speed performance, but I didn't notice a slow down. Also, this mainly affects initializing the epoch, so compared to the actual training it is not too critical. HDFDataset should rather be designed to allow large amounts of data.

HDFDataset.py
```diff
@@ -84,34 +85,38 @@ def add_file(self, filename):
...
-    if len(self._seq_lengths) == 0:
-      self._seq_lengths = numpy.array(seq_lengths)
-    else:
-      self._seq_lengths = numpy.concatenate((self._seq_lengths, seq_lengths), axis=0)
-    if not self._seq_start:
-      self._seq_start = [numpy.zeros((seq_lengths.shape[1],), 'int64')]
     seq_start = numpy.zeros((seq_lengths.shape[0] + 1, seq_lengths.shape[1]), dtype="int64")
     numpy.cumsum(seq_lengths, axis=0, dtype="int64", out=seq_start[1:])
-    self._tags += fin["seqTags"][...].tolist()
-    self.file_seq_start.append(seq_start)
```
patrick-wilken: Btw, this is mostly just reordering. I wanted to keep the time where seq_lengths is in memory as short as possible.

doetsch:
Well the tag function seems to be a very specific solution few people would choose. Why do you even need the tags? Wouldn't be an option to simply not load them sufficient?

albertz:
Or is it not possible to just load the tag on request in `get_tag`? E.g. sth like `return self.h5_files[file_idx]["seqTags"][real_file_seq_idx]`?

patrick-wilken:
Every dataset is supposed to provide tags, no?
But I will try out what @albertz suggested.

patrick-wilken:
Yes, this is much better and also fast.

HDFDataset.py
```diff
@@ -257,10 +272,14 @@ def get_tag(self, sorted_seq_idx):
...
-    return list(map(self._decode, self._tags))
+    tags = []
+    for h5_file in self.h5_files:
+      tags += h5_file["seqTags"][...].tolist()
+
+    return list(map(self._decode, tags))
 
   def get_total_num_seqs(self):
-    return len(self._tags)
+    return self._num_seqs
```
albertz: This is supposed to return the total number of seqs (without partition epoch or so). `_num_seqs` (if it is set) is for one specific epoch. Maybe `len(self._seq_index)` is right here? Not sure.

HDFDataset.py
```diff
@@ -276,6 +295,11 @@ def len_info(self):
                       "sequences: %i" % self.num_seqs,
                       "frames: %i" % self.get_num_timesteps()])
 
+  def _get_file_index(self, real_seq_idx):
+    file_index = 0
+    while file_index < len(self.file_start) - 1 and real_seq_idx >= self.file_start[file_index + 1]:
+      file_index += 1
+    return file_index
```
albertz: Is this function needed? There is `self.file_index` which is used at other places which seems to provide exactly that information.

patrick-wilken: Not anymore, I wanted to save that memory.

albertz: Ah I see. I guess this is not too much overhead...

HDFDataset.py
patrick-wilken: Sadly, in HDFDataset _num_seqs is actually the total number. It's added up in add_file() and never updated. But to use the same name as in _num_seqs in CachedDataset2 is definitively confusing, I agree.

patrick-wilken: Or worse, it's an attribute of Dataset even, but used differently in CachedDataset and CachedDataset2.

albertz: Oh, then `HDFDataset` might actually be an example of a dataset where you cannot rely on `num_seqs` (it can be different, depending e.g. if you use a predefined seq list, or use epoch split, or sth like that). Although, I don't know. `_num_seqs` is supposed to be internal anyway, and as long as `num_seqs` is right, it should not matter from the outside.

albertz:
What is the state here? All fine now? Any objections from anyone? Any drawbacks with these changes? If not, I would merge it in.

doetsch-apptek:
The new approach looks better. A drawback would be that repeated loading of the same tags might be slower now but that should not happen very often. I would say you can merge it.

JackTemaki:
In case `self.cache_byte_size_total_limit == 0` the current code does not work any more, as the h5_file does not exist when the tags are loaded. It seems there is a test case for HDFDataset missing that covers a very basic case.

The config that crashes is:
```
eval = { 'class': 'HDFDataset',
  'files': ['output/eval_conv_output.hdf'],
  'use_cache_manager': True}
```

patrick-wilken:
Yep, sorry for not testing the cached case.
What is the motivation behind closing the files, can we keep them open in any case to fix this?
Otherwise we would have to open them in _get_tag_by_real_idx().

albertz:
Can you add a test case for this?


Patrick Wilken <notifications@github.com> schrieb am Fr., 26. Juli 2019,
18:08:

PR #201, opened 2019-07-19
added additional audio options for feature extraction

JackTemaki:
I added 2 new feature extraction methods, one for linear spectrograms and a modified version of log-mel-spectrograms with additional options and a different scaling. 
Other additions are:

- The peak normalization can be disabled now
- A preemphasis filter can be set
- The features can be stacked as "superframes" (https://arxiv.org/pdf/1507.06947.pdf, section 2.3, https://arxiv.org/abs/1703.10135)
- the normalization can be a single value instead of a file or numpy array representing each feature dimension

The superframes can in theory be implemented as a layer, but with the existing layers I was not able to produce this behavior, and additional layers for conversion would slightly increase the memory load.

### DIFF
--- a/GeneratingDataset.py
+++ b/GeneratingDataset.py
@@ -917,18 +917,23 @@ class ExtractAudioFeatures:
   def __init__(self,
                window_len=0.025, step_len=0.010,
                num_feature_filters=None, with_delta=False, norm_mean=None, norm_std_dev=None,
-               features="mfcc", random_permute=None, random_state=None, raw_ogg_opts=None):
+               features="mfcc", feature_options=None, random_permute=None, random_state=None, raw_ogg_opts=None,
+               peak_normalization=True, preemphasis=None, join_frames=None):
     """
     :param float window_len: in seconds
     :param float step_len: in seconds
     :param int num_feature_filters:
     :param bool|int with_delta:
-    :param numpy.ndarray|str|None norm_mean: if str, will interpret as filename
-    :param numpy.ndarray|str|None norm_std_dev: if str, will interpret as filename
+    :param numpy.ndarray|str|int|float|None norm_mean: if str, will interpret as filename
+    :param numpy.ndarray|str|int|float|None norm_std_dev: if str, will interpret as filename
     :param str features: "mfcc", "log_mel_filterbank", "log_log_mel_filterbank", "raw", "raw_ogg"
+    :param dict[str]|None feature_options: provide additional paramters for the feature function
     :param CollectionReadCheckCovered|dict[str]|bool|None random_permute:
     :param numpy.random.RandomState|None random_state:
     :param dict[str]|None raw_ogg_opts:
+    :param bool peak_normalization: set to False to disable the peak normalization for audio files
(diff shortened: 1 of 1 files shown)


GeneratingDataset.py
```diff
@@ -917,18 +917,23 @@ class ExtractAudioFeatures:
...
     :param float window_len: in seconds
     :param float step_len: in seconds
     :param int num_feature_filters:
     :param bool|int with_delta:
-    :param numpy.ndarray|str|None norm_mean: if str, will interpret as filename
-    :param numpy.ndarray|str|None norm_std_dev: if str, will interpret as filename
+    :param numpy.ndarray|str|int|float|None norm_mean: if str, will interpret as filename
+    :param numpy.ndarray|str|int|float|None norm_std_dev: if str, will interpret as filename
     :param str features: "mfcc", "log_mel_filterbank", "log_log_mel_filterbank", "raw", "raw_ogg"
+    :param dict|None feature_options: provide additional paramters for the feature function
```
albertz: I guess `dict[str]`, right?

GeneratingDataset.py
```diff
@@ -1051,16 +1078,65 @@ def get_audio_features(self, audio, sample_rate):
...
+      else:
+        feature_data /= self.norm_std_dev[None, :]
+
+    if self.join_frames is not None:
+      pad_len = self.join_frames - (feature_data.shape[0] % self.join_frames)
+      pad_len = pad_len % self.join_frames
+      new_len = feature_data.shape[0] + pad_len
+      feature_data = numpy.pad(feature_data, pad_width=((0, pad_len), (0, 0)), mode="edge")
+      feature_data = numpy.reshape(feature_data,
+                                   newshape=(int(new_len/self.join_frames), feature_data.shape[1]*self.join_frames),
```
albertz: There are some spaces missing (around any operators).

albertz: Also, never use `/` for ints. Use `//` instead. And then you also don't need the `int` cast.

GeneratingDataset.py
```diff
@@ -1051,16 +1078,65 @@ def get_audio_features(self, audio, sample_rate):
...
     return feature_data
 
   def get_feature_dimension(self):
     """
     :rtype: int
     """
-    return (self.with_delta + 1) * self.num_feature_filters
+    if self.join_frames is not None:
+      return ((self.with_delta + 1) * self.num_feature_filters) * self.join_frames
+    return ((self.with_delta + 1) * self.num_feature_filters)
```
albertz: I think it's cleaner to not duplicate code, but to write this instead:
`return return ((self.with_delta + 1) * self.num_feature_filters) * (self.join_frames or 1)`

albertz: (Also, remove the unnecessary brackets.)

GeneratingDataset.py
JackTemaki: When being a parameter no spaces are PEP8 conform so there was no error, but I can change it.

albertz: I think the code should be consistent. Also, even if not in PEP8, I think this is standard Python code convention.

PR #202, opened 2019-07-23
added correct sorting for targets only mode in OggZipDataset

JackTemaki:
I am not sure if this is too slow, but I think applying the vocabulary to the whole dataset should be ok.

### DIFF
--- a/GeneratingDataset.py
+++ b/GeneratingDataset.py
@@ -2646,13 +2646,13 @@ class OggZipDataset(CachedDataset2):
 
     def get_seq_len(i):
       """
+      Returns the length based on the duration entry of the dataset,
+      multiplied by 100 to avoid similar rounded durations.
+      It is also used when using the dataset in text-only-mode (`audio` is None).
       :param int i:
       :rtype: int
       """
-      if self.feature_extractor:
-        return int(self._data[i]["duration"] / self.feature_extractor.step_len)
-      else:
-        return int(self._data[i]["duration"])
+      return int(self._data[i]["duration"] * 100)
 
     if seq_list is not None:
       seqs = {seq["file"]: i for i, seq in enumerate(self._data) if seq["file"] in seq_list}


JackTemaki:
related to  d71ef1a77812b0d4ba27b675508654de275c5135

albertz:
This has the problem that the seq ordering will depend on the feature extraction settings. This is bad, it should be independent from that (i.e. whatever you set there, you should always get the same sequence order). Basically, as I wrote in the comments on d71ef1a77812b0d4b, just revert that commit. It was fine before.

albertz:
(Btw, you say you apply the vocab here, but you actually don't do this.)

JackTemaki:
> This has the problem that the seq ordering will depend on the feature extraction settings

it is actually the other way. Just multiplying by 100 makes the ordering depending on the feature extractions settings. Assume an Audio file with 1007 milliseconds (duration = 1.007) and another one with 1000 milliseconds. With a default step_len of 10ms this will result in exactly 100 frames (when not looking at the window_size, but that is not relevant for the problem here) for both audio files, so the original code is correct. But if you would specify a step length of 5ms, one file has 201 and the other 200 frames, but with the original code both still would have the duration of 100, so the ordering is arbitrary.

Also, why is get_seq not applying the dictionary? Am I seeing something wrong there?

JackTemaki:
Or did I get you wrong, and you do not want a correct sequence order, but just any fixed one?

albertz:
I'm saying the ordering should be independent from whatever you set to the feature extractions settings (`audio` option). So no matter whether `feature_extractor is None` or whatever `feature_extractor.step_len` is, if you set the sorting to e.g. "sorted" or e.g. "laplace:10", you should get the same sequence ordering. With my original code, that was the case. With the new code, it is not the case anymore. And that is wrong. It should have exactly the behavior as with my original code, i.e. the sequence ordering should be independent from whatever `feature_extractor` is.

JackTemaki:
Ok, i will just remove everything. Should i revert the commit, or rebase to the previous one and completely remove it? (git revert or git rebase -i and force push?)

albertz:
No, don't alter the Git history. I.e. use `git revert` and add a new commit (maybe to this pull request). As said, you maybe can add a comment to the code as well explaining it.

PR #203, opened 2019-07-26
HDFDataset tag fix

patrick-wilken:
I assume there is a good reason why the files were closed and re-opened, but closing and re-opening to get each tag felt wrong.

I added (and enabled) a test that failed and is fixed by the first commit. I hope enabling the test doesn't cause the next round of trouble ':D


albertz:
Just for reference, this is also related to #200.

PR #204, opened 2019-07-29
added segment_list, seq_name support to OggZipDataset ...

JackTemaki:
... and support None for target.

### DIFF
--- a/GeneratingDataset.py
+++ b/GeneratingDataset.py
@@ -2544,18 +2544,19 @@ class OggZipDataset(CachedDataset2):
   Generic dataset which reads a Zip file containing Ogg files for each sequence.
   """
 
-  def __init__(self, path, audio, targets,
+  def __init__(self, path, audio, targets=None,
                targets_post_process=None,
-               use_cache_manager=False,
+               use_cache_manager=False, segment_file=None,
                fixed_random_seed=None, fixed_random_subset=None,
                epoch_wise_filter=None,
                **kwargs):
     """
     :param str path: filename to zip
     :param dict[str]|None audio: options for :class:`ExtractAudioFeatures`. use {} for default. None means to disable.
-    :param dict[str] targets: options for :func:`Vocabulary.create_vocab` (e.g. :class:`BytePairEncoding`)
+    :param dict[str]|None targets: options for :func:`Vocabulary.create_vocab` (e.g. :class:`BytePairEncoding`)
     :param str|list[str]|((str)->str)|None targets_post_process: :func:`get_post_processor_function`, applied on orth
     :param bool use_cache_manager: uses :func:`Util.cf`
+    :param str|None segment_file: an uncompressed segment whitelisting file containing lines corresponding to seq_name
     :param int|None fixed_random_seed: for the shuffling, e.g. for seq_ordering='random'. otherwise epoch will be used
     :param float|int|None fixed_random_subset:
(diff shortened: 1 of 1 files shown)


GeneratingDataset.py
```diff
@@ -2718,13 +2748,13 @@ def get_tag(self, seq_idx):
     :param int seq_idx:
     :rtype: str
     """
-    return self._data[self._get_ref_seq_idx(seq_idx)]["file"]
+    return self._data[self._get_ref_seq_idx(seq_idx)]["seq_name"]
```
albertz: It should stay compatible to existing datasets. I.e. try to read "seq_name", but if it is not there, then fall back to "file".

GeneratingDataset.py
```diff
@@ -2621,13 +2631,33 @@ def _collect_data(self):
...
-    assert isinstance(first_entry["file"], str)
+
+    if "seq_name" in first_entry.keys():
+      assert isinstance(first_entry["seq_name"], str)
+    else:
+      assert isinstance(first_entry["file"], str)
+
+    segments = None
+    if self.segment_file:
+      segments = [s.strip() for s in open(self.segment_file)]
```
albertz: Minor: Better is: `[s for s in open(...).read().splitlines()]`.

GeneratingDataset.py
```diff
@@ -2621,13 +2631,33 @@ def _collect_data(self):
...
+    data = []
+    for entry in data_input:
+      if "seq_name" not in entry:
+        # use the filename as seq_name if it is not provided
+        entry['seq_name'] = entry['file']
+      if segments:
+        if entry['seq_name'] in segments:
+          data.append(entry)
+      else:
+        data.append(entry)
```
albertz: Please avoid doing that for the default case (i.e. looping over all seqs). If the number of seqs is large (e.g. millions of seqs), this would take too much time, and is not quite necessary.

GeneratingDataset.py
albertz: Sorry, I did not saw that you loop over the whole dataset, and fix that dict. But as written there, better don't do that, and just do it on-the-fly here.

GeneratingDataset.py
JackTemaki: But I need to iterate to check against the segment file, and doing it at access time is not saving time in the end, because the call would be done in each epoch, but the iteration at the beginning is done only once. The bigger problem is that segments is a list, it has to be changed to "set()" for major speed improvements. I created 5 million random strings, and iterated over 10 million random strings to check if being in the 5 million set, which took less than 10 seconds including converting from list to set. So time should not be an issue when using corpora less than 40k hours. (if e.g. 250.000 entries are 1000 hours)

albertz: I'm not talking about that case. If you provide a segment file, yes, you can do the filtering here. However, otherwise, I would not iterate through it.

albertz:
(Btw, a bit meta: It would have been nicer in general, and also simpler, if this would be three separate commits (or even pull requests): segment_list, seq_name support, and optional target. Sometimes it just happens to be combined, but if possible, keep logical separate things in separate commits.)

albertz:
I also needed the same thing now, implemented and pushed it already. See d59508e0f76, cf5bdab5c2dd3ea13f and 0659664580.
So I guess we can close this pull request.

Issue #206, opened 2019-08-04
Intended behavior of SliceLayer for negative slice_end values


ChristopherBrix:
What is the intended behavior of the `SliceLayer` for negative `slice_end' values?

Let's say, we have two sequences in our batch, one of length 3, one of length 6 (so `size_placeholder[0] = [3, 6]` and `placeholder.shape[time_dim] = 6`). We want to remove the last time step. I would assume, this could be done using a `SliceLayer` with `axis='T'` and `slice_end=-1`.
But then `slice_end` is [updated](https://github.com/rwth-i6/returnn/blob/master/TFNetworkLayer.py#L2229) to `tf.shape(self.input_data.placeholder)[axis] + slice_end = 6 + (-1) = 5`
And then the new `size_placeholder` is set to `tf.minimum(tf.shape(self.input_data.placeholder)[axis] - slice_end, self.output.size_placeholder[axis_wo_batch])) = tf.minimum(6 - 5, [3, 6]) = [1, 1]`

Is my math wrong? Why don't we just set `self.output.size_placeholder[axis_wo_batch] += slice_end` for negative `slice_end` values? Maybe with an additional check that the sequence lengths don't become negative.

albertz:
I recently also stumbled upon this, or at least that I wanted to have such a functionality, and saw that `SliceLayer` behaves differently. Although then for some reason I decided to do it differently or that it was not a good idea anyway or so (I don't exactly remember anymore).
I'm not sure which of the two behaviors (the existing one, and the one you describe) is more correct or is the canonical one (i.e. which you would assume as the default one if you don't set it).
I'm also not exactly sure, if we change the default behavior now, if it breaks something for someone. Although probably this is not used too much (I don't know).
But maybe you are right, and we should change this behavior. So I would say, for the beginning without any new option, just change it (i.e. make a pull request with the change). But if anyone ever gets back to this, and wants the old behavior, we should introduce an option to make it configurable (and then maybe we again can discuss the default behavior, or maybe even require this to be an explicit setting, without default).
Maybe @Spotlight0xff has also an opinion on this.

Spotlight0xff:
My code only uses the `SliceNdLayer`, I have never used `SliceLayer`.
That being said, I would also assume the behavior you expected (resulting in `size_pl[0]=[2, 5]`).

albertz:
[...]

PR #207, opened 2019-08-05
Fixed SliceLayer

ChristopherBrix:
This should fix the behavior of the `SliceLayer`, resolving issue #206 

Because shorter sequences have some padding, we cannot rely on the shape of the `placeholder` to determine the new `size_placeholder`. Also, we should ignore the padding when using negative indexing.

This PR should probably also contain some new test cases, that demonstrate where the old code was wrong. So far, I've only checked by hand. I can add tests later.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -2210,9 +2210,9 @@ class SliceLayer(_ConcatInputLayer):
     """
     :param int|str axis:
     :param str|None axis_kind: "T" for time, "B" for batch, "F" for feature
-    :param int|None slice_start:
-    :param int|None slice_end:
-    :param int|None slice_step:
+    :param int|None slice_start: Start index. Currently, only non-negative values are supported
+    :param int|None slice_end: End index. May be negative or positive
+    :param int|None slice_step: Step size. Currently, only positive values are supported
     """
     super(SliceLayer, self).__init__(**kwargs)
     axis = self.input_data.get_axis_from_description(axis)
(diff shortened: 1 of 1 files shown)


ChristopherBrix:
I just noticed I'm not using `size_placeholder` correctly (it's a dict, I'm using it as a tensor). As it is, the could will fail on execution.

Shouldn't be complicated to fix, but I don't have the time just yet. I'll fix this tomorrow.

ChristopherBrix:
I need to use a negative end index to remove the last token. Just fixing the size_placeholder wasn't enough, as the actual selection would still be wrong (for an input sequence of length 4 with 2 padding, the output would still have length 4 but a size_placeholder saying it has length 3). This might be ok if a later layers respect the size_placeholder, but still may cause confusion.
A quick fix would have been to apply masking after the selection, but this would not support negative slice steps (which I don't need right now).

I can change it to this simpler approach, if you don't thing the full functionality is required.

albertz:
The size_placeholder is always respected (if not, that would be a separate bug; and as you say, maybe you can have an explicit TimeMaskingLayer or so to work around this). There can be any data in the padding area. This is totally fine. There are multiple other layers which also leave arbitrary data in there. This is expected.

albertz:
Ping?

ChristopherBrix:
I've reversed my change, and implemented the simpler variant. Now, start and step size must be non-negative and positive, respectively.

albertz:
In the meanwhile, I also fixed it (in a simple way) in 7e912cc36a9bf, which I will soon merge in. So I guess this becomes obsolete now. Looks very similar though.

ChristopherBrix:
Yes, we did it similarly.
In your code, when is 

```
tf.minimum(
  tf.shape(self.input_data.placeholder)[axis] + slice_end,
  self.output.size_placeholder[axis_wo_batch] + slice_end)
```

needed? This should always equal `self.output.size_placeholder[axis_wo_batch] + slice_end)`, or otherwise the `size_placeholder` specified a sequence length longer than the actual `placeholder`. I guess this would be a separate bug?

albertz:
Ah yes I think you are right.

albertz:
Ok, this is now in de2bcfc.
I guess we can close this now. Please open a new pull request if you need anything else.

PR #208, opened 2019-08-13
nce: add log_norm_factor for nce

Gerstenberger:
I propose the following changes to the `sampling_loss`:
* Add `log_norm_factor` to `sampling_loss` for nce; this is fixed to 0.0 in the standard TensorFlow implementation. However, we found that this is worth tuning for Noise Contrastive Estimation.
* As result, it is required to also include `compute_sampled_logits` and `sum_rows` separately, as this is an implementation detail in TensorFlow and not part of the public API. These functions are mostly copied from there. See comments in the doc-string.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -8243,6 +8243,7 @@ class SamplingBasedLoss(Loss):
                use_full_softmax=False,
                remove_accidental_hits=None,
                sampler_args=None,
+               nce_log_norm_term=0.0,
                **kwargs):
     """
     :param int num_sampled: Number of classes to be sampled. For sampled softmax, this is the number of classes to be
@@ -8257,6 +8258,7 @@ class SamplingBasedLoss(Loss):
       and the objective is equal to sampled logistic loss.
     :param dict[str] sampler_args: additional arguments for the candidate sampler. This is most relevant to the fixed_unigram sampler.
       See https://www.tensorflow.org/api_docs/python/tf/random/fixed_unigram_candidate_sampler for details.
+    :param float nce_log_norm_term: The logarithm of the constant normalization term for NCE.
     """
     super(SamplingBasedLoss, self).__init__(**kwargs)
     assert num_sampled >= 1
@@ -8273,6 +8275,69 @@ class SamplingBasedLoss(Loss):
     self.sampler_args = sampler_args
     if self.sampler_args is None:
       self.sampler_args = {}
+    self.nce_log_norm_term = nce_log_norm_term
+
+  def _nce_loss(self,
+                weights,
+                biases,
+                labels,
+                inputs,
+                num_sampled,
+                num_classes,
+                num_true=1,
+                sampled_values=None,
+                remove_accidental_hits=False,
+                partition_strategy="div",
+                name=None,
+                seed=None):
+    """
+    Returns the example-wise NCE losses for the batch.
+
+    This is mostly a copy of
+      https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1729-L1837
+    with an extension which introduces the subtraction by a constant normalization term.
+
+    :param tf.Tensor weights: NCE parameters of shape [num_classes, dim].
+    :param tf.Tensor biases: biases with shape [num_classes]
+    :param tf.Tensor labels: The target classes of shape [batch_size, num_true]
+    :param tf.Tensor inputs: The forward activations of the network of shape [batch_size, dim]
+    :param int num_sampled: Number of sampled classes
+    :param int num_classes: Number of possible classes
+    :param int num_true: Number of target classes per training example
+    :param (tf.Tensor, tf.Tensor, tf.Tensor)|None sampled_values: a tuple of
+      (`sampled_candidates`, `true_expected_count`, `sampled_expected_count`)
+      returned by a `*_candidate_sampler` function (if None, we default to `log_uniform_candidate_sampler`).
+      sampled_candidates is of dtype int64.
+    :param bool remove_accidental_hits: Remove sampled classes that equal on of the target classes
+    :param str partition_strategy: A string specifying the partitioning strategy, relevant
+      if `len(weights) > 1`. Currently `"div"` and `"mod"` are supported.
+      Default is `"mod"`. See `tf.nn.embedding_lookup` for more details.
+    :param str|None name: Name for the operation
+    :param int|None seed: Seed for sampling if no sampled_values are given
+    :return: A [batch_size] tensor of example-wise NCE losses
+    :rtype: tf.Tensor
+    """
+
+    from TFUtil import compute_sampled_logits
+
+    logits, labels = compute_sampled_logits(weights=weights,
+                                            biases=biases,
+                                            labels=labels,
+                                            inputs=inputs,
+                                            num_sampled=num_sampled,
+                                            num_classes=num_classes,
+                                            num_true=num_true,
+                                            sampled_values=sampled_values,
+                                            subtract_log_q=True,
+                                            remove_accidental_hits=remove_accidental_hits,
+                                            partition_strategy=partition_strategy,
+                                            name=name,
+                                            seed=seed)
+    logits -= self.nce_log_norm_term
+    sampled_losses = tf.nn.sigmoid_cross_entropy_with_logits(labels=labels,
+                                                             logits=logits,
+                                                             name="sampled_losses")
+    return tf.reduce_sum(sampled_losses, axis=1)
 
   def get_value(self):
     """
@@ -8336,7 +8401,7 @@ class SamplingBasedLoss(Loss):
                                    range_max=self.target.dim,
                                    **self.sampler_args)
           if self.nce_loss:
-            loss_fn = tf.nn.nce_loss
+            loss_fn = self._nce_loss
           else:
             loss_fn = tf.nn.sampled_softmax_loss
 
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -8517,3 +8517,156 @@ def get_non_deterministic_ops_from_graph():
     # elif ... more non det ops to be added
 
   return non_det_ops
+
+
+def compute_sampled_logits(weights,
+                           biases,
+                           labels,
+                           inputs,
+                           num_sampled,
+                           num_classes,
+                           num_true=1,
+                           sampled_values=None,
+                           subtract_log_q=True,
+                           remove_accidental_hits=False,
+                           partition_strategy="mod",
+                           name=None,
+                           seed=None):
+  """Helper function for nce_loss and sampled_softmax_loss functions.
+  Computes sampled output training logits and labels suitable for implementing
+  e.g. noise-contrastive estimation (see nce_loss) or sampled softmax (see
+  sampled_softmax_loss).
+  Note: In the case where num_true > 1, we assign to each target class
+  the target probability 1 / num_true so that the target probabilities
+  sum to 1 per-example.
+
+  This is a copy of
+    https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1440
+
+  :param tf.Tensor|list[tf.Tensor]|tuple[tf.Tensor] weights: A `Tensor` of shape `[num_classes, dim]`,
+    or a list of `Tensor` objects whose concatenation along dimension 0 has shape `[num_classes, dim]`.
+    The class embeddings.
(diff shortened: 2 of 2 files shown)


TFNetworkLayer.py
```diff
@@ -8257,6 +8258,7 @@ def __init__(self,
       and the objective is equal to sampled logistic loss.
     :param dict[str] sampler_args: additional arguments for the candidate sampler. This is most relevant to the fixed_unigram sampler.
       See https://www.tensorflow.org/api_docs/python/tf/random/fixed_unigram_candidate_sampler for details.
+    :param float log_norm_factor=0.0: The logarithm of the constant normalization term for NCE.
```
albertz: I think the format syntax is wrong (remove the `=0.0` here).

TFNetworkLayer.py
```diff
@@ -8273,6 +8275,68 @@ def __init__(self,
...
+      https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1729-L1837
+
+    :param tf.Tensor weights: NCE parameters of shape [num_classes, dim].
+    :param tf.Tensor biases: biases with shape [num_classes]
+    :param tf.Tensor labels: The target classes of shape [batch_size, num_true]
+    :param tf.Tensor inputs: The forward activations of the network of shape [batch_size, dim]
+    :param int num_sampled: Number of sampled classes
+    :param int num_classes: Number of possible classes
+    :param int num_true: Number of target classes per training example
+    :param (tf.Tensor[int64], tf.Tensor, tf.Tensor) sampled_values: a tuple of
```
albertz: I think `tf.Tensor[int64]` is invalid syntax. It should just be `tf.Tensor`. Write in the description that `sampled_candidates` is of dtype int64.

TFNetworkLayer.py
```diff
@@ -8273,6 +8275,68 @@ def __init__(self,
...
+    :param int num_true: Number of target classes per training example
+    :param (tf.Tensor[int64], tf.Tensor, tf.Tensor) sampled_values: a tuple of 
+      (`sampled_candidates`, `true_expected_count`, `sampled_expected_count`)
+      returned by a `*_candidate_sampler` function.
+      (if None, we default to `log_uniform_candidate_sampler`)
+    :param bool remove_accidental_hits: Remove sampled classes that equal on of the target classes
+    :param str partition_strategy: A string specifying the partitioning strategy, relevant
+      if `len(weights) > 1`. Currently `"div"` and `"mod"` are supported.
+      Default is `"mod"`. See `tf.nn.embedding_lookup` for more details.
+    :param str name: Name for the operation
```
albertz: Should be `str|None`.

TFNetworkLayer.py
```diff
@@ -8273,6 +8275,68 @@ def __init__(self,
...
+    :param int seed: Seed for sampling if no sampled_values are given
```
albertz: Should be `int|None`.

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+
+
+def sum_rows(x):
+  """
+  Returns a vector summing up each row of the matrix x.
+
+  This is mostly a copy of
+    https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1427
+
+  :param tf.Tensor x: input tensor
```
albertz: What shape do you expect for `x`? What dtype? Please specify.

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+  :return: Vector, which rows are summed along axis 1 with shape [shape(x)[0]]
+  :rtype: tf.Tensor
+  """
+  # sum_rows(x) is equivalent to math_ops.reduce_sum(x, 1) when x is
+  # a matrix.  The gradient of _sum_rows(x) is more efficient than
+  # reduce_sum(x, 1)'s gradient in today's implementation. Therefore,
+  # we use _sum_rows(x) in the nce_loss() computation since the loss
+  # is mostly used for training.
+  cols = tf.shape(x)[1]
+  ones_shape = tf.stack([cols, 1])
```
albertz: You don't need the `tf.stack` here, or do you?

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+
+  :param tf.Tensor x: input tensor
+  :return: Vector, which rows are summed along axis 1 with shape [shape(x)[0]]
+  :rtype: tf.Tensor
+  """
+  # sum_rows(x) is equivalent to math_ops.reduce_sum(x, 1) when x is
+  # a matrix.  The gradient of _sum_rows(x) is more efficient than
+  # reduce_sum(x, 1)'s gradient in today's implementation. Therefore,
+  # we use _sum_rows(x) in the nce_loss() computation since the loss
+  # is mostly used for training.
```
albertz: `sum_rows` or `_sum_rows`? Be consistent.
Also, maybe this description would actually be also helpful in the docstring itself, not just as a code comment? So move it there.

albertz: Btw, are you sure that the gradient is more efficient? Have you measured that? Or do they write that in the TF comment?
I remember seeing people using `matmul` instead of `reduce_sum`, but the reason was that in earlier TF versions, `matmul` was deterministic and `reduce_sum` was not (in later TF versions, also `reduce_sum` should be deterministic).

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+  e.g. noise-contrastive estimation (see nce_loss) or sampled softmax (see
+  sampled_softmax_loss).
+  Note: In the case where num_true > 1, we assign to each target class
+  the target probability 1 / num_true so that the target probabilities
+  sum to 1 per-example.
+
+  This is mostly a copy of
+    https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1440
+
+  :param weights tf.Tensor|[tf.Tensor]: A `Tensor` of shape `[num_classes, dim]`, or a list of `Tensor`
```
albertz: The syntax is wrong. It should be `:param tf.Tensor|list[tf.Tensor| weights: ...`.

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+  :return:
+    out_logits: `Tensor` object with shape
+        `[batch_size, num_true + num_sampled]`, for passing to either
+        `nn.sigmoid_cross_entropy_with_logits` (NCE) or
+        `nn.softmax_cross_entropy_with_logits` (sampled softmax).
+    out_labels: A Tensor object with the same shape as `out_logits`.
+  :rtype: (tf.Tensor, tf.Tensor)
+  """
+
+  if not isinstance(weights, list):
```
albertz: Also allow `tuple`. (i.e. `isinstance(weights, (list, tuple))`).

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+          validate_indices=False)
+
+    if subtract_log_q:
+      true_logits -= tf.log(true_expected_count)
+      sampled_logits -= tf.log(sampled_expected_count)
+
+    out_logits = tf.concat([true_logits, sampled_logits], 1)
+
+    out_labels = tf.concat([
+        tf.ones_like(true_logits) / num_true,
```
albertz: Use `//` for integer division.

TFUtil.py
albertz: I just checked; that comment is copied from the TF code. So they claim that. But that code is at least 4 years old, probably even older, and dates back to before TF was public (see [here](https://github.com/tensorflow/tensorflow/blame/ef2a926ec05dfd337d84279aafa58b22f0f36123/tensorflow/python/ops/nn.py#L1084)). So I'm not sure if that is still true today. I would verify that if you have time (just try out with `tf.reduce_sum` instead, and see if there is a difference). In any case, replace "today", and be more specific (write the date of the initial TF commit).

TFUtil.py
```diff
@@ -8517,3 +8517,177 @@ def get_non_deterministic_ops_from_graph():
...
+                           seed=None):
+  """Helper function for nce_loss and sampled_softmax_loss functions.
+  Computes sampled output training logits and labels suitable for implementing
+  e.g. noise-contrastive estimation (see nce_loss) or sampled softmax (see
+  sampled_softmax_loss).
+  Note: In the case where num_true > 1, we assign to each target class
+  the target probability 1 / num_true so that the target probabilities
+  sum to 1 per-example.
+
+  This is mostly a copy of
```
kazuki-irie: > This is **mostly** a copy of

What are the changes?

TFUtil.py
Gerstenberger: I see, thanks.
I removed sum_rows now and replaced this with reduce_sum. I did some small scale tests and on the V100 there is no difference in runtime.

TFUtil.py
Gerstenberger: I think this should be correct as is. If num_true > 1 and we do `floordiv` then the results should be `0.0` which is not what we want. This is just to ensure that the per-example labels sum to `1.0`

Gerstenberger:
I did make the requested changes. Please let me know if anything is missing.

albertz:
Btw, please clean up the commits in the end.

TFUtil.py
```diff
@@ -8517,3 +8517,155 @@ def get_non_deterministic_ops_from_graph():
...
+  e.g. noise-contrastive estimation (see nce_loss) or sampled softmax (see
+  sampled_softmax_loss).
+  Note: In the case where num_true > 1, we assign to each target class
+  the target probability 1 / num_true so that the target probabilities
+  sum to 1 per-example.
+
+  This is a copy of
+    https://github.com/tensorflow/tensorflow/blob/e19c354920c3b246dda6598229210a582caaa1a9/tensorflow/python/ops/nn_impl.py#L1440
+
+  :param tf.Tensor|list(tf.Tensor)|tuple(tf.Tensor) weights: A `Tensor` of shape `[num_classes, dim]`,
```
albertz: The syntax is wrong. It should be `list[tf.Tensor]` and `tuple[tf.Tensor]`. (You can just check whether PyCharm correctly infers the type for `weights`.)

TFUtil.py
```diff
@@ -8517,3 +8517,155 @@ def get_non_deterministic_ops_from_graph():
...
+    all_ids = tf.concat([labels_flat, sampled], 0)
+
+    all_w = tf.nn.embedding_lookup(
+        weights, all_ids, partition_strategy=partition_strategy)
+    if all_w.dtype != inputs.dtype:
+      all_w = tf.cast(all_w, inputs.dtype)
+
+    true_w = tf.slice(all_w,
+                      [0, 0],
+                      tf.stack([tf.shape(labels_flat)[0], -1]))
```
albertz: I think you don't need the `tf.stack` here.

TFUtil.py
```diff
@@ -8517,3 +8517,155 @@ def get_non_deterministic_ops_from_graph():
...
+
+    sampled_w = tf.slice(
+        all_w, tf.stack([tf.shape(labels_flat)[0], 0]), [-1, -1])
```
albertz: I think you don't need the `tf.stack` here.

TFUtil.py
albertz: Ah, sorry, I got confused. I thought that are labels (because you call it `out_labels`), i.e. integers. But why do you call it "labels" then? Better rename it, so that it is more clear what it is.

TFUtil.py
```diff
@@ -8517,3 +8517,155 @@ def get_non_deterministic_ops_from_graph():
...
+    Default is `"mod"`. See `tf.nn.embedding_lookup` for more details.
+  :param str|None name: A name for the operation.
+  :param int|None seed: random seed for candidate sampling. Default to None, which doesn't set
+    the op-level random seed for candidate sampling.
+  :return:
+    out_logits: `Tensor` object with shape
+        `[batch_size, num_true + num_sampled]`, for passing to either
+        `nn.sigmoid_cross_entropy_with_logits` (NCE) or
+        `nn.softmax_cross_entropy_with_logits` (sampled softmax).
+    out_labels: A Tensor object with the same shape as `out_logits`.
```
albertz: What is in `out_labels`? Make more clear. Also write the dtype (from "labels", I would infer that it must be integer).

TFUtil.py
albertz: I'm not sure if "targets" is better... Maybe "soft targets"? Well, maybe good enough like that.

PR #209, opened 2019-08-14
added possibility to use another layer as initial state for layers in…

JackTemaki:
… TFNetworkLayer

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -456,6 +456,13 @@ class LayerBase(object):
             # Also, if we are inside a rec layer, and doing search, we also cannot do that.
             if not network.is_inside_rec_layer() or not network.search_flag:
               network.used_data_keys.add(target)
+    # support that the initial_output of any layer inside a recurrent unit can be a layer from base
+    if d.get("initial_output", None):
+      initial_output = d["initial_output"]
+      if isinstance(initial_output, str):
+        if initial_output not in ["zeros", "ones", "var", "keep_over_epoch", "keep_over_epoch_no_init", "apply(0)"]:
+          # if initial_output is not a reserverd keyword, assume it is a layer
+          d['initial_output'] = get_layer(initial_output)
     if "n_out" not in d and targets and network.eval_flag:
       # Must be done here now because loss might be set to None later.
       target = targets[0]  # guess using first target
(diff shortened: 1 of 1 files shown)


albertz:
Btw, for commit messages, your title line should not exceed 80 chars (or so; don't remember the exact number; you see that in PyCharm or here on GitHub).

TFNetworkLayer.py
```diff
@@ -456,6 +456,11 @@ def transform_config_dict(cls, d, network, get_layer):
             # Also, if we are inside a rec layer, and doing search, we also cannot do that.
             if not network.is_inside_rec_layer() or not network.search_flag:
               network.used_data_keys.add(target)
+    # support that the initial_output of any layer inside a recurrent unit can be a layer from base
+    if d.get("initial_output", None):
+      initial_output = d["initial_output"]
+      if isinstance(initial_output, str) and initial_output.startswith('layer:'):
+        d['initial_output'] = get_layer(d['initial_output'][6:])
```
albertz: Don't write `[6:]` here. It's not clear where the number comes from. Use e.g. `[len("layer:"):]` instead.

albertz: But in any case, this should be consistent to the logic for initial states. See `RnnCellLayer.transform_initial_state`. We don't use a prefix like `"layer:"` there.

JackTemaki: Should i copy the function  `transform_initial_state` from TFNetworkRecLayer and add the apply(0)?

albertz: You should never copy code. You could directly call it instead, or share the code in some other way. But if that is too complicated for now (I'm pretty sure that you actually find that comment from me somewhere about this), just do it for this case now, i.e. what I said, just make it consistent (remove that `"layer:"` prefix).

JackTemaki: I can not remove it, as this would break multiple test cases involving "apply(0)". I could catch this case, but this is what the other function does already. So i will try to share the code.

albertz: Sure you can. Just check if it is not `"apply(0)"` (and some other cases, similar as in `transform_initial_state`), and if so, then assume it is a layer name (just as in `transform_initial_state`).
This is probably the simplest way for now. I'm not sure how much effort it is to share the code there (if that would be easy, I would have done it already).

PR #210, opened 2019-08-15
Feature cache synchronization for bliss-to-ogg-zip tool

vieting:
If a sprint feature cache is given, the generated ogg files are synchronized to the feature caches to overcome possible mismatches, e.g. by dc-detection.

### DIFF
--- a/tools/bliss-to-ogg-zip.py
+++ b/tools/bliss-to-ogg-zip.py
@@ -9,10 +9,17 @@ from argparse import ArgumentParser
 from decimal import Decimal
 import tempfile
 import gzip
+import numpy
 import xml.etree.ElementTree as ElementTree
 import zipfile
 import shutil
 from subprocess import check_call
+
+my_dir = os.path.dirname(os.path.abspath(__file__))
+returnn_dir = os.path.dirname(my_dir)
+sys.path.insert(0, returnn_dir)
+
+import SprintCache
 
 
 class BlissItem:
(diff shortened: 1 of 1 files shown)


tools/bliss-to-ogg-zip.py
```diff
@@ -9,10 +9,12 @@
 from decimal import Decimal
 import tempfile
 import gzip
+import numpy
 import xml.etree.ElementTree as ElementTree
 import zipfile
 import shutil
 from subprocess import check_call
+from returnn import SprintCache
```
albertz: You cannot do the import like this.
You would first need to add sth like this:
```
my_dir = os.path.dirname(os.path.abspath(__file__))
returnn_dir = os.path.dirname(my_dir)
sys.path.insert(0, returnn_dir)
```
And then you can do:
```
import SprintCache
```
See the other tools.

tools/bliss-to-ogg-zip.py
```diff
@@ -201,6 +311,23 @@ def main():
...
+             "-loglevel", args.ffmpeg_loglevel]
+      print("$ %s" % " ".join(cmd))
+      check_call(cmd)
+      import soundfile  # pip install pysoundfile
+      audio, sample_rate = soundfile.read(wav_tmp_filename)
+      assert sample_rate == args.raw_sample_rate
+      data = sprint_cache_handler.feature_post_process(numpy.expand_dims(audio, axis=1), seq.segment_name)
+      soundfile.write(wav_tmp_filename, data, args.raw_sample_rate)
+      source_filename = wav_tmp_filename
+      start_time = 0
```
albertz: I think you should also reset `duration`, or not? Maybe you can set `duration = 'inf'` or so.

tools/bliss-to-ogg-zip.py
```diff
@@ -209,9 +336,12 @@ def main():
       if os.path.exists(dest_filename):
         print("already exists, delete: %s" % os.path.basename(dest_filename))
         os.remove(dest_filename)
-      cmd = ["ffmpeg", "-i", rec_filename, "-ss", str(seq.start_time), "-t", str(duration), dest_filename]
+      cmd = ["ffmpeg", "-i", source_filename, "-ss", str(start_time), "-t", str(duration), dest_filename,
+             "-loglevel", args.ffmpeg_loglevel]
       print("$ %s" % " ".join(cmd))
       check_call(cmd)
+    if args.sprint_cache:
+      os.remove(wav_tmp_filename)
```
albertz: As an additional check, maybe you load the Ogg file now, and verify again that the number of frames is correct. Because I'm not 100% whether the wav->ogg ffmpeg encoding keeps exactly the same number of frames. (@JackTemaki maybe you know this?)

tools/bliss-to-ogg-zip.py
vieting: I added the import as suggested.

tools/bliss-to-ogg-zip.py
vieting: ffmpeg does not accept inf as a duration. Instead in that case it is possible not to specify a duration at all, which I did now.

However, the duration should be exactly the same as it is the duration of the wav file that is converted, right?

albertz: I thought setting `duration = 'inf'` would directly work (without the ugly check below). Did you try?
If it does not work, rather than setting it to `"inf"`, `inf` (or `Decimal('Infinity')` in your case, which is more ugly), you should just set it to `None`.

tools/bliss-to-ogg-zip.py
vieting: I added a check and in a short test it looks like the number of frames is always correct.

tools/bliss-to-ogg-zip.py
albertz: There is also nothing like -1 or so? But if there is just no way to do it, then, as I say, just set it to `None`.

vieting: `duration = 'inf'` does not work as described above. I'll set it to `None` then.

tools/bliss-to-ogg-zip.py
```diff
@@ -209,9 +342,16 @@ def main():
       if os.path.exists(dest_filename):
         print("already exists, delete: %s" % os.path.basename(dest_filename))
         os.remove(dest_filename)
-      cmd = ["ffmpeg", "-i", rec_filename, "-ss", str(seq.start_time), "-t", str(duration), dest_filename]
+      cmd = ["ffmpeg", "-i", source_filename, "-ss", str(start_time), dest_filename, "-loglevel", args.ffmpeg_loglevel]
+      if not duration == Decimal('Infinity'):
+        cmd[5:5] = ["-t", str(duration)]
```
albertz: Don't do that. This is error prone (and ugly) for many reasons:
* You should never write `not x == y` but rather `x != y` (btw, PyCharm tells you that, or not?).
* The check is error prone and usual, as you usually would have set something to `None` to declare that it is unset/unused.
* Never hardcode fixed numbers (5 here).

Rather, the code better should be written that way:
```
cmd = ["ffmpeg", "-i", source_filename]
if start_time:
  cmd += ["-ss", str(start_time)]
if duration is not None:
  cmd += ["-t", str(duration)]
cmd += [dest_filename, "-loglevel", args.ffmpeg_loglevel]
```

albertz:
See my further comments. Also, please clean up the commits (merge it into a single commit or so; or at least remove those commits which just fix previous commits).

PR #213, opened 2019-08-26
problem with CombineLayer after ConvLayer

squarenabla:
test Data.get_common_data for two sources with equal shape. Expecting to get the same shape. 
The behavior is required by CombineLayer with param `kind=add`. Common shape has to be the same as the shape of two inputs to do addition.

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -384,6 +384,48 @@ def test_activation_layer_net_construct_two_out():
     assert_equal(v.tolist(), [[[0, 0], [0, 0], [2, 2]]])
 
 
+def test_cnn_building_block():
+  with make_scope() as session:
+    num_inputs = 192
+    channel_num = 32
+    feature_dim = 6
+    filters = 32
+    filter_size = (3, 3)
+    config = Config()
+    config.update({
+      "num_inputs": num_inputs,
+      "num_outputs": filters,
+      "network": {
+        "split": {"class": "split_dims", "axis": "f", "dims": (channel_num, feature_dim), "from": ["data"]},
+        "swap_axes": {"class": "swap_axes", "axis1": "s:1", "axis2": "f", "from": ["split"]},
+        "c1": {"class": "conv", "n_out": filters, "filter_size": filter_size, "auto_use_channel_first": False,
+               "strides": (1, 1), "dilation_rate": (1, 1), "padding": "SAME", "activation": None, "with_bias": False,
+               "from": "swap_axes"},
+        "bn1": {"class": "batch_norm", "from": "c1"},
+        "y1": {"class": "activation", "activation": "relu", "batch_norm": False, "from": "bn1"},
+        "c2": {"class": "conv", "n_out": filters, "filter_size": filter_size, "auto_use_channel_first": False,
+               "strides": (1, 1), "dilation_rate": (1, 1), "padding": "SAME", "activation": None, "with_bias": False,
+               "from": "y1"},
+        "p": {"class": "combine", "kind": "add", "from": ["c2", "swap_axes"]},
+        "bn2": {"class": "batch_norm", "from": "p"},
+        "y2": {"class": "activation", "activation": "relu", "batch_norm": False, "from": "bn2"},
+
+        "out_pool": {"class": "reduce", "mode": "avg", "axes": "s:1", "keep_dims": False, "from": "y2"},
+        "output": {"class": "copy", "from": ["out_pool"], "is_output_layer": True}
+      }})
+    network = TFNetwork(config=config, train_flag=True)
+    network.construct_from_dict(config.typed_value("network"))
+    session.run(tf.global_variables_initializer())
+    out = network.layers["output"].output.placeholder
+    n_batch = 5
+    seq_len = 10
+    seq_lens = numpy.array([10, 10, 10, 10, 10], dtype=numpy.int32)
+    feed = {network.extern_data.get_default_input_data().placeholder:
+            numpy.random.rand(n_batch, seq_len, num_inputs).astype('f'),
+            network.extern_data.get_default_input_data().size_placeholder[0]: seq_lens}
+    v = session.run(out, feed_dict=feed)
+
+
 def test_combine_layer_net_construct():
   with make_scope() as session:
     net_dict = {
--- a/tests/test_TFUtil.py
+++ b/tests/test_TFUtil.py
@@ -469,6 +469,22 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data_extra_static_spatial():
+  d1 = Data(name='t', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2.get_size_dim_tag(0).declare_same_as(d1.get_size_dim_tag(0))
+  common = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  assert d1.shape == common.shape
+
+
+def test_Data_get_common_data_extra2_static_spatial():
+  d1 = Data(name='t', shape=(None, 32, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2.get_size_dim_tag(0).declare_same_as(d1.get_size_dim_tag(0))
+  common = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  assert d1.shape == common.shape
+
+
 def test_Data_copy_compatible_to_get_common_data_auto_feature_non_sparse():
   d1 = Data(name='t', shape=(None,), dtype='int32', batch_dim_axis=None, feature_dim_axis=None,
             auto_create_placeholders=True)  # placeholder for specific spatial dim-tag
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -2173,7 +2173,7 @@ class Data(object):
       common = common.copy()
       # Note: we don't use copy_extend_with_beam because we don't want to create any ops in the TF graph at this point.
       common.beam_size = max([s.beam_size or 0 for s in sources])
-    is_equal_opts = dict(ignore_feature_dim=True)
+    is_equal_opts = dict(ignore_feature_dim=True, allow_same_spatial_dim=True)
     all_dim_tags, tags_dict = DimensionTag.get_all_dimension_tags(sources, is_equal_opts=is_equal_opts)
     # Note: We cannot compare len(all_dims_tags) to len(shape) as e.g. shape (B,1,1,D) would have only 3 dim tags.
     largest_dim_tags, _ = DimensionTag.get_all_dimension_tags([common], is_equal_opts=is_equal_opts)


albertz:
Do I have access to your repo? So that I can push a fix to this branch?
Or do you fix it yourself?

squarenabla:
I think you don't have the access, but I can pull from master if you do any changes there

tests/test_TFUtil.py
```diff
@@ -469,6 +469,15 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data():
+  d1 = Data(name='t', shape=(None, 32, 128), dtype='float32', batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3,
+            auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 128), dtype='float32', batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3,
+            auto_create_placeholders=True)
+  common = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  assert d1.shape == common.shape
```
albertz: The test is actually wrong. Because you create individual placeholders for each, it will create two different independent sizes. So the common shape should be `(None, None, 32, 128)`.

squarenabla: But that's the case for the combine layer, where I have two placeholders as an input.

albertz: If you want that you get your expected result, you would need to add this:
```
d2.get_size_dim_tag(0).declare_same_as(d1.get_size_dim_tag(0))
```

albertz: I'm speaking about the size placeholders, and their dimension tags. They should be the same. Or if they are not, the question is, why are they not the same. Anyway, `Data.get_common_shape` behaves correct here.

squarenabla:
https://github.com/rwth-i6/returnn/blob/da95d40a5c1affd54bc1c57d9820786f1fd12d8e/TFNetworkLayer.py#L5608
simply doesn't work because the `common_data.shape` is different from the shape of the inputs

albertz:
> I think you don't have the access, but I can pull from master if you do any changes there

But that's not the purpose of pull request. The commit should end up here.

albertz:
> https://github.com/rwth-i6/returnn/blob/da95d40a5c1affd54bc1c57d9820786f1fd12d8e/TFNetworkLayer.py#L5608
> 
> 
> simply doesn't work because the `common_data.shape` is different from the shape of the inputs

It does work correctly. See my comment on `get_common_data`.

tests/test_TFUtil.py
squarenabla: But I need that behavior here
 https://github.com/rwth-i6/returnn/blob/da95d40a5c1affd54bc1c57d9820786f1fd12d8e/TFNetworkLayer.py#L5608

squarenabla:
> > https://github.com/rwth-i6/returnn/blob/da95d40a5c1affd54bc1c57d9820786f1fd12d8e/TFNetworkLayer.py#L5608
> > 
> > simply doesn't work because the `common_data.shape` is different from the shape of the inputs
> 
> It does work correctly. See my comment on `get_common_data`.

You mean this? 
https://github.com/rwth-i6/returnn/blob/da95d40a5c1affd54bc1c57d9820786f1fd12d8e/TFUtil.py#L2190
    # Ok, there is some other axis (or multiple), or we cannot identify/compare them because of incomplete information.
    # Try something more complex: Make all axes unique.
    # Note that this should also work at template construction time,
    # where we do not have access to the size_placeholder,
    # and thus the dimension tags are not reliable (in the current implementation).

tests/test_TFUtil.py
albertz: As I explained, that code is correct there. See my comments on `get_common_data`. It's correct. (Or where do you think it is wrong? Please read my comments.)

albertz:
> > It does work correctly. See my comment on `get_common_data`.
> 
> You mean this?

No, I mean what I just wrote here:

> The test is actually wrong. Because you create individual placeholders for each, it will create two different independent sizes. So the common shape should be `(None, None, 32, 128)`.

> If you want that you get your expected result, you would need to add this: ...

Please read what I write here.

squarenabla:
> > > It does work correctly. See my comment on `get_common_data`.
> > 
> > 
> > You mean this?
> 
> No, I mean what I just wrote here:
> 
> > The test is actually wrong. Because you create individual placeholders for each, it will create two different independent sizes. So the common shape should be `(None, None, 32, 128)`.

I made the test, which is more real. There's an addition in the building block, which fails due to the new implementation.

squarenabla:
Actually, I see what's the problem. There's a new system with so called "tags" for axes. But I see the changes only in the layers, which you use, but e.g. there's no adaptation in "conv" layer. Why did you switched to the new variant  without adapting all layers?

albertz:
> I think you don't have the access

Do I have access now?

Please also rename the commit. The commit message does not make sense. You just add the test `test_cnn_building_block` in the commit, so the commit message should be sth like just `test_cnn_building_block`. (And as I explained already, the bug is likely not in `CombineLayer`. Please read what I wrote.)

albertz:
Why have you removed the other earlier test? Can you add it again? And please modify as I explained.

albertz:
> Actually, I see what's the problem. There's a new system with so called "tags" for axes. But I see the changes only in the layers, which you use, but e.g. there's no adaptation in "conv" layer. Why did you switched to the new variant without adapting all layers?

I adapted all layers. See the commits in the last few months. There are already many test cases to cover lots of cases. However, there are an infinite number of possible cases, so it's hard to verify that everything always works as intended. But it is your responsibility to make sure your cases work correct. Both by checking all the commits, and by having test cases for them.

squarenabla:
I add you to my repository https://github.com/squarenabla/returnn/invitations

squarenabla:
> Why have you removed the other earlier test? Can you add it again? And please modify as I explained.

I added the old test with the midification `d2.get_size_dim_tag(0).declare_same_as(d1.get_size_dim_tag(0))`
There's still an assertion error.

albertz:
I just see this in the error:

```
Exception creating layer <network via test_cnn_building_block>/'p' of class CombineLayer with opts:
{'kind': 'add',
 'name': 'p',
 'network': <TFNetwork '<network via test_cnn_building_block>' train>,
 'output': Data(name='p_output', shape=(None, 64, 64, 64), batch_shape_meta=[B,T|'time:var:extern_data:data',64,64,F|64]),
 'sources': [<ConvLayer 'c2' out_type=Data(shape=(None, 64, 64), batch_shape_meta=[B,T|'time:var:extern_data:data',64,F|64])>,
             <SwapAxesLayer 'swap_axes' out_type=Data(shape=(None, 64, 3), batch_shape_meta=[B,T|'time:var:extern_data:data',64,F|3])>]}
```
The test case is broken, or not? How would you want to add a tensor of shape `(None, 64, 64)` to a tensor of shape `(None, 64, 3)`?

tests/test_TFUtil.py
```diff
@@ -469,6 +469,16 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data():
```
albertz: Please rename this function to something more specific. `test_Data_get_common_data_extra_static_spatial` or so.

tests/test_TFNetworkLayer.py
```diff
@@ -384,6 +384,46 @@ def test_activation_layer_net_construct_two_out():
     assert_equal(v.tolist(), [[[0, 0], [0, 0], [2, 2]]])
 
 
+def test_cnn_building_block():
+  with make_scope() as session:
+    num_inputs = 192
+    channel_num = 3
+    feature_dim = 64
+    filters = 64
```
albertz: Better choose another `filters` here, such that it is easier to debug.

squarenabla:
> I just see this in the error:
> 
> ```
> Exception creating layer <network via test_cnn_building_block>/'p' of class CombineLayer with opts:
> {'kind': 'add',
>  'name': 'p',
>  'network': <TFNetwork '<network via test_cnn_building_block>' train>,
>  'output': Data(name='p_output', shape=(None, 64, 64, 64), batch_shape_meta=[B,T|'time:var:extern_data:data',64,64,F|64]),
>  'sources': [<ConvLayer 'c2' out_type=Data(shape=(None, 64, 64), batch_shape_meta=[B,T|'time:var:extern_data:data',64,F|64])>,
>              <SwapAxesLayer 'swap_axes' out_type=Data(shape=(None, 64, 3), batch_shape_meta=[B,T|'time:var:extern_data:data',64,F|3])>]}
> ```
> 
> The test case is broken, or not? How would you want to add a tensor of shape `(None, 64, 64)` to a tensor of shape `(None, 64, 3)`?

I fixed the number of channels, but I have an error before:
`  File "/u/bozheniuk/crnn/TFNetworkLayer.py", line 5609, in _op_dense_fn
    line: x = sources[0].output.copy_compatible_to(common_data).placeholder
    locals:
      x = <not found>
      sources = <local> [<ConvLayer 'c2' out_type=Data(shape=(None, 6, 32), batch_shape_meta=[B,T|'time:var:extern_data:data',6,F|32])>, <SwapAxesLayer 'swap_axes' out_type=Data(shape=(None, 6, 32), batch_shape_meta=[B,T|'time:var:extern_data:data',6,F|32])>]
      output = <not found>
      output.copy_compatible_to = <not found>
      common_data = <local> Data(name='c2_output', shape=(None, 6, 6, 32), batch_shape_meta=[B,T|'time:var:extern_data:data',6,6,F|32])
      placeholder = <not found>
  File "/u/bozheniuk/crnn/TFUtil.py", line 1030, in copy_compatible_to
    line: assert v.batch_ndim == data.batch_ndim
    locals:
      v = <local> Data(name='c2_output', shape=(None, 6, 32), feature_dim_axis=3, batch_shape_meta=[B,T|'time:var:extern_data:data',6,F|32])
      v.batch_ndim = <local> 4
      data = <local> Data(name='c2_output', shape=(None, 6, 6, 32), batch_shape_meta=[B,T|'time:var:extern_data:data',6,6,F|32])
      data.batch_ndim = <local> 5 `

albertz:
Yes, `test_Data_get_common_data_extra_static_spatial` is a valid test now, and it indeed demonstrates some bug. Do you want to try to fix it yourself? Otherwise you would need to wait a bit.

squarenabla:
Yeah, I need it now, so I'll try to fix it

TFNetworkLayer.py
```diff
@@ -4127,8 +4128,10 @@ def __init__(self, n_out, filter_size, padding, strides=1, dilation_rate=1,
       input_data = input_data.copy_add_feature_dim()
     if self.output.is_batch_feature_major:
       input_data = input_data.copy_as_batch_feature_major()
+      x = check_input_dim(input_data.get_placeholder_as_batch_major(), 1, self.input_data.dim)
     else:
       input_data = input_data.copy_with_feature_dim_axis(-1)
+      x = check_input_dim(input_data.get_placeholder_as_batch_major(), -1, self.input_data.dim)
```
albertz: Please remove that. That's both incorrect, and also does not belong into this pull request.

tests/test_TFUtil.py
```diff
@@ -469,6 +469,16 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
...
+def test_Data_get_common_data_extra_static_spatial():
+  d1 = Data(name='t', shape=(None, 32, 128), dtype='float32', batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3,
+            auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 128), dtype='float32', batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3,
+            auto_create_placeholders=True)
+  d2.get_size_dim_tag(0).declare_same_as(d1.get_size_dim_tag(0))
+  common = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  assert d1.shape == common.shape
+
+
```
albertz: Can you also add another further test, called `test_Data_get_common_data_extra2_static_spatial` or sth like that, where `d1/d2` has `shape=(None, 32, 32, 128)`? I'm not exactly sure if this behaves strange now with your change.

TFNetworkLayer.py
squarenabla: Done

tests/test_TFUtil.py
albertz: (Also, btw, you set `batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3` explicitly, but I think you don't need that, or do you? Maybe remove that, to keep the test simple.)

squarenabla: Done, that works fine

squarenabla: > (Also, btw, you set `batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=3` explicitly, but I think you don't need that, or do you? Maybe remove that, to keep the test simple.)

Done

PR #214, opened 2019-08-27
problem with tags in the combine layer

squarenabla:
Inputs in combine layer may contain different tags for time_dim => an error is thrown.
Solution - force tags to be equal if they contain 'time' in the description.

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -261,6 +261,66 @@ def test_LinearLayer_batch_feature_major():
     session.run(layer.output.placeholder, feed_dict=feed_dict)
 
 
+def test_CombineLayer_after_WindowLayer_and_SliceLayer():
+  import numpy as np
+  from TFNetwork import TFNetwork, ExternData
+  with make_scope() as session:
+    net = TFNetwork(extern_data=ExternData())
+    src_data = Data(name='t', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+    window_size = 5
+    slice_size = window_size - 1
+
+    rnd = np.random.RandomState(42)
+    input_data = rnd.rand(10, 11, 32, 128)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
+
+    with tf.variable_scope("src"):
+      src = InternalLayer(name="src", network=net, out_type={"dim": 128, "shape": (None, 32, 128),
+        "batch_dim_axis": 0, "time_dim_axis": 1, "feature_dim_axis": 3, "sparse": False})
+      src.output = src_data
+      feed_dict = {src_data.placeholder: input_data,
+        src_data.size_placeholder[0]: seq_lens}
+    with tf.variable_scope("win"):
+      win = WindowLayer(
+        name="win", network=net, window_size=window_size, sources=[src], padding="valid",
+        output=WindowLayer.get_out_data_from_opts(
+          name="win", network=net, window_size=window_size,
+          padding="valid", sources=[src]))
+    with tf.variable_scope("red"):
+      red = ReduceLayer(
+        name="red", network=net, mode="AVG", axes="s:1", keep_dims=False, sources=[win],
+        output=ReduceLayer.get_out_data_from_opts(
+          name="red", network=net, mode="AVG",
+          axes="s:1", keep_dims=False, sources=[win]))
+    with tf.variable_scope("slice"):
+      sl = SliceLayer(
+        name="slice", axis="t", slice_start=slice_size // 2, slice_end=-(slice_size // 2),
+        sources=[src], network=net,
+        output=SliceLayer.get_out_data_from_opts(
+          name="slice", axis="t", slice_start=slice_size // 2,
+          slice_end=-(slice_size // 2), sources=[src], network=net))
+    with tf.variable_scope("combine"):
+      combine = CombineLayer(
+        name="combine", network=net, kind="add", sources=[sl, red],
+        output=CombineLayer.get_out_data_from_opts(
+          name="combine", network=net,
+          kind="add", sources=[sl, red]))
+    tf.global_variables_initializer().run()
+    print("sl:", sl)
+    print("win:", win)
+    print("red:", red)
+    print("combine:", combine)
+    in1, in2 = session.run([
+      combine.sources[0].output.placeholder,
+      combine.sources[1].output.placeholder],
+      feed_dict=feed_dict)
+    assert in1.shape == (10, 7, 32, 128)
+    assert in2.shape == (10, 7, 32, 128)
+
+    res = session.run(combine.output.placeholder, feed_dict=feed_dict)
+    assert res.shape == (10, 7, 32, 128)
+
+
 def test_batch_norm_vars():
   with make_scope() as session:
     n_in, n_out = 2, 3
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -152,6 +152,85 @@ def test_concat_sources_missing_dim():
     session.run(out.output.placeholder)
 
 
+def test_self_att_time_dim_tags():
+  with make_scope() as session:
+    n_in = 127
+    channels = 17
+    n_time = 119
+    n_batch = 11
+    context_size = 15
+    rnd = numpy.random.RandomState(42)
+    querry_dim = 31
+    querry_dim += context_size
+    key_dim = 31
+    value_dim = 43
+    num_heads = 5
+    key_scale = 1.0 / (float(key_dim) ** .5)
+
+    crop_left = (context_size - 1) // 2
+    crop_right = (context_size - 1) - crop_left
+
+    net_dict = {
+      "att_q": {"class": "linear", "activation": None, "with_bias": False, "n_out": querry_dim * num_heads, "from": "data:in0"},
+      "att_k": {"class": "linear", "activation": None, "with_bias": False, "n_out": key_dim * num_heads, "from": "data:in0"},
+      "att_v": {"class": "linear", "activation": None, "with_bias": False, "n_out": value_dim * num_heads, "from": "data:in0"},
+
+      "att_q_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, querry_dim), "from": "att_q"},
+      "att_k_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, key_dim), "from": "att_k"},
+      "att_v_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, value_dim), "from": "att_v"},
+
+      "att_queries_crop": {"class": "slice", "axis": "t", "slice_start": crop_left, "slice_end": -crop_right, "from": "att_q_split"},
+
+      "att_queries_key_part": {"class": "slice", "axis": "f", "slice_start": 0, "slice_end": querry_dim - context_size, "from": "att_queries_crop"},
+      "att_queries_context_part": {"class": "slice", "axis": "f", "slice_start": querry_dim - context_size, "slice_end": querry_dim, "from": "att_queries_crop"},
+
+      "att_key_window": {"class": "window", "window_size": context_size, "padding": "valid", "from": "att_k_split"},
+      "att_energy_dot": {"class": "dot", "red1": "f", "red2": "f", "var1": None, "var2": "s:1", "from": ["att_queries_key_part", "att_key_window"]},
+      "att_energy_scaled": {"class": "eval", "eval": "source(0) * %f" % key_scale, "from": "att_energy_dot"},
+
+      "att_energy_biased": {"class": "combine", "kind": "add", "from": ["att_energy_scaled", "att_queries_context_part"]},
+
+      "att_scales": {"class": "activation", "activation": "softmax", "from": "att_energy_biased"},
+
+      "att_value_window": {"class": "window", "window_size": context_size, "padding": "valid", "from": "att_v_split"},
+      "att_out_dot": {"class": "dot", "red1": "f", "red2": "s:1", "var1": None, "var2": "f", "from": ["att_scales", "att_value_window"]},
+
+      "att_out_combine": {"class": "merge_dims", "axes": ["s:1", "f"], "from": "att_out_dot"},
+      "output": {"class": "copy", "from": ["att_out_combine"]}
+    }
+
+    config = Config({"debug_print_layer_output_template": True})
+    in0 = Data(
+      name="in0", shape=(None, n_in // channels, channels), batch_dim_axis=0, auto_create_placeholders=True)
+    extern_data = ExternData()
+    extern_data.register_data(in0)
+
+    network = TFNetwork(config=config, extern_data=extern_data, train_flag=True)
+    network.construct_from_dict(net_dict)
+
+    in_data = rnd.normal(size=(n_batch, n_time, n_in // channels, channels)).astype("float32")
+    in_seq_len = numpy.empty(n_batch)
+    in_seq_len.fill(n_time)
+
+    feed_dict = {
+      in0.placeholder: in_data,
+      in0.size_placeholder[0]: in_seq_len}
+    fetches = []
+
+    for _, l in network.layers.items():
+      l.output.placeholder = tf.Print(
+        l.output.placeholder, [l.name, "shape:", tf.shape(l.output.placeholder)],
+        summarize=10, name="debug_print_layer_output_shape")
+      fetches.append(l.output.placeholder)
+    output = network.get_default_output_layer().output
+
+    tf.global_variables_initializer().run()
+    out = session.run(fetches=fetches, feed_dict=feed_dict)
+    inp1, inp2 = network.layers["att_energy_dot"].sources[0].output.placeholder, network.layers["att_energy_dot"].sources[1].output.placeholder
+    res = network.layers["att_energy_dot"].output.placeholder
+    inp1, inp2, res = session.run([inp1, inp2, res], feed_dict=feed_dict)
+
+
 def test_LinearLayer_batch_feature_major():
   with make_scope() as session:
     network = TFNetwork(config=Config(), extern_data=ExternData(), train_flag=True)


TFNetworkLayer.py
```diff
@@ -2219,13 +2219,14 @@ class SliceLayer(_ConcatInputLayer):
...
 
-  def __init__(self, axis, slice_start=None, slice_end=None, slice_step=None, **kwargs):
+  def __init__(self, axis, slice_start=None, slice_end=None, slice_step=None, force_old_tag=False, **kwargs):
     """
     :param int|str axis:
     :param str|None axis_kind: "T" for time, "B" for batch, "F" for feature
     :param int|None slice_start:
     :param int|None slice_end:
     :param int|None slice_step:
+    :param bool force_old_tag:
```
albertz: No, please don't do that. It should be more generic. Please remove this.

TFNetworkLayer.py
```diff
@@ -2250,9 +2251,12 @@ def __init__(self, axis, slice_start=None, slice_end=None, slice_step=None, **kw
...
       if not DimensionTag.get_tag_from_size_tensor(self.output.size_placeholder[axis_wo_batch]):
-        tag = DimensionTag(
-          description="slice%i:%s" % (axis_wo_batch, self.get_absolute_name()),
-          kind=DimensionTag.Types.Spatial)
+        if not force_old_tag:
+          tag = DimensionTag(
+            description="slice%i:%s" % (axis_wo_batch, self.get_absolute_name()),
+            kind=DimensionTag.Types.Spatial)
+        else:
+          tag = DimensionTag.get_tag_from_size_tensor(self.input_data.size_placeholder[axis_wo_batch])
```
albertz: Please remove this.

TFUtil.py
```diff
@@ -194,6 +194,8 @@ def is_equal(self, other, ignore_feature_dim=False, allow_same_feature_dim=False
         return True
     if self.description == other.description:
       return True
+    elif "time" in self.description and "time" in other.description:
+      return True
```
albertz: Please remove this.

albertz:
> Inputs in combine layer may contain different tags for time_dim => an error is thrown.

Yes, that's correct, and that's how it should be.

> Solution - force tags to be equal if they contain 'time' in the description.

No. Nothing is needed. If you really want to force that, you could use `ReinterpretDataLayer`.

But I don't think you need to force anything. Why? Can you explain better where in your test case it goes wrong? Which combine layer? What is the error exactly? What different time dim tags do you get there?

squarenabla:
> > Inputs in combine layer may contain different tags for time_dim => an error is thrown.
> 
> Yes, that's correct, and that's how it should be.
> 
> > Solution - force tags to be equal if they contain 'time' in the description.
> 
> No. Nothing is needed. If you really want to force that, you could use `ReinterpretDataLayer`.
> 
> But I don't think you need to force anything. Why? Can you explain better where in your test case it goes wrong? Which combine layer? What is the error exactly? What different time dim tags do you get there?

In case of self attention layer, there're two flows. One is modified with Window layer with padding. As a result time dim is reduced. Another flow is modified with a Slice layer, to match the time dim size. But due to different tags they can't be combined (see "att_energy_biased" layer in the test)

squarenabla:
File "/u/bozheniuk/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/framework/ops.py", line 1983, in __init__
    line: self._c_op = _create_c_op(self._graph, node_def, grouped_inputs,
                                    control_input_ops)
    locals:
      self = <local> !AttributeError: 'Operation' object has no attribute '_c_op'
      self._c_op = <local> !AttributeError: 'Operation' object has no attribute '_c_op'
      _create_c_op = <global> <function _create_c_op at 0x7fc585562aa0>
      self._graph = <local> <tensorflow.python.framework.ops.Graph object at 0x7fc56d598890>
      node_def = <local> name: "att_energy_biased/Add"
                         op: "Add"
                         attr {
                           key: "T"
                           value {
                             type: DT_FLOAT
                           }
                         }

      grouped_inputs = <local> [<tf.Tensor 'att_energy_biased/att_energy_scaled_output_add_spatial_dim:0' shape=(?, ?, 8, 4, 1, 15) dtype=float32>, <tf.Tensor 'att_energy_biased/att_queries_context_part_output_add_spatial_dim:0' shape=(?, 1, ?, 8, 4, 15) dtype=float32>]
      control_input_ops = <local> []
  File "/u/bozheniuk/tensorflow/local/lib/python2.7/site-packages/tensorflow/python/framework/ops.py", line 1822, in _create_c_op
    line: raise ValueError(str(e))
    locals:
      ValueError = <builtin> <type 'exceptions.ValueError'>
      str = <builtin> <type 'str'>
      e = <local> InvalidArgumentError()
     ValueError: Dimensions must be equal, but are 4 and 8 for 'att_energy_biased/Add' (op: 'Add') with input shapes: [?,?,8,4,1,15], [?,1,?,8,4,15].

This is the error, which I have. Different tags in time dim make `get_common_data` create an extra dim of size 1 in both inputs.

albertz:
The error just says that they are not compatible, which is correct. `[?,?,8,4,1,15]` is not compatible to `[?,1,?,8,4,15]`. Probably some `copy_compatible_to` is missing, or there is a bug in `copy_compatible_to` for this case.

But maybe you mean something different anyway. What you describe has nothing to do with the error you posted here. You described something about different time dim tags, but the error is unrelated. Adding a broadcast dim (1) should be not a problem.

> In case of self attention layer, there're two flows. One is modified with Window layer with padding. As a result time dim is reduced. Another flow is modified with a Slice layer, to match the time dim size. But due to different tags they can't be combined (see "att_energy_biased" layer in the test)

Please be much more specific what you mean. Just post the debug output, don't make it so complicated. I don't know what you mean by "flow", or "result time dim is reduced", or "different tags". What is different? Just show it.

squarenabla:
> As a result time dim is reduced
=> time dim is reduced. 

> "flow"
two data flows in the net topology. 

> "different tags"
Tags are not equal because `is_equal` returns `False` for tags with different description.
I don't know, why that should be the problem. I just want to combine two tensors, no matter if their tags are equal or not. 
I know the steps which lead to the error: tensors have different tags => common data have an extra dim => inputs are casted and extra dim is pasted => TF error. I just fix the first reason. If you insist, that all tensor should have the unique tag, then I don't understand the lines in the CombineLayer
https://github.com/rwth-i6/returnn/blob/a1be6760a556cbfd4276669a090962956837f567/TFNetworkLayer.py#L5608
https://github.com/rwth-i6/returnn/blob/a1be6760a556cbfd4276669a090962956837f567/TFNetworkLayer.py#L5609
Why should we create a common data?
Why should we broadcast the inputs according to it?

albertz:
Please be more specific. Just answer my questions. I don't really understand why you are making it so complicated.

> > As a result time dim is reduced
> 
> => time dim is reduced.

What do you mean by "time dim is reduced"? Again. Where. As I said, just show what you mean. What is the input. What is the output.

> > "flow"
> 
> two data flows in the net topology.

There is nothing like "flow". I have no idea what you are talking about. Do you mean layers? Do you mean tensors? Do you mean an instance of `Data`? What do you mean? Just be specific. Just show it. What is the object. What is the output.

> > "different tags"
> 
> Tags are not equal because `is_equal` returns `False` for tags with different description.
> I don't know, why that should be the problem. I just want to combine two tensors, no matter if their tags are equal or not.

Please be specific what you mean. Just show it! What is the input, what do you expect as an output.

> I know the steps which lead to the error: tensors have different tags => common data has an extra dim => inputs are casted and an extra dim is pasted => TF error.

As I explained (please read what I wrote), that error is not relevant. The extra dim is fine. The TF error is something unrelated (might be another bug though).

> I just fix the first reason. If you insist, that all tensors should have the unique tag, then I don't understand the lines in the CombineLayer

Please read the code more carefully. I think it's pretty clear. If not, ask specific questions. But please read the code first before asking questions.

> Why should we create a common data?

To make it compatible. Such that you can do e.g. `tf.add(x,y)`. For that, `x` and `y` must be compatible.

> Why should we broadcast the inputs according to it?

To make it compatible.

> Why can't we rely on TF to perform the combination operation?

Because TF does not automatically make it compatible. As you see e.g. in the error, which is one example. [?,?,8,4,1,15] is not compatible to [?,1,?,8,4,15], so it crashes. The code is there to fix it.

squarenabla:
I've made a test case which emulates the desired behavior and throws an error. To make it simple, what should be done to fix the error?

tests/test_TFUtil.py
```diff
@@ -469,6 +469,43 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data_modified_sources():
+  from TFNetworkLayer import WindowLayer, InternalLayer, ReduceLayer, SliceLayer
+  from TFNetwork import TFNetwork, ExternData
```
albertz: All tests in `test_TFUtil.py` should just depend on `TFUtil`, nothing else. Please simplify this test case. Or move it to `test_TFNetworkLayer.py` or so. Or better both.

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,88 @@ def test_concat_sources_missing_dim():
...
+    net_dict = {
+      "att_q": {"class": "linear", "activation": None, "with_bias": False, "n_out": querry_dim * num_heads, "from": "data:in0"},
+      "att_k": {"class": "linear", "activation": None, "with_bias": False, "n_out": key_dim * num_heads, "from": "data:in0"},
+      "att_v": {"class": "linear", "activation": None, "with_bias": False, "n_out": value_dim * num_heads, "from": "data:in0"},
+
+      "att_q_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, querry_dim), "from": "att_q"},
+      "att_k_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, key_dim), "from": "att_k"},
+      "att_v_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, value_dim), "from": "att_v"},
+
+      "att_queries_crop": {"class": "slice", "force_old_tag": True, "axis": "t", "slice_start": crop_left, "slice_end": -crop_right, "from": "att_q_split"},
```
albertz: `force_old_tag` is invalid.

tests/test_TFUtil.py
squarenabla: Done

tests/test_TFNetworkLayer.py
squarenabla: Done

tests/test_TFUtil.py
```diff
@@ -469,6 +469,24 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
...
+  from TFUtil import DimensionTag
+  tag1 = DimensionTag(description="slice1",
+                      kind=DimensionTag.Types.Spatial)
+  tag1.set_tag_on_size_tensor(d1.size_placeholder[0])
+  tag2 = DimensionTag(description="slice2",
+                      kind=DimensionTag.Types.Spatial)
+  tag2.set_tag_on_size_tensor(d2.size_placeholder[0])
+  common = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  assert d1.shape == d2.shape
+  assert common.shape == d1.shape
```
albertz: This assert is wrong. They have different time dims, and thus the `common.shape` should be `(None, None, 32, 128)`.

squarenabla: But that's my case, I have d1 and d2, with different time dims. And I want to combine them in the CombineLayer. And I don't want  `common.shape` have an extra dim. You see what I mean?

albertz: That does not make sense. If d1 is e.g. of shape (3,T1=4,5), and d2 is of shape (3,T2=7,5), you cannot combine T1 and T2. It has to become (3,T1,T2,5).

squarenabla: But I take care of that. I don't need that check. Do you see more high level tests?

squarenabla: My topology guarantees that T1 and T2 are equal despite they have different tags

albertz: But then this is a bug/problem with your setup, or elsewhere. They should have the same tag if they are the same. If they have different tags, that means they are different. That's the whole point of having this.

squarenabla: I added a test with the attention layer `test_self_att_config`. What's wrong with it? Indeed the input is split into key, querry, and value. Each is processed differently and your new tagging system assigns them different tags. And that makes impossible to combine them later.
It worked fine with old version of returnn, so there's no bugs in the setups.

squarenabla: As I see the problem, the tagging doesn't work properly with all layers, which modify the data shape. E.g. window layer and slice layer. Maybe some others.

albertz: This thread here is about `test_Data_get_common_data_modified_time_dim`, nothing else. I'm explaining you now multiple times that this test is invalid, and the behavior is correct, and as intended.
For the other tests, or in general, I already asked you, or told you, that you should show more specifically, where you think it goes wrong. Where exactly. What is the input there, what is the output. Exactly at what step is it wrong?

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,125 @@ def test_concat_sources_missing_dim():
...
+      "att_queries_key_part": {"class": "slice", "axis": "f", "slice_start": 0, "slice_end": querry_dim - context_size, "from": "att_queries_crop"},
+      "att_queries_context_part": {"class": "slice", "axis": "f", "slice_start": querry_dim - context_size, "slice_end": querry_dim, "from": "att_queries_crop"},
+
+
+      "att_key_window": {"class": "window", "window_size": context_size, "padding": "valid", "from": "att_k_split"},
+      "att_energy_dot": {"class": "dot", "red1": "f", "red2": "f", "var1": None, "var2": "s:1", "from": ["att_queries_key_part", "att_key_window"]},
+      "att_energy_scaled": {"class": "eval", "eval": "source(0) * %f" % key_scale, "from": "att_energy_dot"},
+
+
+      "att_energy_biased": {"class": "combine", "kind": "add", "from": ["att_energy_scaled", "att_queries_context_part"]},
```
squarenabla: Here, two inputs `att_energy_scaled`, `att_queries_context_part` have modified time dims. `att_energy_scaled` was windowed and then dot product was applied, `att_queries_context_part` was just sliced. The inputs can't be combined since they have different tags in time dim.

albertz: Please show what they have (just show the output where we can see that).

tests/test_TFUtil.py
squarenabla: > This thread here is about `test_Data_get_common_data_modified_time_dim`, nothing else. I'm explaining you now multiple times that this test is invalid, and the behavior is correct, and as intended.

Ok

tests/test_TFNetworkLayer.py
squarenabla: Exception creating layer <network via test_self_att_config>/'att_energy_biased' of class   CombineLayer with opts:
{'kind': 'add',
 'name': 'att_energy_biased',
 'network': <TFNetwork '<network via test_self_att_config>' train>,
 'output': Data(name='att_energy_biased_output', shape=(None, 8, 4, None, 15), batch_shape_meta=[B,T|u"time:var:u'att_key_window/Maximum:0':att_key_window/att_energy_scaled_output",8,4,'slice0:att_queries_crop',F|15]),
 'sources': [<EvalLayer 'att_energy_scaled' out_type=Data(shape=(None, 8, 4, 15), batch_shape_meta=[B,T|u"time:var:u'att_key_window/Maximum:0':att_key_window/att_energy_scaled_output",8,4,F|15])>,
             <SliceLayer 'att_queries_context_part' out_type=Data(shape=(None, 8, 4, 15), batch_shape_meta=[B,T|'slice0:att_queries_crop',8,4,F|15])>]}

squarenabla: Also, you can see from the above error, that the output template is wrong, extra `None` dim is pasted.

albertz: No, I just see that it is exactly correct, as intended. The inputs have two different time dims (`T|u"time:var:u'att_key_window/Maximum:0':att_key_window/att_energy_scaled_output"` vs `T|'slice0:att_queries_crop'`), thus, as already explained multiple times now, the common data shape will have both time dims.

squarenabla: So, you say that everything works correct there. Ok, then I'll assume the tagging system as a black box. The question is, how to make my config work? It worked before, but now it doesn't.

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,141 @@ def test_concat_sources_missing_dim():
...
+                             output=CombineLayer.get_out_data_from_opts(name="combine", network=net,
+                                                                        kind="add", sources=[sl, red]))
+    tf.global_variables_initializer().run()
+    in1, in2 = session.run([combine.sources[0].output.placeholder,
+                            combine.sources[1].output.placeholder],
+                           feed_dict=feed_dict)
+    assert in1.shape == (10, 7, 32, 128)
+    assert in2.shape == (10, 7, 32, 128)
+
+    res = session.run(combine.output.placeholder, feed_dict=feed_dict)
```
squarenabla: There's an error here:
      message = <local> u'Incompatible shapes: [10,7,32,1,128] vs. [10,1,7,32,128]\n\t [[node combine/Add (defined at /u/bozheniuk/crnn-dev/TFNetworkLayer.py:5612) ]]\n\nErrors may have originated from an input operation.\nInput Source operations connected to node combine/Add:\n combine/red_output_add_spatial_dim (defin..., len = 331
     InvalidArgumentError: Incompatible shapes: [10,7,32,1,128] vs. [10,1,7,32,128]

Previous assert are passed.

tests/test_TFNetworkLayer.py
albertz: > So, you say that everything works correct there.

No. Please read carefully what I write. I just wrote that the output template is correct, exactly as intended. I did not say anything else. I just explained again why the output template is correct here.

> Ok, then I'll assume the tagging system as a black box.

No, don't do that. Why? It's not a black box. Read the code and please understand it (ask questions if sth is unclear). It's really simple, though. I already explained it multiple times now.

* If you have two different time dims, you also end up with two different time dims.
* If you have the same time dim, you end up with a single (the same) time dim.

That behavior is very simple, and that is exactly as it should be, and that works correct.
If the outcome of this is not what you want (having two different time dims), then by logical deduction, what you really want is that the time dim is not different but the same (for `att_energy_scaled` and `att_queries_context_part`). That is totally a different topic now. Why do you keep insisting that the logic of `get_common_data` is wrong? It's not. It's correct. The issue is somewhere else. Please try to understand that.

Thus, the real question is, why do `att_energy_scaled` and `att_queries_context_part` have different time dims?
And the next question is, how do you fix it (your config, or sth in Returnn) such that they have the same time dim?

> The question is, how to make my config work? It worked before, but now it doesn't.

Please just read what I wrote, and understand the real issue.

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,141 @@ def test_concat_sources_missing_dim():
     session.run(out.output.placeholder)
 
 
+def test_self_att_config():
```
albertz: Please use more meaningful names. The function name should point out what issue you are testing here (e.g. `test_self_att_time_dim_tags` or so; later when you understood what the problem is, you can maybe even use a better name).

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,141 @@ def test_concat_sources_missing_dim():
...
+
+    tf.global_variables_initializer().run()
+    out = session.run(fetches=fetches,
+      feed_dict=feed_dict)
+    inp1, inp2 = network.layers["att_energy_dot"].sources[0].output.placeholder, network.layers["att_energy_dot"].sources[1].output.placeholder
+    res = network.layers["att_energy_dot"].output.placeholder
+    inp1, inp2, res = session.run([inp1, inp2, res], feed_dict=feed_dict)
+
+
+def test_CombineLayer():
```
albertz: Please use more meaningful names. The function name should point out what issue you are testing here (e.g. `test_WindowLayer_ReduceLayer_dim_tags` or so). Later when you understood what the problem is, you can maybe even use a better name. In any case, `CombineLayer` should not be part of the name, as this is likely not the problem (please read again all my comments here in this issue to understand why).

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,141 @@ def test_concat_sources_missing_dim():
...
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
+
+    with tf.variable_scope("src"):
+      src = InternalLayer(name="src", network=net, out_type={"dim": 128,
+                                                             "shape": (None, 32, 128),
+                                                             "batch_dim_axis": 0,
+                                                             "time_dim_axis": 1,
+                                                             "feature_dim_axis": 3,
+                                                             "sparse": False
+                                                             })
```
albertz: Please don't format the code in this way. Never indent so far to the right. Rather, just use new line, such that every further indent is *always* exactly 2 spaces.
Also, for debugging purpose, it would be helpful to have sth like this:
```print("src:", src)```
(The same also for the following layers.)

File: tests/test_TFNetworkLayer.py (hunk 8 above)
Comment (albertz) [reply tests/test_TFNetworkLayer.py]: Can you try to reproduce this problem just by using `Data` (or other things from `TFUtil`), and create a test case in `test_TFUtil.py`?

File: tests/test_TFNetworkLayer.py (hunk 9 above)
Comment (squarenabla) [reply tests/test_TFNetworkLayer.py]: Done

File: tests/test_TFNetworkLayer.py (hunk 10 above)
Comment (squarenabla) [reply tests/test_TFNetworkLayer.py]: Done

File: tests/test_TFNetworkLayer.py (hunk 8 above)
Comment (squarenabla) [reply tests/test_TFNetworkLayer.py]: I committed a new test in the test_TFUtit.py

File: tests/test_TFNetworkLayer.py [hunk 12]
```diff
@@ -150,6 +150,137 @@ def test_concat_sources_missing_dim():
     session.run(out.output.placeholder)
 
 
+def test_self_att_time_dim_tags():
+  with make_scope() as session:
+    n_in = 128
+    channels = 16
+    n_time = 128
```
albertz: Use a different number, not the same as `n_in`, such that debugging becomes easier.
(Also, better use prime numbers, such that certain bugs are not hidden.)

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,137 @@ def test_concat_sources_missing_dim():
...
+    n_batch = 10
+    context_size = 15
+    rnd = numpy.random.RandomState(42)
+    querry_dim = 32
+    querry_dim += context_size
+    key_dim = 32
+    value_dim = 32
```
albertz: Use different numbers for `querry_dim`, `key_dim` and `value_dim`.

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,137 @@ def test_concat_sources_missing_dim():
...
+      "att_q_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, querry_dim), "from": "att_q"},
+      "att_k_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, key_dim), "from": "att_k"},
+      "att_v_split": {"class": "split_dims", "axis": "f", "dims": (num_heads, value_dim), "from": "att_v"},
+
+      "att_queries_crop": {"class": "slice", "axis": "t", "slice_start": crop_left, "slice_end": -crop_right, "from": "att_q_split"},
+
+      "att_queries_key_part": {"class": "slice", "axis": "f", "slice_start": 0, "slice_end": querry_dim - context_size, "from": "att_queries_crop"},
+      "att_queries_context_part": {"class": "slice", "axis": "f", "slice_start": querry_dim - context_size, "slice_end": querry_dim, "from": "att_queries_crop"},
+
+
```
albertz: Don't use two empty lines (btw, I think PyCharm warns about that, or not?).

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,137 @@ def test_concat_sources_missing_dim():
...
+    slice_size = window_size - 1
+
+    rnd = np.random.RandomState(42)
+    input_data = rnd.rand(10, 11, 32, 128)
+    seq_lens = np.array([11, 11, 11, 11, 11, 11, 11, 11, 11, 11])
+
+    with tf.variable_scope("src"):
+      src = InternalLayer(name="src", network=net, out_type={"dim": 128, "shape": (None, 32, 128),
+                                                             "batch_dim_axis": 0, "time_dim_axis": 1,
+                                                             "feature_dim_axis": 3, "sparse": False})
```
albertz: Can you reformat that, such that the indent is only 2 spaces?

tests/test_TFNetworkLayer.py
```diff
@@ -150,6 +150,137 @@ def test_concat_sources_missing_dim():
...
+                        output=ReduceLayer.get_out_data_from_opts(name="red", network=net, mode="AVG",
+                                                                  axes="s:1", keep_dims=False, sources=[win]))
+    with tf.variable_scope("slice"):
+      sl = SliceLayer(name="slice", axis="t", slice_start=slice_size // 2, slice_end=-(slice_size // 2),
+                      sources=[src], network=net,
+                      output=SliceLayer.get_out_data_from_opts(name="slice", axis="t", slice_start=slice_size // 2,
+                                                               slice_end=-(slice_size // 2), sources=[src],
+                                                               network=net))
+    with tf.variable_scope("combine"):
+      combine = CombineLayer(name="combine", network=net, kind="add", sources=[sl, red],
```
albertz: Please add some print statements, for debugging, such as:
```
print("sl:", sl)
print("red:", red)
print("combine:", combine)
```

tests/test_TFUtil.py
```diff
@@ -469,6 +469,30 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data_modified_time_dim():
+  d1 = Data(name='t', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  # simulating the changes in the time dim
+  d1.size_placeholder[0] -= 1
+  d2.size_placeholder[0] -= 1
```
albertz: This is not really needed. The time dims of `d1` and `d2` are already different from each other (why do assume that they are the same?).

tests/test_TFUtil.py
```diff
@@ -469,6 +469,30 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
...
+
+  from TFUtil import DimensionTag
```
albertz: You don't need to import anything from `TFUtil` here. Remove that.

tests/test_TFUtil.py
```diff
@@ -469,6 +469,30 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
...
+  # doing tagging a la SliceLayer for d1.size_placeholder
+  if not DimensionTag.get_tag_from_size_tensor(d1.size_placeholder[0]):
```
albertz: Why this check? You are constructing the test case. And before the `size_placeholder[0] -= 1`, it already has a valid dim tag, so you don't need to set a new one. Just remove the `size_placeholder[0] -= 1`. And then also this whole `if` branch.

tests/test_TFUtil.py
```diff
@@ -469,6 +469,30 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
...
+    tag = DimensionTag(
+      description="slice1",
+      kind=DimensionTag.Types.Spatial)
+    tag.set_tag_on_size_tensor(d1.size_placeholder[0])
+
+  # WindowLayer does no tagging on size_placeholder
+  # => d2.size_placeholder untagged
```
albertz: Remove this comment. This is just about `TFUtil`.
Also, the comment is wrong. `d2.size_placeholder` is tagged (when you remove the `d2.size_placeholder[0] -= 1`).
What case are you actually trying to produce here? That `d1` and `d2` have different time dim tags? You already have that from the very beginning.
Or that `d1.size_placeholder` is tagged but `d2.size_placeholder` is not tagged?

Btw, if you are suggesting that `WindowLayer` does not correctly set the time dim tag, then you should add a test case which shows and demonstrates and checks for exactly that: I.e. just using and testing `WindowLayer` (nothing else; keep focused in a test case) (this would go to `test_TFNetworkLayer.py` again).

tests/test_TFNetworkLayer.py
squarenabla: Done

tests/test_TFNetworkLayer.py
squarenabla: Already done

tests/test_TFUtil.py
squarenabla: Data constructs tags for the size_placeholders. After the operation (-=1) it's another placeholder. And there's no tag. i.e.
`DimensionTag.get_tag_from_size_tensor(d1.size_placeholder[0])` becomes `None`.

tests/test_TFUtil.py
squarenabla: > Or that `d1.size_placeholder` is tagged but `d2.size_placeholder` is not tagged

You asked to reproduce the error within TFUtil.py. 

> Can you try to reproduce this problem just by using Data (or other things from TFUtil), and create a test case in test_TFUtil.py?

Here's the case.

tests/test_TFUtil.py
albertz: I know what it does and what the effect is. I just say, you don't need that. But maybe I misunderstood what situation exactly you want to construct here. But in any case, whatever situation you try to construct (see my other question also), you are doing it too complicated.
Please also add a comment, saying what situation you want to construct (e.g. `# We construct it such that d1 and d2 have different time dim tags.`).

tests/test_TFUtil.py
squarenabla: > Btw, if you are suggesting that `WindowLayer` does not correctly set the time dim tag, then you should add a test case which shows and demonstrates and checks for exactly that: I.e. just using and testing `WindowLayer` (nothing else; keep focused in a test case) (this would go to `test_TFNetworkLayer.py` again).

From the very beginning of the discussion I suggested, that tagging doesn't work properly for WindowLayer and SliceLayer.

albertz: I'm not sure what you are saying. Please answer my comment / question.

tests/test_TFUtil.py
squarenabla: I updated the test. Indeed there's no need in modifying them, the error remains.

tests/test_TFNetworkLayer.py
squarenabla: Done

tests/test_TFNetworkLayer.py
squarenabla: Done

tests/test_TFUtil.py
squarenabla: > Btw, if you are suggesting that `WindowLayer` does not correctly set the time dim tag, then you should add a test case which shows and demonstrates and checks for exactly that: I.e. just using and testing `WindowLayer` (nothing else; keep focused in a test case) (this would go to `test_TFNetworkLayer.py` again).

From the very beginning I suggested that the tagging system doesn't work for some layer. In particular, SliceLayer assigns one tag, Window layer assigns no tag. And as a result the outputs of these two layers can't be processed in the CombineLayer. So, I do the test on this particular situation.

tests/test_TFNetworkLayer.py
squarenabla: Done

tests/test_TFUtil.py
albertz: > From the very beginning of the discussion I suggested, that tagging doesn't work properly for WindowLayer and SliceLayer.

Where? Make a very specific test case for exactly / only that. One for `WindowLayer` only. Another one for `SliceLayer` only. Just show directly where sth is wrong. Keep focused in a test case. Keep it simple.

Also, you did not answer/addressed my comments here:

Remove this comment. This is just about TFUtil.
Also, the comment is wrong. d2.size_placeholder is tagged (when you remove the d2.size_placeholder[0] -= 1).

What case are you actually trying to produce here?
That d1 and d2 have different time dim tags? You already have that from the very beginning.
Or that d1.size_placeholder is tagged but d2.size_placeholder is not tagged?
What case exactly? (Don't just answer "the error I have". Say *what error*, *what situation exactly*.)

tests/test_TFUtil.py
```diff
@@ -469,6 +469,15 @@ def test_Data_copy_compatible_to_batch_feature_is_dynamic():
   assert_equal(t.get_size_dim_tag(0), dec.get_time_dim_tag())
 
 
+def test_Data_get_common_data_different_time_dim():
+  d1 = Data(name='t', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  d2 = Data(name='r', shape=(None, 32, 128), dtype='float32', auto_create_placeholders=True)
+  common_data = Data.get_common_data([d1, d2], warnings_out=sys.stdout)
+  op1 = d1.copy_compatible_to(common_data).placeholder
+  op2 = d2.copy_compatible_to(common_data).placeholder
+  assert op1.shape == op2.shape
```
albertz: The `assert` is wrong. In one case, you get sth like (None, 1, 32, 128), in the other case, you get sth like (1, None, 32, 128), so they will not be the same (the purpose of `copy_compatible_to` is to make them compatible, not to make them the same).
Btw, is it on purpose that you take the `placeholder` here? Why not just the `Data` instance directly? Please do both in the code, and add a comment why you do the `placeholder` check.

tests/test_TFUtil.py
squarenabla: > > From the very beginning of the discussion I suggested, that tagging doesn't work properly for WindowLayer and SliceLayer.
> 
> Where? Make a very specific test case for exactly / only that. One for `WindowLayer` only. Another one for `SliceLayer` only. Just show directly where sth is wrong. Keep focused in a test case. Keep it simple.

`WindowLayer` and `SliceLayer` work ok when separate. The main issue is that the tags on the placeholders are different, when processed by different layers. And they can't be processed by the `CombineLayer`. I think, one of the solution, is to make `CombineLayer` less discriminative. Maybe by modifying `Data.get_common_data()`, s.t. `common_data` has a single time_dim even if the inputs have different tags on it.

> Remove this comment. This is just about TFUtil.
> Also, the comment is wrong. d2.size_placeholder is tagged (when you remove the d2.size_placeholder[0] -= 1).
I removed `d2.size_placeholder[0] -= 1`


> What case are you actually trying to produce here?
> That d1 and d2 have different time dim tags? You already have that from the very beginning.
> Or that d1.size_placeholder is tagged but d2.size_placeholder is not tagged?
> What case exactly? (Don't just answer "the error I have". Say _what error_, _what situation exactly_.)

I reproduce the test s.t. two Data instances have different tags for the time dim. Then, I create a `common_data`. Then make a copy for d1 and d2. Then compare the shapes. The same procedure as in the `CombineLayer`. Because you asked to reproduce the problem by using the `Data`.

tests/test_TFUtil.py
squarenabla: See comment above, this is the behavior of the `CombineLayer`. In particular, this method:
https://github.com/rwth-i6/returnn/blob/68ee06301cd7597ba3909d8a5baebfbc27a72308/TFNetworkLayer.py#L5598
`  def _op_dense_fn(sources, fn):
    """
    :param ((x1,x2) -> y) fn: function to perform on x1 and x2
    :rtype: tf.Tensor
    """
    # Earlier we checked for the same dense dim.
    # Now, we completely rely on Data.get_common_data and Data.copy_compatible_to.
    # That should fail if they are not compatible. Otherwise it would add any needed broadcast dimensions.
    # All the dense element-wise functions should be able to deal with broadcasting.
    common_data = Data.get_common_data([s.output for s in sources])
    x = sources[0].output.copy_compatible_to(common_data).placeholder
    for source in sources[1:]:
      x2 = source.output.copy_compatible_to(common_data).placeholder
      x = fn(x, x2)
    return x`

tests/test_TFNetworkLayer.py
squarenabla: > Why do you keep insisting that the logic of `get_common_data` is wrong? It's not. It's correct. The issue is somewhere else. Please try to understand that.
 
I agree, that `get_common_data` works fine. That's why I assume it as a black box. Then I do the tests with config, layer and imitate the behavior by means of Data.py.

tests/test_TFUtil.py
albertz: The assert is wrong. In one case, you get sth like (None, 1, 32, 128), in the other case, you get sth like (1, None, 32, 128), so they will not be the same (the purpose of copy_compatible_to is to make them compatible, not to make them the same).
Btw, is it on purpose that you take the placeholder here? Why not just the Data instance directly? Please do both in the code, and add a comment why you do the placeholder check.

tests/test_TFUtil.py
albertz: > `WindowLayer` and `SliceLayer` work ok when separate. The main issue is that the tags on the placeholders are different, when processed by different layers.

This is a contradiction. I'm not sure what you are saying. The tags are the same before these layers, and then become different? (If you mean that, why are you not directly saying and showing that?) And this is wrong, or not? If this is wrong, you can clearly just make an individual test for that (checking that the time dim is the same as it was before the individual layer). If it is correct, then I don't see where the problem is, because CombineLayer is also correct. I.e. this is a contradiction.

> And they can't be processed by the `CombineLayer`. I think, one of the solution, is to make `CombineLayer` less discriminative.

No, I explained already, many times now, the behavior of `CombineLayer` (or more specifically, `Data.get_common_data`) is perfectly correct (for all the cases which you have shown so far).

> > What case are you actually trying to produce here?
> > That d1 and d2 have different time dim tags? You already have that from the very beginning.
> > Or that d1.size_placeholder is tagged but d2.size_placeholder is not tagged?
> > What case exactly? (Don't just answer "the error I have". Say _what error_, _what situation exactly_.)
> 
> I reproduce the test s.t. two Data instances have different tags for the time dim. ...

You are not answering my question. I gave multiple options in my question:
1. That d1 and d2 have different time dim tags?
2. Or that d1.size_placeholder is tagged but d2.size_placeholder is not tagged?
3. Something else.

Please make it simple. Just directly answer my question (sentence by sentence, direct answer). So, do I interpret it correctly, it is case 1?

tests/test_TFNetworkLayer.py
albertz: > I agree, that `get_common_data` works fine. That's why I assume it as a black box. Then I do the tests with config, layer and imitate the behavior by means of Data.py.

I don't understand. Why is it a black box? It's not. Just read the code. Also, I think it's actually pretty simple, what it does, and how it works.
Also, if you finally understand now that `get_common_data` (and also `CombineLayer`) work just fine, then maybe you can start searching where the real problem is.

albertz:
I just pushed some fixes which might also fix your issues.
One of the test cases I pushed is also similar to your `test_Data_get_common_data_different_time_dim`, so I guess you can remove that now.
Please rebase this to the latest master, and see what problems you still have.

squarenabla:
> I just pushed some fixes which might also fix your issues.
> One of the test cases I pushed is also similar to your `test_Data_get_common_data_different_time_dim`, so I guess you can remove that now.
> Please rebase this to the latest master, and see what problems you still have.

Got it

squarenabla:
After rebasing, tests still fail. E.g. 
    Executing: test_CombineLayer_after_WindowLayer_and_SliceLayer
    2019-09-02 12:49:46.851264: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU 
    supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX
    ('sl:', <SliceLayer 'slice' out_type=Data(shape=(None, 32, 128), batch_shape_meta=
    [B,T|'slice0:slice',32,F|128])>)
    ('red:', <ReduceLayer 'red' out_type=Data(shape=(None, 32, 128), batch_shape_meta=
    [B,T|u"time:var:u'win/Maximum:0':red/red_output",32,F|128])>)
    ('combine:', <CombineLayer 'combine' out_type=Data(shape=(None, None, 32, 128), 
    batch_shape_meta=[B,T|'slice0:slice',u"time:var:u'win/Maximum:0':red/red_output",32,F|128])>)
    EXCEPTION
    File "./tests/test_TFNetworkLayer.py", line 2683, in <module>
    line: globals()[arg]()  # assume function and execute
    locals:
      globals = <builtin> <built-in function globals>
      arg = <local> 'test_CombineLayer_after_WindowLayer_and_SliceLayer', len = 50
     File "./tests/test_TFNetworkLayer.py", line 320, in test_CombineLayer_after_WindowLayer_and_SliceLayer
    line: assert res.shape == (10, 7, 32, 128)
    locals:
      res = <local> array([[[[[0.67056465, 0.4985959 , 0.745118  , ..., 1.3219421 ,
                               1.577019  , 1.036398  ],
                              [0.42261794, 1.2963672 , 0.73444337, ..., 1.1170576 ,
                               1.1277343 , 1.4695563 ],
                              [0.5920633 , 1.3517644 , 1.3206363 , ..., 0.5968997 ,
                               1.3976564 , 0.44143954..., len = 10, _[0]: {len = 7, _[0]: {len = 7, _[0]: {len = 32}}}
      res.shape = <local> (10, 7, 7, 32, 128)
AssertionError

albertz:
But the shape (10, 7, 7, 32, 128) is correct. The assert is wrong. Why should it be (10, 7, 32, 128)? It combines two different time dims (`'slice0:slice'` and `u"time:var:u'win/Maximum:0':red/red_output"`), thus you get both in the output.
What is the output of the `win` layer? You miss that `print` there.

tests/test_TFNetworkLayer.py
squarenabla: Once again, this is a test of a simplified topology:
                        1) ->  WindowLayer ->  ReduceLayer ->
InternalLayer                                                                          CombineLayer
                        2)         -> SliceLayer ->

In the example above, InternalLayer outputs the tensor of size [10, 11, 32, 128]
The branch 1) takes the output [10, 11, 32, 128] and first applies a window of size 5(the result is [10, 7, 5, 32, 128]) then a reduce layer (the result is [10, 7, 32, 128]). In general case, if the input is [b, t, s:1, f] the output of 1) is [b, t-(win_size-1), s:1,  f]
The branch 2) just slices the input with a SliceLayer (the result is [10, 7, 32, 128]). In general case, if the input is [b, t, s:1, f] the output of 1) is [b, t-slice_size, s:1,  f] = [b, t-(win_size-1), s:1,  f]
The CombineLayer tries to add the outputs from 1) and from 2), which have the same size. But fails.

albertz:
What is the output of the win layer? You miss that print there. Can you add that? And write the output?

tests/test_TFNetworkLayer.py
albertz: Is this related to the error you write? You get still `Incompatible shapes` here?

tests/test_TFUtil.py
albertz: Is this test case still needed?

tests/test_TFNetworkLayer.py
```diff
@@ -182,6 +261,65 @@ def test_LinearLayer_batch_feature_major():
...
+          name="slice", axis="t", slice_start=slice_size // 2,
+          slice_end=-(slice_size // 2), sources=[src], network=net))
+    with tf.variable_scope("combine"):
+      combine = CombineLayer(
+        name="combine", network=net, kind="add", sources=[sl, red],
+        output=CombineLayer.get_out_data_from_opts(
+          name="combine", network=net,
+          kind="add", sources=[sl, red]))
+    tf.global_variables_initializer().run()
+    print("sl:", sl)
```
albertz: Print for `win` is missing here.

tests/test_TFNetworkLayer.py
albertz: You cannot know in general that branch 1 and branch 2 (and whatever other branch of modifying the size / variable-dim shape) are the same. How would you? (This is not just a rhetorical question. I don't really understand how you get that idea.)

So, what you want is, that you can explicitly mark in some way that they are in fact the same (that is the function `DimensionTag.declare_same_as`). Or, you could use `ReinterpretDataLayer` with `size_base`, like this:
```
"red1": {"class": "reinterpret_data", "from": "red", "size_base": "sl"},
"combine": {"class": "combine", "kind": "add", "from": ["sl", "red1"]},
```
That will reset the `size_placeholder`, and thus explicitly defines it to have the same time-dim.

albertz: (Btw, better write the whole test case using a dictionary to define the layers. That's simpler than constructing it manually like you do.)

squarenabla: yes, there's still an error: 
https://github.com/rwth-i6/returnn/pull/214#issuecomment-527103109

tests/test_TFUtil.py
squarenabla: I think no, I'll delete it

tests/test_TFNetworkLayer.py
albertz: I thought this discussion here was about `Incompatible shapes`, but now you write about a different issue.
Anyway, my suggestion was now specifically about what I think you actually want to have Although you still did not specifically said what you really want. I assume now that you want that the time dimensions (and their tags) are the same. (Please write explicitly if that is what you want.)

tests/test_TFNetworkLayer.py
squarenabla: Done

squarenabla:
With the win output:

(tensorflow) bozheniuk@cobalt:~/crnn-dev$ python ./tests/test_TFNetworkLayer.py test_CombineLayer_after_WindowLayer_and_SliceLayer
/u/bozheniuk/.local/lib/python2.7/site-packages/h5py/__init__.py:34: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Limited tf.compat.v2.summary API due to missing TensorBoard installation
('TF version:', '1.12.0')
Executing: test_CombineLayer_after_WindowLayer_and_SliceLayer
2019-09-02 16:20:23.373188: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX
('sl:', <SliceLayer 'slice' out_type=Data(shape=(None, 32, 128), batch_shape_meta=[B,T|'slice0:slice',32,F|128])>)
('win:', <WindowLayer 'win' out_type=Data(shape=(None, 5, 32, 128), batch_shape_meta=[B,T|'window0:win',5,32,F|128])>)
('red:', <ReduceLayer 'red' out_type=Data(shape=(None, 32, 128), batch_shape_meta=[B,T|'window0:win',32,F|128])>)
('combine:', <CombineLayer 'combine' out_type=Data(shape=(None, None, 32, 128), batch_shape_meta=[B,T|'slice0:slice','window0:win',32,F|128])>)
EXCEPTION
  File "./tests/test_TFNetworkLayer.py", line 2684, in <module>
    line: globals()[arg]()  # assume function and execute
    locals:
      globals = <builtin> <built-in function globals>
      arg = <local> 'test_CombineLayer_after_WindowLayer_and_SliceLayer', len = 50
  File "./tests/test_TFNetworkLayer.py", line 321, in test_CombineLayer_after_WindowLayer_and_SliceLayer
    line: assert res.shape == (10, 7, 32, 128)
    locals:
      res = <local> array([[[[[0.67056465, 0.4985959 , 0.745118  , ..., 1.3219421 ,
                               1.577019  , 1.036398  ],
                              [0.42261794, 1.2963672 , 0.73444337, ..., 1.1170576 ,
                               1.1277343 , 1.4695563 ],
                              [0.5920633 , 1.3517644 , 1.3206363 , ..., 0.5968997 ,
                               1.3976564 , 0.44143954..., len = 10, _[0]: {len = 7, _[0]: {len = 7, _[0]: {len = 32}}}
      res.shape = <local> (10, 7, 7, 32, 128)
AssertionError

albertz:
Please always properly format the output using Markdown.

tests/test_TFNetworkLayer.py
squarenabla: I have two outputs, from different sources and I want them to be added together. If their tags have to be equal, than yes, I want the tags to be equal as well.

albertz: They don't have to be equal. If they are not equal, i.e. you have two different time dims, you will just end up with two different time dims (as you would expect).
But as I understand you, you say, you want that they are considered as the same time dim.

albertz: I wrote you above one way how you can get that. Did you try?

squarenabla: > They don't have to be equal. If they are not equal, i.e. you have two different time dims, you will just end up with two different time dims (as you would expect).
> But as I understand you, you say, you want that they are considered as the same time dim.

Yes, I assure and consider, that they are the same.  

> I wrote you above one way how you can get that. Did you try?

You mean `ReinterpretDataLayer`? No, I didn't try

albertz:
What is the state here?

squarenabla:
The state is the same. I'll be back to the pull request in a few days

tests/test_TFNetworkLayer.py
albertz: Did you try now? Is this resolved?

albertz:
Still no update here. But as far as I understand, there is no real bug, this is more like a usage error on your side. So I'm closing this now.

PR #215, opened 2019-09-03
Collection of small changes

JackTemaki:
I wrote some small commits that I did not want to push directly into the master branch without review. I thought it is easier to bundle it in a single pull request.

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -2071,6 +2071,10 @@ class LossHolder:
         tf.summary.scalar("loss_%s" % name, self._loss_value * self._norm_factor)
         if self._network.get_config().bool("calculate_exp_loss", False):
           tf.summary.scalar("exp_loss_%s" % name, tf.exp(self._loss_value * self._norm_factor))
+        if self._network.get_config().bool("debug_unnormalized_loss_summaries", False):
+          tf.summary.scalar("unnormalized_loss_%s" % name, self._loss_value)
+        if self._network.get_config().bool("debug_objective_loss_summaries", False):
+          tf.summary.scalar("objective_loss_%s" % name, self._loss_value_for_objective)
     if self._error_value is not None:
       if self._error_value.get_shape().ndims == 0:
         tf.summary.scalar("error_%s" % name, self._error_value * self._norm_factor)
@@ -2102,7 +2106,6 @@ class LossHolder:
         "layer %r loss %r return None for loss and error" % (self._layer, self.loss))
     if self._norm_factor is None:
       self._norm_factor = self.loss.get_normalization_factor()
-    self._tf_summary()
     loss_value = self._loss_value
     if loss_value is not None:
       if self._network.get_config().bool("debug_add_check_numerics_on_output", False):
@@ -2120,6 +2123,7 @@ class LossHolder:
     if self.loss.use_normalized_loss and loss_value is not None:
       loss_value *= self._norm_factor
     self._loss_value_for_objective = loss_value
+    self._tf_summary()
     self._is_prepared = True
 
   def copy_new_base(self, name=None, layer=None, network=None, reduce_func=None):
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -1149,6 +1149,8 @@ class HDFDatasetWriter:
     print("Data keys:", data_keys, file=log.v3)
     if "orth" in data_keys:  # special workaround for now, not handled
       data_keys.remove("orth")
+    if "raw" in data_keys:
+      data_keys.remove("raw")
     data_target_keys = [key for key in dataset.get_target_list() if key in data_keys]
     data_input_keys = [key for key in data_keys if key not in data_target_keys]
     assert len(data_input_keys) > 0 and len(data_target_keys) > 0
--- a/docs/configuration_reference/debugging.rst
+++ b/docs/configuration_reference/debugging.rst
@@ -7,10 +7,20 @@ Debugging
 debug_add_check_numerics_on_output
     If set to ``True``, should assert for ``inf`` and ``nan``.
 
+debug_grad_summaries
+    If set to ``True``, adds additional information about the gradients to the TensorBoard.
+
 debug_print_layer_output_template
     If set to ``True``, print the layer template information during network construction.
 
 debug_print_layer_output_shape
     If set to ``True``, print the layer shape information while the graph is executed.
+
+debug_objective_loss_summaries
+    If set to ``True``, adds the objective loss values (normalization only when activated, including loss scaling)
+    to the TensorBoard.
+
+debug_unnormalized_loss_summaries
+    If set to ``True``, adds the unnormalized loss values to the TensorBoard
 
 Also see :ref:`debugging`.
--- a/docs/configuration_reference/generation_search.rst
+++ b/docs/configuration_reference/generation_search.rst
@@ -14,6 +14,10 @@ forward_override_hdf_output
     Per default, Returnn will give an error when trying to overwrite an existing output. If this flag is set to true,
     the check is disabled.
 
+output_file
+    When the task is "forward", specifies the output path for the resulting hdf. If not specified,
+    the name will be "dump-fwd-epoch-%i.hdf" % epoch.
+
 search_output_layer
     TODO...
 


TFNetwork.py
```diff
@@ -985,6 +985,9 @@ def inv_reduce_sum(x, name):
           d["cost:%s" % loss_name] = reduce_sum(loss.get_loss_value_for_fetch(), name="cost:%s" % loss_name)
         if loss.get_error_value() is not None:
           d["error:%s" % loss_name] = reduce_sum(loss.get_error_value(), name="error:%s" % loss_name)
+        if config.bool("debug_display_loss_per_layer", False):
+          if loss.get_loss_value_for_objective() is not None:
+            d["loss:%s" % loss_name] = reduce_sum(loss.get_loss_value_for_objective(), name="loss:%s" % loss_name)
```
albertz: The key name is confusing. It's already confusing that we have sort of "cost" == "loss", but this makes it more confusing.
Better name it `"cost_objective:..."` or so.
But anyway, is this really needed? I think adding it via `tf.summary.scalar` is probably better, right?
Note also that the difference is not necessarily that one is normalized and the other is not.

TFEngine.py
```diff
@@ -276,7 +276,8 @@ def _collect_eval_info(self, fetches_results):
     """
     # See see self._get_fetches_dict() for the keys.
     # keys are e.g. "cost:output", "error:output" or "loss".
-    keys = [k for k in fetches_results.keys() if k.startswith("cost:") or k.startswith("error:") or k == "loss"]
+    keys = [k for k in fetches_results.keys() if k.startswith("cost:") or k.startswith("error:") or
+            k.startswith("loss:") or k == "loss"]
```
albertz: The formatting is confusing. `if ...` should be on the same line with all the conditions (all the `or`). If the line becomes too long, make the newline before the `if`.

TFNetworkRecLayer.py
```diff
@@ -4670,10 +4670,13 @@ def _weights_time_axis_to_reduce(cls, weights, base):
     # SoftmaxOverSpatialLayer by default uses time_dim_axis.
     # We also should maybe check that this matches the base time dim axis.
     dyn_axes = weights.get_dynamic_axes()
+
```
albertz: Why this extra empty line?

TFNetworkRecLayer.py
```diff
@@ -4670,10 +4670,13 @@ def _weights_time_axis_to_reduce(cls, weights, base):
...
     # for static time-dim
-    if weights.time_dim_axis not in dyn_axes:
+    if not dyn_axes or weights.time_dim_axis not in dyn_axes:
+      assert weights.time_dim_axis is not None and weights.time_dim_axis is not NotSpecified
       return weights.time_dim_axis
     assert dyn_axes, "no dynamic axes in %r" % weights
+
```
albertz: Why this extra empty line?

TFNetworkRecLayer.py
```diff
@@ -4670,10 +4670,13 @@ def _weights_time_axis_to_reduce(cls, weights, base):
     # SoftmaxOverSpatialLayer by default uses time_dim_axis.
     # We also should maybe check that this matches the base time dim axis.
     dyn_axes = weights.get_dynamic_axes()
+
     # for static time-dim
-    if weights.time_dim_axis not in dyn_axes:
+    if not dyn_axes or weights.time_dim_axis not in dyn_axes:
```
albertz: This change (adding `not dyn_axes`) does not change anything. Why do you change it then? It's confusing now, because `not dyn_axes` is always False if the other condition is False, and if the other condition is True, it does not matter.

TFNetworkRecLayer.py
```diff
@@ -4670,10 +4670,13 @@ def _weights_time_axis_to_reduce(cls, weights, base):
...
+      assert weights.time_dim_axis is not None and weights.time_dim_axis is not NotSpecified
```
albertz: It never can be `NotSpecified`. That check does not make sense.
Also, I don't quite understand how this fixes a problem.

JackTemaki: This seems to be a merge error, as the check for static time axis did not exist when I made the change a long time ago (but never did a commit for it before), so the commit can be removed

albertz: Ok, then remove it, and update the PR.

albertz: And resolve this when done.

TFNetwork.py
JackTemaki: I used this to be able to compare the actual loss of two different objectives and tune my loss scaling accordingly. The "cost" values do not include the loss scaling and seemed to be normalized with respect to the sequence length. I did not see another way to solve my issue.

albertz: This does not address any of my points?

* The key name is confusing. It's already confusing that we have sort of "cost" == "loss", but this makes it more confusing.
Better name it "cost_objective:..." or so.
* Is `tf.summary.scalar` maybe better? (That would have the advantage that the code is much simpler.)
* Note also that the difference is not necessarily that one is normalized and the other is not.

JackTemaki: I refered to "But anyway, is this really needed?". Besides of how to use tf.summary.scalar (the tensorflow page did not help), I understand your points.

albertz: Do you know TensorBoard? See TF documentation about that. Putting stuff on stdout is somewhat limited, and I think for more extra info, TensorBoard (via `tf.summary`) should always be used. That's what we used already a lot (e.g. stuff like gradient norms, or basically anything).

JackTemaki: Yes, I used Tensorboard once but it did not show more than the logs/learning-rate file itself for me, so I never used it again (are there flags for enabling more values?). But if you want it do be added only there it is ok, and I guess the option can be removed. The question is then how to distinguish the losses? "score" (in the learning rates file), "cost" (log) and "loss" (tensorboard) are all the same, so how should I call the actual loss? "objective_loss"? Also, they would be equal when `loss.use_normalized_loss` is `True`, but that should be ok.

albertz: TensorBoard should always show all losses.
See all `tf.summary.scalar` usages.
You will find (among other things) this:
```
tf.summary.scalar("loss_%s" % name, self._loss_value * self._norm_factor)
```
If you have your debug option enabled, you could add this:
```
tf.summary.scalar("loss_unnormalized_%s" % name, self._loss_value)
```

JackTemaki: Yes, I already checked the Tensorboard again, and also the function `_tf_summary` of the `LossHolder`. My question was how you can enable more than the all losses that are already part of the logs. (e.g, the gradient norms you mentioned)

Also unnormalized might be the wrong term, as (if I understand correctly) the actual loss can be normalized. Or do you want to display normalized and un-normalized, and let the user have to know which is used for gradient computation and which is not?

albertz: Just see all usages of `tf.summary.scalar`. E.g. there is `debug_grad_summaries`.
`_loss_value` at that point is not normalized. In any case, the only difference is that you have factor `self._norm_factor` in one case and not in the other case, so I think it is a reasonable name (or `self._loss_value * self._norm_factor` would be wrong then, right?).

JackTemaki: I am just saying that my initial code had a slightly different behaviour. But yes, I can add it as always normalized and un-normalized. I will also add `debug_grad_summaries` to the docs.

JackTemaki:
what do you think about this? I now added two flags, one for unnormalized losses and one for the actual objective loss (including the scaling, and normalized only if activated).

TFNetwork.py
```diff
@@ -2071,6 +2071,10 @@ def _tf_summary(self):
         tf.summary.scalar("loss_%s" % name, self._loss_value * self._norm_factor)
         if self._network.get_config().bool("calculate_exp_loss", False):
           tf.summary.scalar("exp_loss_%s" % name, tf.exp(self._loss_value * self._norm_factor))
+        if self._network.get_config().bool("debug_unnormalized_loss_summaries", False):
+          tf.summary.scalar("unnormalized_loss_%s" % name, self._loss_value)
+        if self._network.get_config().bool("debug_objective_loss_summaries", False):
+          tf.summary.scalar("objective_loss_%s" % name, self._loss_value_for_objective)
```
albertz: Did you test this? Do you actually use this? I think this does not work, as `_loss_value_for_objective` is not set at this point. I would just remove it, if you don't use it. Don't just add random features.
See where `_tf_summary` is called. The logic currently makes sure that `_loss_value` at that point is actually exactly the loss used for the objective. I think this option is confusing.

albertz: Ah, I'm just seeing, you moved the `_tf_summary` call.
But anyway, I'm not really sure why you would need `objective_loss_...`. How is this useful? You either have the information in `loss_...` or `unnormalized_loss_...`.

TFEngine.py
albertz: Please resolve this thread if it is resolved (or outdated).

TFNetworkRecLayer.py
albertz: I guess this is resolved?

TFNetwork.py
JackTemaki: Normalized and unnormalized do not include the loss scaling. I use pretraining to modify the loss scaling over time, and i want to now the actual relation of the losses during the training. And for me it seemed only `_loss_value_for_objective` contains the exact loss for gradient computation in all cases.

albertz: Ah I see, ok.

PR #216, opened 2019-09-03
PositionalEncodingLayer, changes to support offset

kazuki-irie:
Changes to support `offset` option in `PositionalEncodingLayer`.

This enables using another layer as offsets to shift positions.

In addition, this removes the previously available option to specify `batch_dim_axis` independently of the input when `add_to_input` is `False`. The output shape is now always set to be the same as the input shape. Please let me know if anyone is against this change.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -5293,15 +5293,14 @@ class SelfAttentionLayer(_ConcatInputLayer):
 
 class PositionalEncodingLayer(_ConcatInputLayer):
   """
-  Provides positional encoding in the form of (batch, time, n_out),
-  where n_out is the number of channels,
-  if it is run outside a :class:`RecLayer`,
-  or (batch, n_out)
+  Provides positional encoding in the form of (batch, time, n_out) or (time, batch, n_out) same as the source
+  where n_out is the number of channels, if it is run outside a :class:`RecLayer`,
+  and (batch, n_out) or (n_out, batch) same as the source
   if run inside a :class:`RecLayer`, where it will depend on the current time frame.
 
   Assumes one source input with a time dimension if outside a :class:`RecLayer`.
-  By default ("from" key not provided), it would either use "data", or ":i".
   With `add_to_input`, it will calculate `x + input`.
+  The output shape is the same as the input in all cases.
 
   The positional encoding is the same as in Tensor2Tensor.
   See :func:`TFUtil.get_positional_encoding`.
@@ -5309,33 +5308,54 @@ class PositionalEncodingLayer(_ConcatInputLayer):
   layer_class = "positional_encoding"
   recurrent = True
 
-  def __init__(self, add_to_input=False, constant=-1, **kwargs):
+  def __init__(self, add_to_input=False, constant=-1, offset=None, **kwargs):
     """
     :param bool add_to_input: will add the signal to the input
     :param int constant: if positive, always output the corresponding positional encoding.
+    :param None|LayerBase offset: Specify the offset to be added to positions. Expect shape (batch, time) or (batch,).
     """
     super(PositionalEncodingLayer, self).__init__(**kwargs)
     assert len(self.sources) == 1, "%s: expect a single source" % self
     source = self.input_data
+    assert source.feature_dim_axis == source.batch_ndim - 1  # Must be last
     if add_to_input:
       assert source.dim == self.output.dim
+    output_templ_wo_feat = self.output.copy_template_excluding_axis(self.output.feature_dim_axis)
+    offset_data = None
+    if offset:
+      offset_data = offset.output.copy_compatible_to(output_templ_wo_feat, check_dtype=False)
     from TFUtil import get_positional_encoding
     if source.have_time_axis():
-      length = tf.shape(source.placeholder)[source.time_dim_axis]
       if constant > -1:
-        position = constant * tf.ones([length], tf.int32)
-        signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (len,n_out)
+        position = constant * tf.ones([1] * output_templ_wo_feat.batch_ndim, tf.int32)
+        if offset_data:
+          position += offset_data.placeholder  # (batch, len)
+        # signal has shape (1, len) or (batch, len) or (1, 1) or more ones
+        signal = get_positional_encoding(num_channels=self.output.dim, position=position)
+        if not add_to_input and not offset_data:  # Need to tile the time dimension
+          tiles = [1] * self.output.batch_ndim
+          tiles[self.output.time_dim_axis] = tf.shape(source.placeholder)[source.time_dim_axis]
+          signal = tf.tile(signal, tiles)
       else:
-        signal = get_positional_encoding(num_channels=self.output.dim, length=length)  # (len,n_out)
+        length = tf.shape(source.placeholder)[source.time_dim_axis]
+        position = tf.range(length)  # (len,)
+        # Expand dims e.g. (1, len)
+        position = tf.reshape(
+          position,
+          [length if i == output_templ_wo_feat.time_dim_axis else 1 for i in range(output_templ_wo_feat.batch_ndim)])
+        if offset_data:
+          position += offset_data.placeholder
+        # signal has shape (1,len,n_out) or (batch,len,n_out)
+        signal = get_positional_encoding(num_channels=self.output.dim, position=position)
     else:
       if constant > -1:
         position = tf.convert_to_tensor([constant])
       else:
         position = tf.convert_to_tensor([self.network.get_rec_step_index()])
-      signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1,n_out)
-      signal = tf.squeeze(signal, axis=0)  # (n_out,)
-    if self.output.batch_dim_axis is not None:
-      signal = tf.expand_dims(signal, axis=self.output.batch_dim_axis)  # e.g. (len,batch,n_out)
+      if offset_data:
+        position += offset_data.placeholder  # (batch,)
+      signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1,n_out) or (batch,n_out)
+
     if add_to_input:
       signal += source.placeholder
     else:
@@ -5358,6 +5378,8 @@ class PositionalEncodingLayer(_ConcatInputLayer):
       else:
         d["from"] = ["data"]
     super(PositionalEncodingLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+    if d.get("offset", None):
+      d["offset"] = get_layer(d["offset"])
 
   @classmethod
   def get_out_data_from_opts(cls, name, network, add_to_input=False, sources=(), **kwargs):
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -1249,6 +1249,51 @@ class Data(object):
       kwargs["dtype"] = dtype
     return Data(**kwargs)
 
+  def copy_template_excluding_axis(self, exclude_axis, name=None):
+    """
+    :param int exclude_axis: axis to be removed.
+    :param str|None name: if set, this will be the new name.
+    :return: copy of myself excluding remove_axis_num axis, without placeholder.
+    :rtype: Data
+    """
+    kwargs = self.get_kwargs()
+    if exclude_axis < 0:
+      exclude_axis += self.batch_ndim
+      assert exclude_axis >= 0
+    assert 0 <= exclude_axis < self.batch_ndim
+    axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)  # None if exclude_axis == batch_dim_axis
+    if exclude_axis == self.feature_dim_axis:
+      del kwargs["dim"]
+
+    other_special_axes = self.get_special_axes_dict(
+      counted_with_batch_dim=True, only_available=True, include_batch_dim_axis=True)
+    for axis_name, axis in other_special_axes.items():
+      assert axis_name in kwargs
+      if axis == exclude_axis:
+        del kwargs[axis_name]
+      else:
+        kwargs[axis_name] = axis if (axis < exclude_axis) else (axis - 1)
+    if exclude_axis == self.batch_dim_axis:
+      kwargs["batch_dim_axis"] = None
+
+    new_shape = list(self.shape)
+    if axis_to_exclude_wo_b is not None:
+      del new_shape[axis_to_exclude_wo_b]
+    kwargs["shape"] = new_shape
+
+    if self.size_placeholder is not None:
+      size_placeholder = {}
+      for i, size in self.size_placeholder.items():
+        if i == axis_to_exclude_wo_b:
+          continue
+        if axis_to_exclude_wo_b is not None and i > axis_to_exclude_wo_b:
+          i -= 1
+        size_placeholder[i] = size
+      kwargs["size_placeholder"] = size_placeholder
+    if name:
+      kwargs["name"] = name
+    return Data(**kwargs)
+
   def copy_template_excluding_spatial_dim(self, spatial_axis_num, name=None):
     """
     :param int spatial_axis_num: index in self.get_spatial_batch_axes()
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -8625,7 +8625,7 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
   memory inputs to attention.
 
   The use of relative position is possible because sin(x+y) and cos(x+y) can be
-  experessed in terms of y, sin(x) and cos(x).
+  expressed in terms of y, sin(x) and cos(x).
 
   In particular, we use a geometric sequence of timescales starting with
   min_timescale and ending with max_timescale.  The number of different
@@ -8638,11 +8638,11 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
 
   :param int num_channels: scalar, size of timing embeddings to create. The number of
     different timescales is equal to channels / 2.
-  :param tf.Tensor|None length: scalar, length of timing signal sequence. if not given, is shape(position)[0]
-  :param tf.Tensor|None position: could be provided directly. int32, shape (length,)
-  :param float min_timescale: a float
-  :param float max_timescale: a float
-  :return: a Tensor of timing signals (length, channels)
+  :param tf.Tensor|None length: scalar, length of timing signal sequence.
+  :param tf.Tensor|None position: could be provided directly. int32. Can have any shape.
+  :param float min_timescale: a float.
+  :param float max_timescale: a float.
+  :return: a Tensor of timing signals of shape (length, channels) or (batch, length, channels).
   :rtype: tf.Tensor
   """
   import math
@@ -8651,16 +8651,17 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
     position = tf.range(length)
   else:
     assert length is None
-    position.set_shape((None,))
   position = tf.to_float(position)
   num_timescales = num_channels // 2
   log_timescale_increment = (
     math.log(float(max_timescale) / float(min_timescale)) / (float(num_timescales - 1)))
   inv_timescales = min_timescale * tf.exp(
     tf.to_float(tf.range(num_timescales)) * -log_timescale_increment)
-  scaled_time = tf.expand_dims(position, 1) * tf.expand_dims(inv_timescales, 0)
-  signal = tf.concat([tf.sin(scaled_time), tf.cos(scaled_time)], axis=1)
-  signal = tf.pad(signal, [[0, 0], [0, num_channels % 2]])  # (length, channels)
+  scale = tf.reshape(inv_timescales, [1] * len(position.shape) + [num_timescales])  # Usually (1, D//2) or (1, 1, D//2).
+  scaled_time = tf.expand_dims(position, -1) * scale
+  signal = tf.concat([tf.sin(scaled_time), tf.cos(scaled_time)], axis=-1)
+  # (length, channels) or (batch, length, channels).
+  signal = tf.pad(signal, [[0, 0]] * len(position.shape) + [[0, num_channels % 2]])
   return signal
 
 
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -4161,6 +4161,121 @@ def test_trafo_search_lm():
       assert all(out_seqs[i, :input_seq_lens[i]] == input_seqs[i, :input_seq_lens[i]])
 
 
+def test_PositionalEncodingLayer_offset_no_rec():
+  # Test `offset` option when `PositionalEncodingLayer` is out of loop.
+  rnd = numpy.random.RandomState(42)
+  n_batch, n_in, n_time, size = 3, 5, 7, 8
+  n_out = n_in
+
+  net_dict = {
+    "input": {"class": "linear", "activation": None, "n_out": size,},
+    "offset": {"class": "constant", "from": [], "value": 1},
+    "raw_pos_enc": {"class": "positional_encoding", "add_to_input": False, "from": ["input"], "n_out": size},
+    "const_pos_enc": {"class": "positional_encoding", "add_to_input": False, "constant": 42,
+                      "from": ["input"], "n_out": size},
+    "pos_enc": {"class": "positional_encoding", "add_to_input": True, "from": ["input"], "offset": "offset"},
+    "output": {"class": "softmax", "target": "data", "from": ["pos_enc", "raw_pos_enc", "const_pos_enc"], "loss": "ce"},
+  }
+
+  config = Config()
+  config.update({
+    "extern_data": {"data": {"dim": n_out, "sparse": True}},
+    "debug_print_layer_output_template": True})
+
+  with make_scope() as session:
+    network = TFNetwork(config=config, train_flag=True)
+    pprint(network.extern_data.data)
+    network.construct_from_dict(net_dict)
+
+    fetches = network.get_fetches_dict()
+    data_input = network.extern_data.data["data"]
+    assert data_input.batch_shape == (None, None)
+
+    train_out = network.get_layer("output").output
+    session.run(tf.variables_initializer(tf.global_variables() + [network.global_train_step]))
+    rand_data = rnd.randint(0, n_out, size=(n_batch, n_time,), dtype="int32")
+    outputs = [train_out.placeholder]
+    info, out = session.run(
+      (fetches, outputs),
+      feed_dict={
+        data_input.placeholder: rand_data,
+        data_input.size_placeholder[0]: numpy.array([n_time] * n_batch, dtype="int32"),
+      })
+    print(info)
+    print(out)  # random...
+
+
+def test_PositionalEncodingLayer_offset_in_rec():
+  # Test `offset` option when `PositionalEncodingLayer` is in loop.
+  rnd = numpy.random.RandomState(42)
+  n_batch, n_in, n_time, size = 3, 5, 7, 8
+  n_out = n_in
+
+  net_dict = {
+    "output": {
+      "class": "rec",
+      "from": "data",
+      'target': "data",
+      'unit': {
+        "input": {"class": "linear", "activation": None, "from": "data:source", "n_out": size},
+        "in_prefix": {  # Just a dummy switch which turns off once it sees an input with the ID 0.
+          "class": "eval", "from": ["data:source", "prev:in_prefix"], "initial_output": True,
+          "out_type": {'time_dim_axis': None, "dtype": "bool", "dim": 2},
+          "eval": "tf.logical_and(tf.not_equal(source(0, "
+              "auto_convert=False, as_data=True).copy_as_batch_major().placeholder, 0), source(1))"},
+        # Increment by one as long as `in_prefix` is on.
+        "counter": {
+          "class": "eval", "from": ["in_prefix", "prev:counter"], "initial_output": 0,
+          "out_type": {'time_dim_axis': None, "shape": (), "dtype": "int32", "dim": None, "sparse": False},
+          "eval": "where_bc(source(0, auto_convert=False, as_data=True).copy_as_batch_major().placeholder,"
+                  "source(1, auto_convert=False, as_data=True).copy_as_batch_major().placeholder + "
+                  "tf.ones(tf.shape(source(0, auto_convert=False, as_data=True).copy_as_batch_major().placeholder),"
+                           "dtype=tf.int32),"
+                  "source(1, auto_convert=False, as_data=True).copy_as_batch_major().placeholder)"},
+        # 0 when `in_prefix` is on. Minus `counter` otherwise.
+        "offset": {
+          "class": "eval", "from": ["in_prefix", "counter"],
+          "out_type": {'time_dim_axis': None, "shape": (), "dtype": "int32", "sparse": False, "dim": None},
+          "eval": "where_bc(source(0, auto_convert=False, as_data=True).copy_as_batch_major().placeholder,"
+                  "tf.zeros(tf.shape(source(1, auto_convert=False, as_data=True).copy_as_batch_major().placeholder),"
+                            "dtype=tf.int32),"
+                  "-source(1, auto_convert=False, as_data=True).copy_as_batch_major().placeholder)"},
+        "pos_enc": {"class": "positional_encoding", "add_to_input": True, "from": ["input"], "offset": "offset"},
+        "rnn_h": {"class": "linear", "activation": "tanh", "from": ["prev:rnn_h", "pos_enc"],
+                  "n_out": size, "initial_output": 0},
+        "output": {"class": "softmax", "from": ["rnn_h"], "loss": "ce", "target": "data"},
+      }
+    }
+  }
+
+  config = Config()
+  config.update({
+    "extern_data": {"data": {"dim": n_out, "sparse": True}},
+    "debug_print_layer_output_template": True})
+
+  with make_scope() as session:
+    network = TFNetwork(config=config, train_flag=True)
+    pprint(network.extern_data.data)
+    network.construct_from_dict(net_dict)
+
+    fetches = network.get_fetches_dict()
+    data_input = network.extern_data.data["data"]
+    assert data_input.batch_shape == (None, None)
+
+    train_out = network.get_layer("output").output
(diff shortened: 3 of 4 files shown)


tests/test_TFNetworkRecLayer.py
```diff
@@ -4161,6 +4161,118 @@ def test_trafo_search_lm():
       assert all(out_seqs[i, :input_seq_lens[i]] == input_seqs[i, :input_seq_lens[i]])
 
 
+def test_offset_no_rec_positional_encoding():
```
albertz: The function should be called `test_ PositionalEncodingLayer_offset_no_rec` or so.

tests/test_TFNetworkRecLayer.py
```diff
@@ -4161,6 +4161,118 @@ def test_trafo_search_lm():
...
+      (fetches, outputs),
+      feed_dict={
+        data_input.placeholder: rand_data,
+        data_input.size_placeholder[0]: numpy.array([n_time] * n_batch, dtype="int32"),
+      })
+    print(info)
+    print(out)  # random...
+
+
+def test_offset_rec_positional_encoding():
```
albertz: The function should be called `test_PositionalEncodingLayer_offset_in_rec` or so.

tests/test_TFUtil.py
```diff
@@ -3186,6 +3186,20 @@ def cond(self, i, x):
...
+  position1 = tf.range(5, 10)
+  position2 = tf.range(5)
+  position = tf.stack([position0, position1, position2])  # (3, 5).
+  signal = get_positional_encoding(num_channels=num_channels, position=position)  # (3, 5, 8).
+  signal1 = get_positional_encoding(num_channels=num_channels, position=position1)  # (5, 8).
+  signal_np, signal1_np = session.run([signal, signal1])
+  assert signal_np.shape == (3, 5, 8)
+  numpy.array_equal(signal1_np, signal_np[1])
+
+
```
albertz: In addition, it would be helpful if you put an explicit test for the values itself.
Basically, just add exactly this test:
```
def test_get_position_encoding():
  y = get_positional_encoding(num_channels=4, position=tf.range(3))
  values = session.run(y)
  numpy.testing.assert_almost_equal(
    values,
    [[0.0, 0.0, 1.0, 1.0],
     [0.8414709568023682, 9.99999901978299e-05, 0.5403022766113281, 1.0],
     [0.9092974066734314, 0.0001999999803956598, -0.416146844625473, 1.0]])
```
Note: I calculated these values with the old code. But that's exactly the point. It should be the same with the new code.

TFUtil.py
```diff
@@ -8651,16 +8651,26 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
     position = tf.range(length)
   else:
     assert length is None
-    position.set_shape((None,))
+    if len(position.shape) == 1:
+      position.set_shape((None,))
```
albertz: This does not make sense. This is redundant.

TFUtil.py
```diff
@@ -8651,16 +8651,26 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
   num_timescales = num_channels // 2
   log_timescale_increment = (
     math.log(float(max_timescale) / float(min_timescale)) / (float(num_timescales - 1)))
   inv_timescales = min_timescale * tf.exp(
     tf.to_float(tf.range(num_timescales)) * -log_timescale_increment)
-  scaled_time = tf.expand_dims(position, 1) * tf.expand_dims(inv_timescales, 0)
-  signal = tf.concat([tf.sin(scaled_time), tf.cos(scaled_time)], axis=1)
-  signal = tf.pad(signal, [[0, 0], [0, num_channels % 2]])  # (length, channels)
+  if len(position.shape) == 1:
+    scale = tf.expand_dims(inv_timescales, 0)
```
albertz: Can you add comments about the shapes? (Here: `# (1, D//2)`)

TFUtil.py
```diff
@@ -8651,16 +8651,26 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+  else:
+    assert len(position.shape) == 2
+    scale = tf.expand_dims(tf.expand_dims(inv_timescales, 0), 0)
```
albertz: Use `expand_multiple_dims` or `tf.reshape`. This is ugly otherwise.
Also, comment about the shape. (`# (1, 1, D//2)`)
Also, why don't you simplify the code, and cover any number of axes in `position`?
E.g. like this:
```
scale = tf.reshape(inv_timescales, [1] * len(position.shape) + [num_timescales])
```

TFUtil.py
```diff
@@ -8651,16 +8651,26 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
...
+  scaled_time = tf.expand_dims(position, -1) * scale
+  signal = tf.concat([tf.sin(scaled_time), tf.cos(scaled_time)], axis=-1)
+  if len(position.shape) == 1:
+    signal = tf.pad(signal, [[0, 0], [0, num_channels % 2]])  # (length, channels).
+  else:
+    assert len(position.shape) == 2
+    signal = tf.pad(signal, [[0, 0], [0, 0], [0, num_channels % 2]])  # (batch, length, channels).
```
albertz: Why don't you simplify this code? You don't need the `if len(position.shape)`. And why not generic for any number?
E.g. like:
```
signal = tf.pad(signal, [[0, 0]] * len(position.shape) + [[0, num_channels % 2]])
```

TFUtil.py
```diff
@@ -8638,11 +8638,11 @@ def get_positional_encoding(num_channels, length=None, position=None, min_timesc
 
   :param int num_channels: scalar, size of timing embeddings to create. The number of
     different timescales is equal to channels / 2.
-  :param tf.Tensor|None length: scalar, length of timing signal sequence. if not given, is shape(position)[0]
-  :param tf.Tensor|None position: could be provided directly. int32, shape (length,)
-  :param float min_timescale: a float
-  :param float max_timescale: a float
-  :return: a Tensor of timing signals (length, channels)
+  :param tf.Tensor|None length: scalar, length of timing signal sequence.
+  :param tf.Tensor|None position: could be provided directly. int32, shape (length,) or (batch, length).
```
albertz: It does not really matter at all what the axes are, does it? Then you can just write "can have any shape". It's a bit confusing that you write it must be "(batch,length)", while it does not need to be that way, and I think even you yourself also use it differently.

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
       else:
-        signal = get_positional_encoding(num_channels=self.output.dim, length=length)  # (len,n_out)
+        position = tf.range(length)  # (len,).
+        position = tf.expand_dims(position, axis=0)  # (1, len).
+        if offset:
+          position += offset.output.placeholder
+        signal = get_positional_encoding(num_channels=self.output.dim, position=position)
+      # signal has shape (1,len,n_out) or (batch,len,n_out).
+      if source.batch_dim_axis == 1:
+        assert source.time_dim_axis == 0,\
```
albertz: Don't use `\`. Rather, use `()`. See other code.

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
+        assert offset.output.time_dim_axis == 1, "Expect time_dim_axis to be 1 for offset."
     if add_to_input:
       assert source.dim == self.output.dim
     from TFUtil import get_positional_encoding
     if source.have_time_axis():
       length = tf.shape(source.placeholder)[source.time_dim_axis]
       if constant > -1:
-        position = constant * tf.ones([length], tf.int32)
-        signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (len,n_out)
+        position = constant * tf.ones([1, length], tf.int32)  # (1, len).
```
albertz: Do we still need the `length`? Why not just `tf.ones([1] * source.batch_ndim)`?

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
+        if offset:
+          position += offset.output.placeholder  # (batch, len).
+        signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1, len) or (batch, len).
       else:
-        signal = get_positional_encoding(num_channels=self.output.dim, length=length)  # (len,n_out)
+        position = tf.range(length)  # (len,).
+        position = tf.expand_dims(position, axis=0)  # (1, len).
```
albertz: Better:
```
position = tf.reshape(position, [length if i == source.time_dim_axis else 1 for i in range(source.batch_ndim)])
```

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
     assert len(self.sources) == 1, "%s: expect a single source" % self
     source = self.input_data
+    if self.output.batch_dim_axis:
+      assert self.output.batch_dim_axis == source.batch_dim_axis
+    if self.output.time_dim_axis:
+      assert self.output.time_dim_axis == source.time_dim_axis
+    if offset:  # Assumptions on the shape of `offset`.
+      assert offset.output.batch_dim_axis == 0, "Expect batch_dim_axis to be 0 for offset."
+      if offset.output.time_dim_axis is not None:
+        assert offset.output.time_dim_axis == 1, "Expect time_dim_axis to be 1 for offset."
```
albertz: I think you can just remove these assertions, and put this:
```
offset = offset.copy_compatible_to(source)
```

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
+        position = tf.range(length)  # (len,).
+        position = tf.expand_dims(position, axis=0)  # (1, len).
+        if offset:
+          position += offset.output.placeholder
+        signal = get_positional_encoding(num_channels=self.output.dim, position=position)
+      # signal has shape (1,len,n_out) or (batch,len,n_out).
+      if source.batch_dim_axis == 1:
+        assert source.time_dim_axis == 0,\
+          "Unexpected input form: batch_dim_axis 1 and time_dim_axis %s." % source.time_dim_axis
+        signal = tf.transpose(signal, [1, 0, 2])
```
albertz: Better avoid the `tf.transpose`, and construct it with the right shape/axes in the first place.

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
-      signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1,n_out)
-      signal = tf.squeeze(signal, axis=0)  # (n_out,)
-    if self.output.batch_dim_axis is not None:
-      signal = tf.expand_dims(signal, axis=self.output.batch_dim_axis)  # e.g. (len,batch,n_out)
+      if offset:
+        position += offset.output.placeholder  # (batch,).
+      signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1,n_out) or (batch,n_out).
+      if source.batch_dim_axis != 0:
+        assert source.batch_dim_axis == 1 and source.feature_dim_axis == 0, "Unexpected input form."
+        signal = tf.transpose(signal, [1, 0])
```
albertz: Better avoid the `tf.transpose`, and construct it with the right shape/axes in the first place.

TFNetworkRecLayer.py
albertz: Sorry. `source` here has to be without the feature axis. So make a copy of `source`, remove that axis (e.g. `source.copy_remove_axis(source.feature_dim_axis)`).
Also, you still need this check:
```
assert source.feature_dim_axis == source.batch_ndim - 1  # must be last
```

albertz: Assign that, as we reuse it later.
```
source_wo_feat = source.copy_remove_axis(source.feature_dim_axis)
```
(Note: `copy_remove_axis` does not exist yet. Has to be added.)

TFNetworkRecLayer.py
albertz: `source_wo_feat` instead of `source`.

TFNetworkRecLayer.py
albertz: `source_wo_feat` instead of `source`.

TFNetworkRecLayer.py
```diff
@@ -5293,56 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
+
     if add_to_input:
       signal += source.placeholder
-    else:
-      # tile to batch dimension explicitly, as batch_dim=1 will not be automatically unbroadcasted
-      tiles = [1] * self.output.batch_ndim
-      tiles[self.output.batch_dim_axis] = tf.shape(source.placeholder)[source.batch_dim_axis]
-      signal = tf.tile(signal, tiles)
```
albertz: Why did you remove this? I think this is still needed (for `not add_to_input`).

TFNetworkRecLayer.py
```diff
@@ -5369,10 +5388,7 @@ def get_out_data_from_opts(cls, name, network, add_to_input=False, sources=(), *
     :rtype: Data
     """
     assert len(sources) > 0, "%s %r: must have one source" % (cls, name)
-    if add_to_input:
-      return get_concat_sources_data_template(sources, name="%s_output" % name)  # just the same as the input
-    return super(PositionalEncodingLayer, cls).get_out_data_from_opts(
-      name=name, network=network, sources=sources, **kwargs)
+    return get_concat_sources_data_template(sources, name="%s_output" % name)  # just the same as the input.
```
albertz: I think this is wrong for `not add_to_input`, or not?

tests/test_TFNetworkRecLayer.py
kazuki-irie: Done.

tests/test_TFNetworkRecLayer.py
kazuki-irie: Done.

tests/test_TFUtil.py
kazuki-irie: Done.

TFUtil.py
kazuki-irie: Removed.

TFUtil.py
kazuki-irie: Done.

TFUtil.py
kazuki-irie: Done.

TFUtil.py
kazuki-irie: Done.

TFUtil.py
kazuki-irie: Done.

TFNetworkRecLayer.py
kazuki-irie: Done.

TFNetworkRecLayer.py
kazuki-irie: Done.

TFNetworkRecLayer.py
kazuki-irie: I'm adding `copy_template_excluding_axis` in the next commit.

TFNetworkRecLayer.py
kazuki-irie: Done.

TFNetworkRecLayer.py
kazuki-irie: OK. This can be removed if we check:
`assert source.feature_dim_axis == source.batch_ndim - 1`

TFNetworkRecLayer.py
kazuki-irie: Putting this back.
In one of the previous versions, I was handling this upper in the code so I had removed this at some point.

TFNetworkRecLayer.py
kazuki-irie: When we talked off-line, I thought we said that we could always make the output format the same as the input in all cases (the input was by the way only used to get the time dimension when `add_to_input` was `False`).

Otherwise, some of your suggestions would be also wrong like this one:
`position = tf.reshape(position, [length if i == source_wo_feat.time_dim_axis else 1 for i in range(source_wo_feat.batch_ndim)])`.
This assumes the output shape to be the same as the input, correct?

TFNetworkRecLayer.py
kazuki-irie: But for the moment I would put:
`tiles[self.source.batch_dim_axis]` instead of `tiles[self.output.batch_dim_axis]` .
We have to first decide whether we want to leave the output shape flexible or make it the same as the input, right?

TFNetworkRecLayer.py
albertz: For all the code using `source`, you would just use `self.output` instead, which always has the right format (as defined here).
I think I meant it just for `add_to_input`. For `not add_to_input`, I don't know what makes most sense. Maybe you are right though. Not sure if anyone anyway uses this case.

TFNetworkRecLayer.py
albertz: Make sure that the commits have the right order (so that all tests would run through after every individual commit).

TFNetworkRecLayer.py
kazuki-irie: > I think I meant it just for `add_to_input`.

I see. Then let's keep this flexible as before for the `add_to_input` `False` case.

TFNetworkRecLayer.py
kazuki-irie: Done.

kazuki-irie:
Thanks again for all the comments!
I have pushed the corresponding updates.
Please take another look.

albertz:
I see some "wip" in the commits (at least in the commit message). Can you cleanup the commits? (Also make sure the order is correct, e.g. you add `copy_template_excluding_axis` in a commit *before* it is being used in a later commit.)

TFNetworkRecLayer.py
```diff
@@ -5293,53 +5293,73 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
     """
     super(PositionalEncodingLayer, self).__init__(**kwargs)
     assert len(self.sources) == 1, "%s: expect a single source" % self
     source = self.input_data
+    assert source.feature_dim_axis == source.batch_ndim - 1  # Must be last.
     if add_to_input:
       assert source.dim == self.output.dim
+      template_wo_feat = source.copy_template_excluding_axis(source.feature_dim_axis)
+    else:
+      template_wo_feat = self.output.copy_template_excluding_axis(self.output.feature_dim_axis)
```
albertz: Why do you call it `template_wo_feat`? Better call it `output_templ_wo_feat` or so.
Also, you don't need the `else:`. The other code is actually wrong. In both cases, it should be `self.output.copy_template_excluding_axis(...)`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+    kwargs = self.get_kwargs()
+    if exclude_axis < 0:
+      exclude_axis += self.batch_ndim
+      assert exclude_axis >= 0
+    assert 0 <= exclude_axis < self.batch_ndim
+    exclude_batch_dim_axis = exclude_axis == self.batch_dim_axis
+    if not exclude_batch_dim_axis:
+      axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)
+    if exclude_axis == self.feature_dim_axis:
+      kwargs["dim"] = None
```
albertz: Better use `NotSpecified`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+    :rtype: Data
+    """
+    kwargs = self.get_kwargs()
+    if exclude_axis < 0:
+      exclude_axis += self.batch_ndim
+      assert exclude_axis >= 0
+    assert 0 <= exclude_axis < self.batch_ndim
+    exclude_batch_dim_axis = exclude_axis == self.batch_dim_axis
+    if not exclude_batch_dim_axis:
+      axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)
```
albertz: Never assign a variable only conditionally (in `if` but not in `else`).
Anyway, you don't need the `if` at all. Just remove the `if`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+    :param int exclude_axis: axis to be removed.
+    :return: copy of myself excluding remove_axis_num axis, without placeholder.
+    :rtype: Data
+    """
+    kwargs = self.get_kwargs()
+    if exclude_axis < 0:
+      exclude_axis += self.batch_ndim
+      assert exclude_axis >= 0
+    assert 0 <= exclude_axis < self.batch_ndim
+    exclude_batch_dim_axis = exclude_axis == self.batch_dim_axis
```
albertz: You don't need this extra variable. In other code, just either write `exclude_axis == self.batch_dim_axis` directly (easier to understand), or maybe `axis_to_exclude_wo_b is None`.

TFUtil.py
albertz: Or better: Just `del kwargs["dim"]`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+    if not exclude_batch_dim_axis:
+      axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)
+    if exclude_axis == self.feature_dim_axis:
+      kwargs["dim"] = None
+
+    other_special_axes = self.get_special_axes_dict(
+      counted_with_batch_dim=True, only_available=True, include_batch_dim_axis=True)
+    for axis_name, axis in other_special_axes.items():
+      if axis == exclude_axis:
+        kwargs[axis_name] = None
```
albertz: Better: `del kwargs[axis_name]`.
(E.g. when you remove the feature dim axis, or time dim axis, maybe some other spatial axis would automatically become the new time dim axis. `Data.__init__` should handle this.)

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+      else:
+        kwargs[axis_name] = axis if (axis < exclude_axis) else (axis - 1)
+
+    new_shape = list(self.shape)
+    if not exclude_batch_dim_axis:
```
albertz: Here better `if axis_to_exclude_wo_b is not None:`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+      del new_shape[axis_to_exclude_wo_b]
+
+    size_placeholder = None
+    if self.size_placeholder is not None:
+      if exclude_batch_dim_axis:
```
albertz: Remove that `if`. You don't need to have conditional code here. Keep the code short if possible (easier to read).

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+        size_placeholder = self.size_placeholder
+      else:
+        size_placeholder = {}
+        for i, size in self.size_placeholder.items():
+          if i == axis_to_exclude_wo_b:
+            continue
+          if i > axis_to_exclude_wo_b:
```
albertz: `if axis_to_exclude_wo_b is not None and i > axis_to_exclude_wo_b`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+      if axis == exclude_axis:
+        kwargs[axis_name] = None
+      else:
+        kwargs[axis_name] = axis if (axis < exclude_axis) else (axis - 1)
+
+    new_shape = list(self.shape)
+    if not exclude_batch_dim_axis:
+      del new_shape[axis_to_exclude_wo_b]
+
+    size_placeholder = None
```
albertz: Remove that. You don't need it.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+    exclude_batch_dim_axis = exclude_axis == self.batch_dim_axis
+    if not exclude_batch_dim_axis:
+      axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)
+    if exclude_axis == self.feature_dim_axis:
+      kwargs["dim"] = None
+
+    other_special_axes = self.get_special_axes_dict(
+      counted_with_batch_dim=True, only_available=True, include_batch_dim_axis=True)
+    for axis_name, axis in other_special_axes.items():
+      if axis == exclude_axis:
```
albertz: Add: `and axis_name in kwargs`.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
       kwargs["dtype"] = dtype
     return Data(**kwargs)
 
+  def copy_template_excluding_axis(self, exclude_axis, name=None):
+    """
+    :param str|None name: if set, this will be the new name.
+    :param int exclude_axis: axis to be removed.
```
albertz: Reorder the doc params to be the same order as the params.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+      counted_with_batch_dim=True, only_available=True, include_batch_dim_axis=True)
+    for axis_name, axis in other_special_axes.items():
+      if axis == exclude_axis:
+        kwargs[axis_name] = None
+      else:
+        kwargs[axis_name] = axis if (axis < exclude_axis) else (axis - 1)
+
+    new_shape = list(self.shape)
+    if not exclude_batch_dim_axis:
+      del new_shape[axis_to_exclude_wo_b]
```
albertz: Add the `kwargs["shape"] = new_shape` directly here.

TFUtil.py
albertz: You can remove that.

TFUtil.py
```diff
@@ -1249,6 +1249,54 @@ def copy_template(self, name=None, dtype=None):
...
+      else:
+        size_placeholder = {}
+        for i, size in self.size_placeholder.items():
+          if i == axis_to_exclude_wo_b:
+            continue
+          if i > axis_to_exclude_wo_b:
+            i -= 1
+          size_placeholder[i] = size
+    kwargs["shape"] = new_shape
+    kwargs["size_placeholder"] = size_placeholder
```
albertz: Do the `kwargs["size_placeholder"] = size_placeholder` in the `if self.size_placeholder is not None:`.

TFNetworkRecLayer.py
albertz: Resolved?

TFNetworkRecLayer.py
kazuki-irie: > Better call it output_wo_feat (or maybe output_templ_wo_feat) or so.

OK.

> Also, you don't need the else:..

I see.

> The other code is actually wrong.

I'm not sure what is wrong here. Isn't it same to use `source` or `output` if `add_to_input`?
Plus, the code with `source` was your suggestion :)

albertz: Maybe "wrong" is the wrong word. What you want here is that `output_wo_feat` is always exactly as `self.output` (wo. feature axis).
Because of how `get_out_data_from_opts` is defined currently, this luckily is correct here, but would become incorrect if you change `get_out_data_from_opts`. In any case, it's confusing, as a reader of this code might not know how `get_out_data_from_opts` is defined, and also, should not need to know (the less a reader needs to know about other independent places in the code, the easier it is to understand some code).

TFUtil.py
kazuki-irie: with `del kwargs[axis_name]`, the test fails with:
`axis time_dim_axis (0) must be different from batch_dim_axis (0)`
when the batch_dim_axis is removed.
Am I doing something wrong in my test?
(also maybe I should split the current test into separate tests testing the removal of each special axis?).

TFNetworkRecLayer.py
kazuki-irie: Now I see. Thank you for the explanation.

kazuki-irie: Now I see. Thank you for the explanation.

TFUtil.py
albertz: What is the full error exactly? And the relevant stack?

kazuki-irie:
```
Executing: test_Data_copy_template_excluding_axis
EXCEPTION
  File "tests/test_TFUtil.py", line 3242, in <module>
    line: globals()[arg]()  # assume function and execute
    locals:
      globals = <builtin> <built-in function globals>
      arg = <local> 'test_Data_copy_template_excluding_axis', len = 38
  File "tests/test_TFUtil.py", line 310, in test_Data_copy_template_excluding_axis
    line: data_wo_batch = data.copy_template_excluding_axis(data.batch_dim_axis)
    locals:
      data_wo_batch = <not found>
      data = <local> Data(name='data', shape=(None, 8), feature_dim_axis=2, batch_shape_meta=[B,T|?,F|8])
      data.copy_template_excluding_axis = <local> <bound method Data.copy_template_excluding_axis of Data(name='data', shape=(None, 8), feature_dim_axis=2, batch_shape_meta=[B,T|?,F|8])>
      data.batch_dim_axis = <local> 0
  File "/u/irie/src/returnn_ki/TFUtil.py", line 1293, in copy_template_excluding_axis
    line: return Data(**kwargs)
    locals:
      Data = <global> <class 'TFUtil.Data'>
      kwargs = <local> {'name': 'data', 'shape': [None, 8], 'dtype': 'float32', 'sparse': False, 'dim': 8, 'time_dim_axis': 0, 'feature_dim_axis': 1}, len = 7
  File "/u/irie/src/returnn_ki/TFUtil.py", line 477, in __init__
    line: self.sanity_check()
    locals:
      self = <local> Data(name='data', shape=(None, 8), feature_dim_axis=1, batch_shape_meta=[B|T,F|?,8])
      self.sanity_check = <local> <bound method Data.sanity_check of Data(name='data', shape=(None, 8), feature_dim_axis=1, batch_shape_meta=[B|T,F|?,8])>
  File "/u/irie/src/returnn_ki/TFUtil.py", line 499, in sanity_check
    line: assert axis != self.batch_dim_axis, "%s: axis %s (%i) must be different from batch_dim_axis (%i)" % (
            self, axis_name, axis, self.batch_dim_axis)
    locals:
      axis = <local> 0
      self = <local> Data(name='data', shape=(None, 8), feature_dim_axis=1, batch_shape_meta=[B|T,F|?,8])
      self.batch_dim_axis = <local> 0
      axis_name = <local> 'time_dim_axis', len = 13
AssertionError: Data(name='data', shape=(None, 8), feature_dim_axis=1, batch_shape_meta=[B|T,F|?,8]): axis time_dim_axis (0) must be different from batch_dim_axis (0)
```

albertz: Ah I see. The default handling for `batch_dim_axis` is different. So it means you need to treat it different.
You need this addition code (after this loop):
```
if exclude_axis == self.batch_dim_axis:
  kwargs["batch_dim_axis"] = None
```

kazuki-irie:
Thank you for further comments @albertz !
I cleaned up the commits.

TFNetworkRecLayer.py
```diff
@@ -5293,53 +5293,72 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
+          tiles = [1] * self.output.batch_ndim
+          tiles[self.output.time_dim_axis] = tf.shape(source.placeholder)[source.time_dim_axis]
+          signal = tf.tile(signal, tiles)
       else:
-        signal = get_positional_encoding(num_channels=self.output.dim, length=length)  # (len,n_out)
+        length = tf.shape(source.placeholder)[source.time_dim_axis]
+        position = tf.range(length)  # (len,).
+        # Expand dims e.g. (1, len).
+        position = tf.reshape(position, [length if i == output_templ_wo_feat.time_dim_axis
+                                         else 1 for i in range(output_templ_wo_feat.batch_ndim)])
```
albertz: Don't indent code like this. There are multiple issues:
* The indentation is far too much to the right. Try to make it max 2 spaces if possible (and that is possible in almost all cases). You can get that by making a newline just before `position`, and then again before `[length ...]`.
* Don't make a newline just before `else`. This is strange. Rather, you can make a newline before `for` or `if` or both.

TFNetworkRecLayer.py
```diff
@@ -5293,53 +5293,72 @@ def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
...
-      signal = tf.expand_dims(signal, axis=self.output.batch_dim_axis)  # e.g. (len,batch,n_out)
+      if offset_data:
+        position += offset_data.placeholder  # (batch,).
+      signal = get_positional_encoding(num_channels=self.output.dim, position=position)  # (1,n_out) or (batch,n_out).
+
     if add_to_input:
       signal += source.placeholder
     else:
-      # tile to batch dimension explicitly, as batch_dim=1 will not be automatically unbroadcasted
+      # tile to batch dimension explicitly, as batch_dim=1 will not be automatically unbroadcasted.
```
albertz: Why this change?

TFUtil.py
```diff
@@ -1249,6 +1249,50 @@ def copy_template(self, name=None, dtype=None):
...
+      assert exclude_axis >= 0
+    assert 0 <= exclude_axis < self.batch_ndim
+    axis_to_exclude_wo_b = self.get_batch_axis_excluding_batch(exclude_axis)  # None if exclude_axis == batch_dim_axis.
+    if exclude_axis == self.feature_dim_axis:
+      del kwargs["dim"]
+
+    other_special_axes = self.get_special_axes_dict(
+      counted_with_batch_dim=True, only_available=True, include_batch_dim_axis=True)
+    for axis_name, axis in other_special_axes.items():
+      if axis == exclude_axis and axis_name in kwargs:
```
albertz: Ah sorry, `axis_name in kwargs` does not make sense here (if you have that `else:` part there), because in the `else:` part, you assume that it is always in `kwargs` (otherwise you should not touch it).
As you have `only_available=True` above, I guess it is anyway always already in. You can just check by:
`assert axis_name in kwargs` (before the `if`)

kazuki-irie:
Pushed the requested changes. Thanks again!

kazuki-irie:
Great! Thank you!

PR #217, opened 2019-09-07
Fix softmax window masking

Spotlight0xff:
Hi,

this PR adds a test for the `SoftmaxOverSpatial` layer in the case of window masking.
Previously this functionality was broken due to some previous changes in the "default" case.
Some of the changes were already in some branch, but got lost I think.

Also, see
```python
indices.set_shape((None, None))  # required for older TF versions
```
which is a workaround for an error I got for TF 1.4.1, where the tensor shape could not get inferred automatically (and thus `nd_indices` failed). Hope this is ok.

The test is also good to verify various edge cases for the masking.

Cheers

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -2855,7 +2855,8 @@ class SoftmaxOverSpatialLayer(_ConcatInputLayer):
         energy_mask = tf.logical_and(energy_mask, tf.greater_equal(idxs, start_data.placeholder))
       if window_start:
         assert window_size, "set window_size explicitly"
-        from TFUtil import nd_indices, expand_dims_unbroadcast
+        assert window_start.output.batch_ndim == 1
+        from TFUtil import nd_indices, expand_dims_unbroadcast, move_axis
         # handle edge cases correctly:
         # 1. if the energy time-dim is less than `window_size`, we adjust the window size.
         # 2. for each seq, we adjust the window so that no elements after the seq-len are indexed.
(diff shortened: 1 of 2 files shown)


TFNetworkLayer.py
```diff
@@ -2867,14 +2868,24 @@ def __init__(self, axis=None, energy_factor=None,
...
-        mask_shape[axis] = window_len
-        energy_mask_window = tf.scatter_nd(idxs, tf.ones(shape=mask_shape), energy_shape[:2])
+        indices = expand_dims_unbroadcast(tf.range(window_len), 0, n_batch)  # (B, W)
+        # (B, W) + (B, 1)
+        indices += tf.expand_dims(tf.to_int32(window_start), axis=1)  # + (B, 1)
+        indices.set_shape((None, None))  # required for older TF versions
+        idxs = nd_indices(indices)  # (B, W, 2)
+        mask_shape = [n_batch, window_len]  # (B, W)
+        rem_energy_data = energy_data.copy_with_time_dim_axis(time_dim_axis=1)  # (B, T, ...)
+        rem_energy_shape = tf.shape(rem_energy_data.placeholder, name="rem_energy_shape")
```
albertz: Note that I prefer to use `TFUtil.get_shape` instead of `tf.shape` in many cases now.

TFNetworkLayer.py
```diff
@@ -2867,14 +2868,24 @@ def __init__(self, axis=None, energy_factor=None,
...
+        rem_energy_shape = [rem_energy_shape[i] for i in range(rem_energy_data.batch_ndim)]
+        rem_axes = list(energy_shape)
+        rem_axes.pop(energy_data.batch_dim_axis)
+        rem_axes.pop()  # (B, ..., T) -> (...)
```
albertz: `pop` removes by default the first or last? Maybe better make that explicit.
Is the comment correct? Above you do `rem_energy_data. copy_with_time_dim_axis(1)`, so time is at axis 1, not at the end?
Also, does this code assume that batch dim axis is 0 or 1? What if it is bigger?

TFNetworkLayer.py
Spotlight0xff: Cool, I changed that.

TFNetworkLayer.py
Spotlight0xff: `pop` removes the last item by default. I made it more explicit.
`rem_axes.pop()` operates on a copy of `energy_shape`, and not `rem_energy_data`.
I've moved the code a bit further down, so that it is not confusing at first glance.

I haven't tested it with different batch-dim axes, but it does convert it into batch-major.

Spotlight0xff:
Thanks, I addressed your comments, you can merge if it still looks good.

PR #221, opened 2019-09-12
Fix debug_grad_summaries

Spotlight0xff:
Hi,

this PR aims to fix the `debug_grad_summaries` config-option.
Currently it does not add any summaries to the event file.
This is due to the order of the operations `engine.network.get_fetches_dict` and `engine.updater.get_optim_op`.

To test this, I had to add an option to provide a `callable` (so a lambda or so) to the `extra_fetches` parameter of the `Runner` class, so that the operation which gets the summary-tensor is called once it is actually created.

This may lead to issues later, maybe we have to implement this more generic (e.g. it breaks if `Data` becomes callable). But maybe this is enough.

Also the actual fix is a bit ugly, maybe you have a better suggestion?

Cheers,
André

### DIFF
--- a/TFEngine.py
+++ b/TFEngine.py
@@ -109,13 +109,8 @@ class Runner(object):
     :return: values and actions which should be calculated and executed in self.run() by the TF session for each step
     :rtype: dict[str,tf.Tensor|tf.Operation]
     """
-    # Note that it is important that we do not recreate graph nodes for every call to this function.
-    # Thus everything which we access here should be cached.
-    d = self.engine.network.get_fetches_dict(
-      config=self.engine.config,
-      should_train=self._should_train, should_eval=self._should_eval,
-      with_summary=True, with_size=True)
-
+    d = {}
+    optim_op = None
     if self._should_train:
       assert self.engine.updater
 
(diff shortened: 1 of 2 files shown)


TFEngine.py
```diff
@@ -59,7 +59,8 @@ def __init__(self, engine, dataset, batches, train, eval=True, train_flag=None,
     :param bool train: whether to do updates on the model
     :param bool|None train_flag: normally just as train. but e.g. maybe you want to have the train_flag but not train
     :param bool eval: whether to evaluate (i.e. calculate loss/error)
-    :param dict[str,tf.Tensor|TFUtil.Data|TFNetworkLayer.LayerBase]|None extra_fetches: additional fetches per step.
+    :param dict[str,tf.Tensor|TFUtil.Data|TFNetworkLayer.LayerBase|lambda]|None extra_fetches:
```
albertz: Note: `lambda` is not a valid type. Use `()->tf.Tensor`.

TFEngine.py
```diff
@@ -127,7 +122,17 @@ def callback_on_new():
         self.engine._checked_uninitialized_vars = False
         self.engine.updater.init_optimizer_vars(session=self.engine.tf_session)
 
-      d["optim_op"] = self.engine.updater.get_optim_op(callback_on_new=callback_on_new)
+      optim_op = self.engine.updater.get_optim_op(callback_on_new=callback_on_new)
```
albertz: Can you add a comment and explain why this should be called before `get_fetches_dict`?

tests/test_TFEngine.py
```diff
@@ -2094,6 +2094,62 @@ def make_network(num_layers):
...
+  forwarder = Runner(
+    engine=engine, dataset=train_data, batches=batches,
+    train=True, eval=False,
+    extra_fetches={
+      "summary_proto": lambda: engine.network._get_all_merged_summaries(),
+    },
+    extra_fetches_callback=extra_fetches_cb)
+  forwarder.run(report_prefix="test_grad_summaries")
+  if not forwarder.finalized:
+    raise Exception("Error happened. Exit now.")
```
albertz: I'm pretty sure the test worked already before your change in `_get_fetches_dict`. Why do you think it did not?

Spotlight0xff: Because now the sequence of methods is:
1. `engine.updater.get_optim_op()` calls `engine.updater.create_optim_op()` -> which adds the debug grad summaries to the graph
2. `engine.network.get_fetches_dict()` calls `net._get_all_merged_summaries()`, collects all summaries from the graph (`tf.summary.merge_all()`).

Before the changes, 1. and 2. were reversed. So the summaries were collected before they were created.

Also, I ran the test without the fix and it did fail.

TFEngine.py
Spotlight0xff: Fixed.

TFEngine.py
Spotlight0xff: Done.

tests/test_TFEngine.py
albertz: Ah sorry, yes you are right.

PR #222, opened 2019-09-13
Dataset: Allow explicit left/right context window

vieting:
I added a way to explicitly specify left and right context for the `context_window` in `Dataset.py`. To do that, `context_window` should be a dict with the keys "left" and "right".

There some code in `TFDataPipeline.py` in `TFChunkingQueueRunner` which also uses a context window but not exactly in the same way as done in `Dataset.py`. Not sure if that should be updated, too...

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -92,7 +92,7 @@ class Dataset(object):
     :param str name: e.g. "train" or "eval"
     :param int window: features will be of dimension window * feature_dim, as we add a context-window around.
       not all datasets support this option.
-    :param None|int|dict|NumbersDict context_window: will add this context for each chunk
+    :param None|int|dict|NumbersDict|(dict,dict) context_window: will add this context for each chunk
     :param None|str|int|(int,int)|dict|(dict,dict) chunking: "chunk_size:chunk_step"
     :param str seq_ordering: "batching"-option in config. e.g. "default", "sorted" or "random".
       See self.get_seq_order_for_epoch() for more details.
@@ -152,14 +152,31 @@ class Dataset(object):
       assert sorted(chunk_step.keys()) == sorted(chunk_size.keys())
       assert chunk_step.max_value() > 0, "chunking step must be positive (for some key)"
     self.chunk_step = chunk_step
-    if context_window is None:
-      context_window = NumbersDict(0)
-    elif isinstance(context_window, int):
-      context_window = NumbersDict(broadcast_value=0, numbers_dict={"data": context_window})
-    elif isinstance(context_window, dict):
-      context_window = NumbersDict(broadcast_value=0, numbers_dict=context_window)
-    assert isinstance(context_window, NumbersDict)
-    self.context_window = context_window
+    if isinstance(context_window, (tuple, list)):
+      assert len(context_window) == 2
+      for elem in context_window:
+        assert isinstance(elem, dict)
+      # assume that first element of tuple|list is for left context and second element for right context
+      self.ctx_left = NumbersDict(numbers_dict=context_window[0])
+      self.ctx_right = NumbersDict(numbers_dict=context_window[1])
+    else:
+      if context_window is None:
+        context_window = NumbersDict()
+      elif isinstance(context_window, int):
+        context_window = NumbersDict(numbers_dict={"data": context_window})
+      elif isinstance(context_window, dict):
+        context_window = NumbersDict(numbers_dict=context_window)
+      assert isinstance(context_window, NumbersDict)
+      # ctx_total is how much frames we add additionally.
+      # One less because the original frame also counts, and context_window=1 means that we just have that single frame.
+      ctx_total = NumbersDict.max([context_window, 1]) - 1
+      # In case ctx_total is odd / context_window is even, we have to decide where to put one more frame.
+      # To keep it consistent with e.g. 1D convolution with a kernel of even size, we add one more to the right.
+      # See test_tfconv1d_evensize().
+      self.ctx_left = ctx_total // 2
+      self.ctx_right = ctx_total - self.ctx_left
+    assert isinstance(self.ctx_left, NumbersDict)
+    assert isinstance(self.ctx_right, NumbersDict)
     self.shuffle_frames_of_nseqs = shuffle_frames_of_nseqs
     self.epoch = None
 
@@ -909,24 +926,6 @@ class Dataset(object):
             break
       s += 1
 
-  def _get_context_window_left_right(self):
-    """
-    :return: (ctx_left, ctx_right)
-    :rtype: None|(NumbersDict,NumbersDict)
-    """
-    if self.context_window:
-      # One less because the original frame also counts, and context_window=1 means that we just have that single frame.
-      # ctx_total is how much frames we add additionally.
-      ctx_total = NumbersDict.max([self.context_window, 1]) - 1
-      # In case ctx_total is odd / context_window is even, we have to decide where to put one more frame.
-      # To keep it consistent with e.g. 1D convolution with a kernel of even size, we add one more to the right.
-      # See test_tfconv1d_evensize().
-      ctx_left = ctx_total // 2
-      ctx_right = ctx_total - ctx_left
-      return ctx_left, ctx_right
-    else:
-      return None
-
   def get_start_end_frames_full_seq(self, seq_idx):
     """
     :param int seq_idx:
@@ -935,10 +934,8 @@ class Dataset(object):
     """
     end = self.get_seq_length(seq_idx)
     start = NumbersDict.constant_like(0, numbers_dict=end)
-    ctx_lr = self._get_context_window_left_right()
-    if ctx_lr:
-      start -= ctx_lr[0]
-      end += ctx_lr[1]
+    start -= self.ctx_left
+    end += self.ctx_right
     return start, end
 
   def sample(self, seq_idx):
@@ -996,7 +993,6 @@ class Dataset(object):
         print("Non-recurrent network, chunk size %s:%s ignored" % (chunk_size, chunk_step), file=log.v4)
         chunk_size = 0
     batch = Batch()
-    ctx_lr = self._get_context_window_left_right()
     total_num_seqs = 0
     last_seq_idx = -1
     avg_weight = sum([v[0] for v in self.weights.values()]) / (len(self.weights.keys()) or 1)
@@ -1009,9 +1005,8 @@ class Dataset(object):
         continue
       if total_num_seqs > max_total_num_seqs:
         break
-      if ctx_lr:
-        t_start -= ctx_lr[0]
-        t_end += ctx_lr[1]
+      t_start -= self.ctx_left
+      t_end += self.ctx_right
       if recurrent_net:
         length = t_end - t_start
         if length.any_compare(max_seq_length, (lambda a, b: a > b)):


Dataset.py
```diff
@@ -157,7 +157,12 @@ def __init__(self, name=None,
     elif isinstance(context_window, int):
       context_window = NumbersDict(broadcast_value=0, numbers_dict={"data": context_window})
     elif isinstance(context_window, dict):
-      context_window = NumbersDict(broadcast_value=0, numbers_dict=context_window)
+      if set(context_window.keys()) == {"left", "right"}:
```
albertz: This is not a good idea. This is incompatible with the behavior it had before. Before, the keys were the data keys. Now you have a special behavior for this specific case. What e.g. if your data keys are "left" and "right"?

I would suggest to make your case if `isinstance(context_window, (tuple, list))`.

Also, it is even uglier that you have the same strange check below.
To avoid this, directly set `ctx_left` and `ctx_right` here.

Dataset.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, name=None,
     :param str name: e.g. "train" or "eval"
     :param int window: features will be of dimension window * feature_dim, as we add a context-window around.
       not all datasets support this option.
-    :param None|int|dict|NumbersDict context_window: will add this context for each chunk
+    :param None|int|dict|NumbersDict|tuple|list context_window: will add this context for each chunk
```
albertz: I would just write explicitly `(int,int)|(dict,dict)|(NumbersDict,NumbersDict)`. No need to specifically also mention list here, even if we handle it, and also not all the possible combinations (like `(int,dict)`, even though you should also handle that). But more importantly, it's better to specify that you expect two elements, not just a list/tuple of any length.

albertz: Actually, maybe just supporting `(dict,dict)` is fine for now, to not make the code too complicated, for cases which you don't use currently.

Dataset.py
```diff
@@ -158,6 +158,12 @@ def __init__(self, name=None,
       context_window = NumbersDict(broadcast_value=0, numbers_dict={"data": context_window})
     elif isinstance(context_window, dict):
       context_window = NumbersDict(broadcast_value=0, numbers_dict=context_window)
+    elif isinstance(context_window, (tuple, list)):
+      assert len(context_window) == 2
+      # assume that first element of tuple|list is for left context and second element for right context
+      self.ctx_left = NumbersDict(broadcast_value=0, numbers_dict=context_window[0])
+      self.ctx_right = NumbersDict(broadcast_value=0, numbers_dict=context_window[1])
```
albertz: Never conditionally assign attributes. That very easily can lead to unexpected behavior and bugs.
Instead, move this case (`instance .. tuple`) to the top, as the first `if`, then do this logic, and then put a single `else:` branch, where you put all the other `context_window` logic, and instead of assigning `self.context_window` there, you also calculate `ctx_left`/`ctx_right` (copy the logic from `_get_context_window_left_right` here), and so you only have these attributes, and no `self.context_window` at all anymore.

albertz:
There are failing test cases now.

Dataset.py
```diff
@@ -152,14 +152,31 @@ def __init__(self, name=None,
       assert sorted(chunk_step.keys()) == sorted(chunk_size.keys())
       assert chunk_step.max_value() > 0, "chunking step must be positive (for some key)"
     self.chunk_step = chunk_step
-    if context_window is None:
-      context_window = NumbersDict(0)
```
vieting: Do you know why `context_window` was set to `NumbersDict(0)` here and not to `NumbersDict()`? Using the latter resolves the issues in the test cases (in line 164 now).

albertz: `NumbersDict(0)` and `NumbersDict()` is not the same. We explicitly want that the `broadcast_value` is 0 (i.e. the default fallback value). Just `NumbersDict()` would not have any default fallback value.

Dataset.py
```diff
@@ -152,14 +152,31 @@ def __init__(self, name=None,
...
+    if isinstance(context_window, (tuple, list)):
+      assert len(context_window) == 2
+      for elem in context_window:
+        assert isinstance(elem, dict)
+      # assume that first element of tuple|list is for left context and second element for right context
+      self.ctx_left = NumbersDict(broadcast_value=0, numbers_dict=context_window[0])
+      self.ctx_right = NumbersDict(broadcast_value=0, numbers_dict=context_window[1])
+    else:
+      if context_window is None:
+        context_window = NumbersDict()
```
albertz: That's wrong, it should be `NumbersDict(0)`.

Dataset.py
albertz: (Btw, you should have posted the exception somewhere. I don't remember what it was. I assume the problem was maybe related but in fact somewhere else.)

vieting: ======================================================================
FAIL: test_Dataset.test_batches_recurrent_1
----------------------------------------------------------------------
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    line: self.test(*self.arg)
    locals:
      self = <local> test_Dataset.test_batches_recurrent_1
      self.test = <local> <function test_batches_recurrent_1 at 0x7fb689320510>
      self.arg = <local> ()
  File "/work/asr4/vieting/programs/returnn/tests/test_Dataset.py", line 108, in test_batches_recurrent_1
    line: assert_equal(all_batches[0].seqs[0].seq_end_frame, 10)
    locals:
      assert_equal = <global> <bound method TestCase.assertEqual of <nose.tools.trivial.Dummy testMethod=nop>>
      all_batches = <local> [<Batch start_seq:0, len(seqs):1>, <Batch start_seq:0, len(seqs):1>, <Batch start_seq:0, len(seqs):1>, <Batch start_seq:1, len(seqs):1>, <Batch start_seq:1, len(seqs):1>, <Batch start_seq:1, len(seqs):1>], len = 6
      seqs = <not found>
      seq_end_frame = <not found>
  File "/usr/lib/python3.5/unittest/case.py", line 820, in assertEqual
    line: assertion_func(first, second, msg=msg)
    locals:
      assertion_func = <local> <bound method TestCase._baseAssertEqual of <nose.tools.trivial.Dummy testMethod=nop>>
      first = <local> NumbersDict(numbers_dict={'data': 10, 'classes': 10}, broadcast_value=0)
      second = <local> 10
      msg = <local> None
  File "/usr/lib/python3.5/unittest/case.py", line 813, in _baseAssertEqual
    line: raise self.failureException(msg)
    locals:
      self = <local> <nose.tools.trivial.Dummy testMethod=nop>
      self.failureException = <local> <class 'AssertionError'>
      msg = <local> "NumbersDict(numbers_dict={'data': 10, 'classes': 10}, broadcast_value=0) != 10", len = 78
AssertionError: NumbersDict(numbers_dict={'data': 10, 'classes': 10}, broadcast_value=0) != 10

======================================================================
FAIL: test_Dataset.test_batches_non_recurrent_1
----------------------------------------------------------------------
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    line: self.test(*self.arg)
    locals:
      self = <local> test_Dataset.test_batches_non_recurrent_1
      self.test = <local> <function test_batches_non_recurrent_1 at 0x7fb689320598>
      self.arg = <local> ()
  File "/work/asr4/vieting/programs/returnn/tests/test_Dataset.py", line 168, in test_batches_non_recurrent_1
    line: assert_equal(all_batches[0].seqs[0].seq_end_frame, 5)
    locals:
      assert_equal = <global> <bound method TestCase.assertEqual of <nose.tools.trivial.Dummy testMethod=nop>>
      all_batches = <local> [<Batch start_seq:0, len(seqs):1>, <Batch start_seq:0, len(seqs):1>, <Batch start_seq:0, len(seqs):2>, <Batch start_seq:1, len(seqs):1>, <Batch start_seq:1, len(seqs):1>]
      seqs = <not found>
      seq_end_frame = <not found>
  File "/usr/lib/python3.5/unittest/case.py", line 820, in assertEqual
    line: assertion_func(first, second, msg=msg)
    locals:
      assertion_func = <local> <bound method TestCase._baseAssertEqual of <nose.tools.trivial.Dummy testMethod=nop>>
      first = <local> NumbersDict(numbers_dict={'data': 5, 'classes': 5}, broadcast_value=0)
      second = <local> 5
      msg = <local> None
  File "/usr/lib/python3.5/unittest/case.py", line 813, in _baseAssertEqual
    line: raise self.failureException(msg)
    locals:
      self = <local> <nose.tools.trivial.Dummy testMethod=nop>
      self.failureException = <local> <class 'AssertionError'>
      msg = <local> "NumbersDict(numbers_dict={'data': 5, 'classes': 5}, broadcast_value=0) != 5", len = 75
AssertionError: NumbersDict(numbers_dict={'data': 5, 'classes': 5}, broadcast_value=0) != 5

----------------------------------------------------------------------
Ran 9 tests in 0.049s

FAILED (failures=2)

Process finished with exit code 1

vieting: I think the problem is that since I removed the check `if ctx_lr:` in line 1012, now `start -= self.ctx_left` is executed even if we do not want a context window. 

In that case, the `broadcase_value` of `self.ctx_left` (which is 0) is used and eventually ends up in `all_batches[0].seqs[0].seq_end_frame` which is why the test fails. 

I could include a check again whether `self.ctx_left` and `self.ctx_right` are both zero.

albertz: (Please properly format using Markdown. I edited your post.)

So, maybe what we want here is actually that `start -= self.ctx_left` would not add a `broadcast_value` to `start`, I guess. I'm not sure if that is always true, or only in this case. Actually, I think calculating `start - self.ctx_left` should result in `NumbersDict(numbers_dict={'data': 5, 'classes': 5}, broadcast_value=0)`, so `NumbersDict.__sub__` (`-`) is correct. However, about the logic for `-=`, I'm not exactly sure... That logic is implemented in `NumbersDict.__isub__`.

albertz: Btw, just adding such a check is not the right solution. That would be just a workaround.

albertz: I pushed b8c79207a203c9 (to a test branch albert-numbersdict-better-inplace), let's see if all the tests passes. If so, I will push that to master. This should solve this problem with `NumbersDict`.

albertz: Ok, this is now failing somewhere else. I have to look into that.

vieting: This way, `test_batches_non_recurrent_1()` fails because in `EngineBatch.py` -> `add_frames()` -> `self.max_num_frames_per_slice += length`, `NumbersDict(0) += NumbersDict({'data': 5, 'classes': 5})` results in `NumbersDict(0)` which eventually leads to the error. We could replace `a += b` by `a = a + b` there and the tests pass.

However, I am not sure what the most sensible behaviour would be for adding two `NumbersDicts`, one having just the `broadcast_value` and one having just the `dict`. So, should it be `NumbersDict(0) + NumbersDict({'data': 5, 'classes': 5}) = NumbersDict({'data': 5, 'classes': 5}, broadcast_value=0)`? The 0 could be misleading then.

albertz: Maybe you are right, maybe using `context_window = NumbersDict()` is actually a good solution. But if we use that, for the other code where we set `broadcast_value=0`, we also should remove that.

Dataset.py
```diff
@@ -152,14 +152,31 @@ def __init__(self, name=None,
...
-      context_window = NumbersDict(broadcast_value=0, numbers_dict=context_window)
-    assert isinstance(context_window, NumbersDict)
-    self.context_window = context_window
+    if isinstance(context_window, (tuple, list)):
+      assert len(context_window) == 2
+      for elem in context_window:
+        assert isinstance(elem, dict)
+      # assume that first element of tuple|list is for left context and second element for right context
+      self.ctx_left = NumbersDict(broadcast_value=0, numbers_dict=context_window[0])
+      self.ctx_right = NumbersDict(broadcast_value=0, numbers_dict=context_window[1])
```
albertz: Here, we should remove `broadcast_value=0`.

Dataset.py
```diff
@@ -152,14 +152,31 @@ def __init__(self, name=None,
...
+    else:
+      if context_window is None:
+        context_window = NumbersDict()
+      elif isinstance(context_window, int):
+        context_window = NumbersDict(broadcast_value=0, numbers_dict={"data": context_window})
```
albertz: Here as well.

Dataset.py
```diff
@@ -152,14 +152,31 @@ def __init__(self, name=None,
...
+      elif isinstance(context_window, dict):
+        context_window = NumbersDict(broadcast_value=0, numbers_dict=context_window)
```
albertz: And here.

Dataset.py
vieting: I removed `broadcast_value=0` in the other cases now.

albertz:
Hm, there are failing tests, but I think there are unrelated. But I will fix them first, but probably merge this in afterwards.

PR #224, opened 2019-09-19
Fixes for ChoiceLayer sub-layers

patrick-wilken:
Some recent changes did not reflect sub-layers, in particular those of the ChoiceLayer in case of two targets.

Apart from these minor changes, thanks to correctly backtracking the beams of all output layers of the rec layer via _opt_search_resolve, ChoiceLayer with two targets is finally fully working in the master branch, nice! ':D

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -2353,8 +2353,8 @@ class _SubnetworkRecCell(object):
   def _opt_search_resolve(self, layer_name, acc_ta, final_net_vars):
     """
     This assumes that we have frame-wise accumulated outputs of the specific layer (acc_ta).
-    If that layer depends on frame-wise search choices, i.e. it the batch dim includes a search beam,
-    and each frame will has different beams,
+    If that layer depends on frame-wise search choices, i.e. if the batch dim includes a search beam,
+    and each frame has different beams,
     then we will resolve that to the common final best search choices,
     which is determined by the latest search choice given by the layer.
     This assumes that we also have `self.final_acc_tas_dict["choice_%s" % choice_base.name]` available.
@@ -2369,9 +2369,11 @@ class _SubnetworkRecCell(object):
     from TFUtil import nd_indices, assert_min_tf_version, expand_dims_unbroadcast, get_valid_scope_name_from_str
     from TFUtil import get_shape_dim
     rec_layer = self.parent_rec_layer
-    if layer_name not in self.net.layers:  # inside loop
+    try:
+      layer = self.net.get_layer(layer_name)
+    except LayerNotFound:  # layer is not inside loop
       return acc_ta, None
-    layer_choice = self.net.get_search_choices(src=self.net.layers[layer_name])
+    layer_choice = self.net.get_search_choices(src=layer)
     if not layer_choice:
       return acc_ta, None
     is_prev_choice = False
@@ -2430,8 +2432,8 @@ class _SubnetworkRecCell(object):
 
     new_acc_output_ta = tf.TensorArray(
       name="search_resolved_%s" % os.path.basename(acc_ta.handle.op.name),
-      dtype=self.net.layers[layer_name].output.dtype,
-      element_shape=tf.TensorShape(self.net.layers[layer_name].output.batch_shape),
+      dtype=layer.output.dtype,
+      element_shape=tf.TensorShape(layer.output.batch_shape),
       size=seq_len,
       infer_shape=True)
 
@@ -2445,8 +2447,8 @@ class _SubnetworkRecCell(object):
       """
       with tf.name_scope("transform_output_%s" % get_valid_scope_name_from_str(layer_name)):
         output_ = acc_ta.read(i)  # (batch * beam, [n_out])
-        assert self.net.layers[layer_name].output.batch_dim_axis == 0
-        out_shape = list(self.net.layers[layer_name].output.batch_shape)
+        assert layer.output.batch_dim_axis == 0
+        out_shape = list(layer.output.batch_shape)
         output_.set_shape(tf.TensorShape(out_shape))
         for i_, d in list(enumerate(out_shape)):
           if i_ == 0:
@@ -2829,7 +2831,7 @@ class _SubnetworkRecCell(object):
       if name in loop_acc_layers:
         return loop_acc_layers[name]
       with tf.name_scope(self.layer_data_templates[name].layer_class_type.cls_get_tf_scope_name(name)):
-        inner_layer = self.net.layers[name]
+        inner_layer = self.net.get_layer(name)
         acc_ta = loop_accumulated["output_%s" % name]
         acc_ta, search_choices = self._opt_search_resolve(
           layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars)
@@ -3049,7 +3051,7 @@ class _TemplateLayer(LayerBase):
       return self._cell
     rec_layer = self.network.parent_layer
     assert isinstance(rec_layer, RecLayer)
-    cell = self.network.parent_layer
+    cell = self.network.parent_layer.cell
     assert isinstance(cell, _SubnetworkRecCell)
     return cell
 
@@ -4429,6 +4431,23 @@ class ChoiceLayer(BaseChoiceLayer):
     super(ChoiceLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
 
   @classmethod
+  def _create_search_beam(cls, name, beam_size, sources, network):
+    """
+    :param str name:
+    :param int beam_size:
+    :param list[LayerBase] sources:
+    :param TFNetwork.TFNetwork network:
+    :rtype: TFUtil.SearchBeam
+    """
+    from TFUtil import SearchBeam
+    search_dep = NotSpecified
+    if sources and sources[0] and not sources[0].output.undefined:
+      search_dep = sources[0].output.beam
+    return SearchBeam(
+      beam_size=beam_size, dependency=search_dep,
+      name="%s%s" % (network.get_absolute_name_prefix(), name))
+
+  @classmethod
   def get_out_data_from_opts(cls, name, sources, target, network,
                              beam_size, search=NotSpecified, scheduled_sampling=False, cheating=False, **kwargs):
     """
@@ -4457,13 +4476,7 @@ class ChoiceLayer(BaseChoiceLayer):
       del shape[out_data.batch_dim_axis]
       out_data = Data(name="%s_output" % name, shape=shape, sparse=True, dim=out_data.dim)
     if search:
-      from TFUtil import SearchBeam
-      search_dep = NotSpecified
-      if sources and sources[0] and not sources[0].output.undefined:
-        search_dep = sources[0].output.beam
-      out_data.beam = SearchBeam(
-        beam_size=beam_size, dependency=search_dep,
-        name="%s%s" % (network.get_absolute_name_prefix(), name))
+      out_data.beam = cls._create_search_beam(name=name, beam_size=beam_size, sources=sources, network=network)
     elif sources and sources[0]:
       out_data.beam = sources[0].output.beam
     if cheating or scheduled_sampling or not search:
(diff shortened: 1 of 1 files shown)


TFNetworkLayer.py
```diff
@@ -2036,7 +2036,7 @@ def get_out_data_from_opts(cls, name, sources=(), extra_deps=(), out_type=None,
     :param int|None|NotSpecified n_out:
     :rtype: Data
     """
-    if out_type or n_out is not NotSpecified:
+    if out_type or (n_out is not NotSpecified and sources and n_out != sources[0].output.dim):
```
albertz: This is about `CopyLayer`. Why this change?
(Maybe actually this whole `if` branch could be removed; it's a bit historical. But anyway, it seems unrelated to the remaining PR.)

TFNetworkRecLayer.py
```diff
@@ -2369,9 +2369,10 @@ def _opt_search_resolve(self, layer_name, acc_ta, final_net_vars):
     from TFUtil import nd_indices, assert_min_tf_version, expand_dims_unbroadcast, get_valid_scope_name_from_str
     from TFUtil import get_shape_dim
     rec_layer = self.parent_rec_layer
-    if layer_name not in self.net.layers:  # inside loop
+    layer = self.net.get_layer(layer_name)
+    if not layer:  # inside loop
```
albertz: This looks wrong. `self.net.get_layer` never returns `None`. If the layer does not exists, it would throw a `LayerNotFound` exception.

TFNetworkRecLayer.py
```diff
@@ -3049,7 +3050,7 @@ def _get_cell(self):
       return self._cell
     rec_layer = self.network.parent_layer
     assert isinstance(rec_layer, RecLayer)
-    cell = self.network.parent_layer
+    cell = self.network.parent_layer.cell
```
albertz: Whoops. How was it actually possible that we/I did not run into this bug before?

TFNetworkRecLayer.py
```diff
@@ -4501,12 +4502,23 @@ def get_sub_layer_out_data_from_opts(cls, layer_name, parent_layer_kwargs):
...
+    parent_layer_name = parent_layer_kwargs["name"]
+    network = parent_layer_kwargs["network"]
+
     # The sub-layer with index n will output the n-th target. The out_data is taken directly
     # from the target as it is done in self.get_out_data_from_opts().
-    sub_layer_out_data = cls.get_out_data_from_opts(target=targets[index], network=parent_layer_kwargs["network"],
+    sub_layer_out_data = cls.get_out_data_from_opts(name=parent_layer_name + "/" + layer_name,
+                                                    sources=parent_layer_kwargs["sources"],
+                                                    target=targets[index],
+                                                    network=network,
```
albertz: Don't indent code like this. Make a newline and the indent by 2 spaces.

TFNetworkRecLayer.py
```diff
@@ -4501,12 +4502,23 @@ def get_sub_layer_out_data_from_opts(cls, layer_name, parent_layer_kwargs):
...
                                                     beam_size=parent_layer_kwargs["beam_size"])
+
+    if network.search_flag:
+        # Set name of beam to parent_layer, so that they compare to equal.
+        sub_layer_out_data.beam.name = "%s%s" % (network.get_absolute_name_prefix(), parent_layer_name)
```
albertz: We indent by 2 spaces.

albertz: Btw, regarding the code: Instead of overwriting the `beam.name`, maybe better resetting the `beam` itself, to `parent_layer.beam` or so? I feel that this is a bit ugly, and has certain assumptions hard-coded about the name of the beam, which might easily break later.

TFNetworkLayer.py
patrick-wilken: Sorry, I actually made some comments on Friday, but somehow they weren't posted.

I ran into the problem that `n_out` is set via `transform_config_dict` if the layer has a target. Using `super(CopyLayer, cls).get_out_data_from_opts` in this case fails to set `sparse=True`. (This happened copying from a layer in the rec unit to outside, so it's somewhat related.)

With this, I only use the base method if `n_out` is "unexpected", i.e. different than what get_concat_sources_data_template would result in.

I would be happy to remove the `if`. Because in a CopyLayer we should always know the shape from the inputs, so I don't see when setting `out_type` should be necessary.

TFNetworkRecLayer.py
patrick-wilken: We always set the cell when creating a TemplateLayer in `GetLayer.__call__()`, such that `self._cell` is returned. I didn't do that in `_TemplateLayer.get_sub_layer` (I guess I should/could have...).

TFNetworkRecLayer.py
patrick-wilken: I agree this is not clean. But `get_sub_layer_out_data_from_opts()` has no access to the parent layer (as it's a classmethod), only its arguments.
I think I will create a new `SearchBeam` object in this place using the parent layer name, this should be better.

TFNetworkLayer.py
albertz: Anyway. This seems unrelated to this pull request. So don't mix it in here.

But maybe create a new pull request, where you remove the whole `if` branch, and then we can see if any test case breaks.

patrick-wilken:
Ok, I removed the CopyLayer change and tried to address the other issues.

TFNetworkRecLayer.py
albertz: Did you resolve this? Can you comment?

TFNetworkRecLayer.py
```diff
@@ -4428,6 +4430,22 @@ def transform_config_dict(cls, d, network, get_layer):
         [get_layer(name) for name in d["explicit_search_sources"]] if network.search_flag else [])
     super(ChoiceLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
 
+  @classmethod
+  def _create_search_beam(cls, name, beam_size, sources, network):
+    """
+    :param str name:
+    :param int beam_size:
+    :param list[LayerBase] sources:
+    :param TFNetwork.TFNetwork network:
```
albertz: `:rtype:` not specified.

TFNetworkRecLayer.py
```diff
@@ -4428,6 +4430,22 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    """
+    from TFUtil import SearchBeam
+    search_dep = NotSpecified
+    if sources and sources[0] and not sources[0].output.undefined:
+      search_dep = sources[0].output.beam
+    return SearchBeam(
+      beam_size=beam_size, dependency=search_dep,
+      name="%s%s" % (network.get_absolute_name_prefix(), name))
```
albertz: This has still the hardcoded solution. This is not nice and can potentially break later.

TFNetworkRecLayer.py
patrick-wilken: Yes, it was wrong. I'm now catching `LayerNotFound` exceptions and return the original tensor array in the except block.

TFNetworkRecLayer.py
patrick-wilken: Uhm, what exactly do you mean is hardcoded here? The fact that the same SearchBeam is created twice, relying on the name? 

Well. the idea of `get_sub_layer_out_data_from_opts` was to create `out_data` of the sub-layer without depending on the `out_data` or an instance of the parent layer. Do you want to change that?

Another option would be to create the parent layer beam by calling `get_out_data_from_opts` using the `parent_layer_kwargs` and then just use `out_data.beam` and discard the rest. But this is the most generic you could do here, right?

albertz: No, for `name` you assume that it matches to other code.

patrick-wilken: I'm still not sure I get what you mean. An instance of SearchBeam is characterised by the owner layer name and comparison of beams relies on the name, this is nothing I introduced. Also I did not change the name format `"%s%s" % (network.get_absolute_name_prefix(), name))`.
All I need to do is to use `parent_layer_kwargs["name"]` instead of `layer_name` when creating the beam of the sub layer. How does this depend on other code? Well, I'm assuming that the parent layer beam has indeed the name of the parent layer, but that's it.

albertz: You assume that we have this specific name format.

patrick-wilken: Yes, because it was there before ':D As of now, it is a local property of ChoiceLayer. And the class methods have access to it.
I want to go from kwargs to SearchBeam. How would I do this without using some kind of naming convention?

albertz: I'm just saying that you have this assumption here about the name format, and this could break at some point. Better write code in a way that it will not break.

patrick-wilken: I really don't understand what my PR changed about the naming. As before, there is one, and only one, place in the code that *defines* the naming of SearchBeams.
Again, how do I get from kwargs to SearchBeam? In kwargs there is no SearchBeam, so I have to create one. For this I need a name.

albertz: I thought there are two places now? Where was the `SearchBeam` created before?

TFNetworkRecLayer.py
```diff
@@ -4457,13 +4475,7 @@ def get_out_data_from_opts(cls, name, sources, target, network,
...
       out_data = Data(name="%s_output" % name, shape=shape, sparse=True, dim=out_data.dim)
     if search:
-      from TFUtil import SearchBeam
-      search_dep = NotSpecified
-      if sources and sources[0] and not sources[0].output.undefined:
-        search_dep = sources[0].output.beam
-      out_data.beam = SearchBeam(
-        beam_size=beam_size, dependency=search_dep,
-        name="%s%s" % (network.get_absolute_name_prefix(), name))
+      out_data.beam = cls._create_search_beam(name, beam_size, sources, network)
```
albertz: You should always use named kwargs if possible, if you have more than 1 param or there can be any ambiguity from the function name about the params.

TFNetworkRecLayer.py
```diff
@@ -4457,13 +4475,7 @@ def get_out_data_from_opts(cls, name, sources, target, network,
       del shape[out_data.batch_dim_axis]
       out_data = Data(name="%s_output" % name, shape=shape, sparse=True, dim=out_data.dim)
     if search:
-      from TFUtil import SearchBeam
-      search_dep = NotSpecified
-      if sources and sources[0] and not sources[0].output.undefined:
-        search_dep = sources[0].output.beam
-      out_data.beam = SearchBeam(
-        beam_size=beam_size, dependency=search_dep,
-        name="%s%s" % (network.get_absolute_name_prefix(), name))
```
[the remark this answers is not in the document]
patrick-wilken: Ah, ok, sorry ':D In general comments get a bit messy when I force push. Should I better commit fixups and then rebase when approved?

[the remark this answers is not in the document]
patrick-wilken: Or is it the same with fixups...?

patrick-wilken:
Added result type and using kwargs now.

TFNetworkRecLayer.py
albertz: So, is this resolved? (If you resolve sth, and the old code gets outdated, I cannot see the new code, so I don't directly see it. Please click yourself on "resolve".)

TFNetworkRecLayer.py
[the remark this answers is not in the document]
albertz: There is no difference in the final diff (here shown on GitHub), no matter if you rebase (force push) or just add commits. So this has no effect.
It's just hard in general to see from a diff (or easy to miss), that you changed the only place where `SearchBeam` was created.

TFNetworkRecLayer.py
```diff
@@ -4428,6 +4430,23 @@ def transform_config_dict(cls, d, network, get_layer):
...
     super(ChoiceLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
 
+  @classmethod
+  def _create_search_beam(cls, name, beam_size, sources, network):
+    """
+    :param str name:
+    :param int beam_size:
+    :param list[LayerBase] sources:
+    :param TFNetwork.TFNetwork network:
+    :rtype: SearchBeam
```
albertz: I think the type is wrong, if it is not globally imported. (Did you check the type inference for the function?) I guess it must be `TFUtil.SearchBeam`.

patrick-wilken: True, fixed now.

albertz:
Merged now. Maybe it makes sense to add a simple short test case for this later.

PR #226, opened 2019-09-23
CopyLayer: always get out_type from sources if possible

patrick-wilken:
`get_concat_sources_data_template` should be the right thing to do in all cases to create the output template.

Logic depending on `n_out` can lead to problems, e.g.:
`n_out` is set via `LayerBase.transform_config_dict()` if a layer has a target. Using super(CopyLayer, cls).get_out_data_from_opts in this case fails to set `sparse=True`.

Should we add a check that, if `out_type` or `n_out` are given, they match the result of  `get_concat_sources_data_template`?

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -2036,10 +2036,15 @@ class CopyLayer(_ConcatInputLayer):
     :param int|None|NotSpecified n_out:
     :rtype: Data
     """
-    if out_type or n_out is not NotSpecified:
-      return super(CopyLayer, cls).get_out_data_from_opts(
-        name=name, out_type=out_type, n_out=n_out, sources=sources, **kwargs)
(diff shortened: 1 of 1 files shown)


albertz:
Yes, at least for `n_out` there should be a check, maybe also for `out_type`.

albertz:
Tests are failing.

patrick-wilken:
Hmm, ok, I will look into it.

patrick-wilken:
Ok, the test was failing in a case where the copy layer was the only layer in the net and the only input was an "initial_output". Here we need the `out_type`. All layers ignoring `out_type` would fail under this condition (and there are quite a few).

So now I simply changed the order: first try to get the shape from the sources (which always works if they are defined) and only if necessary resort to the out_type. This solves my problem with `sparse` not being set and also passes the test.

PR #227, opened 2019-09-24
Create directory if it doesn't already exist

ChristopherBrix:
If the initial state of the model should be saved, the directory might not yet exist. The code that creates it is called in `log_runtime_info_to_dir`, which only runs during the training.

### DIFF
--- a/TFNetwork.py
+++ b/TFNetwork.py
@@ -1371,6 +1371,8 @@ class TFNetwork(object):
     """
     import os
     filename = os.path.abspath(filename)  # TF needs absolute path
+    from Util import maybe_make_dirs
+    maybe_make_dirs(os.path.dirname(filename))
     if not self.saver:
       self._create_saver()
     # We add some extra logic to try again for DiskQuota and other errors.
--- a/Util.py
+++ b/Util.py
@@ -2975,6 +2975,23 @@ def get_utc_start_time_filename_part():
   return time.strftime("%Y-%m-%d-%H-%M-%S", time.gmtime(start_time))
 
 
+def maybe_make_dirs(dirname):
+  """
+  Creates the directory if it does not yet exist.
+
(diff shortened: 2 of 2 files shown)


TFNetwork.py
```diff
@@ -1371,6 +1371,10 @@ def save_params_to_file(self, filename, session):
     """
     import os
     filename = os.path.abspath(filename)  # TF needs absolute path
+    dirname = os.path.dirname(filename)
+    from Util import maybe_create_folder
+    maybe_create_folder(dirname)
+
```
albertz: Don't randomly add empty lines. Also, avoid creating too much local variables (here: `dirname`), esp if it is only used once. You can directly use `os.path.dirname(filename)`.

Util.py
```diff
@@ -2975,6 +2975,22 @@ def get_utc_start_time_filename_part():
   return time.strftime("%Y-%m-%d-%H-%M-%S", time.gmtime(start_time))
 
 
+def maybe_make_dirs(dirname):
+  """
+  Creates the directory if it does not yet exist.
+  :param dirname: The absolute path of the directory
```
albertz: There should be a type in the docstring (`:param str dirname:`).
Also better make a newline in the docstring after the description, before the params.

ChristopherBrix:
Done. Also fixed the description, the path doesn't have to be absolute. 

But actually, couldn't we just use the `exist_ok` flag of `os.makedirs`? It also accounts for possible race conditions due to parallel threads, so this should cover our use case, right?
Then we could remove the whole `maybe_make_dirs` function and call `os.makedirs` directly.

albertz:
> But actually, couldn't we just use the exist_ok flag of os.makedirs?

This is only supported since Python 3. We have not dropped that support yet.

PR #228, opened 2019-09-24
SubnetworkRecCell: resolve beams of seq_len

### LINKED ISSUE #225: SubnetworkRecCell: size_placeholder of outputs not resolved
I have the following configuration in my recurrent unit:

Choice layers C1 and C2, end layer E
C2 depends on C1, but not on E
E depends on C1

C2 is the last choice layer, thus defines the beam to which all outputs are resolved to (via `_opt_search_resolve()`). C1 is resolved to C2, this works.
However, the size_placeholder of all outputs is set to `seq_len` (in `get_loop_acc_layer()`), which itself comes from the length information of the end layer. In my case, the end layer has beams according to C1, not C2. Resolving these beams is currently missing, giving a shape mismatch in case of different beam sizes or, even worse, outputs of wrong length in case of same beam sizes.

I'm trying to figure out where to do this, but maybe you, @albertz, have an idea right away.

patrick-wilken:
This fixes  #225

Two things to discuss:
1. Resolving beams in `body()` is now pretty inconsistent (`translate_to_this_search_beam`, `SelectSearchSourcesLayer.transform_func` and `select_src_beams`). I think we should refactor the code, such that there is one utility function, which translates beams taking a sequence of search choices, and `SelectSearchSourcesLayer` should use this function.

2. It is a bit vague how exactly the latest search choice in the recurrent unit should be determined. I now calculate it globally, comparing all search choices of `needed_outputs`. Is there any downside to this? It should give the same result as the first "prev" layer in the choice sequence and it should be enough to do it only once, but there is simply no obvious layer to start the choice sequence from ("output" can be outside the loop, "end" doesn't always exists). 
And for the initial shape of `dyn_seq_len` I use the "output" layer template, which assumes it to have the resolved output beam size... Maybe we should also use the latest choice here, but this would require getting it from the templates.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -8,6 +8,7 @@ from __future__ import print_function
 import tensorflow as tf
 import typing
 from tensorflow.python.ops.nn import rnn_cell
+from TFNetwork import LayerNotFound
 from TFNetworkLayer import LayerBase, _ConcatInputLayer, SearchChoices, get_concat_sources_data_template, Loss
 from TFUtil import Data, SearchBeam, reuse_name_scope, get_random_seed, select_src_beams
 from Util import NotSpecified
@@ -293,7 +294,6 @@ class RecLayer(_ConcatInputLayer):
           # We only need this to resolve any other layer dependencies in the main network.
           cl.transform_config_dict(sub.copy(), network=subnet, get_layer=sub_get_layer)
     if isinstance(d.get("max_seq_len"), str):
-      from TFNetwork import LayerNotFound
 
       def max_len_from(src):
         """
@@ -2216,10 +2216,10 @@ class _SubnetworkRecCell(object):
       [tf.TensorShape(None) for _ in init_acc_tas])
     if not have_known_seq_len:
       # See body().
-      out_batch_dim = self.layer_data_templates["end"].get_batch_dim()
+      end_layer_batch_dim = self.layer_data_templates["end"].get_batch_dim()
       init_seq_len_info = (
-        constant_with_shape(False, shape=[out_batch_dim], name="initial_end_flag"),
-        constant_with_shape(0, shape=[out_batch_dim], name="initial_seq_len"))
+        constant_with_shape(False, shape=[end_layer_batch_dim], name="initial_end_flag"),
+        constant_with_shape(0, shape=[end_layer_batch_dim], name="initial_seq_len"))
       init_loop_vars += (init_seq_len_info,)
       shape_invariants += ((tf.TensorShape([None]), tf.TensorShape([None])),)
     if self.layers_in_loop:
@@ -2352,7 +2352,7 @@ class _SubnetworkRecCell(object):
 
     return output, output_search_choices
 
-  def _opt_search_resolve(self, layer_name, acc_ta, final_net_vars):
+  def _opt_search_resolve(self, layer_name, acc_ta, final_net_vars, seq_len):
     """
     This assumes that we have frame-wise accumulated outputs of the specific layer (acc_ta).
     If that layer depends on frame-wise search choices, i.e. if the batch dim includes a search beam,
@@ -2360,27 +2360,30 @@ class _SubnetworkRecCell(object):
     then we will resolve that to the common final best search choices,
     which is determined by the latest search choice given by the layer.
     This assumes that we also have `self.final_acc_tas_dict["choice_%s" % choice_base.name]` available.
+    In addition, we resolve the sequence lengths (whose beams correspond to the end layer) to the final search choices.
 
     :param str layer_name:
     :param tf.TensorArray acc_ta: accumulated outputs of that layer
     :param final_net_vars:
-    :return: (new acc_ta, final search choices)
-    :rtype: (tf.TensorArray,SearchChoices|None)
+    :param tf.Tensor seq_len: shape (batch * beam,), has beam of the "end" layer in case of dynamic sequence lengths,
+      otherwise beam of rec_layer.output
+    :return: (new acc_ta, final search choices, resolved seq_len)
+    :rtype: (tf.TensorArray,SearchChoices|None,tf.Tensor)
     """
     import os
     from TFUtil import nd_indices, assert_min_tf_version, expand_dims_unbroadcast, get_valid_scope_name_from_str
     from TFUtil import get_shape_dim
-    from TFNetwork import LayerNotFound
     rec_layer = self.parent_rec_layer
     try:
       layer = self.net.get_layer(layer_name)
     except LayerNotFound:  # layer is not inside loop
-      return acc_ta, None
-    layer_choice = self.net.get_search_choices(src=layer)
-    if not layer_choice:
-      return acc_ta, None
-    if layer_choice.search_choices.keep_raw:
-      return acc_ta, layer_choice.search_choices
+      return acc_ta, None, seq_len
+    search_choices = layer.get_search_choices()
+    if not search_choices:
+      return acc_ta, None, seq_len
+    if search_choices.keep_raw:
+      return acc_ta, search_choices, seq_len
+    layer_choice = search_choices.owner
     is_prev_choice = False
     if isinstance(layer_choice, _TemplateLayer):
       assert layer_choice.is_prev_time_frame
@@ -2418,11 +2421,11 @@ class _SubnetworkRecCell(object):
     latest_layer_choice = self.net.layers[latest_layer_choice.name[len("prev:"):]]
     assert latest_layer_choice.search_choices
 
-    # The seq_len might actually be one more, as it includes the EOS, but that does not matter;
+    # The max_seq_len might actually be one more, as it includes the EOS, but that does not matter;
     # we just want to create a new acc_ta with the same length.
     # (Cutting off the EOS is handled elsewhere.)
-    seq_len = acc_ta.size()
-    initial_i = tf.identity(seq_len - 1, name="search_resolve_initial_i")  # we go backwards
+    max_seq_len = acc_ta.size()
+    initial_i = tf.identity(max_seq_len - 1, name="search_resolve_initial_i")  # we go backwards
     latest_beam_size = latest_layer_choice.output.beam.beam_size
     batch_dim = rec_layer.network.get_data_batch_dim()
 
@@ -2435,11 +2438,45 @@ class _SubnetworkRecCell(object):
     assert prev_frame_choice.name == "prev:%s" % latest_layer_choice.name
     assert layer_choice in choice_seq_in_frame
 
+    # Translate the beams of seq_lens to correspond to the latest choice
+    try:
+      end_layer = self.net.get_layer("end")
+    except LayerNotFound:
+      # This means have_known_seq_len=True.
+      end_layer = None
+
+    if end_layer:
+      end_layer_choice = self.net.get_search_choices(src=end_layer)
+      assert end_layer_choice in choice_seq_in_frame, (
+        "End layer must not have a beam independent from output layer '{}'.".format(layer_name))
+
+      end_layer_choice_index = choice_seq_in_frame.index(end_layer_choice)
+      choices_seq_until_end_layer = choice_seq_in_frame[:end_layer_choice_index]
+
+      for choice_ in reversed(choices_seq_until_end_layer):
+        src_choice_beams = self.final_acc_tas_dict["choice_%s" % choice_.name].read(
+          max_seq_len - 1, name="ta_read_choice")  # (batch, beam) -> beam_in idx
+        seq_len = select_src_beams(seq_len, src_choice_beams)
+    else:
+      # Here we don't need to resolve anything, as the sequence length is the same for all hyps in the beam.
+      # However, beam size for the current output may be different from the "output" layer.
+      # Therefore take the first len in the beam and tile it to the desired beam size.
+
+      # Separate batch and beam dims
+      seq_len_beam_size = rec_layer.output.beam.beam_size
+      seq_len = tf.reshape(seq_len, [batch_dim, seq_len_beam_size], name="split_batch_beam")
+
+      seq_len = seq_len[:, 0:1]
+      seq_len = tf.tile(seq_len, [1, latest_beam_size], name="resize_seq_len_beam")
+
+      # Recombine batch and beam dims
+      seq_len = tf.reshape(seq_len, [batch_dim * latest_beam_size], name="merge_batch_beam")
+
     new_acc_output_ta = tf.TensorArray(
       name="search_resolved_%s" % os.path.basename(acc_ta.handle.op.name),
       dtype=layer.output.dtype,
       element_shape=tf.TensorShape(layer.output.batch_shape),
-      size=seq_len,
+      size=max_seq_len,
       infer_shape=True)
 
     def transform(i, idxs_exp, new_acc_output_ta_):
@@ -2475,7 +2512,7 @@ class _SubnetworkRecCell(object):
       This loops goes backwards through time.
       Similar to tf.contrib.seq2seq.GatherTree.
 
-      :param tf.Tensor i: starts at seq_len - 1
+      :param tf.Tensor i: starts at max_seq_len - 1
       :param tf.Tensor choice_beams: from the previous step, shape (batch, beam_out) -> beam idx,
         just before prev_frame_choice
       :param tf.TensorArray|None new_acc_output_ta_: shape (batch * beam, n_out) per frame
@@ -2533,7 +2570,7 @@ class _SubnetworkRecCell(object):
       beam_choices = tf.zeros_like(final_beam_choices)
       with tf.name_scope("search_resolve_first_frame"):
         new_acc_output_ta = tf.cond(
-          tf.less(0, seq_len),
+          tf.less(0, max_seq_len),
           lambda: transform(0, nd_indices(beam_choices), new_acc_output_ta),
           lambda: new_acc_output_ta)
 
@@ -2545,7 +2582,7 @@ class _SubnetworkRecCell(object):
       layer_name=latest_layer_choice.name)
     acc_search_choices.set_beam_from_rec(final_choice_rec_vars)
 
-    return new_acc_output_ta, acc_search_choices
+    return new_acc_output_ta, acc_search_choices, seq_len
 
   def _input_layer_used_inside_loop(self, layer_name):
     """
@@ -2804,10 +2841,8 @@ class _SubnetworkRecCell(object):
     from TFNetworkLayer import InternalLayer
 
     if seq_len is not None:
-      max_len = tf.reduce_max(seq_len)
       time_dim_tag = DimensionTag.get_tag_from_size_tensor(seq_len)
     else:
-      max_len = None
       time_dim_tag = None
     self.output_layers_net = TFNetwork(
       name="%s/%s:rec-subnet-output" % (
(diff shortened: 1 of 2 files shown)


TFNetworkRecLayer.py
```diff
@@ -2214,9 +2268,10 @@ def cond(i, net_vars, acc_tas, seq_len_info=None):
       [tf.TensorShape(None) for _ in init_acc_tas])
     if not have_known_seq_len:
       # See body().
-      out_batch_dim = self.layer_data_templates["end"].get_batch_dim()
+      end_batch_dim = self.layer_data_templates["end"].get_batch_dim()
+      out_batch_dim = self.layer_data_templates["output"].get_batch_dim()
       init_seq_len_info = (
-        constant_with_shape(False, shape=[out_batch_dim], name="initial_end_flag"),
+        constant_with_shape(False, shape=[end_batch_dim], name="initial_end_flag"),
         constant_with_shape(0, shape=[out_batch_dim], name="initial_seq_len"))
```
albertz: I don't think it is a good idea to mix it. As said, the end layer should just be resolved automatically to the latest choice. And the same for the "output" layer.

TFNetworkRecLayer.py
```diff
@@ -2135,13 +2178,24 @@ def maybe_transform(layer):
...
+              # Translate the beams of end_flags to the latest choices in the net.
+              last_choices = self._latest_choice_layer.get_search_choices()
+              resolved_end_layer = last_choices.translate_to_this_search_beam(cur_end_layer)
+              end_flag_resolved = resolved_end_layer.transform_func(
+                cur_end_layer.output.placeholder if rec_layer.include_eos else end_flag)
             with tf.name_scope("dyn_seq_len"):
-              dyn_seq_len = cur_end_layer.transform_func(dyn_seq_len)
+              # Translate beams of dyn_seq_len from previous to current frame (going through all choices)
+              for choice_layer in reversed(self._choice_seq_in_frame):
+                dyn_seq_len = select_src_beams(dyn_seq_len, src_beams=choice_layer.search_choices.src_beams)
```
albertz: Don't call `select_src_beams` here, and esp do not replicate the logic about unrolling over the choice seq. Just use the existing logic (e.g. call `transform_func`).
In any case, as said, seq len and end flag, both should be resolved in the same way.

albertz:
> Resolving beams in body() is now pretty inconsistent (translate_to_this_search_beam, SelectSearchSourcesLayer.transform_func and select_src_beams). I think we should refactor the code, such that there is one utility function, which translates beams taking a sequence of search choices, and SelectSearchSourcesLayer should use this function.

But there is only one such place. This is what the `SelectSearchSourcesLayer` does. You basically never should call `select_src_beams` explicitly, this should be all via `SelectSearchSourcesLayer`. `SelectSearchSourcesLayer.transform_func` was just an export of that functionality.

albertz:
> It is a bit vague how exactly the latest search choice in the recurrent unit should be determined.

Why? This is perfectly defined currently. See `_opt_search_resolve`, where we define `latest_layer_choice`.

> I now calculate it globally, comparing all search choices of needed_outputs. 

This would be wrong. E.g. if there are two separate choices C1 and C2, which are not dependent at all on each other. Then there is not a single global search choice. Every single layer can basically have it's own latest layer choice.

albertz:
Btw, in any case, irrelevant how you fix it (#225), you should also add a simple test case here which fails before the fix, but works after the fix.

TFNetworkRecLayer.py
patrick-wilken: Yeah, I was aware of this, but the alternative here would be to get the "prev:_latest_choice_layer", transform it to the current frame with `translate_to_this_search_beam` and then never use the output, but only use `transform_function` to transform `dyn_seq_len`. This didn't feel right either... Or is there a better way?

patrick-wilken:
> This would be wrong. E.g. if there are two separate choices C1 and C2, which are not dependent at all on each other. Then there is not a single global search choice. Every single layer can basically have it's own latest layer choice.

Oh wow, true, but this makes it complicated. So then there is no such thing as a single beam of seq_lens which can be used to set size_placeholder of the outputs as we currently do it in `_construct_output_layers_moved_out`...

albertz:
Even with multiple independent choices, there is still a single ending condition (must be), which depends (or might depend) on one of the choices. And seq len relates to the end flag.

patrick-wilken:
Rewrote basically the whole PR and added test cases.

TFNetworkRecLayer.py
```diff
@@ -2350,33 +2351,34 @@ def cond(i, net_vars, acc_tas, seq_len_info=None):
...
     """
     import os
     from TFUtil import nd_indices, assert_min_tf_version, expand_dims_unbroadcast, get_valid_scope_name_from_str
     from TFUtil import get_shape_dim
-    from TFNetwork import LayerNotFound
     rec_layer = self.parent_rec_layer
     try:
       layer = self.net.get_layer(layer_name)
     except LayerNotFound:  # layer is not inside loop
       return acc_ta, None
```
albertz: This is wrong now.

patrick-wilken:
Hmm, somehow I cannot add inline comments at the moment...

What I do now is keep the sequence lengths on the end layer beam and only translate the beam afterwards to the different outputs. This allows different beams for different outputs.

The part I added is in part copy pasted from `search_resolve_body`, maybe we should avoid this code duplication.

The test `test_attention_two_dependent_targets()` fails before this commit. The other test fails before 0e23bcd.

TFNetworkRecLayer.py
```diff
@@ -2834,19 +2879,20 @@ def get_loop_acc_layer(name):
...
-          layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars)
+        acc_ta, search_choices, resolved_seq_len = self._opt_search_resolve(
+          layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars, seq_len=seq_len)
         output = self.layer_data_templates[name].output.copy_template_adding_time_dim(time_dim_axis=0)
         output.beam = search_choices.get_beam_info() if search_choices else None
+        max_len = tf.reduce_max(resolved_seq_len) if resolved_seq_len is not None else None
         # We should have accumulated it.
         output.placeholder = tensor_array_stack(acc_ta, stop=max_len)  # e.g. (time,batch,dim)
-        output.size_placeholder = {0: seq_len}
+        output.size_placeholder = {0: resolved_seq_len}
```
albertz: Can it be `None`? (You check for that above.) If yes, than this is wrong. If no, then the code above does not make sense.

TFNetworkRecLayer.py
```diff
@@ -2431,11 +2433,56 @@ def get_choice_seq(choice_base):
...
+
+      # Separate batch and beam dims
+      seq_len_beam_size = end_layer.output.beam.beam_size if end_layer else rec_layer.output.beam.beam_size
+      seq_len = tf.reshape(
+        seq_len, [batch_dim, seq_len_beam_size],
+        name="split_batch_beam")
+
+      if end_layer:
+        end_layer_choice = self.net.get_search_choices(src=end_layer)
+        assert end_layer_choice in choice_seq_in_frame,\
```
albertz: Don't use `\`. Rather `assert ..., (\n...)`

TFNetworkRecLayer.py
```diff
@@ -8,6 +8,7 @@
 import tensorflow as tf
 import typing
 from tensorflow.python.ops.nn import rnn_cell
+from TFNetwork import LayerNotFound
```
patrick-wilken: I imported it here, as I use it twice now.

tests/test_TFEngine.py
```diff
@@ -1020,6 +1020,149 @@ def make_net_dict(task):
...
+  data.init_seq_order()
+
+  dec_time = DimensionTag(kind=DimensionTag.Types.Spatial, description="dec time")
+
+  config = Config()
+  config.update({
+    "model": "%s/model" % _get_tmp_dir(),
+    "batch_size": 5000,
+    "max_seqs": 2,
+    "extern_data": {"data": {"dim": n_data_dim, "sparse": False},
```
patrick-wilken: (Ah, works again)
This should be an issue on it's own. When using MetaDataset I have to mark the time dims as equal, also for in real configs. Is this expected?

TFNetworkRecLayer.py
```diff
@@ -2431,11 +2433,56 @@ def get_choice_seq(choice_base):
...
+
+        choice_beams = initial_beam_choices
+        if choices_seq_until_end_layer:  # otherwise end layer already corresponds to latest choice
+          for choice_ in choices_seq_until_end_layer:
+            assert_min_tf_version((1, 1), "gather_nd")
+            idxs_exp = nd_indices(choice_beams)  # (batch, beam_out, 2) -> (batch idx, beam idx)
+            src_choice_beams = self.final_acc_tas_dict["choice_%s" % choice_.name].read(
+              -1, name="ta_read_choice")  # (batch, beam) -> beam_in idx
+            assert src_choice_beams.get_shape().ndims == 2
+            choice_beams = tf.gather_nd(src_choice_beams, idxs_exp)  # (batch, beam_out)
```
albertz: I don't think it's a good idea to replicate the logic of `SelectSearchSourcesLayer`. Although it might be a bit tricky to resolve that in an easy and clean way (without making it too abstract/complicated).

I also don't think it's a good idea to use `tf.gather_nd` and `nd_indices` directly, and replicate the logic of `select_src_beams`.

TFNetworkRecLayer.py
patrick-wilken: Ah, it cannot be None. It's None in `get_output` if there are no layers in the loop, but then we would never get here.

TFNetworkRecLayer.py
```diff
@@ -2350,33 +2351,34 @@ def cond(i, net_vars, acc_tas, seq_len_info=None):
...
     which is determined by the latest search choice given by the layer.
     This assumes that we also have `self.final_acc_tas_dict["choice_%s" % choice_base.name]` available.
+    In addition, we resolve the sequence lengths (whose beams correspond to the end layer) to the final search choices.
 
     :param str layer_name:
     :param tf.TensorArray acc_ta: accumulated outputs of that layer
     :param final_net_vars:
-    :return: (new acc_ta, final search choices)
-    :rtype: (tf.TensorArray,SearchChoices|None)
+    :param tf.Tensor|None seq_len: shape (batch * beam,)
```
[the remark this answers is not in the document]
patrick-wilken: There are two cases: If `have_known_seq_len`, we get the beam from the "output" layer, otherwise from the "end" layer. I will document it.

TFNetworkRecLayer.py
patrick-wilken: I just made it work using `select_src_beams` (this could also be done in `transform()`). But using `SelectSearchSourcesLayer` is not easily possible, because we need to get the search choices from the TensorArray.

tests/test_TFEngine.py
albertz: Yes, this is expected. How/why should it assume that they are the same?

TFNetworkRecLayer.py
```diff
@@ -2834,19 +2879,20 @@ def get_loop_acc_layer(name):
       with tf.name_scope(self.layer_data_templates[name].layer_class_type.cls_get_tf_scope_name(name)):
         inner_layer = self.net.get_layer(name)
         acc_ta = loop_accumulated["output_%s" % name]
-        acc_ta, search_choices = self._opt_search_resolve(
-          layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars)
+        acc_ta, search_choices, resolved_seq_len = self._opt_search_resolve(
+          layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars, seq_len=seq_len)
         output = self.layer_data_templates[name].output.copy_template_adding_time_dim(time_dim_axis=0)
         output.beam = search_choices.get_beam_info() if search_choices else None
+        max_len = tf.reduce_max(resolved_seq_len) if resolved_seq_len is not None else None
```
albertz: You said it cannot be `None`. Then this does not make sense.

TFNetworkRecLayer.py
albertz: Where? I still see `tf.gather_nd` in that code.

patrick-wilken: Now it's pushed.

TFNetworkRecLayer.py
patrick-wilken: Fixed.

tests/test_TFEngine.py
patrick-wilken: So the idea is that each data key has its own time dim unless configured otherwise, right? But it's quite common that multiple data keys are on the same time dim (at least for me ;) ). This should be easier to configure, without importing DimensionTag into the config.

TFNetworkRecLayer.py
albertz: Still wrong.

TFNetworkRecLayer.py
```diff
@@ -2431,11 +2434,45 @@ def get_choice_seq(choice_base):
...
+        src_choice_beams = self.final_acc_tas_dict["choice_%s" % choice_.name].read(
+          max_seq_len - 1, name="ta_read_choice")  # (batch, beam) -> beam_in idx
+        seq_len = select_src_beams(seq_len, src_choice_beams)
+    else:
+      # Here we don't need to resolve anything, as the sequence length is the same for all hyps in the beam.
+      # However, beam size for the current output may be different from the "output" layer.
+      # Therefore take the first len in the beam and tile it to the desired beam size.
+
+      # Separate batch and beam dims
+      seq_len_beam_size = end_layer.output.beam.beam_size if end_layer else rec_layer.output.beam.beam_size
```
albertz: The whole branch is for `not end_layer`, so I don't understand the `if` here.

TFNetworkRecLayer.py
patrick-wilken: I moved the import to line 11, because I need it here and in line 2439. I can also change it back to local imports, but it's working.

TFNetworkRecLayer.py
patrick-wilken: Yep, copy-paste error, fixed it.

TFNetworkRecLayer.py
```diff
@@ -2834,21 +2868,22 @@ def get_loop_acc_layer(name):
...
           if self.parent_rec_layer.output.beam:
             # seq_len should have already a beam, same as rec_layer.output
             assert output.beam == self.parent_rec_layer.output.beam
           else:
-            output.size_placeholder[0] = tile_transposed(seq_len, axis=0, multiples=output.beam.beam_size)
-            if time_dim_tag:
-              time_dim_tag.set_tag_on_size_tensor(output.size_placeholder[0])
+            output.size_placeholder[0] = tile_transposed(resolved_seq_len, axis=0, multiples=output.beam.beam_size)
+        if time_dim_tag:
+          time_dim_tag.set_tag_on_size_tensor(output.size_placeholder[0])
```
patrick-wilken: I was missing this change to set the tag again after a new seq_len tensor is created. Now tests are passing.

TFNetworkRecLayer.py
albertz: I'm not speaking about the `import`. I mean exactly the line where I put the comment. The `return` is wrong.

TFNetworkRecLayer.py
```diff
@@ -2834,21 +2868,22 @@ def get_loop_acc_layer(name):
...
+        output.size_placeholder = {0: resolved_seq_len}
         if output.beam:
           if self.parent_rec_layer.output.beam:
             # seq_len should have already a beam, same as rec_layer.output
             assert output.beam == self.parent_rec_layer.output.beam
           else:
-            output.size_placeholder[0] = tile_transposed(seq_len, axis=0, multiples=output.beam.beam_size)
-            if time_dim_tag:
-              time_dim_tag.set_tag_on_size_tensor(output.size_placeholder[0])
+            output.size_placeholder[0] = tile_transposed(resolved_seq_len, axis=0, multiples=output.beam.beam_size)
```
albertz: Does this make sense? If you have logic for resolving `seq_len` in `_opt_search_resolve`, shouldn't it be tiled correctly already?
In any case, the logic should not be split; `_opt_search_resolve` should correctly resolve `seq_len`.

TFNetworkRecLayer.py
patrick-wilken: Ahh, fixed.

TFNetworkRecLayer.py
patrick-wilken: I think you're right, it is not needed anymore. The assert
```
assert output.beam == self.parent_rec_layer.output.beam
```
is definitively wrong now for the case that different outputs have different latest_choices and therefore maybe different beam sizes.

As far as I understand the case distinction added in e442199, it is for the case that there is a beam that is decided inside the rec layer, right? I think with the seq_lens of each output having the beam of its latest_search_choice this case should be handled now correctly implicitly.

albertz:
Sorry, I made too much changes in the meantime, and there is a small conflict now. Can you rebase? Thanks!

patrick-wilken:
I tried to rebase, but please double check.

TFNetworkRecLayer.py
```diff
@@ -2838,33 +2872,23 @@ def get_loop_acc_layer(name):
       with tf.name_scope(self.layer_data_templates[name].layer_class_type.cls_get_tf_scope_name(name)):
         inner_layer = self.net.get_layer(name)
         acc_ta = loop_accumulated["output_%s" % name]
-        if inner_layer.get_search_choices() is None:
-          search_choices = None
-        else:
-          acc_ta, search_choices = self._opt_search_resolve(
-            layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars)
+        acc_ta, search_choices, resolved_seq_len = self._opt_search_resolve(
+          layer_name=name, acc_ta=acc_ta, final_net_vars=final_net_vars, seq_len=seq_len)
```
patrick-wilken: Why did you add the if statement here? Shouldn't `_opt_search_resolve` return `search_choices =None` and the original `acc_ta` in that case?

TFNetworkRecLayer.py
```diff
@@ -2435,11 +2437,45 @@ def get_choice_seq(choice_base):
...
+      # Here we don't need to resolve anything, as the sequence length is the same for all hyps in the beam.
+      # However, beam size for the current output may be different from the "output" layer.
+      # Therefore take the first len in the beam and tile it to the desired beam size.
+
+      # Separate batch and beam dims
+      seq_len_beam_size = rec_layer.output.beam.beam_size
+      seq_len = tf.reshape(seq_len, [batch_dim, seq_len_beam_size], name="split_batch_beam")
+
+      seq_len = seq_len[:, 0:1]
+      seq_len = tf.tile(seq_len, [1, latest_beam_size], name="resize_seq_len_beam")
```
patrick-wilken: I also updated this. Tiling was incorrect before, luckily @talkhouli ran into this ':D

TFNetworkRecLayer.py
albertz: No. In `_opt_search_resolve`, we use `layer_choice = self.net.get_search_choices(src=layer)`, which might not be the same, as I derived `LayerBase.get_search_choices()` for some layers. Although I think I undid that later. But maybe we can just use `layer_choice = layer. get_search_choices()` inside `_opt_search_resolve`.

TFNetworkRecLayer.py
```diff
@@ -2352,35 +2352,38 @@ def cond(i, net_vars, acc_tas, seq_len_info=None):
...
-      return acc_ta, None
-    if layer_choice.search_choices.keep_raw:
-      return acc_ta, layer_choice.search_choices
+      return acc_ta, None, seq_len
+    search_choices = layer.get_search_choices()
+    if not search_choices:
+      return acc_ta, None, seq_len
+    if search_choices.keep_raw:
+      return acc_ta, search_choices, seq_len
+    layer_choice = search_choices.owner
```
patrick-wilken: Like this? I checked that search choices are the same as before for all test cases...

albertz: Yes exactly. Yes I think I once had a case where this was more important, but changed that later on. But still, maybe this is better.

PR #230, opened 2019-10-04
Dataset, add sort_bin_shuffle and sort_bin_shuffle_x2 seq ordering

kazuki-irie:
sort_bin_shuffle first sorts sequences by length, creates bins, and
shuffles the bins.

sort_bin_shuffle_x2 does an extra shuffling within each bin after
sort_bin_shuffle.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -377,6 +377,38 @@ class Dataset(object):
     elif self.seq_ordering == "sorted_reverse":
       assert get_seq_len
       seq_index.sort(key=get_seq_len, reverse=True)  # sort by length, in reverse, starting with longest
+    elif self.seq_ordering.startswith('sort_bin_shuffle'):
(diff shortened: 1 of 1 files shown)


Dataset.py
```diff
@@ -377,6 +377,38 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
     elif self.seq_ordering == "sorted_reverse":
       assert get_seq_len
       seq_index.sort(key=get_seq_len, reverse=True)  # sort by length, in reverse, starting with longest
+    elif self.seq_ordering.startswith('sort_bin_shuffle') or self.seq_ordering.startswith('sort_bin_shuffle_x2'):
```
albertz: The `or` here is redundant.
```suggestion
    elif self.seq_ordering.startswith('sort_bin_shuffle'):
```
is enough

kazuki-irie: Ah yes, thank you.

PR #238, opened 2019-10-23
added multi zip-file support to OggZipDataset

JackTemaki:


GeneratingDataset.py
```diff
@@ -2677,6 +2681,24 @@ def _collect_data(self):
     else:
       assert self.feature_extractor, "feature extraction is enabled, but no audio files are specified"
       assert isinstance(first_entry["seq_name"], str)
+    # add index to data list
+    for entry in data:
+      entry['index'] = index
```
albertz: This is not a good idea, for multiple reasons:
* `"index"` is too ambiguous. It could mean multiple things. I first assumed is the seq index, but it's not of course.
* `"index"` looks like a key which comes from the user, but it is an auxiliary key which you add here. There should be some distinction; e.g. you could use a prefix `_`.
* It's not a good idea to have this loop here. This makes it slow for lots of entries. Basically I would avoid doing this at all.

What you actually need is the mapping ref_seq_idx -> name/path/zip index. You can simply calculate this mapping, by saving how much entries every zip has. Then you don't need this.

JackTemaki: I see that the name was not chosen well, but I do not see the speed issue. I think it will be negligible in comparison to the time the `eval` command will need to parse the data. But if you want I can go for an alternative.

JackTemaki:
I am not saying that it is slow, I am just talking about the relative change. Or is this somehow optimized to be faster than O(n)?

data parsing: 10.827578 seconds
loop: 0.032600 seconds

GeneratingDataset.py
albertz: Hm, you are right. I did not expect this. (And I think that there is a lot of chance for CPython to optimize this. I think the `eval` should actually be faster.)
Ok, then maybe leave the logic like it is, but change the key.

albertz:
Well, it cannot be faster than O(n) (it has to read through the whole string). But it could be really fast. The actual parsing is actually fast, I assume (I see no reason why parsing Python source code is complicated). Probably there is quite some overhead happening in `eval` though (not exactly sure what, though). And then, maybe the on-the-fly `list` and `dict` creation is not well optimized. E.g. it could see from the parsed code how much entries there are in the list, and allocate a list which is exactly big enough, so no resize needed later. I actually would wonder if that is not the case. Maybe creating a new dict is slow, though. But also here I thought there might be a few options to optimize it when you know in advance all keys/values. The whole `eval` actually could in principle be optimized away, and all the objects directly created, and just returned. I mean, there is no actual Python bytecode which needs to be executed (in principle). And exactly this is different for your loop, where you always have Python bytecode which is interpret.

GeneratingDataset.py
```diff
@@ -2651,21 +2652,25 @@ def __init__(self, path, audio, targets,
...
+  def _collect_data_part(self, zip_index):
     """
-    :return: entries
+    collect all the entries of a single zip-file or txt file
+    :param int zip_index: index of the zip-file in self._zip_files, unused when loading without zip
+    :return: data entries
     :rtype: list[dict[str]]
     """
-    data = eval(self._read("%s.txt" % self._name))  # type: typing.List[typing.Dict[str]]
+    import time
```
albertz: Remove unused imports (check PyCharm warnings).

JackTemaki: unused import is unfortunately not displayed as a warning in my case, I am still searching for where to enable it.

PR #239, opened 2019-10-24
SelfAttentionLayer, share_kv option

kazuki-irie:
Has SelfAttentionLayer already too many options? or is this still acceptable?

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -5480,7 +5480,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
   recurrent = True
 
   def __init__(self, num_heads, total_key_dim,
-               key_shift=None,
+               key_shift=None, share_kv=False,
                forward_weights_init="glorot_uniform", attention_dropout=0.0,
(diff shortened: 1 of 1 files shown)


albertz:
Hm, I don't like it too much... I mean, this specific option on its own is maybe ok, but it really becomes too complicated.
Two maybe better options:
* How much effort do you think it is to rewrite your configs that it doesn't use `SelfAttentionLayer` at all? What we discussed.
* You can have a copy of this modified/extended layer in your config, call it `ExtendedSelfAttentionLayer` or so.

kazuki-irie:
> ... it really becomes too complicated.

I understand.

> How much effort do you think it is to rewrite your configs that it doesn't use `SelfAttentionLayer` at all? What we discussed.

I haven't yet checked what exactly is missing for that. I will take a look.

Thanks for the feedback!

Closing.

albertz:
We can do it together when I'm back.

Best,
Albert


Kazuki Irie <notifications@github.com> schrieb am Fr., 25. Okt. 2019, 20:30:

PR #240, opened 2019-10-29
change dtype of ActivationLayer output according to placeholder

mennetob:
Some activation functions have a different `dtype` at the output compared to the input.
E.g. `tf.abs` converts `tf.complex64` to `tf.float32`.

Thus the `ActivationLayer` should change the `dtype` of the `self.output` accordingly

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -2314,6 +2314,7 @@ class ActivationLayer(CopyLayer):
     else:
       self.output_before_activation = OutputWithActivation(x)
     self.output.placeholder = self.output_before_activation.y
+    self.output.dtype = self.output_before_activation.y.dtype.base_dtype.name
 
 
 class BatchNormLayer(CopyLayer):


TFNetworkLayer.py
```diff
@@ -2314,6 +2314,7 @@ def __init__(self, activation, **kwargs):
     else:
       self.output_before_activation = OutputWithActivation(x)
     self.output.placeholder = self.output_before_activation.y
+    self.output.dtype = self.output_before_activation.y.dtype.base_dtype.name
```
albertz: The `__init__` should never change the type/format of `self.output`. This should be done via `get_out_data_from_opts`. The `dtype` (and shape, etc) must be known before any computation graph is actually constructed (via `__init__`).

albertz:
Btw, it would also be good to add two test cases for this: One for some standard activation function (`tanh` or whatever), another one for e.g. `tf.abs`.

albertz:
Are you going to update this? Otherwise I will close it for now.

albertz:
Closing in favor of #312.

PR #241, opened 2019-11-06
TF model paths are allowed to have .meta postfix

JackTemaki:
Sisyphus can only handle existing paths, so .meta has to be part of the path.

### DIFF
--- a/EngineBase.py
+++ b/EngineBase.py
@@ -73,10 +73,14 @@ class EngineBase(object):
       assert start_epoch >= 1
 
     load_model_epoch_filename = config.value('load', '')
+    if load_model_epoch_filename.endswith(".meta"):
+      load_model_epoch_filename = load_model_epoch_filename[:-len(".meta")]
     if load_model_epoch_filename:
       assert os.path.exists(load_model_epoch_filename + get_model_filename_postfix())
 
     import_model_train_epoch1 = config.value('import_model_train_epoch1', '')
+    if import_model_train_epoch1.endswith(".meta"):
+      import_model_train_epoch1 = import_model_train_epoch1[:-len(".meta")]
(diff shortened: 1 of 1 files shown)


EngineBase.py
```diff
@@ -74,11 +74,15 @@ def get_epoch_model(cls, config):
 
     load_model_epoch_filename = config.value('load', '')
     if load_model_epoch_filename:
-      assert os.path.exists(load_model_epoch_filename + get_model_filename_postfix())
+      if not load_model_epoch_filename.endswith(get_model_filename_postfix()):
+        load_model_epoch_filename = load_model_epoch_filename + get_model_filename_postfix()
+      assert os.path.exists(load_model_epoch_filename)
```
albertz: The variable `load_model_epoch_filename` should **not** contain the filename postfix. That variable might get returned by this function `get_epoch_model`, and that is being used for multiple purpose, in many places. And many other code expects that it gets the model name there (as path, but **excluding** any filename postfix).

So, better would be to just remove the postfix in case it is there.

EngineBase.py
```diff
@@ -74,11 +74,15 @@ def get_epoch_model(cls, config):
...
 
     import_model_train_epoch1 = config.value('import_model_train_epoch1', '')
     if import_model_train_epoch1:
-      assert os.path.exists(import_model_train_epoch1 + get_model_filename_postfix())
+      if not import_model_train_epoch1.endswith(get_model_filename_postfix()):
+        import_model_train_epoch1 = import_model_train_epoch1 + get_model_filename_postfix()
+      assert os.path.exists(import_model_train_epoch1)
```
albertz: Same problem here.

JackTemaki:
This is how I had it in my old codebase, just removing it.

PR #243, opened 2019-11-12
ReduceLayer, allow reduce_mean of time dimension without flattening

vieting:
A time mask is applied to take different sequence lengths into account, the elements are divided by the sequence length and then tf.reduce_sum is applied. Flattening is removed.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -4735,7 +4735,7 @@ class ReduceLayer(_ConcatInputLayer):
     x_ = x.placeholder
     # Check if we should ignore some frames, e.g. via masking.
     if use_time_mask:
-      if f in (tf.reduce_sum, tf.reduce_min, tf.reduce_max):
+      if f in (tf.reduce_sum, tf.reduce_min, tf.reduce_max) or (f == tf.reduce_mean and axes == [x.time_dim_axis]):
         # For sum, the fastest and simplest way is masking.
         for axis in axes:
           if axis == x.batch_dim_axis:
@@ -4750,12 +4750,19 @@ class ReduceLayer(_ConcatInputLayer):
           mask = tf.logical_and(mask, tf.ones_like(x_, dtype=mask.dtype))
 
           replacement_value = {
+            tf.reduce_mean: tf.zeros_like(x.placeholder),
             tf.reduce_sum: tf.zeros_like(x.placeholder),
             tf.reduce_min: tf.zeros_like(x.placeholder) + x.placeholder.dtype.max,
             tf.reduce_max: tf.zeros_like(x.placeholder) + x.placeholder.dtype.min
           }
 
           x_ = tf.where(mask, x_, replacement_value[f], "x_masked_axis_%i" % axis)
+          if f == tf.reduce_mean:
+            seq_len_bc = tf.reshape(
+              x.get_sequence_lengths(),
+              [1 if (i != x.batch_dim_axis) else -1 for i in range(x.batch_ndim)])  # (1,..B..,1)
+            x_ = x_ / tf.cast(seq_len_bc, tf.float32)
+            f = tf.reduce_sum
       elif f == tf.reduce_mean:
         # Flattening.
         if x.time_dim_axis in axes:
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -3065,6 +3065,63 @@ def test_CrossEntropyLoss_masked_inf_fake_upper_bound():
       last_var_v = var_v
 
 
+def test_reduce_mean_in_time():
+  with make_scope() as session:
+    n_out = 5
+    config = Config({
+      "debug_print_layer_output_template": True,
+      "extern_data": {
+        "data": {"dim": n_out},
+      }})
+    net = TFNetwork(config=config, train_flag=True)
+    net.construct_from_dict({
+      "output": {"class": "reduce", "mode": "mean", "axis": "T", "from": ["data"]}
+    })
+    session.run(tf.global_variables_initializer())
+    out = net.layers["output"].output.placeholder
+    n_batch = 3
+    max_seq_len = 10
+    feed = make_feed_dict(net.extern_data.data.values(), n_batch=n_batch, n_time=max_seq_len, same_time=True)
+    v = session.run(out, feed_dict=feed)
+    input_len = feed[net.extern_data.data["data"].size_placeholder[0]]
+    input_data = feed[net.extern_data.data["data"].placeholder]
+
+    ref = numpy.zeros([n_batch, n_out])
+    for batch, seq_len in enumerate(input_len):
+      ref[batch, :] = numpy.mean(input_data[batch, :seq_len, :], axis=0)
+
+    numpy.testing.assert_allclose(ref, v, rtol=1e-5)
+
+
+def test_reduce_mean_batch_time():
+  with make_scope() as session:
+    n_out = 5
+    config = Config({
+      "debug_print_layer_output_template": True,
+      "extern_data": {
+        "data": {"dim": n_out},
+      }})
+    net = TFNetwork(config=config, train_flag=True)
+    net.construct_from_dict({
+      "output": {"class": "reduce", "mode": "mean", "axis": ["B", "T"], "from": ["data"]}
+    })
+    session.run(tf.global_variables_initializer())
+    out = net.layers["output"].output.placeholder
+    n_batch = 3
+    max_seq_len = 10
+    feed = make_feed_dict(net.extern_data.data.values(), n_batch=n_batch, n_time=max_seq_len, same_time=True)
+    v = session.run(out, feed_dict=feed)
+    input_len = feed[net.extern_data.data["data"].size_placeholder[0]]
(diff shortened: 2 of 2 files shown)


tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
 
+def test_reduce_mean_in_time():
+  with make_scope() as session:
+    import numpy as np
+    n_out = 5
+    config = Config({
+      "debug_print_layer_output_template": True,
+      "extern_data": {
+        "data": {"dim": n_out},
+        "classes": {"dim": n_out},
```
albertz: You don't need classes. Do not add it then.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
       last_var_v = var_v
 
 
+def test_reduce_mean_in_time():
+  with make_scope() as session:
+    import numpy as np
```
albertz: Just use `numpy` directly. No need for the `np` alias. And `numpy` is already imported.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+    config = Config({
+      "debug_print_layer_output_template": True,
+      "extern_data": {
+        "data": {"dim": n_out},
+        "classes": {"dim": n_out},
+      }})
+    net = TFNetwork(config=config, train_flag=True)
+    net.construct_from_dict({
+      "reduce_mean": {"class": "reduce", "mode": "mean", "axis": "T", "keep_dims": True, "from": ["data"],
+                      "is_output_layer": True}
```
albertz: If you call it `"output"` directly, you would not need `"is_output_layer": True`. But as you like.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+    })
+    session.run(tf.global_variables_initializer())
+    out = net.layers["reduce_mean"].output.placeholder
+    n_batch = 3
+    max_seq_len = 10
+    feed = make_feed_dict(net.extern_data.data.values(), n_batch=n_batch, n_time=max_seq_len, same_time=True)
+    v = session.run(out, feed_dict=feed)
+    input_len = feed[list(feed.keys())[0]]
```
albertz: This is not a good idea. The order in `feed.keys()` is non-deterministic (if you have more than 1 entry). But anyway, you assume it is `"data"`, right? So just make it more explicit, sth like `feed[net.extern_data.data["data"].size_placeholder[0]]`.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+    input_data = feed[list(feed.keys())[1]]
```
albertz: Same here.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+
+    ref = np.zeros([n_batch, n_out])
+    for batch, len in enumerate(input_len):
```
albertz: Do not overwrite Python builtins (`len`). (PyCharm should warn about that, or not?)

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+      ref[batch, :] = np.mean(input_data[batch, :len, :], axis=0)
+    ref = np.expand_dims(ref, axis=1)
```
albertz: Why don't you just use `"keep_dims": False`?

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,71 @@ def mask_func(source, **kwargs):
...
+    ref = ref.mean(axis=0, keepdims=True)
```
albertz: This is not correct. This is the mean of the mean. Which is not the same as the mean over all frames directly. I wonder why you test still passes. Maybe `rtol=1e-5` is too large? Check that.

tests/test_TFNetworkLayer.py
vieting: I had it in the network config with `keep_dims` before, probably it doesn't really matter. I changed it anyway.

tests/test_TFNetworkLayer.py
vieting: You are right. I guess then it is easier to just keep the flattening for cases where `axes` contains the time dimension plus other dimensions. What do you think?

Regarding the tolerance, `rtol=1e-6` still passes, `1e-7` does not. I took `1e-5` from the other usage of `assert_allclose()` in this file. Not sure what a sensible threshold is. Anyway, it should not be related to calculating the mean of the mean because currently that is also what is done in the layer.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,63 @@ def mask_func(source, **kwargs):
...
+    max_seq_len = 10
+    feed = make_feed_dict(net.extern_data.data.values(), n_batch=n_batch, n_time=max_seq_len, same_time=True)
+    v = session.run(out, feed_dict=feed)
+    input_len = feed[net.extern_data.data["data"].size_placeholder[0]]
+    input_data = feed[net.extern_data.data["data"].placeholder]
+
+    ref = numpy.zeros([n_batch, n_out])
+    for batch, seq_len in enumerate(input_len):
+      ref[batch, :] = numpy.mean(input_data[batch, :seq_len, :], axis=0)
+    ref = ref.mean(axis=0)
```
albertz: Still the problem as explained earlier.

tests/test_TFNetworkLayer.py
albertz: > You are right. I guess then it is easier to just keep the flattening for cases where `axes` contains the time dimension plus other dimensions. What do you think?

Are you talking about the test, or the layer code?

I actually did not realize, the whole logic in the layer for `tf.reduce_mean` is wrong, as you do it the same way there.

Yes,  maybe the flattening is nicer and simpler. I also thought about numerical stability, which is probably better that way.

> Regarding the tolerance, `rtol=1e-6` still passes, `1e-7` does not. I took `1e-5` from the other usage of `assert_allclose()` in this file. Not sure what a sensible threshold is. Anyway, it should not be related to calculating the mean of the mean because currently that is also what is done in the layer.

I think it's ok. I thought that you do it correct in the layer, but wrong here, so I wondered that it did not catch that. But I see now that you do it just as wrong in the layer.

TFNetworkLayer.py
```diff
@@ -4750,21 +4750,23 @@ def __init__(self, mode, axes=None, axis=None, keep_dims=False, enforce_batch_di
...
-          assert x.batch_dim_axis in axes, "not yet implemented otherwise"
-          axes = [a if (a < x.time_dim_axis) else (a - 1)
-                  for a in axes if a != x.time_dim_axis]
-          x = x.copy_time_flattened()
-          x_ = x.placeholder
+          if f == tf.reduce_mean:
+            seq_len_bc = tf.reshape(
+              x.get_sequence_lengths(),
+              [1 if (i != x.batch_dim_axis) else -1 for i in range(x.batch_ndim)])  # (1,..B..,1)
+            x_ = x_ / tf.cast(seq_len_bc, tf.float32)
```
albertz: This should not be done multiple times (if this code is run for multiple axes). And this only really make sense if `axis == x.time_dim_axis`.
I think I really would write specific code for `axes == [x.time_dim_axis]`. Otherwise this is a bit too unreadable.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,63 @@ def mask_func(source, **kwargs):
...
+    max_seq_len = 10
+    feed = make_feed_dict(net.extern_data.data.values(), n_batch=n_batch, n_time=max_seq_len, same_time=True)
+    v = session.run(out, feed_dict=feed)
+    input_len = feed[net.extern_data.data["data"].size_placeholder[0]]
+    input_data = feed[net.extern_data.data["data"].placeholder]
+
+    input_data_masked = numpy.copy(input_data)
+    for batch, seq_len in enumerate(input_len):
+      input_data_masked[batch, seq_len:, :] = numpy.nan
+    ref = numpy.nanmean(input_data_masked, axis=(0, 1))
```
albertz: Ah, interesting, I did not know there is such a function.

TFNetworkLayer.py
albertz: It's still not so nice. If you read only this code, and not the check a couple of lines above (`axes == [x.time_dim_axis]`), you would think this is a bug. Code should be locally bug-free, as far as possible, e.g. when you remove the check `axes == [x.time_dim_axis]` above, this should still be correct. (This is for easier readability, but also for maintainability, e.g. when changes to the code are made, e.g. someone might think, why do we have this check `axes == [x.time_dim_axis]` there, it should also work without, and then removes it...)
Anyway, I only see the diff here, and don't really know whether other ways to write this are more clear, so let's leave it like that for now.

PR #244, opened 2019-11-12
Added MLM (BERT):  ProbabilisticMaskLayer and GatherPositionsLayer

ArneNx:
I implemented two layers that are meant to be used jointly to obtain the BERT-style Masked Language Model (MLM) on the fly without relying on a fixed preprocessing.

`ProbabilisticMaskLayer`: masks a given sequence and has as outputs the masked sequence, the indices which were masked and the labels for each of those positions
`GatherPositionsLayer`: takes a sequence and a tensor of indices (positions from `ProbabilisticMaskLayer`) and selects the contents from those positions in the primary input

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -32,7 +32,7 @@ class LayerBase(object):
               " docstring, document the args! "
               super(YourOwnLayer, self).__init__(**kwargs)
               # Now we need to set self.output, which must be of type :class:`Data`.
-              # It is set at this point to whatever we got from `selfget_out_data_from_opts()`,
+              # It is set at this point to whatever we got from `self.get_out_data_from_opts()`,
               # so it is enough if we set self.output.placeholder and self.output.size_placeholder,
               # but we could also reset self.output.
               self.output.placeholder = self.input_data.placeholder + 42  # whatever you want to do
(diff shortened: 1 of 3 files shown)


TFNetworkLayer.py
```diff
@@ -9411,3 +9411,344 @@ def get_layer_class_name_list():
...
+  """
+  Gathers vectors at the specified positions on time axis.
+  """
+  layer_class = "gather_positions"
+
+  def __init__(self, **kwargs):
+    """
+    Gathers input data at given positions.
+    Output has shape (B,max(num_positions),None) and is zero-padded for each sequence.
+    Inputs: [LayerBase,LayerBase] [input_data,positions]: with
```
albertz: You should not assume specific meaning in the order or length of `self.sources`. You should rather have specific named separate arguments, like `positions`.

TFNetworkLayer.py
```diff
@@ -9411,3 +9411,344 @@ def get_layer_class_name_list():
...
+    input_data: shape=(B,T,None)
```
albertz: What is `None` here? You mean some dimension `D`?

TFNetworkLayer.py
```diff
@@ -9411,3 +9411,344 @@ def get_layer_class_name_list():
...
+class GatherPositionsLayer(LayerBase):
+  """
+  Gathers vectors at the specified positions on time axis.
+  """
+  layer_class = "gather_positions"
+
+  def __init__(self, **kwargs):
+    """
+    Gathers input data at given positions.
+    Output has shape (B,max(num_positions),None) and is zero-padded for each sequence.
```
albertz: Again, what is `None` here? The same dim `D` as in `input_data`?

TFNetworkLayer.py
```diff
@@ -9411,3 +9411,344 @@ def get_layer_class_name_list():
   if not _LayerClassDictInitialized:
     _init_layer_class_dict()
   return sorted(_LayerClassDict.keys())
+
+
+class GatherPositionsLayer(LayerBase):
+  """
+  Gathers vectors at the specified positions on time axis.
+  """
+  layer_class = "gather_positions"
```
albertz: This layer looks very much like the `GatherNdLayer`. What is the difference? This should be documented.
Also, if it is basically the same, then you should remove this, and just use `GatherNdLayer` instead. Or maybe it is easy to extend `GatherNdLayer` in such a way that you can use it.

Remember, the layers in Returnn should be simple, straight-forward, and not be specific for some specific task. See the Returnn documentation and introduction video lecture.

TFNetworkLayer.py
```diff
@@ -9411,3 +9411,344 @@ def get_layer_class_name_list():
...
+      feature_dim_axis=sources[0].output.feature_dim_axis_or_unspecified,
+      dtype=sources[0].output.dtype)
+    return data
+
+
+class ProbabilisticMaskLayer(LayerBase):
+  """
+  This layer performs MLM (BERT-style) masking on the fly for a given source sequence.
+  """
+  layer_class = "probabilistic_mask"
```
albertz: You should not add layers which are too specific (like this one).
If you want to write it that way, just copy that to your config, but this should not be part of Returnn. This is useful maybe only for a handful of experiments, and will likely be changed again and again, adding new options all the time, and becomes ugly and non-maintainable, until no-one will ever use it anymore. I explain that in the Returnn video lecture.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
     self.most_recent_value = value
     self.callback_count += 1
     return 0
+
+
+def merge_tensor_array_with_padding(arr, row_lengths, feature_dims):
+  """
+  Merges a tensor array of multiple tensors into a single tensor, while padding them to given length
+  :param tf.TensorArray arr: array to merge with tensor i of form (T[i],None)
+  :param tf.Tensor[int32] row_lengths: shape=(B,) length of each row
```
albertz: This is not a valid type.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
     self.most_recent_value = value
     self.callback_count += 1
     return 0
+
+
+def merge_tensor_array_with_padding(arr, row_lengths, feature_dims):
+  """
+  Merges a tensor array of multiple tensors into a single tensor, while padding them to given length
+  :param tf.TensorArray arr: array to merge with tensor i of form (T[i],None)
```
albertz: What is `None`?

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
     self.most_recent_value = value
     self.callback_count += 1
     return 0
+
+
+def merge_tensor_array_with_padding(arr, row_lengths, feature_dims):
+  """
+  Merges a tensor array of multiple tensors into a single tensor, while padding them to given length
```
albertz: Add a newline after that.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
...
+
+  _, arr_pad = tf.while_loop(while_condition, body, [i, arr_pad], back_prop=True)
+  return arr_pad.stack()
+
+
+def sample_without_replacement(logits, k, seed=None):
+  """
+  This is equivalent to sampling `k` times without replacement from a categorial distribution with logits `logits`.
+  Discussion on Github: https://github.com/tensorflow/tensorflow/issues/9260#issuecomment-437875125
+  Details here: https://timvieira.github.io/blog/post/2014/08/01/gumbel-max-trick-and-weighted-reservoir-sampling/
```
albertz: Add a newline here.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
...
+    non_pad_pos = tf.expand_dims(tf.range(row_lengths[i]), -1)
+    arr_i_padded = tf.scatter_nd(non_pad_pos, arr.read(i), [pad_to_length] + feature_dims)
+    arr_pad = arr_pad.write(i, arr_i_padded)
+    return [tf.add(i, 1), arr_pad]
+
+  _, arr_pad = tf.while_loop(while_condition, body, [i, arr_pad], back_prop=True)
+  return arr_pad.stack()
+
+
+def sample_without_replacement(logits, k, seed=None):
```
albertz: I already use the same code inside `ChoiceLayer`. Maybe `ChoiceLayer` should make use of this function then.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
...
+  """
+  This is equivalent to sampling `k` times without replacement from a categorial distribution with logits `logits`.
+  Discussion on Github: https://github.com/tensorflow/tensorflow/issues/9260#issuecomment-437875125
+  Details here: https://timvieira.github.io/blog/post/2014/08/01/gumbel-max-trick-and-weighted-reservoir-sampling/
+  :param tf.Tensor[tf.float32] logits: log probability for each possible sample
```
albertz: This is not a valid type.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
...
+  :param tf.Tensor[tf.int32] k: number of samples to draw
```
albertz: This is not a valid type.

TFUtil.py
```diff
@@ -9322,3 +9322,48 @@ def _callback(self, value):
...
+  :param int|None seed: random seed
+  :return: k indices sampled from the distribution
+  :rtype: tf.Tensor[tf.int32]
```
albertz: This is not a valid type.

TFUtil.py
albertz: Actually, excluding the `tf.nn.top_k` though. So maybe not.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,266 @@ def mask_func(source, **kwargs):
       last_var_v = var_v
 
 
+def _create_data(n_batch, n_time, n_feature, network, dtype=tf.int32, offset=0, sparse=True):
```
albertz: Note that there is `make_feed_dict` in that function, which has some similar logic. Maybe it makes more sense to reuse that, and not duplicating the logic.

tests/test_TFNetworkLayer.py
```diff
@@ -3065,6 +3065,266 @@ def mask_func(source, **kwargs):
...
+  _test_ProbabilisticMaskLayer(with_prob=False, mask_factor=(0.0, 1.0))
+  _test_ProbabilisticMaskLayer(with_prob=True)
+  _test_ProbabilisticMaskLayer(with_prob=False)
+  _test_ProbabilisticMaskLayer(with_prob=False, n_feature=100)
+  _test_ProbabilisticMaskLayer(fixed_mask=True, mask_factor=1.0, with_prob=False)
+  # test gathering on its own:
+  _test_GatherPositionsLayer()
+  # test them together:
+  _test_mask_and_gather(with_prob=False)
+  _test_mask_and_gather(with_prob=True)
```
albertz: You should make 3 separate test cases for these. You should not combine all into a single test case `test_MLM`.
(But anyway, see discussion above. Some of the layers probably should be removed.)

TFNetworkLayer.py
ArneNx: I thought there was some demand in a layer like this (but I can see why you find this too specific). Also, I don't see a way to make this more generic and putting it in my config is just very ugly in combination with sisyphus. Therefore I'm going to keep this in my own fork for anyone to reproduce my MLM experiments.
Also, I will close this pull-request, since the rest is mainly useful in combination with `ProbabilisticMaskLayer`.

TFNetworkLayer.py
ArneNx: I actually wasn't aware of the `GatherNdLayer` (since it was introduced long after I developed `GatherPositionsLayer`). Yes, they are quite similar, however, this layer allows for different indexing for each sequence by adding an offset to the indices before performing the gather operation on the flattened batch. And more importantly, it recovers the batch structure after the gather operation.
I'm not sure if this (especially the latter) can be done in a more generic way.
Also, I'm not sure if this functionality will be useful at all without `ProbabilisticMaskLayer`.

albertz: But this sounds exactly like what `GatherNdLayer` does. It allows for different indexing for each sequence.

TFNetworkLayer.py
albertz: It's not a good idea to keep your own fork around. Never. Everyone should always be able to reproduce any Returnn experiment with the current official Returnn master. (Or you say all these experiments are not interesting to anyone anymore, and no-one actually will care.)

I don't really know what problem there is with Sisyphus. This sounds like a very artificial problem, or not? If Sisyphus does not really allow this, you could also just put the layer code into a separate file, and then load that file in Returnn (either via `import`, or maybe just `eval`; see also `Util.custom_exec`).

But in any case, all what you do within this very specific layer was already possible before, or not? So you could have also written it in the config, using the existing layers. (Or what specifically was missing to allow this? See the Returnn video lecture. If any basic operation is missing, that is what a new layer is for.)

PR #245, opened 2019-11-13
ChoiceLayer: support for more than two targets

patrick-wilken:
I generalised the implementation of target factors in the ChoiceLayer to an arbitrary number of targets.

Also I changed the tests to 3 instead of 2 targets. A single choice layer with 3 targets works, however the other test, where there are 3 separate ChoiceLayers depending on each other, fails. Something seems to go wrong in `get_search_choices`, I will try to look into this, but help is appreciated.

The two things are only somewhat related, I could also make separate PRs...

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -4055,6 +4055,15 @@ class ChoiceLayer(BaseChoiceLayer):
   Assume that each batch is already a choice via search.
   In search with a beam size of N, we would output
   sparse (batch=N,) and scores for each.
+
+  In case of multiple sources, this layer computes the top-k combinations of choices. The score of such a combination
+  is determined by adding up the (log-space) scores of the choices for the individual sources. In this case, the
+  'target' parameter of the layer has to be set to a list of targets corresponding to the sources respectively. Because
+  computing all possible combinations of source scores is costly, the sources are pruned beforehand using the beam
+  sizes set by the 'source_beam_sizes' parameter. The choices made for the different sources can be accessed via the
+  sublayers '<choice layer name>/out_0', '<choice layer name>/out_1' and so on.
+  Note, that the way scores are combined assumes the sources to be independent. If you want to model a dependency,
+  use separate ChoiceLayers and let the input of one depend on the output of the other.
   """
   layer_class = "choice"
 
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -4129,7 +4129,6 @@ class ChoiceLayer(BaseChoiceLayer):
         self.search_choices.set_src_beams(tf.zeros((net_batch_dim, 1), dtype=tf.int32))
         self.search_choices.set_beam_scores(self.search_choices.src_layer.search_choices.beam_scores)
       else:
-        assert len(self.sources) <= 2, "Combining more than two sources not implemented yet."
         net_batch_dim = self.network.get_data_batch_dim()
         if len(self.sources) > 1:
           # If no separate beam sizes for the sources are given, use the final beam size also for pruning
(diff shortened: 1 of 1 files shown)


albertz:
No, don't change the test. I said that you should copy it, as a separate test.

albertz:
Also, I already forgot what it means to have multiple targets. This is somewhat counter intuitive. Can you explain again? (Or better extend the documentation.)

patrick-wilken:
I've added some documentation.

I altered the test case because checking 3 targets implies checking 2 targets. But ok, now I'm just adding the one failing test case.

tests/test_TFEngine.py
```diff
@@ -1163,6 +1163,126 @@ def test_attention_two_dependent_targets():
...
+        "orth_embed_1": {'class': 'linear', 'activation': None, 'from': ['output_1'], "n_out": 7},
+        "orth_embed_2": {'class': 'linear', 'activation': None, 'from': ['output_2'], "n_out": 7},
+        "orth_embed": {"class": "copy", "from": ["orth_embed_0", "orth_embed_1", "orth_embed_2"]},
+        "s": {"class": "rnn_cell", "unit": "LSTMBlock", "from": ["prev:c", "prev:orth_embed"], "n_out": 7},
+        "c_in": {"class": "linear", "activation": "tanh", "from": ["s", "prev:orth_embed"], "n_out": 5},
+        "c": {"class": "dot_attention", "from": ["c_in"], "base": "base:encoder", "base_ctx": "base:encoder"},
+        "output_prob_0": {"class": "softmax", "from": ["prev:s", "c"], "target": "classes_0", "loss": "ce"},
+        "output_prob_1": {"class": "softmax", "from": ["prev:s", "c", "orth_embed_0"],
+          "target": "classes_1", "loss": "ce"},
+        "output_prob_1": {"class": "softmax", "from": ["prev:s", "c", "orth_embed_0", "orth_embed_1"],
```
patrick-wilken: Oops, this should be `output_prob_2`. With this change it runs. But maybe this only means this test does not reproduce my error, looking into it...

albertz: This test is also not really about multiple targets (and I don't remember what that means anyway), but just about multiple choices.

patrick-wilken:
Ok, sorry for the confusion, so let's keep it separate. I removed the test, which was about using 3 separate ChoiceLayers. I will make another PR as soon as I have a proper test that reproduces my problem.

So this PR is just about making multiple choices at the same time with one ChoiceLayer, a.k.a. using multiple sources/targets.

patrick-wilken:
The use case for this is if you have several targets which should be modelled as independent. E.g. target tokens t1, t2 (at a given timestep), source context `C`,  then:
p(t1, t2) = p(t1|C) * p(t2|C)
rather than:
p(t1, t2) = p(t1|C) * p(t2|C, t1)

The latter can be modelled by having a softmax + ChoiceLayer for t1, and after that the same for t2. In the first case we cannot have two separate ChoiceLayers as this would result in two independent beams that cannot be combined afterwards.

albertz:
Ah, I very briefly remember our discussion about that. But does it even make sense?
Btw, one workaround: In p(t2|C, t1), you can just mark t1 as a dependency (e.g. use EvalLayer which gets C and t1, and simply ignore t1; or use CopyLayer with the extra_deps option, which does exactly that), so that you effectively model p(t2|C).
This should work. Does this solve the whole thing? Or what would be different in that multi-target support or whatever else you plan to extend there?

albertz:
I remember more about this. This was about how to combine the beams then, right? In p(t2|C, t1), it would already select only those beams from the choices of t1. I don't remember how you picked the choices of p(t1|C),p(t2|C) when they are independent.
Wouldn't it make more sense to have a separate layer which is just about that, to combine some independent beams? It feels ugly to extend ChoiceLayer and make it more complicated.

patrick-wilken:
First of all: actually since 0895497 three (and I guess more) dependent ChoiceLayers work in the master, nice! I was working on an older branch and thought I had reproduced the error in the master, but it was just a similar looking error due to this typo. So no need to add any new tests.

Regarding this PR: Now that I'm more familiar with it all I tend to agree. This workaround you describe should really do the same thing*. Maybe it is possible to detect parallel choices in the graph and add this dependency automatically? Might be pretty complicated though.
Writing a layer that combines independent beams after the choices are made should be easier, basically by copying `_prune_and_combine_sources()` and `_get_combined_labels()` from the current ChoiceLayer. What it does is basically compute the "cross product" of the beams, i.e. adding up scores for all possible combination of beam entries. With an additional pruning step in the beginning to reduce the number of combinations.

Maybe you want to merge this PR anyway as it gets rid of the "TODO" in the code ;) I just finished that and wanted to have it committed. But it's not something I currently rely on as we are currently only using the dependent variant of target factors for our systems.

*in combination with resolving beams of all outputs and other things that were not implemented back then...

albertz:
> Regarding this PR: Now that I'm more familiar with it all I tend to agree. This workaround you describe should really do the same thing*. Maybe it is possible to detect parallel choices in the graph and add this dependency automatically? Might be pretty complicated though.

But it cannot be automatically. Depending whether you do p(t2|C,t1) or p(p1|C,t2) will have different behavior.

> Writing a layer that combines independent beams after the choices are made should be easier, basically by copying `_prune_and_combine_sources()` and `_get_combined_labels()` from the current ChoiceLayer. What it does is basically compute the "cross product" of the beams, i.e. adding up scores for all possible combination of beam entries. With an additional pruning step in the beginning to reduce the number of combinations.

Yes, exactly such a combine behavior is what I thought would be good to have in another layer. I'm not sure currently if there are other problems with this approach. I think the dependency handling of the search choices also need some update here, as it always assumes a single dependency, not multiple. Not sure how difficult this is to extend.

The prune behavior can also be a separate layer, and this should be simple and straight forward, right?

albertz:
Sure, I can anyway merge this is. I guess our discussion is better for a separate pull request / issue.

PR #248, opened 2019-12-05
ConstantLayer: array constants

patrick-wilken:
As far as I understand, so far, there is no way to create a non-scalar constant. Taking the shape from the constant if it is an array is the only change necessary to support this.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -3394,6 +3394,7 @@ class ConstantLayer(LayerBase):
     :param bool with_batch_dim:
     :rtype: Data
     """
+    import numpy
     if dtype is None:
       if isinstance(value, int):
         dtype = "int32"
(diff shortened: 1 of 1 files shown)


TFNetworkLayer.py
```diff
@@ -3403,9 +3403,13 @@ def get_out_data_from_opts(cls, name, value=0., dtype=None, with_batch_dim=False
         dtype = "bool"
       else:
         raise TypeError("cannot handle value %r of type %r" % (value, type(value)))
+
+    import numpy
```
albertz: Make that import at the beginning of the function, and avoid the extra empty lines around it.

TFNetworkLayer.py
```diff
@@ -3403,9 +3403,13 @@ def get_out_data_from_opts(cls, name, value=0., dtype=None, with_batch_dim=False
...
+    shape = value.shape if isinstance(value, numpy.ndarray) else ()
+
```
albertz: Also avoid the extra empty line here.

patrick-wilken:
Done.

PR #249, opened 2019-12-05
TranslationDataset: Factors

patrick-wilken:
This is my implementation of translation factors. It allows adding arbitrary additional token-level information to the source and target data by using a simple separator symbol, i.e. the input tokens look like 'lemma|factor1|factor2'. This is also the format used by Moses, Sockeye and OpenNMT.

Unfortunately, the diff is quite large, in part because there is no longer a one to one correspondence between data key and corpus file.

The changes in ConfusionNetworkDataset are just to keep it alive, no features added there. I also updated the hdf dump tool accordingly.

I'm afraid it's time to write a test case for TranslationDataset... ':D

### DIFF
--- a/LmDataset.py
+++ b/LmDataset.py
@@ -1038,13 +1038,17 @@ class TranslationDataset(CachedDataset2):
       target.vocab.pkl
   """
 
-  MapToDataKeys = {"source": "data", "target": "classes"}  # just by our convention
-  _main_data_key = None
-  _main_classes_key = None
+  source_file_prefix = "source"
+  target_file_prefix = "target"
+  main_source_data_key = "data"
+  main_target_data_key = "classes"
 
   def __init__(self, path, file_postfix, source_postfix="", target_postfix="",
                source_only=False,
                unknown_label=None,
+               source_factors=None,
+               target_factors=None,
+               factor_separator='|',
                seq_list_file=None,
                use_cache_manager=False,
                **kwargs):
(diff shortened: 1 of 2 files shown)


albertz:
Can you explain again what "factors" mean here, in more detail?
(And btw, not only explain here, but esp also in the code documentation.)
Why does the dataset need to know about this? I don't think it's a good idea to extend one specific dataset (`TranslationDataset` here) for very specific functionality. This should be handled in a generic way which works for all datasets (something like `MetaDataset`).

Btw, I'm not using `TranslationDataset` so often recently, and also I think there are no test cases at all for `TranslationDataset`, so I would not notice if there is anything wrong here, or it breaks any old setups or so. Did you check with other colleagues whether this is all ok?

LmDataset.py
```diff
@@ -1038,72 +1038,116 @@ class TranslationDataset(CachedDataset2):
...
+    :param None|str target_postfix: will append this token to the target words (for all target factors).
       You might want to add some sentence-end symbol.
     :param bool source_only: if targets are not available
-    :param str|None unknown_label: "UNK" or so. if not given, then will not replace unknowns but throw an error
+    :param str|dict[str,str]|None unknown_label: Label to replace out-of-vocabulary words with, e.g. "<UNK>".
+      If not given, will not replace unknowns but throw an error. Can also be a dict data_key -> unknown_label
+      to configure for each data key separately (default for each key is None).
+    :param list[str]|None source_factors: Names of source factors (excluding first factor, which is always
+      called 'data'). Words in source file have to have that many factors. Also, a vocabulary
+      "<factor_name>.voc.pkl" has to exist for each factor.
```
albertz: What does that mean? These are becoming data stream then?
These data streams are always the same length as `data` (or `classes` respectively)? This is important information, which should be documented here. Also, you should give some actual examples, and maybe ref to papers, and other software (you said other software uses the same format).

LmDataset.py
```diff
@@ -1038,72 +1038,116 @@ class TranslationDataset(CachedDataset2):
...
     self.path = path
     self.file_postfix = file_postfix
     self.seq_list = [int(n) for n in open(seq_list_file).read().splitlines()] if seq_list_file else None
-    if self._main_data_key is None:
-      self._main_data_key = "data"
-    if self._main_classes_key is None:
-      self._main_classes_key = "classes"
-    self._add_postfix = {self._main_data_key: source_postfix, self._main_classes_key: target_postfix}
-    self._keys_to_read = [self._main_data_key, self._main_classes_key]
+    self._add_postfix = {self.source_file_prefix: source_postfix.strip(),
```
albertz: Add a newline after `{`. This would look nicer visually, as it then aligns with the next line.

LmDataset.py
```diff
@@ -1038,72 +1038,116 @@ class TranslationDataset(CachedDataset2):
...
+    self._data_keys = self._source_data_keys + self._target_data_keys
+
+    # Read vocs for main data
+    self._vocabs = {data_key: self._get_vocab(prefix) for (prefix, data_key) in self._main_data_key_map.items()}
+    # Read vocs for all factors
+    self._vocabs.update({data_key: self._get_vocab(data_key)
+      for data_key in self._source_data_keys[1:] + self._target_data_keys[1:]})
+
+    self._data_files = {prefix: self._get_data_file(prefix) for prefix in self._main_data_key_map}
+    self._data = {data_key: []
```
albertz: Add a newline after `{`.

LmDataset.py
```diff
@@ -1038,72 +1038,116 @@ class TranslationDataset(CachedDataset2):
...
+
     self._seq_order = None  # type: typing.Optional[typing.List[int]]  # seq_idx -> line_nr
     self._tag_prefix = "line-"  # sequence tag is "line-n", where n is the line number
     self._thread = Thread(name="%r reader" % self, target=self._thread_main)
     self._thread.daemon = True
     self._thread.start()
 
-  def _extend_data(self, k, data_strs):
-    vocab = self._vocabs[k]
+  def _extend_data(self, file_prefix, data_strs):
```
albertz: Always document types!

LmDataset.py
```diff
@@ -1433,46 +1497,52 @@ def _load_single_confusion_net(self, words, vocab, postfix):
       words_confs[len(words)][0] = 1
     return words_idxs, words_confs
 
-  def _data_str_to_sparse_inputs(self, vocab, s, postfix=None):
+  def _data_str_to_sparse_inputs(self, key, s, postfix=None):
     """
-    :param dict[str,int] vocab:
+    :param str key:
```
albertz: What is the key? Give some example.

LmDataset.py
```diff
@@ -1433,46 +1497,52 @@ def _load_single_confusion_net(self, words, vocab, postfix):
...
-    if key == self._main_data_key:  # the sparse inputs and weights
+
+    if file_prefix == self.source_file_prefix:  # the sparse inputs and weights
+      key = self._source_data_keys[0]
       idx_data = []
       conf_data = []
       for s in data_strs:
         (words_idxs, words_confs) = self._data_str_to_sparse_inputs(
-          vocab, s.decode("utf8").strip(), self._add_postfix[key])
+          key, s.decode("utf8").strip(), self._add_postfix[file_prefix])
```
albertz: Use keyword args.

LmDataset.py
```diff
@@ -1038,72 +1038,116 @@ class TranslationDataset(CachedDataset2):
...
+    if file_prefix == self.source_file_prefix:
+      data_keys = self._source_data_keys
+    else:
+      assert file_prefix == self.target_file_prefix
+      data_keys = self._target_data_keys
+
     data = [
-      self._data_str_to_numpy(vocab, s.decode("utf8").strip() + self._add_postfix[k])
-      for s in data_strs]
+      self._data_str_to_numpy(data_keys, s.decode("utf8").strip(), self._add_postfix[file_prefix])
```
albertz: Use keyword args.

LmDataset.py
```diff
@@ -1433,46 +1497,52 @@ def _load_single_confusion_net(self, words, vocab, postfix):
...
         idx_data.append(words_idxs)
         conf_data.append(words_confs)
       with self._lock:
         self._data[key].extend(idx_data)
         self._data["sparse_weights"].extend(conf_data)
     else:  # the classes
+      key = self._target_data_keys[0]
       data = [
-        self._data_str_to_numpy(vocab, s.decode("utf8").strip() + self._add_postfix[key])
+        self._data_str_to_numpy([key], s.decode("utf8").strip(), self._add_postfix[file_prefix])
```
albertz: Use keyword args.

patrick-wilken:
Before having time to change the code I will answer here:

Factors originally have been used in the literature to split the representation of a word into its base form and its inflections, e.g. something like (in German) "Beispiel|<Plural>|<Akkusativ>" -> "Beispielen" or to provide additional information like part-of-speech tags, see https://www.aclweb.org/anthology/D07-1091.pdf, https://workshop2016.iwslt.org/downloads/IWSLT_2016_paper_2.pdf. 
We use factors to predict casing (as in "beispiel|<all-upper>" -> "BEISPIEL") and subword merges, see https://arxiv.org/pdf/1910.03912.pdf, and also to indicate phrase boundaries, for example. 
Amazon AI uses factors in the context of term overwrite: https://www.aclweb.org/anthology/P19-1294.pdf, which we plan to adopt. So they are a very generic tool.

I agree however that it makes the code of TranslationDataset look very specific. Maybe a derived dataset would be better?
The same thing could also be achieved by having one TranslationDataset per factor and using MetaDataset to combine them. But this introduces overhead, both in RETURNN as well as in the data preparation workflow. It is very convenient to have all the factors in one file (thinking of shuffling, splitting etc.).

We could keep this internal, but I thought it's nice to have, also because a number of other MT frameworks support this. Comments from i6 MT staff appreciated ;)

We used this code, well at least using one target-side factor, to train a number of MT models by now, so its well tested. But independent of whether this will be merged, I will try to invest time to write a test case for the standard case.

albertz:
Well, anyway, I guess you (or some of you) will in the future maintain the `TranslationDataset` (not sure how much I will still use it), so consider it as your code, and just try to keep it maintainable for the future (think into 2 years, or maybe 5 years, or maybe even 10 years into the future). Unless you consider that `TranslationDataset` will be thrown away until then.

albertz:
Is there an update here? As I said, I leave the decision to you. But it would be good to have a decision at some point. (If you merge it, maybe address my minor comments.)

patrick-wilken:
I will have time to work on this now ':D So I first will see how a FactoredTranslationDataset as a subclass would look like and then we can decide.

patrick-wilken:
Closing this in favour of #289.

Issue #87, opened 2018-11-24
TF 1.12: nvcc fails: absl/strings/string_view.h(501): error: constexpr function return is non-constant


abhigarg-iitk:
I am having issues with running returnn experiments with TF1.12 and Python 2. My error is this

`CalledProcessError: Command '['/ml_env/cuda9/cuda-9.0/bin/nvcc', '-shared', '-O2', '-std=c++11', '-I',  '/vol0/ml_env/cuda9/.venv/py2/asr/lib/python2.7/site-packages/tensorflow/include', '-I', '/vol0/ml_env/cuda9/.venv/py2/asr/lib/python2.7/site-packages/tensorflow/include/external/nsync/public', '-I', '/t ool/ml/cuda9/cuda-9.0/include', '-L', '/ml_env/cuda9/cuda-9.0/lib64', '-x', 'cu', '-DGOOGLE_CUDA=1', ' -Xcompiler', '-fPIC', '-arch', 'compute_61', '-D_GLIBCXX_USE_CXX11_ABI=0', '-g', '/tmp/abhinav.garg/ret urnn_tf_cache/ops/NativeLstm2/734b69db29/NativeLstm2.cc', '-o', '/tmp/abhinav.garg/returnn_tf_cache/ops /NativeLstm2/734b69db29/NativeLstm2.so', '-L/vol0/ml_env/cuda9/.venv/py2/asr/lib/python2.7/site-packag es/numpy/.libs', '-lopenblasp-r0-39a31c03.2.18', '-L/vol0/ml_env/cuda9/.venv/py2/asr/lib/python2.7/sit e-packages/tensorflow', '-ltensorflow_framework']' returned non-zero exit status 1 
`

Which is very similar to https://github.com/tensorflow/tensorflow/issues/22766. I found the solution of using `-DNDEBUG` but I'm not sure if it's the correct solution. Can you please tell me how can I run returnn with tf1.12. I have tried applying this patch https://github.com/rwth-i6/returnn/commit/5913f1f975e08f1172217e75dbe5aa4263068a84 but it's still giving me the same error.

Thanks in Advance

patrick-wilken:
For the record, I'm also using `-DNDEBUG` at the moment as a workaround:

```
  def get_compiler_opts(self):
    return [
      "-I", "%s/include" % self.cuda_path,
      "-L", "%s/%s" % (self.cuda_path, self._get_lib_dir_name()),
      "-x", "cu",
      "-v",
[...]

PR #250, opened 2019-12-06
TF 1.15 compatibility

pzelasko:
Work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,7 @@ install:
 # TF tests first, because more relevant (for me), then ordered by alphabet
 env:
   global:
-    # TODO: TF 1.14 needs multiple fixes...
-    - TF_PACKAGE=tensorflow==1.13.1
+    - TF_PACKAGE=tensorflow==1.15.0
   matrix:
     - TEST=TFEngine
     - TEST=TFNativeOp
--- a/NativeOp.cpp
+++ b/NativeOp.cpp
@@ -752,7 +752,12 @@ void* _malloc(size_t num_bytes) {
     //auto* stream = context->op_device_context()->stream();
     Allocator* allocator =
         context->device()->GetAllocator(AllocatorAttributes());
-    void* ptr = (void*)allocator->Allocate<uint8_t>(num_bytes);
+    // Note: Starting with TF 1.14, there is no "Allocate" convenience wrapper anymore.
+#if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 14)
+    void* ptr = allocator->AllocateRaw(Allocator::kAllocatorAlignment, num_bytes);
+#else
+    void* ptr = (void*) allocator->Allocate<uint8_t>(num_bytes);
+#endif
     if(!ptr)
         context->CtxFailure(
             errors::InvalidArgument("NativeOp: cannot allocate ", num_bytes, " bytes on ", allocator->Name()));
--- a/TFNativeOp.py
+++ b/TFNativeOp.py
@@ -243,7 +243,12 @@ class OpMaker(object):
     for in_idx, v in enumerate(in_info):
       out_idx = v.get("want_inplace", -1)
       if out_idx >= 0:
-        code_forward_io += "context->forward_ref_input_to_ref_output(%i, %i);\n" % (in_idx, out_idx)
+        code_forward_io += """
+          // Note: forward_ref_input_to_ref_output expects the input tensor to be a reference, and not a value;
+          //   otherwise it will trigger an assertion.
+          if (IsRefType(context->input_dtype({in_idx})))
+            context->forward_ref_input_to_ref_output({in_idx}, {out_idx});
+          """.format(in_idx=in_idx, out_idx=out_idx)
     code_set_io = ""
     for in_idx, v in enumerate(in_info):
       ndim = len(v["shape"])
@@ -385,7 +390,7 @@ class OpMaker(object):
     if self.description.cpu_support:
       code_cpu_op = """
       %(user_code_kernels)s
-  
+
       class %(op_name)sOp : public OpKernel {
       public:
         explicit %(op_name)sOp(OpKernelConstruction* context) : OpKernel(context) {}
@@ -393,7 +398,7 @@ class OpMaker(object):
           %(code_compute)s
         }
       };
-  
+
       REGISTER_KERNEL_BUILDER(Name("%(op_name)s").Device(DEVICE_CPU), %(op_name)sOp);
       """ % format_args
     else:
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -4785,8 +4785,10 @@ class OpCodeCompiler(NativeCodeCompiler):
     tf_include_nsync = tf_include + "/external/nsync/public"  # https://github.com/tensorflow/tensorflow/issues/2412
     include_paths = list(include_paths) + [tf_include, tf_include_nsync]
     ld_flags = list(ld_flags)
-    if have_min_tf_version((1, 4)):
+    if have_min_tf_version((1, 14)):
       # https://github.com/tensorflow/tensorflow/issues/13607
+      ld_flags += tf.sysconfig.get_link_flags()
+    elif have_min_tf_version((1, 4)):
       ld_flags += ["-L%s" % tf.sysconfig.get_lib(), "-ltensorflow_framework"]
     # noinspection PyUnresolvedReferences
     use_cxx11_abi = hasattr(tf, 'CXX11_ABI_FLAG') and tf.CXX11_ABI_FLAG
@@ -4848,8 +4850,10 @@ class TFNativeUtilCompiler(NativeCodeCompiler):
     tf_include_nsync = tf_include + "/external/nsync/public"  # https://github.com/tensorflow/tensorflow/issues/2412
     include_paths = list(include_paths) + [tf_include, tf_include_nsync]
     ld_flags = list(ld_flags)
-    if have_min_tf_version((1, 4)):
+    if have_min_tf_version((1, 14)):
       # https://github.com/tensorflow/tensorflow/issues/13607
+      ld_flags += tf.sysconfig.get_link_flags()
+    elif have_min_tf_version((1, 4)):
       ld_flags += ["-L%s" % tf.sysconfig.get_lib(), "-ltensorflow_framework"]
     # noinspection PyUnresolvedReferences
     use_cxx11_abi = hasattr(tf, 'CXX11_ABI_FLAG') and tf.CXX11_ABI_FLAG
@@ -7019,7 +7023,7 @@ class GlobalTensorArrayOpMaker:
     #include "tensorflow/core/platform/types.h"
 
     using namespace tensorflow;
-  
+
     // Adopted from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/data_flow_ops.cc.
     REGISTER_OP("GlobalTensorArray")
     .Input("size: int32")
@@ -7041,7 +7045,7 @@ class GlobalTensorArrayOpMaker:
       return Status::OK();
     })
     .Doc("GlobalTensorArray, persistent across runs");
-    
+
     // Copied from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/tensor_array_ops.cc,
     // and https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/resource_op_kernel.h.
     // The original TensorArrayOp used the per-run ("per-step") resource manager container
@@ -7065,7 +7069,7 @@ class GlobalTensorArrayOpMaker:
                                 tensorflow::DT_STRING, tensorflow::TensorShape({2}),
                                 &handle_, alloc_attr));
       }
-    
+
       ~GlobalTensorArrayOp() {
         if (resource_ != nullptr) {
           resource_->Unref();
@@ -7078,7 +7082,7 @@ class GlobalTensorArrayOpMaker:
           }
         }
       }
-    
+
       void Compute(OpKernelContext* ctx) override {
         mutex_lock l(mu_);
         if (resource_ == nullptr) {
@@ -7094,7 +7098,7 @@ class GlobalTensorArrayOpMaker:
         Tensor* handle;
         OP_REQUIRES_OK(ctx, ctx->allocate_output(0, TensorShape({}), &handle));
         handle->flat<ResourceHandle>()(0) =
-            resource_->resource_handle(ctx);            
+            resource_->resource_handle(ctx);
         if (ctx->num_outputs() == 2) {
           // Create the flow output.
           Tensor* flow;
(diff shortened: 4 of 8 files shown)


deps/install_tensorflow_1_15.sh
```diff
@@ -0,0 +1,14 @@
...
+pip install tensorflow==1.15.0
+
+# Find out where do the tensorflow libs and sources reside
+TF_DIR="$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_lib())')"
+
+# Make TF lib detectable for the ldd linker
+ln -s "$TF_DIR/libtensorflow_framework.so.1" "$TF_DIR/libtensorflow_framework.so"
+
+# Work around TF issue: https://github.com/tensorflow/tensorflow/issues/22766
+sed -i 's/return ABSL_ASSERT(len <= kMaxSize), len;/return len;/' "$TF_DIR/include/absl/strings/string_view.h"
```
albertz: As far as I remember, this was only needed for GPU? In that case, we should not do that  for Travis.

deps/install_tensorflow_1_15.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -eou pipefail
+
+pip install tensorflow==1.15.0
+
+# Find out where do the tensorflow libs and sources reside
+TF_DIR="$(python -c 'import tensorflow as tf; print(tf.sysconfig.get_lib())')"
+
+# Make TF lib detectable for the ldd linker
+ln -s "$TF_DIR/libtensorflow_framework.so.1" "$TF_DIR/libtensorflow_framework.so"
```
albertz: Why is this needed? So everyone else also needs to do this? Shouldn't it be simple to just fix the compile/ld flags at compilation to directly find the right so-file? I think this would be the better solution.

.travis.yml
```diff
@@ -26,7 +26,7 @@ install:
   - pip install -qq --upgrade pip setuptools wheel
   - pip install --upgrade numpy scipy | cat  # TF usually needs the latest Numpy
   - pip install -r requirements.txt | cat
-  - pip install $TF_PACKAGE | cat
+  - bash deps/install_tensorflow_1_15.sh
```
albertz: It should install `$TF_PACKAGE` here, **not** TF 1.15.

albertz: Also, because of that, the install Bash script is not so nice.
In any case, I would try to avoid using the Bash script here if possible. Maybe I will just enabled `-DNDEBUG` globally now, so that the patching should not be needed anymore. See also the discussion in #87.

NativeOp.cpp
```diff
@@ -752,7 +752,11 @@ void* _malloc(size_t num_bytes) {
     //auto* stream = context->op_device_context()->stream();
     Allocator* allocator =
         context->device()->GetAllocator(AllocatorAttributes());
-    void* ptr = (void*)allocator->Allocate<uint8_t>(num_bytes);
+#if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 14)
```
albertz: Can you add maybe a comment? `Allocate` was removed from the API since TF 1.14?

TFNativeOp.py
```diff
@@ -243,7 +243,12 @@ def make_dim_str(c):
     for in_idx, v in enumerate(in_info):
       out_idx = v.get("want_inplace", -1)
       if out_idx >= 0:
-        code_forward_io += "context->forward_ref_input_to_ref_output(%i, %i);\n" % (in_idx, out_idx)
+        code_forward_io += """
+if (IsRefType(context->input_dtype({in_idx})))
+  context->forward_ref_input_to_ref_output({in_idx}, {out_idx});
+else
+  context->set_output({out_idx}, context->input({in_idx}));
+\n""".format(in_idx=in_idx, out_idx=out_idx)
```
albertz: Why the `\n`? Also, indent it to the other Python code, which looks a bit nicer in the Python code (I don't care much how the resulting C++ code is indented...).

albertz: Btw, also add a comment here. Did you investigate further what this actually means, and why it happens now, and not earlier?
@curufinwe You said in your compilation of TF 1.15, someone observed some strange behavior in training. Maybe this is related?

deps/install_tensorflow_1_15.sh
albertz: (#87 says this is for nvcc only...)

requirements.txt
```diff
@@ -1,3 +1,10 @@
...
 h5py
+nose
+numpy
+scipy
+setuptools
+six
+typing
 typing
+wheel
+
```
albertz: Don't include this in this pull request.
Also, in general, this should not be needed. Esp it should not contain stuff which are only needed for the tests.
Numpy should also be covered already when you install TF (TF is not in there because some people would use `tensorflow-gpu`, while others use just `tensorflow`).
Scipy is not necessarily needed.
We don't use Six (I think...). Anyway, I guess also already covered implicitly by TF.
Setuptools, Wheels also not needed.
Typing is twice there now.

.travis.yml
albertz: Btw, also check [Travis](https://travis-ci.org/rwth-i6/returnn/builds/621529097): It seems it works fine without this.

deps/install_tensorflow_1_15.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
albertz: See the comment above. I think we don't need the file.

requirements.txt
pzelasko: Sure, I can revert these. I had some test failures when scipy was not installed though, so perhaps it should be included there?

albertz: Where exactly? But anyway, let's move this out of this pull request. Make a separate issue maybe first, including the exception you get/got.

TFNativeOp.py
pzelasko: I'll fix the formatting. 

The `forward_ref...` bit expects to have tensor references both on input and output, and sets `output = input`.  It seems that this assumption was not satisfied in some tests, i.e. the input was actually a tensor value, not a reference, which triggered assertion.

Unfortunately, I don't know why it started happening only with 1.14/1.15... Didn't find any relevant changes in the TF change logs.

NativeOp.cpp
pzelasko: That's right. I'll add a comment.

.travis.yml
pzelasko: OK will do.

tests/test_TFEngine.py
```diff
@@ -391,15 +391,6 @@ def test_engine_train_grad_noise_sparse():
...
 
-  def wrapped_slice_to_tensor(value, dtype=None, name=None, as_ref=False):
-    print("wrapped_slice_to_tensor:", value)
-    #assert "flatten_with_seq_len_mask" in value.name
-    from tensorflow.python.ops import gradients_impl
-    return gradients_impl._IndexedSlicesToTensor(value, dtype=dtype, name=name, as_ref=as_ref)
-
-  from tensorflow.python.framework import ops
-  ops.register_tensor_conversion_function(tf.IndexedSlices, wrapped_slice_to_tensor, priority=1)
-
```
albertz: Why only removing this? How does it fix sth?
Also, there was no problem in Travis. Why is this needed?

deps/install_tensorflow_1_15.sh
pzelasko: Let me check again then - I had no success with this approach earlier.

deps/install_tensorflow_1_15.sh
pzelasko: OK I'll try to remove the Bash script altogether.

tests/test_TFEngine.py
pzelasko: They removed the `_IndexedSlicesToTensor` function from `gradients_impl.py`, I think they replaced it's functionality with something else. The test didn't pass when I checked locally, after the change it works fine.

deps/install_tensorflow_1_15.sh
pzelasko: Ok

TFNativeOp.py
albertz: Yea, don't explain here, but in the code with a comment.
Also, what does it mean to be a reference or a value?

albertz:
I see that the `DebugString()` API changed. It became `const` now. So we need some
```
#if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION <= 13)
const
#endif 
```
there. (I think this is relevant for OpenFst and KenLM only.)

albertz:
Btw, also maybe cleanup the commits already (rebase, squash together a bit).

tests/test_TFEngine.py
albertz: I see, ok.

pzelasko:
Closing this PR; there is a new one with commits squashed: #251

albertz:
> Closing this PR; there is a new one with commits squashed: #251

It's a bad idea to do that. You should rather just rebase and force-push/update this pull request. Otherwise we lose all the discussion here.

pzelasko:
You're right!

TFUtil.py
```diff
@@ -4787,7 +4787,7 @@ def __init__(self, use_cuda_if_available=True, cuda_auto_min_compute_capability=
     ld_flags = list(ld_flags)
     if have_min_tf_version((1, 4)):
       # https://github.com/tensorflow/tensorflow/issues/13607
-      ld_flags += ["-L%s" % tf.sysconfig.get_lib(), "-ltensorflow_framework"]
+      ld_flags += tf.sysconfig.get_link_flags()
```
albertz: Be careful here. I think `get_link_flags()` was introduced only later (later than TF 1.4). So maybe we need another check `have_min_tf_version((1, 14))` (or whatever version) for that.

TFUtil.py
```diff
@@ -8071,6 +8078,19 @@ def tensor_array_is_clear_after_read(ta):
 
 # noinspection PyProtectedMember
 def tensor_array_element_shape(ta):
+  """
+  :param tf.TensorArray ta:
+  :rtype: tf.TensorShape
+  """
+  tf_version = float(tf.version.VERSION.rsplit('.', 1)[0])
+  if tf_version >= 1.14:
```
albertz: Use `have_min_tf_version((1, 14))` instead.

TFUtil.py
```diff
@@ -8085,6 +8105,15 @@ def tensor_array_element_shape(ta):
   return tf.TensorShape(None)
 
 
+def tensor_array_element_shape_py3_tf_1_14_and_later(ta):
+  """
+  :param tf.TensorArray ta:
+  :rtype: tf.TensorShape
+  """
+  assert isinstance(ta.element_shape, tf.TensorShape)
+  return ta.element_shape
```
albertz: Do not create an extra function for this. Just make it inline in `tensor_array_element_shape`. The same also for `tensor_array_element_shape_tf_up_to_1_13`.

requirements.txt
```diff
@@ -1,3 +1,3 @@
-numpy
 h5py
+numpy
```
albertz: Put it back to top, there should be no change here.

tests/test_TFEngine.py
```diff
@@ -391,14 +391,16 @@ def test_engine_train_grad_noise_sparse():
...
-  def wrapped_slice_to_tensor(value, dtype=None, name=None, as_ref=False):
-    print("wrapped_slice_to_tensor:", value)
-    #assert "flatten_with_seq_len_mask" in value.name
-    from tensorflow.python.ops import gradients_impl
-    return gradients_impl._IndexedSlicesToTensor(value, dtype=dtype, name=name, as_ref=as_ref)
-
-  from tensorflow.python.framework import ops
-  ops.register_tensor_conversion_function(tf.IndexedSlices, wrapped_slice_to_tensor, priority=1)
+  tf_version = float(tf.version.VERSION.rsplit('.', maxsplit=1)[0])  # e.g. convert "1.15.0" -> 1.15
+  if tf_version < 1.14:
```
albertz: Use `have_min_tf_version` instead.

TFNativeOp.py
```diff
@@ -243,7 +243,12 @@ def make_dim_str(c):
     for in_idx, v in enumerate(in_info):
       out_idx = v.get("want_inplace", -1)
       if out_idx >= 0:
-        code_forward_io += "context->forward_ref_input_to_ref_output(%i, %i);\n" % (in_idx, out_idx)
+        code_forward_io += """
+          // Note: forward_ref_input_to_ref_output expects the input tensor to be a reference, and not a value;
+          //   otherwise it will trigger an assertion.
+          if (IsRefType(context->input_dtype({in_idx})))
+            context->forward_ref_input_to_ref_output({in_idx}, {out_idx});
```
albertz: Now you removed the `else` part. Is that right?

tests/test_TFEngine.py
```diff
@@ -391,14 +391,16 @@ def test_engine_train_grad_noise_sparse():
...
+    def wrapped_slice_to_tensor(value, dtype=None, name=None, as_ref=False):
+      print("wrapped_slice_to_tensor:", value)
+      #assert "flatten_with_seq_len_mask" in value.name
+      from tensorflow.python.ops import gradients_impl
+      return gradients_impl._IndexedSlicesToTensor(value, dtype=dtype, name=name, as_ref=as_ref)
+
+    from tensorflow.python.framework import ops
+    ops.register_tensor_conversion_function(tf.IndexedSlices, wrapped_slice_to_tensor, priority=1)
```
albertz: I think you can just remove the code. There anyway was just some debug `print` statement.

TFUtil.py
```diff
@@ -4850,7 +4850,7 @@ def __init__(self, include_paths=(), ld_flags=(), **kwargs):
     ld_flags = list(ld_flags)
     if have_min_tf_version((1, 4)):
       # https://github.com/tensorflow/tensorflow/issues/13607
-      ld_flags += ["-L%s" % tf.sysconfig.get_lib(), "-ltensorflow_framework"]
+      ld_flags += tf.sysconfig.get_link_flags()
```
albertz: Be careful here. I think `get_link_flags()` was introduced only later (later than TF 1.4). So maybe we need another check `have_min_tf_version((1, 14))` (or whatever version) for that.

TFNativeOp.py
pzelasko: Yep - otherwise, TF quits with error and complains about more than one allocation on the output tensor. With the current state, all the tests seem to pass.

albertz:
Now you have it all in one single commit, which is not so optimal. It should better be split into multiple commits. (I just requested to cleanup the history, i.e. to squash it a bit, but not into a single commit).

TFNativeOp.py
albertz: Ah ok I see. Still not sure if that `forward_ref_input_to_ref_output` makes sense actually... But anyway, maybe we can leave it like that for now.

pzelasko:
Got it. Do you know if this can be reverted somehow? How many commits would you suggest?

albertz:
Undoing that is a bit tricky. See e.g. [here](https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits).

albertz:
Btw, you also tested that on GPU? It works fine (when you work around the nvcc absl problem manually somehow)?

.travis.yml
```diff
@@ -41,8 +41,7 @@ install:
 # TF tests first, because more relevant (for me), then ordered by alphabet
 env:
   global:
-    # TODO: TF 1.14 needs multiple fixes...
-    - TF_PACKAGE=tensorflow==1.13.1
```
albertz: I would move this change to an own commit, and it should be the last commit in this PR.

pzelasko: No problem!

pzelasko:
Haven't tested on GPU yet. Let me see how it works.

albertz:
Tests looking good, and I don't think there is any reason to not merge this in as it is now.
Thank you very much for the work!

If there are any problems which you (also @curufinwe @Spotlight0xff @pavelgolik @critias and others) observe later for TF 1.14 or TF 1.15, please report a separate issue.

PR #266, opened 2020-01-08
compile tf graph for sprint usage

ZhouW321:
compile tf graph for Spint usage, mainly specific keys and leave some logic to sprint

### DIFF
--- a/tools/compile_tf_graph.py
+++ b/tools/compile_tf_graph.py
@@ -210,13 +210,14 @@ class RecStepByStepLayer(RecLayer):
         info["collections"]["encode_ops"].append(var.init_op().name)
 
     # tile batch and group base ops
-    init_base_ops_list.append( cell.parent_tile_multiples_var.initializer )
+    init_base_ops_list.append(cell.parent_tile_multiples_var.initializer)
     info["collections"]["encode_ops"].append(cell.parent_tile_multiples_var.initializer.name)
     init_base_op = tf.group(*init_base_ops_list, name="rec_step_by_step_init_base_op")
 
     # loop vars
     for name, var in sorted(rec_layer.state_vars.items()):
-      if name.startswith("stochastic_var_") or name.startswith("base_"): continue
+      if name.startswith("stochastic_var_") or name.startswith("base_"):
+        continue
       # global vars (hypothesis independent)
       if name.startswith("global_"):
         global_vars_coll.append(var.var)
@@ -235,7 +236,7 @@ class RecStepByStepLayer(RecLayer):
         info["collections"]["post_update_ops"].append(var.final_op().name)
 
     # group encode_ops
-    with tf.control_dependencies( [init_base_op] ):
+    with tf.control_dependencies([init_base_op]):
       init_op = tf.group(*init_ops_list, name="rec_step_by_step_init_op")
     encode_ops_coll.append( init_op )
  
@@ -330,7 +331,7 @@ class RecStepByStepLayer(RecLayer):
         [d if (d is not None) else 1 for d in self.var_data_shape.batch_shape],
         dtype=self.var_data_shape.dtype)
       zero_initializer.set_shape(self.var_data_shape.batch_shape)
-      # (state) variable collections #
+      # (state) variable collections 
       var_collections = [tf.GraphKeys.GLOBAL_VARIABLES]
       if choice_dependent is False:
         var_collections.append(CollectionKeys.STATE_VARS)
@@ -438,7 +439,7 @@ class RecStepByStepLayer(RecLayer):
     :param str name:
     :param tf.Tensor|None initial_value: assumes batch-major, if data_shape is not given
     :param Data|None data_shape:
-    :param bool|None:
+    :param bool|None choice_dependent: choice dependent or not relevant
     :rtype: tf.Tensor
     """
     assert name not in self.state_vars
@@ -662,6 +663,9 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
   """
 
   def __init__(self, parent_rec_layer, **kwargs):
+    """
+    :param RecStepByStepLayer:
+    """
     self._parent_layers = {}  # type: typing.Dict[str,WrappedInternalLayer]
     with helper_variable_scope():
       self.parent_tile_multiples_var = tf.get_variable(
@@ -697,8 +701,8 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
     """
     make state vars for input from outside the loop (automatic tile_batch)
     :param str name: layer name
-    :param output: layer output data
-    :rtype: converted output data 
+    :param output: Data
+    :rtype: Data
     """
     rec_layer = self.parent_rec_layer
     from TFUtil import tile_transposed
@@ -777,7 +781,7 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
             choice_dependent[-1] = True
         prev_outputs_data.append(self.layer_data_templates[k].output)
 
-      def parse_recursive( nvs ):
+      def parse_recursive(nvs):
         if isinstance(nvs, (tuple, list)):
           return [ parse_recursive(v) for v in nvs ]
         if isinstance(nvs, tf.Tensor):
--- a/tools/compile_tf_graph.py
+++ b/tools/compile_tf_graph.py
@@ -11,6 +11,7 @@ from __future__ import print_function
 import typing
 import os
 import sys
+import contextlib
 import tensorflow as tf
 from tensorflow.python.framework import graph_io
 
@@ -24,7 +25,7 @@ from Config import Config
 import argparse
 import Util
 from Util import NotSpecified
-from TFUtil import Data
+from TFUtil import Data, CollectionKeys
 from TFNetwork import TFNetwork
 from TFNetworkLayer import LayerBase, register_layer_class, WrappedInternalLayer
 from TFNetworkRecLayer import RecLayer, _SubnetworkRecCell, ChoiceLayer
@@ -79,6 +80,16 @@ def create_graph(train_flag, eval_flag, search_flag, net_dict):
   return network
 
 
+@contextlib.contextmanager
+def helper_variable_scope():
+  """
+  :rtype: tf.VariableScope
+  """
+  from TFUtil import reuse_name_scope
+  with reuse_name_scope("IO", absolute=True) as scope:
+    yield scope
+
+
 class RecStepByStepLayer(RecLayer):
   """
   Represents a single step of :class:`RecLayer`.
@@ -114,9 +125,25 @@ class RecStepByStepLayer(RecLayer):
       These are also the state vars which will get updated in every further recurrent step.
       The "base_*" state vars are always kept (although you might need to update the batch dim),
       and the "stochastic_var_*" state vars have the special logic for the stochastic variables.
-      This is "next_step_op" in the info json.
+      This is "update_op" in the info json.
   """
   layer_class = "rec_step_by_step"
+
+  def __init__(self, **kwargs):
+    kwargs = kwargs.copy()
+    kwargs["optimize_move_layers_out"] = False
+    sub_net_dict = kwargs["unit"]
+    assert isinstance(sub_net_dict, dict)
+    for key, layer_dict in list(sub_net_dict.items()):
+      assert isinstance(layer_dict, dict)
+      if layer_dict["class"] == "choice":
+        layer_dict = layer_dict.copy()
+        layer_dict["class"] = ChoiceStateVarLayer.layer_class
+        sub_net_dict[key] = layer_dict
+    kwargs["unit"] = sub_net_dict
+    self.state_vars = {}  # type: typing.Dict[str,RecStepByStepLayer.StateVar]
+    self.stochastic_var_order = []  # type: typing.List[str]
+    super(RecStepByStepLayer, self).__init__(**kwargs)
 
   @classmethod
   def prepare_compile(cls, rec_layer_name, net_dict):
@@ -143,13 +170,37 @@ class RecStepByStepLayer(RecLayer):
     assert rec_layer_name in network.layers
     rec_layer = network.layers[rec_layer_name]
     assert isinstance(rec_layer, RecStepByStepLayer)
-    info = {"state_vars": {}, "stochastic_var_order": [], "stochastic_vars": {}}
-    init_ops = []
-    tile_batch_repetitions = tf.placeholder(name="tile_batch_repetitions", shape=(), dtype=tf.int32)
-    tile_batch_ops = []
-    src_beams = tf.placeholder(name="src_beams", shape=(None, None), dtype=tf.int32)  # (batch,beam)
-    select_src_beams_ops = []
-    next_step_ops = []
+    cell = rec_layer.cell
+    assert isinstance(cell, SubnetworkRecCellSingleStep)
+
+    # tmp structures
+    init_base_ops_list = []
+    init_ops_list = []
+    update_ops_list = []
+    post_update_ops_list = []
+
+    # specific collections needed by sprint #
+    update_ops_coll = tf.get_collection_ref("update_ops")
+    post_update_ops_coll = tf.get_collection_ref("post_update_ops")
+    encode_ops_coll = tf.get_collection_ref("encode_ops")
+    decode_ops_coll = tf.get_collection_ref("decode_ops")
+    decoder_input_vars_coll = tf.get_collection_ref("decoder_input_vars")
+    decoder_output_vars_coll = tf.get_collection_ref("decoder_output_vars")
+    global_vars_coll = tf.get_collection_ref("global_vars")
+
+    # json info #
+    info = { "state_vars" : {}, "stochastic_var_order": [], "stochastic_vars": {},
+             "collections" : { "encode_ops" : [],
+                               "update_ops" : [],
+                               "post_update_ops" : [],
+                               "decode_ops" : [],
+                               "decoder_input_vars" : [],
+                               "decoder_output_vars" : [],
+                               "global_vars" : [] 
+                             }
+           }
+
+    # base vars #
     print("State vars:")
     for name, var in sorted(rec_layer.state_vars.items()):
       assert isinstance(name, str)
@@ -159,21 +210,48 @@ class RecStepByStepLayer(RecLayer):
         "var_op": var.var.op.name,
         "shape": [int(d) if d is not None else None for d in var.var_data_shape.batch_shape],
         "dtype": var.var.dtype.base_dtype.name}
-      if not name.startswith("stochastic_var_"):
-        init_ops.append(var.init_op())
-        tile_batch_ops.append(var.tile_batch_op(tile_batch_repetitions))
-        select_src_beams_ops.append(var.select_src_beams_op(src_beams=src_beams))
-      if not name.startswith("stochastic_var_") and not name.startswith("base_"):
-        next_step_ops.append(var.final_op())
-    info["init_op"] = tf.group(*init_ops, name="rec_step_by_step_init_op").name
-    info["next_step_op"] = tf.group(*next_step_ops, name="rec_step_by_step_update_op").name
-    info["tile_batch"] = {
-      "op": tf.group(*tile_batch_ops, name="rec_step_by_step_tile_batch_op").name,
-      "repetitions_placeholder": tile_batch_repetitions.op.name}
-    info["select_src_beams"] = {
-      "op": tf.group(*select_src_beams_ops, name="rec_step_by_step_select_src_beams_op").name,
-      "src_beams_placeholder": src_beams.op.name}
+      assert var.var not in network.get_saveable_params_list(), 'StateVars are not restorable from the original model check point'
+      if name.startswith("base_"):
+        init_base_ops_list.append(var.init_op())
+        info["collections"]["encode_ops"].append(var.init_op().name)
+
+    # tile batch and group base ops #
+    init_base_ops_list.append( cell.parent_tile_multiples_var.initializer )
+    info["collections"]["encode_ops"].append(cell.parent_tile_multiples_var.initializer.name)
+    init_base_op = tf.group(*init_base_ops_list, name="rec_step_by_step_init_base_op")
+
+    # loop vars #
+    for name, var in sorted(rec_layer.state_vars.items()):
+      if name.startswith("stochastic_var_") or name.startswith("base_"): continue
+      # global vars (hypothesis independent)
+      if name.startswith("global_"):
+        global_vars_coll.append(var.var)
+        info["collections"]["global_vars"].append(var.name)
+      # dependency on init_base_op for each init_op
+      with tf.control_dependencies([init_base_op]):  
+        init_ops_list.append(var.init_op())
+      info["collections"]["encode_ops"].append(var.init_op().name)
+      # choice-dependent loop vars are not I/O stored, but need to be updated after feeding choices
+      if var.choice_dependent:
+        update_ops_list.append(var.final_op())
+        info["collections"]["update_ops"].append(var.final_op().name)
+      # update loop vars after the last decode op
+      elif var.choice_dependent is False:
+        post_update_ops_list.append(var.final_op())
+        info["collections"]["post_update_ops"].append(var.final_op().name)
+
+    # group encode_ops #
+    with tf.control_dependencies( [init_base_op] ):
+      init_op = tf.group(*init_ops_list, name="rec_step_by_step_init_op")
+    encode_ops_coll.append( init_op )
+ 
+    # group post_update_ops #
+    post_update_op = tf.group(*post_update_ops_list, name="rec_step_by_step_post_update_op")
+    post_update_ops_coll.append( post_update_op )
+
+    # stochastic vars # 
     print("Stochastic vars, and their order:")
+    tile_batch_multiple = None
     for name in rec_layer.stochastic_var_order:
       print(" %s" % name)
       info["stochastic_var_order"].append(name)
@@ -181,12 +259,37 @@ class RecStepByStepLayer(RecLayer):
         "calc_scores_op": rec_layer.state_vars["stochastic_var_scores_%s" % name].final_op().name,
         "scores_state_var": "stochastic_var_scores_%s" % name,
         "choice_state_var": "stochastic_var_choice_%s" % name}
+      # collect decode_input_vars
+      input_var = rec_layer.state_vars["stochastic_var_choice_%s" % name].var
+      decoder_input_vars_coll.append(input_var)
+      info["collections"]["decoder_input_vars"].append(input_var.name)
+      # collect decode_output_vars
+      output_var = rec_layer.state_vars["stochastic_var_scores_%s" % name].var
+      decoder_output_vars_coll.append(output_var)
+      info["collections"]["decoder_output_vars"].append(output_var.name)
+      # automatic tile encodings w.r.t. choices
+      if tile_batch_multiple is None:
+        tile_batch_multiple = rec_layer.state_vars["stochastic_var_choice_%s" % name].read_batch_dim()
+      # collect decode_ops
+      decode_op = rec_layer.state_vars["stochastic_var_scores_%s" % name].final_op()
+      decode_ops_coll.append(decode_op)
+      info["collections"]["decode_ops"].append(decode_op.name)
+
+    # automatic tile encodings w.r.t. choices
+    tile_op = cell.set_parent_tile_multiple_op(tile_batch_multiple)
+    update_ops_list.append(tile_op)
+    info["collections"]["update_ops"].append(tile_op.name)
+    
+    # group update ops
+    update_op = tf.group(*update_ops_list, name="rec_step_by_step_update_op")
+    update_ops_coll.append( update_op )
+
     import json
     info_str = json.dumps(info, sort_keys=True, indent=2)
-    print("JSON:")
-    print(info_str)
     if not output_file_name:
       print("No rec-step-by-step output file name specified, not storing this info.")
+      print("JSON:")
+      print(info_str)
     else:
       with open(output_file_name, "w") as f:
         f.write(info_str)
@@ -197,7 +300,7 @@ class RecStepByStepLayer(RecLayer):
     Represents a state variable, i.e. either a state, a choice, or encoder state, etc.
     """
 
-    def __init__(self, parent, name, initial_value, data_shape):
+    def __init__(self, parent, name, initial_value, data_shape, choice_dependent):
       """
       :param RecStepByStepLayer parent:
       :param str name:
@@ -208,9 +311,11 @@ class RecStepByStepLayer(RecLayer):
         Describes the shape of initial_value, and also what we store as self.orig_data_shape,
         and what we return by self.read().
         If it is not a scalar, and batch-dim-axis > 0, the created variable will still be in batch-major.
+      :param bool|None choice_dependent: choice dependent or not relevant
       """
       self.parent = parent
       self.name = name
+      self.choice_dependent = choice_dependent
       self.orig_data_shape = data_shape
       self.var_data_shape = data_shape.copy_as_batch_major() if data_shape.batch_dim_axis is not None else data_shape
       del data_shape
@@ -231,7 +336,12 @@ class RecStepByStepLayer(RecLayer):
         [d if (d is not None) else 1 for d in self.var_data_shape.batch_shape],
         dtype=self.var_data_shape.dtype)
       zero_initializer.set_shape(self.var_data_shape.batch_shape)
-      self.var = tf.get_variable(name=name, initializer=zero_initializer, validate_shape=False)  # type: tf.Variable
+      # (state) variable collections #
+      var_collections = [tf.GraphKeys.GLOBAL_VARIABLES]
+      if choice_dependent is False:
+        var_collections.append(CollectionKeys.STATE_VARS)
+      with helper_variable_scope():
+        self.var = tf.get_variable(name=name, initializer=zero_initializer, validate_shape=False, collections=var_collections)  # type: tf.Variable
       self.var.set_shape(self.var_data_shape.batch_shape)
       assert self.var.shape.as_list() == list(self.var_data_shape.batch_shape)
       print("New state var %r: %s, shape %s" % (name, self.var, self.var_data_shape))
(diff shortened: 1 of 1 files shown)


tools/compile_tf_graph.py
```diff
@@ -143,13 +170,37 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    info = { "state_vars" : {}, "stochastic_var_order": [], "stochastic_vars": {},
+             "collections" : { "encode_ops" : [],
+                               "update_ops" : [],
+                               "post_update_ops" : [],
+                               "decode_ops" : [],
+                               "decoder_input_vars" : [],
+                               "decoder_output_vars" : [],
+                               "global_vars" : [] 
+                             }
+           }
```
albertz: Do not format code like this. It's completely wrong. (PyCharm should warn about this, or not?)
Make a newline after `{` if you need multiple lines. The indentation should not be like that, only 2 spaces more in each line. There are also too much white spaces (e.g. there should be no white spaces around the dict keys).

tools/compile_tf_graph.py
```diff
@@ -143,13 +170,37 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    # specific collections needed by sprint #
+    update_ops_coll = tf.get_collection_ref("update_ops")
+    post_update_ops_coll = tf.get_collection_ref("post_update_ops")
+    encode_ops_coll = tf.get_collection_ref("encode_ops")
+    decode_ops_coll = tf.get_collection_ref("decode_ops")
+    decoder_input_vars_coll = tf.get_collection_ref("decoder_input_vars")
+    decoder_output_vars_coll = tf.get_collection_ref("decoder_output_vars")
+    global_vars_coll = tf.get_collection_ref("global_vars")
+
+    # json info #
```
albertz: Do not make comments like this. Why the ending `#`? Make it consistent to like all the other comments in the code.

tools/compile_tf_graph.py
```diff
@@ -741,9 +911,7 @@ def main(argv):
       from TFUtil import CollectionKeys
       with open(args.output_file_state_vars_list, "w") as f:
         for param in tf.get_collection(CollectionKeys.STATE_VARS):
-          assert param.name[-2:] == ":0"
-          f.write("%s\n" % param.name[:-2])
-
-
+          f.write("%s\n" % param.name)
+
```
albertz: I think there is one too less empty line here. Why did you remove that? It should be two empty lines. (PyCharm should tell you that, or not?)

tools/compile_tf_graph.py
```diff
@@ -741,9 +911,7 @@ def main(argv):
       from TFUtil import CollectionKeys
       with open(args.output_file_state_vars_list, "w") as f:
         for param in tf.get_collection(CollectionKeys.STATE_VARS):
-          assert param.name[-2:] == ":0"
-          f.write("%s\n" % param.name[:-2])
-
-
+          f.write("%s\n" % param.name)
```
albertz: Why this change?

tools/compile_tf_graph.py
```diff
@@ -644,9 +812,10 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
         seq_len_info = None
       else:
         i, net_vars, acc_tas, seq_len_info = res
-    if seq_len_info:
-      rec_layer.set_state_vars_final_values_recursive(["end_flag", "dyn_seq_len"], seq_len_info)
-    rec_layer.set_state_var_final_value("i", i)
+
+    # make it setable by sprint
+    rec_layer.set_state_var_final_value("global_step", i)
+
```
albertz: Please undo this change. This should not be needed. Also end_flag, dyn_seq_len should still be there. There can be models depending on that.

tools/compile_tf_graph.py
```diff
@@ -612,16 +766,32 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
...
+      prev_outputs_data = list()
+      choice_dependent = list()
+      for k in sorted(self._initial_outputs.keys()):
+        choice_dependent.append(False)
+        for source_layer in self.net_dict[k]['from']:
+          if source_layer in self.net_dict.keys() and self.net_dict[source_layer]['class'] == ChoiceStateVarLayer.layer_class:
+            choice_dependent[-1] = True
+        prev_outputs_data.append(self.layer_data_templates[k].output)
+
+      def parse_recursive( nvs ):
```
albertz: The formatting is off. Why the extra white spaces in the params? (PyCharm should tell you that, nor not?)

tools/compile_tf_graph.py
```diff
@@ -612,16 +766,32 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
...
+        if isinstance(nvs, (tuple, list)):
+          return [ parse_recursive(v) for v in nvs ]
+        if isinstance(nvs, tf.Tensor):
+          return False
+
+      assert len(net_vars) == 2
+      state_choice_dependent = parse_recursive( net_vars[1] )
```
albertz: Wrong formatting. (PyCharm should tell you that?)

tools/compile_tf_graph.py
```diff
@@ -612,16 +766,32 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
...
     with tf.name_scope("state"):
       # See _SubnetworkRecCell.GetOutput.body.
       # net_vars is (prev_outputs_flat, prev_extra_flat), and prev_outputs_flat corresponds to self._initial_outputs,
       # where the keys are the layer names.
       # prev_extra is always expected to be batch-major.
-      prev_outputs_data = [self.layer_data_templates[k].output for k in sorted(self._initial_outputs.keys())]
-      net_vars = rec_layer.create_state_vars_recursive(
-        name_prefix="state", initial_values=net_vars, data_shape=(prev_outputs_data, None))
+      prev_outputs_data = list()
+      choice_dependent = list()
```
albertz: This is not idiomatic Python code. (PyCharm should tell you that, or not?)
Use `[]` instead of `list()`.

tools/compile_tf_graph.py
```diff
@@ -612,16 +766,32 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
       seq_len_info = None
     else:
       i, net_vars, acc_tas, seq_len_info = loop_vars
-      seq_len_info = rec_layer.create_state_vars_recursive(["end_flag", "dyn_seq_len"], seq_len_info)
-    i = rec_layer.create_state_var("i", i)
```
albertz: Please leave that in. Also, why did you move `create_state_var("i", i)` down? Leave it just right here.

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
     assert isinstance(rec_layer, RecStepByStepLayer)
     output = layer.output.copy()
-    output.placeholder = rec_layer.create_state_var(
-      name="base_value_%s" % layer_name, initial_value=output.placeholder, data_shape=output)
+    output = self.make_base_state_vars(layer_name, output)
+    layer = WrappedInternalLayer(name=layer_name, network=self.parent_net, output=output, base_layer=layer)
+    self._parent_layers[layer_name] = layer
+    return layer
+
+  # make state vars for input from outside the loop (automatic tile_batch)
```
albertz: Functions are not documented that way. That is what Python doc strings are for.

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+  def make_base_state_vars(self, name, output):
```
albertz: Arguments are not documented (`:param str name:` etc).

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+      new_size = rec_layer.create_state_var(name="base_size%i_%s" % (i, name), initial_value=size)
+      new_size = tile_transposed(new_size, axis=0, multiples=self.parent_tile_multiples_var.read_value())
       dim_tag.set_tag_on_size_tensor(new_size)
       output.size_placeholder[i] = new_size
-    layer = WrappedInternalLayer(name=layer_name, network=self.parent_net, output=output, base_layer=layer)
-    self._parent_layers[layer_name] = layer
-    return layer
+    return output
+ 
+
```
albertz: Why two empty lines? It should be one empty line here. (PyCharm should tell you that, or not?)

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+  def _get_extern_data(self, data, i):
+    """
+    :param dict[str,tf.Tensor] data: All data needed from outside of the loop. see _construct()
+    :param tf.Tensor i: loop counter. scalar, int32, current step (time)
```
albertz: It does not return anything?

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+    for key in data:
+      if not key == 'source':
+        self.net.extern_data.data[key].placeholder = data[key]
+    # create base state vars for encodings
+    if 'source' in data:
+      rec_layer_input = self.parent_rec_layer.input_data.copy_as_batch_major()
+      output = self.make_base_state_vars('source', rec_layer_input)
+      # slice specific time step TODO put before tile batch
+      self.net.extern_data.data['source'].placeholder = tf.gather(output.placeholder, i, axis=output.time_dim_axis)
+
```
albertz: Why an extra empty line?

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+    """
+    for key in data:
+      if not key == 'source':
+        self.net.extern_data.data[key].placeholder = data[key]
+    # create base state vars for encodings
+    if 'source' in data:
+      rec_layer_input = self.parent_rec_layer.input_data.copy_as_batch_major()
+      output = self.make_base_state_vars('source', rec_layer_input)
+      # slice specific time step TODO put before tile batch
+      self.net.extern_data.data['source'].placeholder = tf.gather(output.placeholder, i, axis=output.time_dim_axis)
```
albertz: The function looks incomplete?

tools/compile_tf_graph.py
```diff
@@ -568,21 +695,48 @@ def _get_parent_layer(self, layer_name):
...
+    :param tf.Tensor i: loop counter. scalar, int32, current step (time)
+    """
+    for key in data:
+      if not key == 'source':
+        self.net.extern_data.data[key].placeholder = data[key]
+    # create base state vars for encodings
+    if 'source' in data:
+      rec_layer_input = self.parent_rec_layer.input_data.copy_as_batch_major()
+      output = self.make_base_state_vars('source', rec_layer_input)
+      # slice specific time step TODO put before tile batch
```
albertz: What about this todo? Is this out-dated? Or not relevant? Or totally broken?

albertz:
Also, in the future, try to make much smaller commits. This will probably take now a few iterations and some work and time to get it all right. Smaller more frequent commits would make it much easier for both of us.

tools/compile_tf_graph.py
ZhouW321: this is needed now by sprint to control which input to read. end_flag will be useless, because ending logic is handled by sprint. not sure if dyn_seq_len is used at all

tools/compile_tf_graph.py
ZhouW321: see _SubnetworkRecCell: _get_extern_data

tools/compile_tf_graph.py
ZhouW321: this is nice to have to save memroy, but not critical

tools/compile_tf_graph.py
ZhouW321: stat vars created locally does not have :0 at the end

tools/compile_tf_graph.py
ZhouW321: please check the line before

tools/compile_tf_graph.py
ZhouW321: check line before

tools/compile_tf_graph.py
ZhouW321: I want the end_flag to be always false and dyn_seq_len always 0 as they will be now. Save mem from whatever is useless to be a stat_var

tools/compile_tf_graph.py
albertz: What is the type of `param`? `tf.Tensor.name` always has this. It might be a `tf.Variable` then. But then better check thet, i.e. add an `assert isinstance(param, tf.Variable)`.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
-    info["select_src_beams"] = {
-      "op": tf.group(*select_src_beams_ops, name="rec_step_by_step_select_src_beams_op").name,
-      "src_beams_placeholder": src_beams.op.name}
+      assert var.var not in network.get_saveable_params_list(), 'StateVars are not restorable from the original model check point'
+      if name.startswith("base_"):
+        init_base_ops_list.append(var.init_op())
+        info["collections"]["encode_ops"].append(var.init_op().name)
+
+    # tile batch and group base ops #
+    init_base_ops_list.append( cell.parent_tile_multiples_var.initializer )
```
albertz: There is an extra wrong white space before the bracket.

tools/compile_tf_graph.py
```diff
@@ -143,13 +170,37 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    cell = rec_layer.cell
+    assert isinstance(cell, SubnetworkRecCellSingleStep)
+
+    # tmp structures
+    init_base_ops_list = []
+    init_ops_list = []
+    update_ops_list = []
+    post_update_ops_list = []
+
+    # specific collections needed by sprint #
```
albertz: Wrong comment format.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+      if name.startswith("base_"):
+        init_base_ops_list.append(var.init_op())
+        info["collections"]["encode_ops"].append(var.init_op().name)
+
+    # tile batch and group base ops #
+    init_base_ops_list.append( cell.parent_tile_multiples_var.initializer )
+    info["collections"]["encode_ops"].append(cell.parent_tile_multiples_var.initializer.name)
+    init_base_op = tf.group(*init_base_ops_list, name="rec_step_by_step_init_base_op")
+
+    # loop vars #
```
albertz: Wrong comment format.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    for name, var in sorted(rec_layer.state_vars.items()):
+      if name.startswith("stochastic_var_") or name.startswith("base_"): continue
```
albertz: `continue` should go into another line. (PyCharm should tell you that.)

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+      # choice-dependent loop vars are not I/O stored, but need to be updated after feeding choices
+      if var.choice_dependent:
+        update_ops_list.append(var.final_op())
+        info["collections"]["update_ops"].append(var.final_op().name)
+      # update loop vars after the last decode op
+      elif var.choice_dependent is False:
+        post_update_ops_list.append(var.final_op())
+        info["collections"]["post_update_ops"].append(var.final_op().name)
+
+    # group encode_ops #
```
albertz: Wrong comment format.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    with tf.control_dependencies( [init_base_op] ):
```
albertz: Why the extra white space?

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+      init_op = tf.group(*init_ops_list, name="rec_step_by_step_init_op")
+    encode_ops_coll.append( init_op )
+ 
+    # group post_update_ops #
```
albertz: Wrong comment format.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    post_update_op = tf.group(*post_update_ops_list, name="rec_step_by_step_post_update_op")
+    post_update_ops_coll.append( post_update_op )
+
+    # stochastic vars #
```
albertz: Wrong comment format.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+        info["collections"]["post_update_ops"].append(var.final_op().name)
+
+    # group encode_ops #
+    with tf.control_dependencies( [init_base_op] ):
+      init_op = tf.group(*init_ops_list, name="rec_step_by_step_init_op")
+    encode_ops_coll.append( init_op )
+ 
+    # group post_update_ops #
+    post_update_op = tf.group(*post_update_ops_list, name="rec_step_by_step_post_update_op")
+    post_update_ops_coll.append( post_update_op )
```
albertz: Why the extra white space?

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    info["collections"]["update_ops"].append(tile_op.name)
+    
+    # group update ops
+    update_op = tf.group(*update_ops_list, name="rec_step_by_step_update_op")
+    update_ops_coll.append( update_op )
+
     import json
     info_str = json.dumps(info, sort_keys=True, indent=2)
-    print("JSON:")
-    print(info_str)
```
albertz: I intentionally had this here, always, just to have some nicer output, to see the structure, what is being saved.

tools/compile_tf_graph.py
```diff
@@ -159,34 +210,86 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+      # dependency on init_base_op for each init_op
+      with tf.control_dependencies([init_base_op]):  
+        init_ops_list.append(var.init_op())
+      info["collections"]["encode_ops"].append(var.init_op().name)
+      # choice-dependent loop vars are not I/O stored, but need to be updated after feeding choices
+      if var.choice_dependent:
+        update_ops_list.append(var.final_op())
+        info["collections"]["update_ops"].append(var.final_op().name)
+      # update loop vars after the last decode op
+      elif var.choice_dependent is False:
```
albertz: Such a bool check is wrong Python style. (Please really check the PyCharm comments, it will tell you that.)

tools/compile_tf_graph.py
ZhouW321: this is fixed in the previous commit

tools/compile_tf_graph.py
ZhouW321: fixed in last commit

tools/compile_tf_graph.py
ZhouW321: it's very annoying everytime full screen output. output file can be checked to see the nice output

tools/compile_tf_graph.py
albertz: This takes 8 bytes of memory, so really you can ignore that. Also, there are models which can potentially depend on that, and that would make it more complicated then.

albertz:
When you fixed something, can you mark my comment as resolved?

tools/compile_tf_graph.py
```diff
@@ -231,7 +330,12 @@ def __init__(self, parent, name, initial_value, data_shape):
         [d if (d is not None) else 1 for d in self.var_data_shape.batch_shape],
         dtype=self.var_data_shape.dtype)
       zero_initializer.set_shape(self.var_data_shape.batch_shape)
-      self.var = tf.get_variable(name=name, initializer=zero_initializer, validate_shape=False)  # type: tf.Variable
+      # (state) variable collections #
```
albertz: Wrong format.

tools/compile_tf_graph.py
```diff
@@ -231,7 +330,12 @@ def __init__(self, parent, name, initial_value, data_shape):
...
+      var_collections = [tf.GraphKeys.GLOBAL_VARIABLES]
+      if choice_dependent is False:
```
albertz: Wrong bool check. (Please see PyCharm comments.)

tools/compile_tf_graph.py
```diff
@@ -268,7 +381,7 @@ def init_op(self):
       :rtype: tf.Operation
       """
       assert self.var_initial_value is not None
-      return tf.assign(self.var, self.var_initial_value, name="init_state_var_%s" % self.name).op
+      return tf.assign(self.var, self.var_initial_value, name="init_state_var_%s" % self.name, validate_shape=False).op
```
albertz: Why is that `validate_shape` needed? It should not be needed I think.

tools/compile_tf_graph.py
```diff
@@ -289,7 +402,7 @@ def final_op(self):
         x.placeholder = value
         x = x.copy_compatible_to(self.var_data_shape)
         value = x.placeholder
-      return tf.assign(self.var, value, name="final_state_var_%s" % self.name).op
+      return tf.assign(self.var, value, name="final_state_var_%s" % self.name, validate_shape=False).op
```
albertz: Why is that `validate_shape` needed? It should not be needed I think.

tools/compile_tf_graph.py
```diff
@@ -317,30 +430,15 @@ def select_src_beams_op(self, src_beams):
...
-  def create_state_var(self, name, initial_value=None, data_shape=None):
+  def create_state_var(self, name, initial_value=None, data_shape=None, choice_dependent=None):
     """
     A state var is a variable where the initial value is given by the encoder, or a constant,
     and the final value is determined by one step of this rec layer (usually called the decoder).
 
     :param str name:
     :param tf.Tensor|None initial_value: assumes batch-major, if data_shape is not given
     :param Data|None data_shape:
+    :param bool|None:
```
albertz: There is something wrong here.

tools/compile_tf_graph.py
```diff
@@ -368,11 +466,12 @@ def set_state_var_final_value(self, name, final_value):
     """
     self.state_vars[name].set_final_value(final_value)
 
-  def create_state_vars_recursive(self, name_prefix, initial_values, data_shape=None):
+  def create_state_vars_recursive(self, name_prefix, initial_values, data_shape=None, choice_dependent=None):
```
albertz: Why is this `choice_dependent` needed?

tools/compile_tf_graph.py
```diff
@@ -317,30 +430,15 @@ def select_src_beams_op(self, src_beams):
...
-        layer_dict = layer_dict.copy()
-        layer_dict["class"] = ChoiceStateVarLayer.layer_class
-        sub_net_dict[key] = layer_dict
-    kwargs["unit"] = sub_net_dict
-    self.state_vars = {}  # type: typing.Dict[str,RecStepByStepLayer.StateVar]
-    self.stochastic_var_order = []  # type: typing.List[str]
-    super(RecStepByStepLayer, self).__init__(**kwargs)
-
-  def create_state_var(self, name, initial_value=None, data_shape=None):
+  def create_state_var(self, name, initial_value=None, data_shape=None, choice_dependent=None):
```
albertz: Why is this `choice_dependent` needed? What does it do?

tools/compile_tf_graph.py
```diff
@@ -501,6 +610,7 @@ def __init__(self, beam_size,
     source = self.sources[0]
     assert source.output.is_batch_major and len(source.output.shape) == 1
     scores_in = source.output.placeholder
+    # TODO infer input_type from source layer class 'softmax' or 'linear' with 'log_softmax' activation
```
albertz: What do you mean? There is nothing to be inferred. This is configured.

tools/compile_tf_graph.py
```diff
@@ -551,9 +661,19 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
   Adapts :class:`_SubnetworkRecCell` such that we execute only a single step.
   """
 
-  def __init__(self, **kwargs):
+  def __init__(self, parent_rec_layer, **kwargs):
```
albertz: Param is not documented.

tools/compile_tf_graph.py
```diff
@@ -568,22 +688,52 @@ def _get_parent_layer(self, layer_name):
...
-    output.placeholder = rec_layer.create_state_var(
-      name="base_value_%s" % layer_name, initial_value=output.placeholder, data_shape=output)
+    output = self.make_base_state_vars(layer_name, output)
+    layer = WrappedInternalLayer(name=layer_name, network=self.parent_net, output=output, base_layer=layer)
+    self._parent_layers[layer_name] = layer
+    return layer
+
+  def make_base_state_vars(self, name, output):
+    """
+    make state vars for input from outside the loop (automatic tile_batch)
```
albertz: There should be an empty line after the comment.

tools/compile_tf_graph.py
```diff
@@ -568,22 +688,52 @@ def _get_parent_layer(self, layer_name):
...
+    :param str name: layer name
+    :param output: layer output data
```
albertz: What is the type?

tools/compile_tf_graph.py
```diff
@@ -568,22 +688,52 @@ def _get_parent_layer(self, layer_name):
...
+    :rtype: converted output data
```
albertz: The format/syntax is wrong. `:rtype:` specifies the return type.

tools/compile_tf_graph.py
ZhouW321: forgot why, but it's needed during some test before with the attention model config

tools/compile_tf_graph.py
ZhouW321: same

tools/compile_tf_graph.py
albertz: You probably mean `:return:` here, which describes what is being returned. But then `:rtype:` is missing.

tools/compile_tf_graph.py
```diff
@@ -143,13 +170,31 @@ def post_compile(cls, rec_layer_name, network, output_file_name=None):
...
+    decode_ops_coll = tf.get_collection_ref("decode_ops")
+    decoder_input_vars_coll = tf.get_collection_ref("decoder_input_vars")
+    decoder_output_vars_coll = tf.get_collection_ref("decoder_output_vars")
+    global_vars_coll = tf.get_collection_ref("global_vars")
+
+    # json info
+    info = { "state_vars": {}, "stochastic_var_order": [], "stochastic_vars": {},
+             "collections": { "encode_ops": [], "update_ops": [], "post_update_ops": [],
+               "decode_ops": [], "decoder_input_vars": [], "decoder_output_vars": [],
+               "global_vars": [] } }
```
albertz: The formatting is very wrong (as I previously already explained).

tools/compile_tf_graph.py
ZhouW321: these state vars which can be infered from choice does not need to be saved into Sprint CPU

tools/compile_tf_graph.py
```diff
@@ -197,7 +294,7 @@ class StateVar:
     Represents a state variable, i.e. either a state, a choice, or encoder state, etc.
     """
 
-    def __init__(self, parent, name, initial_value, data_shape):
+    def __init__(self, parent, name, initial_value, data_shape, choice_dependent):
```
albertz: I don't understand why `choice_dependent` is needed? It should not be needed. Keep the code simple for now.

tools/compile_tf_graph.py
albertz: Remove it again. It should not be needed. Or document why it really is needed.

tools/compile_tf_graph.py
ZhouW321: you mean input_type is 'prob' or 'log_prob' is automatically detected when creating this ChoiceStateVarLayer ? I didn't see this

tools/compile_tf_graph.py
albertz: I don't understand. Everything can be inferred from choice and current state?

tools/compile_tf_graph.py
```diff
@@ -612,16 +762,32 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
...
-      prev_outputs_data = [self.layer_data_templates[k].output for k in sorted(self._initial_outputs.keys())]
-      net_vars = rec_layer.create_state_vars_recursive(
-        name_prefix="state", initial_values=net_vars, data_shape=(prev_outputs_data, None))
+      prev_outputs_data = []
+      choice_dependent = []
+      for k in sorted(self._initial_outputs.keys()):
+        choice_dependent.append(False)
+        for source_layer in self.net_dict[k]['from']:
+          if source_layer in self.net_dict.keys() and self.net_dict[source_layer]['class'] == ChoiceStateVarLayer.layer_class:
+            choice_dependent[-1] = True
```
albertz: This code looks very much wrong. There can be *many* different cases how some layer can depend on another layer directly, not just via `from`. And then it can also depend on it indirectly.

tools/compile_tf_graph.py
ZhouW321: why this one param makes it so complicated, it saves a lot of RAM when I use a large beam

tools/compile_tf_graph.py
albertz: What do you mean? The formatting is wrong.

tools/compile_tf_graph.py
albertz: I don't understand. The formatting is wrong.

tools/compile_tf_graph.py
albertz: Then don't write this as TODO, but make some Note instead or so, sth like: `# Note: we could save some memory by doing ...`.

tools/compile_tf_graph.py
albertz: No, input_type is something which is configured. This is not inferred. And this is by design.

tools/compile_tf_graph.py
ZhouW321: only direct ones need to be computed after choices are assigned. agree there maybe nicer way to do this dependency check. but all these are state vars which either have to be assigned by Sprint or infered from others

tools/compile_tf_graph.py
albertz: Is this fixed now? GitHub says outdated. Please mark as resolved when you fixed sth.

tools/compile_tf_graph.py
albertz: I don't understand. Why does it matter for the state var if there is some other layer (or activation function, or whatever) in between or not?
Also, even as a direct layer dependency check, this code is wrong. If the layer is already constructed, you can use `layer.get_dep_layers()`. You should definitely not read from `net_dict`, and hardcode some strange assumptions here. This will definitely easily break for many other configs.

tools/compile_tf_graph.py
albertz: See the other discussion why it becomes complicated. At least if you want to have it correct. But I would definitely be against some incorrect or not generic solution here. This would also become terrible for debugging later, e.g. some new config where some of your assumptions become wrong, but this script still silently returns sth, but the Sprint search behaves buggy then, e.g. also returns sth, but not correctly.

tools/compile_tf_graph.py
albertz: Is this resolved now?

tools/compile_tf_graph.py
albertz: Is this resolved now?

tools/compile_tf_graph.py
```diff
@@ -551,9 +662,22 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
   Adapts :class:`_SubnetworkRecCell` such that we execute only a single step.
   """
 
-  def __init__(self, **kwargs):
+  def __init__(self, parent_rec_layer, **kwargs):
+    """
+    :param RecStepByStepLayer:
```
albertz: The type is missing.

tools/compile_tf_graph.py
```diff
@@ -551,9 +662,22 @@ class SubnetworkRecCellSingleStep(_SubnetworkRecCell):
...
     self._parent_layers = {}  # type: typing.Dict[str,WrappedInternalLayer]
-    super(SubnetworkRecCellSingleStep, self).__init__(**kwargs)
+    with helper_variable_scope():
+      self.parent_tile_multiples_var = tf.get_variable(
+        name="parent_tile_multiples", shape=(), dtype=tf.int32, initializer=tf.ones_initializer())  # type: tf.Variable
+    super(SubnetworkRecCellSingleStep, self).__init__(parent_rec_layer=parent_rec_layer, **kwargs)
+
+  def set_parent_tile_multiple_op(self, multiple):
+    """
+    :param tf.Tensor multiplie:
```
albertz: What is the shape, dtype?

tools/compile_tf_graph.py
albertz: There is no such function. I don't understand what you mean.

albertz:
After offline discussion, I will take over this pull request.
Some notes (mostly for myself, once I get to work on this):
- Choice dependent means: this state-var / output only depends on some (the last?) choice, nothing else, i.e. no (prev) state-var.
- The optimization done for choice dependent is actually not so important. It saves maybe sth like 2% of memory. For our standard global att model, it is only the target_embed which does not need to be stored, which is very small compared to all the other state vars.
- If not choice dependent, there is currently the assumption (on Sprint side) that it (update for state-var) does not depend on the (last? any?) choice but only on (prev) state-vars. This is more important for optimization, mostly for performance. For our standard global att model, this is true for all other state-vars except target_embed. But this is not true in general. We need to make a check for that.
- The missing case given this is some update for state-var which depends both on the (some? last?) choice and some prev state-vars. This is unsupported currently.
- There might be problems if there is any state-var which is not float32.

I still think it makes it rather complicated and ugly to have the distinction between these 3 cases (or more cases once you get to multiple stochastic vars / choices), and also it makes it error/bug prone. But that's how it is implemented on Sprint side now.

@ZhouW321 maybe you can comment if I got sth wrong, or complete this.
Not sure when I have time to do this. Probably after ICML submission only.

albertz:
Unfortunately this is not in a branch here in the main repo, so we need to do a new PR to be able to jointly work on this.

Also, this is long outdated anyway, so all of the necessary changes need to be redone, and in a clean way.

The new PR is #874.

tools/compile_tf_graph.py
albertz: @ZhouW321 I still don't understand this. You never answered here.

tools/compile_tf_graph.py
albertz: Both are used by RETURNN, so both are needed.

tools/compile_tf_graph.py
albertz: Also this is needed.

tools/compile_tf_graph.py
albertz: Also, do we really need this?

PR #268, opened 2020-01-22
WarpRNNT wrapper

Spotlight0xff:
Hi,

This PR implements the wrapper around [warp-transducer](https://github.com/HawkAaron/warp-transducer).

The loss can then be used like this:
`from extern.warprnnt import rnnt_loss`

with the definition `rnnt_loss(acts, labels, input_lengths, label_lengths, blank_label=0)`, defined in `extern/warprnnt/__init__.py`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,3 +14,6 @@
 	path = extern/ParseOggVorbis
 	url = https://github.com/albertz/ParseOggVorbis.git
 	branch = master
+[submodule "extern/HawkAaronWarpTransducer/warp-transducer"]
+	path = extern/HawkAaronWarpTransducer/warp-transducer
+	url = https://github.com/HawkAaron/warp-transducer.git
--- a/tests/test_TFNativeOp.py
+++ b/tests/test_TFNativeOp.py
@@ -3033,6 +3033,95 @@ def test_blocksparse_simple_feature_axis1():
   assert_allclose(result, y_test)
 
 
+def _run_rnnt(acts, labels, input_lengths, label_lengths,
+              expected_costs, expected_grads, blank):
+  from extern.HawkAaronWarpTransducer import rnnt_loss
+  assert_equal(acts.shape, expected_grads.shape)
+  acts_t = tf.constant(acts)
+  labels_t = tf.constant(labels)
+  input_lengths_t = tf.constant(input_lengths)
+  label_lengths_t = tf.constant(label_lengths)
+
+  logits = tf.nn.log_softmax(acts_t)
+  costs = rnnt_loss(logits, labels_t, input_lengths_t, label_lengths_t, blank)
+
+  grads = tf.gradients(costs, [acts_t])[0]
+
+  (tf_costs, tf_grad) = session.run([costs, grads])
+  # Max absolute difference: 2.861023e-06
+  # Max relative difference: 7.264356e-07
+  assert_allclose(tf_costs, expected_costs, atol=1e-6, rtol=1e-6)
+  assert_allclose(tf_grad, expected_grads, atol=1e-6, rtol=1e-6)
+
+
+@unittest.skipIf(not is_gpu_available(), "no gpu on this system")
+def test_warprnnt_forward():
+  from extern.HawkAaronWarpTransducer import rnnt_loss
+  # Softmax activations for the following inputs:
+  import numpy as np
+  acts = np.array([0.1, 0.6, 0.1, 0.1, 0.1, 0.1,
+                   0.1, 0.6, 0.1, 0.1, 0.1, 0.1,
+                   0.2, 0.8, 0.1, 0.1, 0.6, 0.1,
+                   0.1, 0.1, 0.1, 0.1, 0.2, 0.1,
+                   0.1, 0.7, 0.1, 0.2, 0.1, 0.1], dtype=np.float32).reshape(1, 2, 3, 5)
+
+  labels = np.array([[1, 2]], dtype=np.int32)
+  input_lengths = np.array([2], dtype=np.int32)
+  label_lengths = np.array([2], dtype=np.int32)
+
+  acts_t = tf.constant(acts)
+  labels_t = tf.constant(labels)
+  input_lengths_t = tf.constant(input_lengths)
+  label_lengths_t = tf.constant(label_lengths)
+  acts_t = tf.nn.log_softmax(acts_t)  # NOTE cpu
+  costs = rnnt_loss(acts_t, labels_t, input_lengths_t, label_lengths_t)
+  print(costs.eval())
+
+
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -14,3 +14,6 @@
 	path = extern/ParseOggVorbis
 	url = https://github.com/albertz/ParseOggVorbis.git
 	branch = master
+[submodule "extern/warprnnt/warp-transducer"]
+	path = extern/warprnnt/warp-transducer
+	url = git@github.com:HawkAaron/warp-transducer
```
albertz: I think the `git@github.com` url only works if you have an account at GitHub, and configured the GitHub SSH key. The `http` url always works, also without GitHub account. That's why I have used `http` for the others.

extern/warprnnt/__init__.py
```diff
@@ -0,0 +1,103 @@
+"""
+Provides a RETURNN wrapper around `warp-transducer`:
+  https://github.com/HawkAaron/warp-transducer.git
```
albertz: Remove the `.git`.

extern/warprnnt/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    # https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html#Line-Control
+    src_code += "#line 1 \"%s\"\n" % os.path.basename(fn)
+    src_code += f_code
+    src_code += "\n// ------------ %s : END } --------------\n\n" % os.path.basename(fn)
+
+  compiler = OpCodeCompiler(
+    base_name="warprnnt_kernels", code_version=1, code=src_code,
+    include_paths=(submodule_dir + "/include",),
+    c_macro_defines={"WITH_OMP": 1},
+    ld_flags=["-Xcompiler", "-fopenmp"],
```
albertz: It uses OMP? But it is only a GPU implementation?

tests/test_TFNativeOp.py
```diff
@@ -3033,6 +3033,95 @@ def test_blocksparse_simple_feature_axis1():
...
+                             -0.154257, 0.229433, -0.075176, -0.246593, 0.146405, 0.100188,
+                             -0.012918, -0.061593, 0.074512, -0.055986, 0.219831, -0.163845,
+                             -0.497627, 0.209240, 0.288387, 0.013605, -0.030220, 0.016615,
+                             0.113925, 0.062781, -0.176706, -0.667078, 0.367659, 0.299419,
+                             -0.356344, -0.055347, 0.411691, -0.096922, 0.029459, 0.067463,
+                             -0.063518, 0.027654, 0.035863, -0.154499, -0.073942, 0.228441,
+                             -0.166790, -0.000088, 0.166878, -0.172370, 0.105565, 0.066804,
+                             0.023875, -0.118256, 0.094381, -0.104707, -0.108934, 0.213642,
+                             -0.369844, 0.180118, 0.189726, 0.025714, -0.079462, 0.053748,
+                             0.122328, -0.238789, 0.116460, -0.598687, 0.302203, 0.296484], dtype=np.float32)
```
albertz: How did you calculate this?

tests/test_TFNativeOp.py
```diff
@@ -3033,6 +3033,95 @@ def test_blocksparse_simple_feature_axis1():
   assert_allclose(result, y_test)
 
 
+def _run_rnnt(acts, labels, input_lengths, label_lengths,
+              expected_costs, expected_grads, blank):
```
albertz: It would be nice if we can also have a reference implementation. E.g. in pure TF, with `tf.while_loop` over `T+U+1`, how you manually would write down the forward computation (and backward is automatically via backprop), also in log space. But not needed for this initial pull request.

.gitmodules
albertz: Btw, because of this, all Travis tests fail. So we definitely need the `http` url here.

albertz:
Another small thing: You call it `warprnnt`, but the original project name is `warp-transducer`, and there are also other projects which are called `warprnnt` or similar. Maybe we should use a more clear name? Like `HawkAaronWarpTransducer` instead of just `warprnnt`.

extern/warprnnt/__init__.py
albertz: Maybe you can also extend this a bit, by giving also adding refs/links to other RNN-T implementations, just for reference? Just list the other links I gave you. Of course make clear which of these is actually wrapped here.

tests/test_TFNativeOp.py
Spotlight0xff: I just used their test cases. https://github.com/HawkAaron/warp-transducer/blob/master/tensorflow_binding/tests/test_warprnnt_op.py

Should I calculate it myself to verify? (e.g. using a reference implementation)

albertz: For this first pull request, that's ok (but put this link as a comment to the code then).
As a next step, it would be good to have such a reference implementation (also flexible to allow for other cases which we want to study later).

extern/warprnnt/__init__.py
Spotlight0xff: It should work on both CPU and GPU, but I only tested the GPU version.
Their cmake script used omp, so I just added it aswell.

Spotlight0xff:
Ok, I have renamed the module as you suggested, to avoid any confusion.

extern/warprnnt/__init__.py
Spotlight0xff: both comments are addressed in the latest commit.

.gitmodules
Spotlight0xff: Oh, I see. Changed that

extern/warprnnt/__init__.py
albertz: Oh interesting. Can you just try out? It should already work, or not?

.gitmodules
```diff
@@ -14,3 +14,9 @@
 	path = extern/ParseOggVorbis
 	url = https://github.com/albertz/ParseOggVorbis.git
 	branch = master
+[submodule "extern/warprnnt/warp-transducer"]
+	path = extern/warprnnt/warp-transducer
+	url = https://github.com/HawkAaron/warp-transducer.git
```
albertz: This is twice here now. This first submodule can be removed then.

Spotlight0xff: oops, fixed.

albertz:
I just merged it in now.

However, missing was still the link to the [ref test numbers here](https://github.com/HawkAaron/warp-transducer/blob/master/tensorflow_binding/tests/test_warprnnt_op.py) as a comment. But you can later add that.

Also missing: Did you try out whether CPU works? If so, remove the GPU-only check.

And then as a next step, make a reference implementation in pure TF via `tf.while_loop`. We can discuss offline about that.

PR #275, opened 2020-02-04
ChoiceLayer: return length normalized score as expected

patrick-wilken:
Currently the search score depends on the batch and beam size when doing length normalization. This is because the implementation of length normalization in ChoiceLayer introduces a factor of maximum decoding length for all scores, which can vary depending on which sequences are in the batch/beam.

This PR removes the factor, such that ChoiceLayer returns the expected score/seq_lens once search is done.

### DIFF
--- a/TFNetworkRecLayer.py
+++ b/TFNetworkRecLayer.py
@@ -2600,7 +2600,9 @@ class _SubnetworkRecCell(object):
       acc_search_choices = SearchChoices(owner=latest_layer_choice, beam_size=latest_beam_size)
       final_choice_rec_vars = self.get_layer_rec_var_from_loop_vars(
         loop_vars=final_net_vars,
-        layer_name=latest_layer_choice.name)
+        layer_name=latest_layer_choice.name,
+        final_frame=True,
+        seq_len=seq_len)
       acc_search_choices.set_beam_from_rec(final_choice_rec_vars)
       search_choices_cache[latest_layer_choice.name] = acc_search_choices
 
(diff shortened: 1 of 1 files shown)


TFNetworkRecLayer.py
```diff
@@ -2600,7 +2600,9 @@ def search_resolve_cond(i, *args):
       acc_search_choices = SearchChoices(owner=latest_layer_choice, beam_size=latest_beam_size)
       final_choice_rec_vars = self.get_layer_rec_var_from_loop_vars(
         loop_vars=final_net_vars,
-        layer_name=latest_layer_choice.name)
+        layer_name=latest_layer_choice.name,
+        final_frame=True,
```
patrick-wilken: This should have been set before, right? Or am I missing something?

TFNetworkRecLayer.py
```diff
@@ -4770,6 +4773,20 @@ def get_dep_layers(self):
...
+    """
+    :param dict[str,tf.Tensor] rec_vars_outputs:
+    :param tf.Tensor seq_len: shape (batch,)
+    :rtype: dict[str,tf.Tensor]
+    """
+    if self.length_normalization:
+      assert "choice_scores" in rec_vars_outputs
+      # Finalize length normalization. During search we keep an extra factor t (recurrent time step) for efficiency
+      # reasons (see self.get_output()). Remove it here.
+      num_time_steps = tf.reduce_max(seq_len) + 1  # + 1 to include sequence end
```
albertz, line 4786: Note that `+1` is not correct in all cases. You can configure whether the seq/RecLayer includes EOS or not.

TFNetworkRecLayer.py
albertz: Hm yes, good question. Not sure anymore. But looks right. I'm not 100% sure about `seq_len`, if that comes from another beam.

albertz:
I will just merge it now. It's already an improvement. We can improve upon it more later.

Issue #284, opened 2020-04-02
Missing size_placeholder inside of rec-Layer


Zettelkasten:
I am trying to modify the Transformer architecture for MT to include the previous time step's encoder-decoder attention weights to calculate the current weights (a layer called `dec_01_att_weights`).
For this, I take the input `prev:dec_01_att_weights`, and pass it through a self-written layer that performs pooling over a variable sized time dimension (in this case, the source time dimension). However, this custom layer requires information about the source sequence lengths to perform masking, but the source's Data objects lack this info which results in the following exception when calling `get_sequence_mask()`:
```
  File "/u/petrick/att/returnn/TFUtil.py", line 2104, in is_time_axis_dynamic
    line: assert isinstance(self.shape[self.time_dim_axis_excluding_batch], int), (
    locals:
      isinstance = <builtin> <built-in function isinstance>
      self = <local> Data(name='prev:dec_01_att_weights_output', shape=(None, 1, 8), feature_dim_axis=2, beam=SearchBeam(name='output/prev:output', beam_size=12), batch_shape_meta=[B,T|?,F|1,8])
      self.shape = <local> (None, 1, 8)
      self.time_dim_axis_excluding_batch = <local> 0
      int = <builtin> <class 'int'>
AssertionError: Data(name='prev:dec_01_att_weights_output', shape=(None, 1, 8), feature_dim_axis=2, beam=SearchBeam(name='output/prev:output', beam_size=12), batch_shape_meta=[B,T|?,F|1,8]): dynamic time axis dim (None) (axis 1) but size_placeholder {} misses information
```

I noticed that in TFNetworkRecLayer.py in copy_as_prev_time_frame, you set
```
    if prev_output is not None:
      layer.output.placeholder = prev_output
      layer.output.placeholder.set_shape(tf.TensorShape(layer.output.batch_shape))
      layer.output.size_placeholder = {}  # must be set
```
(around line 3125) thus removing the size_placeholder.
I tried to comment out this last line, and that fixes my problem as the size_placeholder is now set. My config now works during training.

However, when using beam search with beam size 12, I run into a problem where the shape of the size_placeholder for the time dimension does not match with the batch dimension of the placeholder itself: For the first sequence, the size_placeholder of the time dim is 6, but the batch dim of my placeholder is 72=12*6.
I guess that somewhere, the batch dim is merged with the beam dimension, and that this does not happen here.

[...]

PR #285, opened 2020-04-04
WIP: Failing test case for cumulated attention weights during search

Zettelkasten:
Contains a failing test case without a fix. ~~This is related to #284.~~

### DIFF
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -5294,6 +5294,7 @@ def test_trafo_search_lm():
       assert out_seq_lens[i] == input_seq_lens[i] * 3 // 2  # we constructed the 'end' layer that way
       assert all(out_seqs[i, :input_seq_lens[i]] == input_seqs[i, :input_seq_lens[i]])
 
+
 def test_cumulated_attention_weights_search():
   rnd = numpy.random.RandomState(42)
   beam_size = 5
@@ -5310,8 +5311,8 @@ def test_cumulated_attention_weights_search():
                   'unit': {
                       'target_embed': { 'class': 'linear', 'activation': None, 'from': ['prev:output'], 'n_out': dim},
                       'att_energy': { 'class': 'dot', 'from': ['base:source_embed', 'target_embed'], 'red1': -1, 'red2': -1, 'var1': 'T', 'var2': 'T?', 'add_var2_if_empty': False},
-                      'cum_att_energy': {'class': 'combine', 'kind': 'add', 'from': ['prev:att_energy', 'att_energy'], 'initial_output': 'att_energy'},
-                      'att_weights': {'class': 'softmax_over_spatial', 'from': ['cum_att_energy']},
+                      'cum_att_energy': {'class': 'combine', 'kind': 'add', 'from': ['prev:att_energy', 'att_energy']},
+                      'att_weights': {'class': 'softmax_over_spatial', 'from': ['cum_att_energy'], 'axis': 'stag:extern_data:data'},
                       'att': {'class': 'generic_attention', 'base': 'base:source_embed', 'weights': 'att_weights', 'is_output_layer': True},
                       'output_prob': { 'class': 'softmax', 'from': ['att'], 'loss': 'ce', 'target': 'classes'},
                       'output': {'beam_size': beam_size, 'class': 'choice', 'from': ['output_prob'], 'initial_output': 0, 'target': 'classes'},
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -5294,6 +5294,79 @@ def test_trafo_search_lm():
       assert out_seq_lens[i] == input_seq_lens[i] * 3 // 2  # we constructed the 'end' layer that way
       assert all(out_seqs[i, :input_seq_lens[i]] == input_seqs[i, :input_seq_lens[i]])
 
+def test_cumulated_attention_weights_search():
+  rnd = numpy.random.RandomState(42)
+  beam_size = 5
+  dim = 7
+
+  # Config works during training, but building graph raises exception during search:
+  # Trying to reshape input tensor with n values into tensor with n * beam_size values
+  net_dict = {
+    'source_embed': { 'class': 'linear', 'activation': None, 'n_out': dim},
+    'output': {   'class': 'rec',
+                  'from': [],
+                  'max_seq_len': "max_len_from('base:source_embed') * 3",
+                  'target': 'classes',
+                  'unit': {
+                      'target_embed': { 'class': 'linear', 'activation': None, 'from': ['prev:output'], 'n_out': dim},
+                      'att_energy': { 'class': 'dot', 'from': ['base:source_embed', 'target_embed'], 'red1': -1, 'red2': -1, 'var1': 'T', 'var2': 'T?', 'add_var2_if_empty': False},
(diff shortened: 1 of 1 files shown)


tests/test_TFNetworkRecLayer.py
```diff
@@ -5294,6 +5294,79 @@ def test_trafo_search_lm():
       assert out_seq_lens[i] == input_seq_lens[i] * 3 // 2  # we constructed the 'end' layer that way
       assert all(out_seqs[i, :input_seq_lens[i]] == input_seqs[i, :input_seq_lens[i]])
```
albertz, line 5295: Note that there should be 2 empty lines before a new function (always check PyCharm warnings).

tests/test_TFNetworkRecLayer.py
```diff
@@ -5294,6 +5294,79 @@ def test_trafo_search_lm():
...
+  net_dict = {
+    'source_embed': { 'class': 'linear', 'activation': None, 'n_out': dim},
+    'output': {   'class': 'rec',
+                  'from': [],
+                  'max_seq_len': "max_len_from('base:source_embed') * 3",
+                  'target': 'classes',
+                  'unit': {
+                      'target_embed': { 'class': 'linear', 'activation': None, 'from': ['prev:output'], 'n_out': dim},
+                      'att_energy': { 'class': 'dot', 'from': ['base:source_embed', 'target_embed'], 'red1': -1, 'red2': -1, 'var1': 'T', 'var2': 'T?', 'add_var2_if_empty': False},
+                      'cum_att_energy': {'class': 'combine', 'kind': 'add', 'from': ['prev:att_energy', 'att_energy'], 'initial_output': 'att_energy'},
```
albertz, line 5313: Does `'initial_output': 'att_energy'` make sense? I think not. I wonder that this works (but maybe actually that's the reason why it does not). `initial_output` specifies the initial output. I.e. when you access `prev: cum_att_energy` in the first frame, it specifies the output. But as you never do that anyway, it's probably ignored anyway. But then better remove this.

tests/test_TFNetworkRecLayer.py
```diff
@@ -5294,6 +5294,79 @@ def test_trafo_search_lm():
...
+                      'att_weights': {'class': 'softmax_over_spatial', 'from': ['cum_att_energy']},
```
albertz, line 5314: You should better specify the `axis`.

albertz:
For reference, the error is:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Input to reshape is a tensor with 27 values, but the requested shape has 135
	 [[node output/rec/att_weights/get_sequence_mask_broadcast/seq_mask_reshape (defined at home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py:1748) ]]
```
Which happens in `SoftmaxOverSpatialLayer` via `energy_mask = SeqLenMaskLayer.build_mask(...)`.

Zettelkasten:
Thanks for all your comments, @albertz.
I think the problem is as follows: When decoding the first token, there is no beam yet, and thus the beam size of `cum_att_energy` is 1 in time step 0. After the `choice` Layer `output`, the best `beam_size` many hypotheses are kept, i.e. the beam of `output` is now the actual beam size.
Then, when calculating `cum_att_energy` for time step 1, you cannot simply add the tensors from `prev:cum_att_energy` and `att_energy`, as here `att_energy` has the real beam size, while `prev:cum_att_energy` still has beam size 1. Thus the reshape error.
For all consecutive time steps, the dimensions should match, however just adding the vectors will match up the current beams with the previous time step's beams wrongly.
Instead, to calculate `prev:cum_att_energy` the following should be done: For every beam entry in the current time step, find the beam entry of the previous time step that it branched off, and create a tensor where for every current beam entry, the previous time step's value for `cum_att_energy`. This tensor then matches the dimensions of `cum_att_energy`, and can be added as usual (or passed to any other tensor).

albertz:
> I think the problem is as follows: When decoding the first token, there is no beam yet, and thus the beam size of `cum_att_energy` is 1 in time step 0. After the `choice` Layer `output`, the best `beam_size` many hypotheses are kept, i.e. the beam of `output` is now the actual beam size.
> Then, when calculating `cum_att_energy` for time step 1, you cannot simply add the tensors from `prev:cum_att_energy` and `att_energy`, as here `att_energy` has the real beam size, while `prev:cum_att_energy` still has beam size 1. Thus the reshape error.

No, RETURNN automatically handles this. See `SelectSearchSourcesLayer` and related code.
Also, if it would not do that correctly, you would get an error at the `tf.add` in your combine-layer, that the shapes do not match. But this seems to be fine.

Zettelkasten:
You are right. However, found out something more:
- If in the `cum_att_energy` combine-layer, you swap the input arguments (i.e. the `att_energy` is first), the error does not occur
- From what I can tell, here `size_placeholder` of `cum_att_energy` is always copied from the first source's output data. Instead, the `size_placeholder`s dim's should be the common multiple of all source's `size_placeholder`s, I think. At least here the `size_placeholder` has the wrong batch-dimension, thus the sequence mask of the softmax layer has the wrong batch dim, and thus the reshape error.

I'm not sure what part of the code should be responsible for this, perhaps `Data.get_common_data`?

albertz:
Unfortunately I don't really have much time right now to investigate this myself further. However, some more answers, maybe they are helpful.

> * If in the `cum_att_energy` combine-layer, you swap the input arguments (i.e. the `att_energy` is first), the error does not occur

See the code of `CombineLayer`.
First it will construct the output template (`Data` instance, which specifies which axes there are, and what spatial axes, including their tag, and `size_placeholder`, if known), via `CombineLayer.get_out_data_from_opts`. This will also determine the order. It uses `Data.get_common_data`. If that gets multiple different time dims (which it knows because `size_placeholder` is different for the sources), it will have both time dims. I.e. (B,T1,D) and (B,T2,D) would end up in (B,T1,T2,D).
Then, the calculation itself (which happens in `CombineLayer.__init__`) will basically call `sources[i].output.copy_compatible_to(self.output)`, and that would add broadcast dimensions.

Swapping the sources can have an effect on the order of the axes, via `Data.get_common_data`. That should not really make a difference in the end but maybe one case triggers a bug while the other does not. Also, it can affect which of the spatial dims will end up marked as `time_dim_axis` in the end.

> * From what I can tell, here `size_placeholder` of `cum_att_energy` is always copied from the first source's output data.

No, it should be copied from both.

> Instead, the `size_placeholder`s dim's should be the common multiple of all source's `size_placeholder`s, I think.

No, there are now two separate spatial dims, and each of them will just have the same size as before.

> At least here the `size_placeholder` has the wrong batch-dimension, thus the sequence mask of the softmax layer has the wrong batch dim, and thus the reshape error.

The key in `size_placeholder` is the axis counted without batch-dim (for some historical reasons...).

Zettelkasten:
Thanks for your time! This is very helpful already. I will investigate further, and for now, I can also use my config by swapping the two layer arguments. Thanks again for your quick help!

albertz:
@Zettelkasten I just checked this, and it seems to run fine now. At least the test passes. There were a couple of changes in the template construction. Maybe this is already fixed? Can you check, and update (rebase) the PR?

Also, make sure the test runs in TF 2.3. You need this:
```
session.run(tf_compat.v1.variables_initializer(tf_compat.v1.global_variables() + [network.global_train_step]))
```

Zettelkasten:
I removed the fork this PR was created on and now I cannot change anything anymore (c.f. [here](https://stackoverflow.com/questions/36071272/fix-unknown-repository-of-an-opened-pr-after-deleted-the-fork)).
I will close this, and create a new PR to make the changes.

PR #290, opened 2020-05-05
CombinedDataset: recovery of general sorting support

patrick-wilken:
Let's follow up on the discussion in 761062b. This commit removed support for the sorting options available via `Dataset.get_seq_order_for_epoch()` in CombinedDataset.

I just boldly put our code here, but it's secondary. Let's discuss the principles first :)

What I want to achieve is to combine several datasets such that they behave exactly as if a dataset would have been created from the merged input files. Our use case is that we already have TranslationDatasets setups (possibly dumped to HDF) on disk for several language pairs and now want to train a multilingual system using those datasets without creating new merged corpora for each new language pair combination. Our bilingual systems use "laplace" sequence ordering and we want to keep that for multilingual systems.

For this, CombinedDataset itself has to call `get_seq_order_for_epoch()`, which requires the total number of sequences and a function to get the sequence lengths as input. The discussion was about how to correctly get those from the subdatasets.

**About num_seqs:**
The check for `know_num_seqs_beforehand` was removed in 761062b because `num_seqs` "is not reliable". However, many dataset types internally are able to provide `num_seqs` to `get_seq_order_for_epoch()`, I just want to do the same from outside the class, so it should be possible in principle. Also, I don't see many cases where it is not reliable. I looked through the code of a lot of datasets and I only found two that don't provide `num_seqs`: ExternSprintDataset only provides it as soon as the reading thread finishes. LMDataset does not provide it at all (because of unknown filtering, I guess). All other datasets that I looked at seem to provide the correct number of sequences. And if they don't they should. Or not provide `num_seqs` at all.
Also, we discussed at which point in time `num_seqs` becomes valid. When initialising a dataset we always call `init_seq_order()` for the first epoch (`init_dataset()`-> `Dataset.initialize()`). And after `init_seq_order()` we should know whether the number of sequences is known, so I claim the `know_num_seqs_beforehand` check was correct. It is clear that if the number of sequences is not known in advance you cannot do a global sort. But I think this was handled correctly before by checking the `seq_ordering` in `_expand_dataset_sec_idxs()`.

**About get_seq_length:**
We discussed that a sequence has to be loaded via `load_seqs` to know its length. Again, internally the subdatasets are able to provide lengths when calling `get_seq_order_for_epoch()`, I just want to do the same from outside the class, so I don't think this is true. Most of the classes use a `get_seq_len` that is different from `CachedDataset2.get_seq_length()` and doesn't load the sequence. So actually, I would have to expose those and use them. However, I found two cases where `get_seq_order_for_epoch()` is called using `CachedDataset2.get_seq_length()`: RawWavDataset and MetaDataset (via `_get_dataset_seq_length`). Are those "wrong" too? Is loading the sequences in `CachedDataset2.get_seq_length()` a hack that should not be used?

Imho, overall, a good design would be:

- `_seq_order` should be a property of the base class Dataset (updated for each epoch)
- all datasets that know the total number of sequences in advance should use `get_seq_order_for_epoch()` to compute `_seq_order` (almost the case I guess)
- `num_seq`should be `len(_seq_order)` **by definition**
- if total number of sequences unknown, `_seq_order` and `num_seq` should be None, or raise
- filtering must be done before creating `_seq_order`, i.e. all sequences in it are really used (confer LMDataset...)
- the `get_seq_len` used for `get_seq_order_for_epoch()` in the different datasets should be exposed like `CachedDataset._get_seq_length_by_real_idx()` to make them available for MetaDatasets

Somewhat unrelated:
In the code we added a `sampling_sizes` parameter such that you can take a fixed amount of sequences from the subdatasets per epoch, still supporting generic sorting of that chunk.
Also, giving a sequence list (list of strings) to the subdatasets was slow and memory intensive for huge datasets, so I introduced this `reinit_indices` parameter to do it by indices. We can discuss those things separately.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -259,6 +259,20 @@ class Dataset(object):
     """
     raise NotImplementedError
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    In contrast to self.get_seq_length(), this method is designed to work for sequences that have not been loaded yet
+    via self.load_seqs().
+    Used by meta-datasets for sequence ordering. Currently we only provide one number, i.e. do not give different
+    estimates for the different data keys (as in get_seq_length()). It is up to the dataset what this number represents
+    and how it is computed.
+
+    :param int seq_idx: for current epoch, not the corpus seq idx
+    :rtype: int
+    :returns sequence length estimate (for sorting)
+    """
+    raise OptionalNotImplementedError
+
   def get_num_timesteps(self):
     """
     :rtype: int
--- a/LmDataset.py
+++ b/LmDataset.py
@@ -1382,6 +1382,17 @@ class TranslationDataset(CachedDataset2):
     self._num_seqs = len(self._seq_order)
     return True
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    :param int seq_idx: for current epoch, not the corpus seq idx
+    :rtype: int
+    :returns sequence length of main source data key ("data"), used for sequence sorting
+    """
+    corpus_seq_idx = self.get_corpus_seq_idx(seq_idx)
+    assert corpus_seq_idx is not None
+
+    return len(self._get_data(key=self.main_source_data_key, line_nr=corpus_seq_idx))
+
   def _collect_single_seq(self, seq_idx):
     if seq_idx >= self._num_seqs:
       return None
--- a/LmDataset.py
+++ b/LmDataset.py
@@ -1334,6 +1334,12 @@ class TranslationDataset(CachedDataset2):
     """
     return True
 
+  def get_all_tags(self):
+    """
+    :rtype: list[str]
+    """
+    return [self._tag_prefix + str(line_nr) for line_nr in range(len(self._data[self.main_source_data_key]))]
+
   def get_corpus_seq_idx(self, seq_idx):
     """
     :param int seq_idx:
--- a/MetaDataset.py
+++ b/MetaDataset.py
@@ -824,7 +824,6 @@ class CombinedDataset(CachedDataset2):
     super(CombinedDataset, self).__init__(**kwargs)
     assert self.shuffle_frames_of_nseqs == 0  # not implemented. anyway only for non-recurrent nets
 
-    self.rnd = Random(self.epoch)
     self.dataset_keys = set([m[0] for m in data_map.keys()])  # type: typing.Set[str]
     self.dataset_idx2key_map = dict(enumerate(sorted(self.dataset_keys)))  # idx -> dataset-key
     self.data_keys = set(data_map.values())  # type: typing.Set[str]
@@ -892,7 +891,6 @@ class CombinedDataset(CachedDataset2):
     need_reinit = self.epoch is None or self.epoch != epoch
     num_seqs_saved = self._num_seqs
     super(CombinedDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list)  # resets self._num_seqs
-    self.rnd.seed(epoch or 1)
 
     if not need_reinit:
       self._num_seqs = num_seqs_saved
@@ -914,8 +912,11 @@ class CombinedDataset(CachedDataset2):
         self.datasets[self.dataset_idx2key_map[dataset_idx]].num_seqs for dataset_idx in range(len(self.datasets))]
       self.dataset_seq_idx_boundaries = self._create_dataset_seq_idx_boundaries()
 
-      seq_order = self.get_seq_order_for_epoch(
-        epoch=epoch, num_seqs=total_num_seqs, get_seq_len=self._get_seq_length)
+      if self.seq_ordering == "random_dataset":
+        seq_order = self._get_random_dataset_seq_order(total_num_seqs)
+      else:
+        seq_order = self.get_seq_order_for_epoch(
+          epoch=epoch, num_seqs=total_num_seqs, get_seq_len=self._get_seq_length)
       self._num_seqs = len(seq_order)
 
       # Apply seq_order to self.dataset_seq_idx.
@@ -980,6 +981,38 @@ class CombinedDataset(CachedDataset2):
     dataset_seq_idx = seq_idx - self.dataset_seq_idx_boundaries[dataset_idx]
 
     return dataset_idx, dataset_seq_idx
+
+  def _get_random_dataset_seq_order(self, total_num_seqs):
+    """
+    Choose datasets randomly but preserve order within each dataset. This sorting method is unique to CombinedDataset.
+
+    :param int total_num_seqs:
+    :param int epoch:
+    :returns: sequence order
+    :rtype: list[int]
+    """
+    # Create a list containing each dataset_idx dataset.num_seqs-times and shuffle it.
+    dataset_ids = [self._seq_idx_to_dataset_seq_idx(seq_idx)[0] for seq_idx in range(total_num_seqs)]
+    rnd = self._get_random_seed_for_epoch(self.epoch)
+    rnd.shuffle(dataset_ids)
+
+    # Create the actual seq_order list.
+    # We want to keep the order within the sub-datasets, thus we assign seq_ids by simply counting up for each dataset.
+    # We however have to account for the different offsets needed when accessing self.dataset_seq_idx_list later.
+    seq_order = []
+    counters = [0] * len(self.datasets)
+    for dataset_idx in dataset_ids:
+      seq_order.append(counters[dataset_idx] + self.dataset_seq_idx_boundaries[dataset_idx])
+      counters[dataset_idx] += 1
+
+    assert sum(counters) == total_num_seqs
+
+    if self.partition_epoch:
+      seq_order = self._apply_partition_epoch(seq_order, self.partition_epoch, self.epoch)
+    if self.repeat_epoch:
+      seq_order = seq_order * self.repeat_epoch
+
+    return seq_order
 
   def _get_seq_length(self, seq_idx):
     """
--- a/MetaDataset.py
+++ b/MetaDataset.py
@@ -847,6 +847,11 @@ class CombinedDataset(CachedDataset2):
     # This will only initialize datasets needed for features occurring in data_map
     self.datasets = {key: init_dataset(datasets[key]) for key in self.dataset_keys}
 
+    self.seq_corpus_tags_per_ds = {}
+    for key, ds in self.datasets.items():
+      assert ds.have_corpus_seq_idx()
+      self.seq_corpus_tags_per_ds[key] = ds.get_all_tags()
+
     self._estimated_num_seqs = sum([self.datasets[k].estimated_num_seqs for k in sorted(self.datasets.keys())])
     self.estimated_num_seq_per_subset = [self.datasets[k].estimated_num_seqs for k in sorted(self.datasets.keys())]
 
@@ -872,8 +877,7 @@ class CombinedDataset(CachedDataset2):
 
     self.data_dtypes = {data_key: _select_dtype(data_key, self.data_dims, data_dtypes) for data_key in self.data_keys}
 
-    self.dataset_seq_idx_list = None  # type: typing.Optional[typing.List[typing.Tuple[int,int]]]
-    self.seq_order = None  # type: typing.Optional[typing.List[int]]
+    self.dataset_seq_idx_boundaries = None  # type: typing.Optional[typing.List[typing.Tuple[int,int]]]
     self.dataset_sorted_seq_idx_list = None  # type: typing.Optional[typing.List[typing.Tuple[int,int]]]
     self.used_num_seqs_per_subset = None  # type: typing.Optional[typing.List[int]]
 
@@ -883,24 +887,109 @@ class CombinedDataset(CachedDataset2):
     :param list[str]|None seq_list:
     :rtype: bool
     """
+
     assert seq_list is None, "seq_list not supported for %s" % self.__class__
     need_reinit = self.epoch is None or self.epoch != epoch
-    super(CombinedDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list)
+    num_seqs_saved = self._num_seqs
+    super(CombinedDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list)  # resets self._num_seqs
     self.rnd.seed(epoch or 1)
 
     if not need_reinit:
+      self._num_seqs = num_seqs_saved
       return False
 
-    # First init sequence order for sub-datasets as usual to get a list of available sequences. This way sorting and
+    # First init sequence order for sub-datasets as usual to get a list of available sequences. This way, sorting and
     # partition epoch of the individual sub-datasets is still supported. Later we will call init_seq_order again with a
     # sequence list to e.g. apply joint sorting or partition epoch of all sequences.
     for dataset in self.datasets.values():
       dataset.init_seq_order(epoch=epoch)
 
-    self.dataset_sorted_seq_idx_list = []  # We will fill this as we go
-    self.used_num_seqs_per_subset = [0] * len(self.datasets)
+    try:
+      total_num_seqs = sum([self.datasets[k].num_seqs for k in sorted(self.datasets.keys())])
+    except Exception:
+      total_num_seqs = None
+
+    if total_num_seqs is not None:
+      self.used_num_seqs_per_subset = [
+        self.datasets[self.dataset_idx2key_map[dataset_idx]].num_seqs for dataset_idx in range(len(self.datasets))]
+      self.dataset_seq_idx_boundaries = self._create_dataset_seq_idx_boundaries()
+
+      seq_order = self.get_seq_order_for_epoch(
+        epoch=epoch, num_seqs=total_num_seqs, get_seq_len=self._get_seq_length)
+      self._num_seqs = len(seq_order)
+
+      # Apply seq_order to self.dataset_seq_idx.
+      # We have to re-calculate the seq_idx's because we will sort the datasets below.
+      self.dataset_sorted_seq_idx_list = []
+      counters = [0] * len(self.datasets)
+      for seq_idx in seq_order:
+        dataset_idx, _ = self._seq_idx_to_dataset_seq_idx(seq_idx)
+        dataset_seq_idx = counters[dataset_idx]
+        counters[dataset_idx] += 1
+        self.dataset_sorted_seq_idx_list.append((dataset_idx, dataset_seq_idx))
+
+      # We only want to load those sequences in the sub-datasets that appear in seq_order. For this, we extract
+      # sequence lists containing the subset of sequences for each dataset from seq_order.
+      seq_lists = [[] for _ in self.datasets]
+      for seq_idx in seq_order:
+        dataset_idx, dataset_seq_idx = self._seq_idx_to_dataset_seq_idx(seq_idx)
+        dataset = self.datasets[self.dataset_idx2key_map[dataset_idx]]
+        dataset_corpus_seq_idx = dataset.get_corpus_seq_idx(dataset_seq_idx)
+        seq_tag = self.seq_corpus_tags_per_ds[self.dataset_idx2key_map[dataset_idx]][dataset_corpus_seq_idx]
+        seq_lists[dataset_idx].append(seq_tag)
+
+      # It may be large, so better delete it early, we don't need it anymore.
+      del seq_order
+
+      # Re-initialize sequence orders of sub-datasets with created sequence list.
+      for dataset_idx, dataset_key in self.dataset_idx2key_map.items():
+        self.datasets[dataset_key].init_seq_order(epoch=epoch, seq_list=seq_lists[dataset_idx])
+
+    else:
+      self.dataset_sorted_seq_idx_list = []  # We will fill this as we go
+      self.used_num_seqs_per_subset = [0] * len(self.datasets)
 
     return True
+
+  def _create_dataset_seq_idx_boundaries(self):
+    """
+    Creates a list storing the positions where the different sub-datasets start within the full sequence list.
+    This happens to be the cumulative sum of the number of sequences in the sub-datasets.
+
+    :rtype: list[int]
+    """
+    dataset_seq_idx_boundaries = [0]
+    for dataset_num_seqs in self.used_num_seqs_per_subset:
+      dataset_seq_idx_boundaries.append(dataset_num_seqs + dataset_seq_idx_boundaries[-1])
+
+    return dataset_seq_idx_boundaries
+
+  def _seq_idx_to_dataset_seq_idx(self, seq_idx):
+    """
+    Maps the sequence index (before sorting) to a dataset and a corresponding sequence index within this dataset.
+
+    :param int seq_idx:
+    :rtype: (int,int)
+    """
+    assert seq_idx < self.dataset_seq_idx_boundaries[-1]
+
+    dataset_idx = 0
+    while self.dataset_seq_idx_boundaries[dataset_idx + 1] <= seq_idx:
+      dataset_idx += 1
+
+    dataset_seq_idx = seq_idx - self.dataset_seq_idx_boundaries[dataset_idx]
+
+    return dataset_idx, dataset_seq_idx
+
+  def _get_seq_length(self, seq_idx):
+    """
+    :param int seq_idx:
+    :rtype: int
+    """
+    dataset_idx, dataset_seq_idx = self._seq_idx_to_dataset_seq_idx(seq_idx)
+    dataset = self.datasets[self.dataset_idx2key_map[dataset_idx]]
+
+    return dataset.get_estimated_seq_length(dataset_seq_idx)
 
   def _expand_dataset_sec_idxs(self, num_values):
     """
@@ -958,6 +1047,19 @@ class CombinedDataset(CachedDataset2):
               self.estimated_num_seq_per_subset[dataset_idx] = self.used_num_seqs_per_subset[dataset_idx]
 
       else:
+        # If we don't need to expand it is ok if the sorting method is not supported.
+        # This can happen if we already sorted everything in advance in `init_seq_order`.
+
+        all_sequences_used = True
+        for dataset_idx, _ in enumerate(self.datasets):
+          if self.datasets[self.dataset_idx2key_map[dataset_idx]].is_less_than_num_seqs(
+                self.used_num_seqs_per_subset[dataset_idx]):
+            all_sequences_used = False
+            break
+
+        if all_sequences_used:
+          return False
+
         raise Exception("The sorting method '{}' is not implemented for the case that number of sequences"
                         "is not known in advance.".format(self.seq_ordering))
 
--- a/HDFDataset.py
+++ b/HDFDataset.py
@@ -300,6 +300,15 @@ class HDFDataset(CachedDataset):
       return super(HDFDataset, self).get_targets(target, sorted_seq_idx)
     return self.get_data(sorted_seq_idx, target)
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    :param int seq_idx: for current epoch, not the corpus seq idx
+    :rtype: int
+    :returns sequence length of "data", used for sequence sorting
+    """
+    real_seq_idx = self._seq_index[self._index_map[seq_idx]]
+    return int(self._get_seq_length_by_real_idx(real_seq_idx)[0])
+
   def _get_seq_length_by_real_idx(self, real_seq_idx):
     """
     :param int real_seq_idx:


albertz:
It is a bit difficult to handle these bigger pull requests with lots of changes... Also difficult to get back to this after such a long time. There also seem to be multiple different, separate, individual things.

About `num_seqs`: I'm not exactly sure what you are arguing. You understand that `num_seqs` is not reliable (i.e. it might be an estimate, and it also might be different for every epoch, it might even change during an epoch), right? There is no way currently to know if this is correct or not. There is no real arguing about that. This was/is the design of the API (that it's optional, and can be inaccurate, and can change per epoch, or during an epoch), and many dataset implementations use it that way. We cannot easily change the definition of the API. (Also, why would we? We can extend the API if needed.)

However, there is `get_all_tags` and `get_total_num_seqs` instead, which are supposed to give you that information, in a reliable way (if sth is not reliable there, then this is a bug). Doesn't that give you what you want? What's missing?

Also, usually none of these are needed anyway, and you would just use `is_less_than_num_seqs` on-the-fly (while loading via `load_seqs`). But I see that for global shuffling, the exact number might be useful.

> Again, internally the subdatasets are able to provide lengths when calling get_seq_order_for_epoch(), I just want to do the same from outside the class, so I don't think this is true.

Unfortunately there is no API for this currently. `get_seq_length` is definitely not the right function. `get_seq_length` is defined such that it can only be used within the epoch, to get the length of current or (nearby) future seqs. Most datasets require that the seq is loaded (in memory) to get that information. The API of `load_seqs` is designed in a way that it will keep all in memory. `self.load_seqs(self.expected_load_seq_start, sorted_seq_idx + 1)` means that it would try to keep the whole dataset in memory.

I don't remember the exact discussion anymore, but are you arguing that you want to use the existing `get_seq_length`, and change the definition of that? Again, we cannot easily change the definition of the existing API. That would break too much things.

It is really not a good idea to just use some function and hope for some specific behavior (`num_seqs` or `get_seq_length`). Your assumption might be wrong, and then it might still all run without explicit errors or exceptions, but produce strange results, which you might even not notice. Which is the worst kind of thing which can happen. You never want to have such hidden traps, which will get unnoticed, and will be extremely hard to debug (if you notice something is strange).

But if you want to have such a function in the API (maybe `get_estimated_seq_length_by_corpus_seq_idx`, or `get_estimated_seq_length_by_tag`; see also `get_corpus_seq_idx` for a related function), then you can add it.
As you correctly noticed, many datasets have something similar (internally), which they use when they call `get_seq_order_for_epoch`. Note that this seq length might not really be the correct seq length but just some number from somewhere. E.g. some datasets would use `round(number_of_seconds * 100.)`, some datasets would use character length (even though you might use BPE in the end), etc. That's why I would strongly suggest to reflect that in the function name.

albertz:
Extending on my comment, more specific to `CombinedDataset`:

When you used `get_all_tags`, `get_total_num_seqs` and `get_estimated_seq_length_by_corpus_seq_idx` and then called `get_seq_order_for_epoch` to get the seq order, you would call `init_seq_order` for all the sub datasets with the argument `seq_list` set to the correct specific sub seq list. I think then you get exactly the behavior which you want, as I understand you, right?

However, part of the previous discussion ([here on the commit](https://github.com/rwth-i6/returnn/commit/761062b4b224cb0bcacfe751197ebd6f61b559bc)) was also about efficiency. For very large datasets, this will be very slow, to calculate the whole sequence order beforehand. That is why other frames (T2T, Lingvo, etc) have implemented something similar, but on-the-fly (see `bucket_by_sequence_length` or so). That behavior is different, but it has a similar effect, and it works on-the-fly. I was proposing to also implement sth like that. I think you can just extend the already existing `CombinedDataset._expand_dataset_sec_idxs` for that.
(But let's not mix up things here. We are already discussing maybe 3 separate issues here. They also definitely should not all be together in a single pull request.)

patrick-wilken:
I guess my problem is, that you say that is how the API is designed, but I don't know the parts of the code where this manifests.

> This was/is the design of the API (that it's optional, and can be inaccurate, and can change per epoch, or during an epoch), and many dataset implementations use it that way. 

Can you point me to an example? (Except maybe for what I said about LMDataset.) Of course I could introduce a `reliable_num_seqs`property to not change the API ;) But I don't see the cases where it *should* have a different value than `num_seqs`.

You suggest to use `get_total_num_seqs()` instead of `num_seqs`. But `get_total_num_seqs()` literally just returns `num_seqs` in general. How can one be reliable while the other isn't? Is this just about design? So we define `get_total_num_seqs` to be reliable?

I think we agree about `get_sequence_length` then. So I should use a new function instead that exposes the sequence lengths to the meta datasets without loading the sequence - if this is possible in the particular dataset. (If not, then global sorting is also not possible internally in that dataset, so there is no chance a meta dataset can do it.)

For me, this discussion was never about efficiency. Again, all I want to do is treat several datasets as one. If sorting is efficient enough in the sub-datasets it will be in the CombinedDataset - if you have access to the sub-dataset sorting method...

In part, this PR is complex just because 761062b deleted much complex code ':D I also added stuff in the meantime, and might better have just opened an issue, but as I said, let's first agree on the design and then look at the code.

albertz:
E.g. there is `ExternSprintDataset`. One single example is already enough here. I don't exactly why this need to be discussed further. The API was defined that way, and at least one dataset use it that way. If you now ignore that, and just assume it works otherwise, which it maybe does 70% of the time (maybe 90% of your cases), you still have sometimes cases where it does not work correctly. You never ever want to have cases where you introduce buggy behavior (which will not get noticed, does not throw an exception).
But there are also more examples. E.g. epoch split. Or epoch wise filter. And maybe more. All these would lead to `num_seqs` changing every epoch. And it might be that `num_seqs` is different initially to what it is later.
Also, maybe I'm forgetting cases now. You would have to go through all existing datasets. But some people even define own datasets in their configs.

What is the problem with `get_total_num_seqs`? This is exactly what you want, or not? You can simply check how that is different from `num_seqs`. This API was designed to stay correct. If `num_seqs` does not return the correct value, that is totally ok, and there are multiple cases. If `get_total_num_seqs` does not return the correct value, that is a bug. I don't think there are any cases where `get_total_num_seqs` returns an incorrect value (if there are, please report). In any case, in your code you can have the valid assumption that `get_total_num_seqs` returns a reliable correct value.
Also, `get_total_num_seqs` is a bit different from `num_seqs`. It returns the total num seqs. Not per epoch.

I still don't exactly understand the discussion here.
Are you arguing that the API is bad? Yes, I agree. That is for historically reasons (and was even not my decision, that was before my time; before my time, there was also no explicit API, I just defined the current behavior at that point to be the API).
Are you arguing that the API definition should be changed? I don't think this is possible now. Or at least without lots of work, and probably introducing very nasty hard to find bugs.
Are you arguing that we need some new API for something? But what exactly? Isn't `get_total_num_seqs` already what you want?
Are you arguing that you want to use some specific function (`num_seqs`), although the API defines that this would be incorrect (although maybe for your case it works correct 90% of the time)? I don't really understand the reasoning behind this?

---

Yes, for the seq length, as said, I would introduce a new API, like `get_estimated_seq_length_by_corpus_seq_idx`. You can go through all `get_seq_order_for_epoch` usages and very easily implement that function for many datasets that way. Although it's totally fine if you just implement it for those 2-3 datasets where you currently need it.

Defining an API means that the base class `Dataset` should add the function, and do some reasonable default behavior. In this case, maybe throw an exception (but not `NotImplementedError` -- that would only be for functions which must be implemented by all datasets).

patrick-wilken:
Ok, then using `get_total_num_seqs` will apparently solve this disussion. At least for the case where partition_epoch=1 for the sub-datasets, which was not required before.

But I still don't understand it. ExternSprintDataset uses `get_total_num_seqs` of the base class. Because of that - assuming  partition_epoch=1 - `ExternSprintDataset.num_seq` and `ExternSprintDataset.get_total_num_seqs()` are the same thing!
How can you say that one is not reliable while the other one never returns an incorrect value?

In the code of CombinedDataset at no point we assumed that Dataset.num_seqs is constant across epochs. We only used the value for the current epoch. (And the two places where EpochWiseFilter is used in the code, num_seqs currently reflects that.)
But ok, if I want to have the exact number of sequences in the current epoch, could I introduce a new function, say `get_exact_num_seqs_current_epoch()`? It would have to raise in the base class because we don't know the number in general. But for many datasets that method could be implemented. I would start with TranslationDataset.

albertz:
> Ok, then using `get_total_num_seqs` will apparently solve this disussion. At least for the case where partition_epoch=1 for the sub-datasets, which was not required before.

What do you mean by partition epoch? When you use `get_total_num_seqs`, that's independent of partition epoch (by purpose). Partition epoch does not matter at all then, so nothing is required here. It will be ignored. After you got the seq order, you would set it via `init_seq_order` for all the sub datasets with the argument `seq_list`.

Or do you actually want to not ignore it (partition epoch of sub datasets)? This would be again a different topic. Then `get_total_num_seqs` is not what you want. (That's why I keep asking whether this is what you want or not...)

> But I still don't understand it. ExternSprintDataset uses `get_total_num_seqs` of the base class. Because of that - assuming partition_epoch=1 - `ExternSprintDataset.num_seq` and `ExternSprintDataset.get_total_num_seqs()` are the same thing!

So maybe `get_total_num_seqs` of `ExternSprintDataset` is wrong then. I never said that there are no bugs and everything is perfectly bug free.
It might also be that `ExternSprintDataset.num_seqs` is actually always correct (if it is set). This is fine (there is no violation to the API if the value is not an estimate but the correct value).
I don't know what it is. Would have to check the code again. But why does that matter?
Do not look at what some existing datasets might do and derive that way what you think might be the behavior for all datasets for some API.

> How can you say that one is not reliable while the other one never returns an incorrect value?

I am speaking of the definition of the API.

> In the code of CombinedDataset at no point we assumed that Dataset.num_seqs is constant across epochs.

You did exactly that. The code before used `know_num_seqs_beforehand`, and that was written in a way that it assumed that it would know the `num_seqs` beforehand. But there was no way to know whether `num_seqs` changes later or not.

> We only used the value for the current epoch. (And the two places where EpochWiseFilter is used in the code, num_seqs currently reflects that.)

No, you used it in `__init__`, where the current epoch was even not set.

> But ok, if I want to have the exact number of sequences in the current epoch, could I introduce a new function, say `get_exact_num_seqs_current_epoch()`? It would have to raise in the base class because we don't know the number in general. But for many datasets that method could be implemented. I would start with TranslationDataset.

Wait, now you are speaking about a very different function / API. Before this, we always spoke about `get_total_num_seqs`. I asked you whether this is what you want. You were not totally clear about that, but I understood you that this is what you want. But now, it sounds like this is not what you want? What exactly do you want?
Num seqs for one epoch, after epoch filter, partition epoch, etc was applied? Or global num seqs, independent of current epoch?
Where do you want to have the sorting / filtering / partition epoch logic? In the sub datasets, or in `CombinedDataset`? Or both? Does it make sense to have it in both? Or only partly in both? I.e. filtering and partition epoch in both, but shuffling/sorting only in `CombinedDataset`?

patrick-wilken:
> You did exactly that.

No :D Setting `know_num_seqs_beforehand` did not require knowing any value for the number of sequences. Not the total and not the dynamic one for each epoch. It just assumed that if num_seq can be *accessed* in some first epoch (which is initialised when creating the sub-dataset!), it can be *accessed* also in any other epoch. Do you criticise this assumption, or using this first initial epoch?

> But there was no way to know whether num_seqs changes later or not.

You mean change between epochs or within an epoch? Well, we assumed that it does not change from being defined to being undefined between epochs and does not change within an epoch. You say both assumptions are invalid? Then, what I want to implement is a function that only returns num_seq if both those assumptions are true.

Sorry if I have been unclear about `get_total_num_seqs`. What I need to have is the number of sequences in a given epoch. More exactly, the range in which `get_data()`should be called for the sub-datasets in a given epoch. So after filtering, partition epoch etc. What I meant is that for partition_epoch=1 this number should be the same as the result of `get_total_num_seqs`, which I see now is wrong because of filtering.

In our current implementation sorting and partition epoch can be done on both the meta- and the sub-dataset level. Of course not all combinations make sense, for example sorting the sub-datasets has no affect when using a random order in the meta-dataset. However, there is no overhead in keeping it flexible. Many combinations are useful, like partition/repeat epoch in the sub-datasets and sorting the meta-dataset. Or maybe in-order in the sub-datasets and then choosing datasets randomly in meta-dataset ("random_dataset" seq_ordering option).

albertz:
> > You did exactly that.
> 
> No :D Setting `know_num_seqs_beforehand` did not require knowing any value for the number of sequences. Not the total and not the dynamic one for each epoch. It just assumed that if num_seq can be _accessed_ in some first epoch (which is initialised when creating the sub-dataset!), it can be _accessed_ also in any other epoch. Do you criticise this assumption, or using this first initial epoch?

I might got confused about this code in `__init__`:

    self._num_seqs = sum([self.datasets[k].num_seqs for k in sorted(self.datasets.keys())])

This code is generally wrong (in `__init__`).
But actually you say now this assignment (what value gets assigned to `self._num_seqs`) is not really the relevant part here, but this is just an indirect check whether `num_seqs` of all sub datasets can be read without exception. And then to set `self.know_num_seqs_beforehand = True`. In that case, this is maybe ok (although there definitely should be a comment explaining that).
It was also not clear to me that this is the meaning which should be reflected by `know_num_seqs_beforehand`. I thought it means that the `num_seqs` is known beforehand. That name should be changed. To `_can_use_num_seqs` or so.

Note that the API does not really define whether your assumption is correct or not (that if you can read `num_seqs` once, you can read it always in the future for all epochs). I guess it might be true for all current datasets, but I could imagine cases where this is not true. E.g. imagine a kind of meta-dataset which iterates through multiple sub-datasets per epoch (epoch 1 takes sub-dataset 1, epoch 2 takes sub-dataset 2, etc).

I'm not really sure why you need this `know_num_seqs_beforehand`/`_can_use_num_seqs` at all, though. Why can't you just do the same logic in `init_seq_order`? That would just avoid this assumption. It is in general bad style to have so much state in an object (so much attributes). This makes it very prone to bugs.

> > But there was no way to know whether num_seqs changes later or not.
> 
> You mean change between epochs or within an epoch? Well, we assumed that it does not change from being defined to being undefined between epochs and does not change within an epoch. You say both assumptions are invalid? Then, what I want to implement is a function that only returns num_seq if both those assumptions are true.

It can change definitely between epochs.
I'm actually not totally sure about within an epoch. Maybe it might also change within an epoch. (The problem is also, this API is not really formally clearly defined somewhere, or only partly. There are lots of implicit assumptions here and there. There might also be bugs or inconsistencies. I'm not sure how `ExternSprintDataset` behaves here exactly.)

> Sorry if I have been unclear about `get_total_num_seqs`. What I need to have is the number of sequences in a given epoch. More exactly, the range in which `get_data()`should be called for the sub-datasets in a given epoch. So after filtering, partition epoch etc. What I meant is that for partition_epoch=1 this number should be the same as the result of `get_total_num_seqs`, which I see now is wrong because of filtering.

I see. Then `get_total_num_seqs` is not what you want. (It's also not only about partition epoch. There is also the epoch filter. And maybe other things. Maybe also some custom `seq_list` is provided. Whatever.)

`num_seqs` is actually quite close to what you want. This should be after the filter, for the specific epoch.

Our discussion about seq length is also somewhat off-topic then. I assumed that `get_total_num_seqs` is what you want. `get_estimated_seq_length_by_corpus_seq_idx` would be for this case. This is what the sub dataset would use for `get_seq_order_for_epoch` (together with `get_total_num_seqs`).
But this is not what you want. You want something like `get_estimated_seq_length` (for seq idx of current epoch), and the API should allow to call this function without needing to load the data in advance.

patrick-wilken:
Ahh, ok ':D To my defence, I did not write that line (which is not a good argument ;) ).

Ok, I will move that logic to init_sequence_order(). Then let me come up with a couple of smaller commits to implement what we discussed so far. Probably start of next week...

albertz:
Btw, I just saw that `MetaDataset._get_dataset_seq_length` has the same issue. It uses `self.datasets[self.default_dataset_key].get_seq_length(seq_idx)`, which has the problems I explained earlier (it will blow up the memory by loading the whole dataset into memory). This is only the fallback code there and most often not used (I would always recommend to use `seq_order_control_dataset`). Maybe actually no-one uses that, or only for cases where this problem does not occur. but we should remember to fix that as well.

patrick-wilken:
Ok, I have now updated the commits to a minimal first step. I moved the `num_seqs` logic into `init_seq_order()`, where it should be ok.

`get_sequence_length` indeed caused the whole epoch to be loaded, thanks for pointing that out! I have now overridden it in TranslationDataset as an example. Similar things could be done in other Datasets. I think this is also fine, as opposed to writing a new function, right?

I will add other commits once we agree on the basics ;)

albertz:
No, that's the wrong approach. `get_sequence_length` is really the wrong API for that. Per API, it is required that this is done for the current loaded seq (via `load_seqs`), or the next after it. It is by design that most dataset will load this seq then. That's even the fallback default implementation. You cannot just hope that some dataset does not do this. It's totally ok (and even the default and very common) that datasets do this. So this is really the wrong function to use. (But that's also what I explained already.)

You would need a new function, sth like `get_estimated_sequence_length` (similar as what other datasets use for `get_seq_order_for_epoch`, but not for the corpus seq idx, but for the filtered/local/per-epoch seq idx) (or so; not sure about a good name; "estimated" because it doesn't need to be exakt, you just need it for seq ordering; although in your case it is more tricky...; the aspect that this can be done offline without loading the dataset is also not reflected in the name, but we don't have a good convention for this). It's actually somewhat problematic because such a function might be inconsistent depending on the dataset implementation (e.g. some dataset would return number of chars, some number of labels, some audio length in secs * 100, etc ...). Although as long as you use the same dataset class, this should still be correct.
As I explained already earlier (for `get_estimated_seq_length_by_corpus_seq_idx`), most datasets already have such an implementation internally (when they call `get_seq_order_for_epoch`).

MetaDataset.py
```diff
@@ -885,23 +886,112 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
     for dataset in self.datasets.values():
       dataset.init_seq_order(epoch=epoch)
 
-    self.dataset_sorted_seq_idx_list = []  # We will fill this as we go
-    self.used_num_seqs_per_subset = [0] * len(self.datasets)
+    try:
+      total_num_seqs = sum([self.datasets[k].num_seqs for k in sorted(self.datasets.keys())])
+      # Whether we can get num_seqs from all sub-datasets should not change between epochs, therefore
+      # (self._can_use_num_seqs is False) should never happen here. However, to be sure we keep it False in such a case
+      # to not change behaviour between epochs.
```
albertz, line 907: Why is that relevant?

MetaDataset.py
```diff
@@ -885,23 +886,112 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+      if self._can_use_num_seqs is None:
+        self._can_use_num_seqs = True
+    except Exception:
+      # Here we cannot work around changing behaviour (see comment above).
+      assert not self._can_use_num_seqs, (
+        "Failed to get num_seqs from all sub-datasets although this was possible in a previous epoch.")
```
albertz, line 913: Why is this a problem?

MetaDataset.py
```diff
@@ -847,6 +847,7 @@ def __init__(self,
     # This will only initialize datasets needed for features occurring in data_map
     self.datasets = {key: init_dataset(datasets[key]) for key in self.dataset_keys}
 
+    self._can_use_num_seqs = None  # whether Dataset._num_seqs of all sub-datasets is available, set in init_seq_order()
```
albertz, line 850: Why do you need the attrib? I thought this is enough as a local variable in `init_seq_oder`? (As I explained, it is always bad to have too much shared state. I don't see why this is needed here.)

MetaDataset.py
```diff
@@ -885,23 +886,112 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+      dataset_seq_idx_boundaries.append(dataset_num_seqs + dataset_seq_idx_boundaries[-1])
+
+    return dataset_seq_idx_boundaries
+
+  def _seq_idx_to_dataset_seq_idx(self, seq_idx):
+    """
+    Maps the sequence index (before sorting) to a dataset and a corresponding sequence index within this dataset.
+
+    :param int seq_idx:
+    :rtype: tuple[int]
```
albertz, line 977: `tuple[int]` is too unspecific. Use `(int,int)`.

MetaDataset.py
```diff
@@ -885,23 +886,112 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+    """
+    assert seq_idx < self.dataset_seq_idx_boundaries[-1]
+
+    dataset_idx = 0
+    while self.dataset_seq_idx_boundaries[dataset_idx + 1] <= seq_idx:
+      dataset_idx += 1
+
+    dataset_seq_idx = seq_idx - self.dataset_seq_idx_boundaries[dataset_idx]
+
+    return (dataset_idx, dataset_seq_idx)
```
albertz, line 987: Parenthesis are redundant.

MetaDataset.py
```diff
@@ -885,23 +886,112 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+
+  def _get_seq_length(self, seq_idx):
```
albertz, line 989: Type should be documented.

MetaDataset.py
```diff
@@ -968,7 +1058,7 @@ def _expand_dataset_sec_idxs(self, num_values):
 
   def _load_seqs(self, start, end):
     # If the segment order is not yet known, fix the next few segments
-    if end > len(self.dataset_sorted_seq_idx_list):
+    if not self._can_use_num_seqs and end > len(self.dataset_sorted_seq_idx_list):
```
albertz, line 1061: Why do you need this check?

MetaDataset.py
```diff
@@ -1017,10 +1107,13 @@ def is_less_than_num_seqs(self, n):
     :param int n:
     :rtype: bool
     """
-    if n < len(self.dataset_sorted_seq_idx_list):
-      return True
+    if self._can_use_num_seqs:
+      return n < self._num_seqs
```
albertz, line 1111: Why do you need this? Wouldn't the other code already cover all cases?

MetaDataset.py
patrick-wilken: Ok, I will allow that then. But note that sequence orderings can mean slightly different things, depending on num_seqs being available. E.g "random_dataset" (which I plan to add back in the next commit for the known num_seqs case). In one case you sample from the "true distribution", in the other you rely on estimated_num_seqs.
Also when using ordering not supported if num_seqs is not known, this will cause an error later anyway.

MetaDataset.py
patrick-wilken: Because calling `self._expand_dataset_sec_idxs` does not work in case of `_can_use_num_seqs`. `used_num_seqs_per_subset` is not created in that case and also not all sequence orderings are covered in the function.

MetaDataset.py
patrick-wilken: Same as above.

MetaDataset.py
albertz: Why don't you just create `used_num_seqs_per_subset` in all cases? It's bad that there is some attrib which is only used in some cases but not in others.

MetaDataset.py
albertz: I don't understand. You are anyway using `dataset_sorted_seq_idx_list` in all cases? So this should work?

Note that `is_less_than_num_seqs` is a function like `get_data`, `get_sequence_length` and many others: This is used while iterating through the data, and supposed to work only for current loaded seqs, or maybe the next following seq. This is not supposed to be used for other seqs. See e.g. `CachedDataset2.is_less_than_num_seqs`.

MetaDataset.py
albertz: I don't quite understand. If you want to check `seq_ordering`, you can check for `seq_ordering`.
But here you are checking `_can_use_num_seqs`. `_can_use_num_seqs ` would be some value from a previous epoch (or `None`). Why does it matter what it was in the previous epoch? That's what I don't understand.

MetaDataset.py
patrick-wilken: ´n < self.dataset_sorted_seq_idx_list` also works for the known num_seqs case, yes. But if n is larger we don't want to call `_expand_dataset_sec_idxs()`. Or at least not without rewriting it. But I think it should not be called in that case because in that mode of operation we never expand the list, so checking at epoch end whether we have to is misleading.

MetaDataset.py
patrick-wilken: There are two in principle very different implementations creating the sequence order. Either globally, if num_seqs is known, or just for the next few sequences, in case num_seqs is unknown. The question is whether we want to allow switching between the two from one epoch to the other. For this we would compare to the previous value of `_can_use_num_seqs`. Switching is only possible if the sequence ordering is implemented for both cases, which is not the case in general (often even not possible). So we can either prohibit it in all cases, or allow it and later `_expand_dataset_sec_idxs`might might fail because of an unsupported ordering.

patrick-wilken: Because it might not be intended that when you set "random_dataset" as a sequence ordering in the config, that this means different things for different epochs. But ok, I will just remove the checks.

patrick-wilken:
Regarding `get_sequence_length`: So I would use a new function (`get_estimated_sequence_length `, maybe `get_sequence_length_for_sorting`). What should be the base class functionality here? Is it allowed to fallback to `get_sequence_length` even if this is inefficient? Because I think that is the only general way that works for all datasets.

MetaDataset.py
patrick-wilken: I could check manually that `is_less_than_num_seqs` is false for all sub-dataset, but then we're back at our discussion about num_seqs. If the sequence order is known in advance, than `is_less_than_num_seqs` <=> n < len(seq_order). So I could write ìf self._num_seqs is not None` instead of `if self._can_use_num_seqs:` if I wanted to avoid using the latter.

MetaDataset.py
albertz: I still don't understand. Why is it not ok if it changes for different epochs, but ok otherwise? I would say, either it is ok that the setting can mean different things, or it is not ok.

I would rather say that it should not be solved like that in any case. There should be distinct settings in the config, for different behaviors. A setting should mean one behavior. E.g. if you have some global random order if `num_seqs` is known, call it `global_random_dataset_order` or so, and otherwise sth else. And then if you use `global_random_dataset_order` but `num_seqs` is not available, you would throw an exception.

albertz:
`get_estimated_sequence_length` would throw an exception `OptionalNotImplementedError`. It doesn't make sense to have a default which should never be used (you don't want that you get out-of-mem because it loads the whole dataset).

MetaDataset.py
albertz: Why is `dataset_sorted_seq_idx_list` not available? As I said, it's bad to have separate logic if this is not really needed. Why can't `_expand_dataset_sec_idxs` just handle both cases? It should be simple to define `dataset_sorted_seq_idx_list` for the case when `num_seqs` is available, or not?

LmDataset.py
```diff
@@ -1378,10 +1378,20 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
       self._seq_order = self.get_seq_order_for_epoch(
-        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: len(self._get_data(key=self.main_source_data_key, line_nr=i)))
+        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: self.get_seq_length(i)[self.main_source_data_key])
     self._num_seqs = len(self._seq_order)
     return True
 
+  def get_seq_length(self, sorted_seq_idx):
+    """
+    :type sorted_seq_idx: int
+    :rtype: Util.NumbersDict
```
albertz, line 1388: The type is just `NumbersDict` (because you imported it).

LmDataset.py
```diff
@@ -1378,10 +1378,20 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
       num_seqs = self._get_data_len()
       self._seq_order = self.get_seq_order_for_epoch(
-        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: len(self._get_data(key=self.main_source_data_key, line_nr=i)))
+        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: self.get_seq_length(i)[self.main_source_data_key])
     self._num_seqs = len(self._seq_order)
     return True
 
+  def get_seq_length(self, sorted_seq_idx):
+    """
+    :type sorted_seq_idx: int
```
albertz, line 1387: Rather than using `type`, I think we should use `:param int sorted_seq_idx:`, which is more consistent to the remaining code, and more flexible.

patrick-wilken:
I added `get_estimated_seq_length()` now.
Also, I removed the `_can_use_num_seqs` attribute and because of that had to check `is_less_than_num_seqs` in the end of `_expand_dataset_sec_idxs()` again. And I really don't like it...

Dataset.py
```diff
@@ -259,6 +259,18 @@ def get_seq_length(self, seq_idx):
...
     raise NotImplementedError
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    In contrast to self.get_seq_length(), this method is designed to work for sequences that have not been loaded yet
+    via self.load_seqs().
+    Used by meta-datasets for sequence ordering.
+
+    :param int seq_idx:
+    :rtype: NumbersDict
```
albertz, line 269: I thought that this function was supposed to return an `int`, to be used by `get_seq_order_for_epoch`. Your function is more generic. Which is nice. But it might be problematic in some cases, e.g. when a dataset can give an estimate for only one data key (e.g. `"classes"` but not `"data"`). How would you handle that? Maybe this can return a `NumbersDict` for only the known data keys. But this should be properly documented then.

Dataset.py
```diff
@@ -259,6 +259,18 @@ def get_seq_length(self, seq_idx):
...
+    :returns the len of the input features and the len of the target sequence.
+    """
+    raise OptionalNotImplementedError
```
albertz, line 272: A better default might be just `return NumbersDict()` (see above).

LmDataset.py
```diff
@@ -1378,10 +1378,21 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    :param seq_idx: int
+    :rtype: NumbersDict
+    """
+    corpus_seq_idx = self.get_corpus_seq_idx(seq_idx)
+    assert corpus_seq_idx is not None
+
+    return NumbersDict({key: len(self._get_data(key=key, line_nr=corpus_seq_idx)) for key in self._data})
```
albertz, line 1394: Note: This will be slightly slower as before, as you load all the keys now. (But it's your dataset, so I don't care too much. If you think this does not matter - then just leave it that way.)

Also note: This will in fact block once the data is available. See the implementation of `_get_data`. `TranslationDataset` is anyway maybe an exception because it **always** loads all data into memory (this is exactly what we wanted to avoid for other datasets; all the problems we discussed are about that).

Btw, maybe you want to reimplement `TranslationDataset` later that it does not keeps everything in memory? E.g. you could go through the whole file and only store the offsets for each line (which takes much less memory, and should be much faster). And then `_load_seqs` (or directly `_get_data`) would load only the relevant parts. And the length can be estimated by the diff of the offset.

MetaDataset.py
```diff
@@ -885,23 +884,106 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
-    self.dataset_sorted_seq_idx_list = []  # We will fill this as we go
-    self.used_num_seqs_per_subset = [0] * len(self.datasets)
+    try:
+      total_num_seqs = sum([self.datasets[k].num_seqs for k in sorted(self.datasets.keys())])
+    except Exception:
+      total_num_seqs = None
+
+    if total_num_seqs is not None:
+      self.used_num_seqs_per_subset = [self.datasets[self.dataset_idx2key_map[dataset_idx]].num_seqs
+                                       for dataset_idx in range(len(self.datasets))]
```
albertz, line 908: The indentation should not be done this way. The indentation should always be 2 spaces (or in certain exceptions maybe 4 spaces). To get that here, just make a line-break after the first `[`.

MetaDataset.py
```diff
@@ -958,6 +1040,17 @@ def _expand_dataset_sec_idxs(self, num_values):
               self.estimated_num_seq_per_subset[dataset_idx] = self.used_num_seqs_per_subset[dataset_idx]
 
       else:
+        # If we don't need to expand it is ok if the sorting method is not supported.
```
albertz, line 1043: Why is that relevant? Why not always throw the exception? Wouldn't you run always into cases where it needs to be expanded at some point?

albertz:
> Also, I removed the `_can_use_num_seqs` attribute and because of that had to check `is_less_than_num_seqs` in the end of `_expand_dataset_sec_idxs()` again. And I really don't like it...

It looks like the amount of changes are much simpler, cleaner and shorter now (i.e. positive in every way).
What exactly don't you like?

MetaDataset.py
albertz: Ah this is because you always call `_expand_dataset_sec_idxs` now?
What `seq_ordering` settings does this `else` branch cover here exactly now? For those `seq_ordering` cases where you know that you must have used `num_seqs` in advance, you might directly `return False`. And all other cases should already be handled above, right?

patrick-wilken: Yes, this is only for the sequence orderings implemented in `get_seq_order_for_epoch()`, i.e. the implementation I use when `num_seqs` is known. But do I really want to check here if the sequence ordering is one of those? (random, sorted, laplace, sort_bin_shuffle...). A check whether we know `num_seqs` and therefore used `get_seq_order_for_epoch()` in the first place would be more intuitive to me. (That's what I said I don't like :D)

LmDataset.py
patrick-wilken: Hmm, that's true that it is slightly slower... Maybe we could make a list of data keys an optional argument. If it is given, only load those. If not, load all that are available. But I'm also not sure if it matters.

I'm aware of the other issues. For that reason we have to use HDFDataset for large corpora.

patrick-wilken: I wanted to keep the possibility to sort by target length, I guess. And keep it similar to `get_sequence_length()`. But actually I'm not sure whether it will be useful.

albertz: Maybe you could also make it an argument to `get_estimated_seq_length`. Although maybe the API should define that this argument can also be ignored. So maybe an argument like `suggested_data_key`.

MetaDataset.py
albertz: I see. I think the current solution is still the easiest one and shortest one.
You might want to extend this comment though like this:
```suggestion
        # If we don't need to expand it is ok if the sorting method is not supported.
        # This can happen if we already sorted everything in advance in `init_seq_order`.
```

patrick-wilken:
I decided to keep it simple for now and just let `get_estimated_seq_length` return an int.
I also added this function for HDFDataset. Maybe it could also be put into CachedDataset, depending on what `_get_seq_length_by_real_idx` is designed for...

LmDataset.py
```diff
@@ -1378,10 +1378,21 @@ def init_seq_order(self, epoch=None, seq_list=None):
     else:
       num_seqs = self._get_data_len()
       self._seq_order = self.get_seq_order_for_epoch(
-        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: len(self._get_data(key=self.main_source_data_key, line_nr=i)))
+        epoch=epoch, num_seqs=num_seqs, get_seq_len=lambda i: self.get_estimated_seq_length(i))
```
albertz, line 1381: First: Why make the `lambda` here? You could just directly pass `self.get_estimated_seq_length`.

But anyway: This usage is wrong. The way you designed `get_estimated_seq_length` is that it works for seq idx **after** it is already sorted (i.e. given the current seq order). 
However, `get_seq_order_for_epoch` expects to get it globally, i.e. for the whole dataset, what I call "real seq idx" in HDFDataset.

Dataset.py
```diff
@@ -259,6 +259,20 @@ def get_seq_length(self, seq_idx):
...
 
+  def get_estimated_seq_length(self, seq_idx):
+    """
+    In contrast to self.get_seq_length(), this method is designed to work for sequences that have not been loaded yet
+    via self.load_seqs().
+    Used by meta-datasets for sequence ordering. Currently we only provide one number, i.e. do not give different
+    estimates for the different data keys (as in get_seq_length()). It is up to the dataset what this number represents
+    and how it is computed.
+
+    :param int seq_idx:
```
albertz, line 270: You should clarify this `seq_idx` further. I think you even have confused this yourself (see my other comment below).
* There is the seq idx related to the current epoch (after you called `init_seq_order`) (this is the same seq idx you would use for `load_seqs` or `get_data`).
* There is a seq idx globally for the dataset (independent of sorting, filtering, etc). In other contexts, I call this "real seq idx", or "corpus seq idx".
This real seq idx might not be well defined for all datasets.
If the function `get_corpus_seq_idx` is defined, it will get epoch seq idx -> corpus seq idx.
If `get_all_tags` is defined, the corpus seq idx is an index in this list.

You should make clear which of these two is expected here.
As I understood your reasoning, you want to have this actually considering the current epoch seq order. So it would use the seq idx related to the current epoch. And not the corpus seq idx.
It is enough if you just write:
```suggestion
    :param int seq_idx: for current epoch, not the corpus seq idx
```

LmDataset.py
albertz: See also my comment above. The `get_seq_len` you pass to `get_seq_order_for_epoch` should match the `num_seqs` (which normally is `self.get_total_num_seqs()`, and not `self.num_seqs`), and in general should match the corpus seq idx.

albertz: (My comment above disappeared, because I applied my suggestion myself. :) So here I copy&paste it again.)

To clarify this `seq_idx` further. I think you even have confused this yourself.
* There is the seq idx related to the current epoch (after you called `init_seq_order`) (this is the same seq idx you would use for `load_seqs` or `get_data`).
* There is a seq idx globally for the dataset (independent of sorting, filtering, etc). In other contexts, I call this "real seq idx", or "corpus seq idx".
This real seq idx might not be well defined for all datasets.
If the function `get_corpus_seq_idx` is defined, it will get epoch seq idx -> corpus seq idx.
If `get_all_tags` is defined, the corpus seq idx is an index in this list.

You should make clear which of these two is expected here.
As I understood your reasoning, you want to have this actually considering the current epoch seq order. So it would use the seq idx related to the current epoch. And not the corpus seq idx.
It is enough if you just write (for `get_estimated_seq_length`) (I already committed this):
```
    :param int seq_idx: for current epoch, not the corpus seq idx
```

patrick-wilken:
Damn, you're right. So the `get_estimated_sequence_length` functions are correct, because in CombinedDataset we provide the `seq_idx` for the current epoch. But reusing the function for `get_seq_order_for_epoch` in TranlsationDataset is wrong.

albertz:
> Damn, you're right. So the `get_estimated_sequence_length` functions are correct, because in CombinedDataset we provide the `seq_idx` for the current epoch. But reusing the function for `get_seq_order_for_epoch` in TranlsationDataset is wrong.

Yes. I think this is also the only remaining open issue for the PR now, right?

patrick-wilken:
I hope so ;)

There are other things I was planning to add:
- adding back "random_dataset" sequence order for known num_seqs. I don't use it, but it was there before.
- a sequence order where you take a fixed amount of sequences from the subdatasets per epoch
- some improvement of efficiency by not providing the sequence order to the subdatasets via a sequence list (strings)

Maybe the first one still belongs here, but the other two should be separate PRs.

patrick-wilken:
And once we have a somewhat final implementation I will double check that everything works with a number of different dataset types, maybe also add some tests.

albertz:
So is this done now? Or you haven't finished double checking yet? I assume you comment when you consider this done, right?

patrick-wilken:
Sorry, was busy with IWSLT. I added 'random_dataset' sequence ordering, so now all features that had been there are back. For me, ok to merge.

MetaDataset.py
```diff
@@ -886,7 +886,8 @@ def init_seq_order(self, epoch=None, seq_list=None):
     need_reinit = self.epoch is None or self.epoch != epoch
     num_seqs_saved = self._num_seqs
     super(CombinedDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list)  # resets self._num_seqs
-    self.rnd.seed(epoch or 1)
+    full_epoch = (epoch - 1) // self.partition_epoch + 1 if epoch else 1
+    self.rnd.seed(full_epoch)
```
patrick-wilken, line 890: This seed was unused and now is only used in ´ _get_random_dataset_seq_order()´. Should we move it there? If not, is it ok to do this partition epoch logic here and assume this is what we want also in other cases?

albertz: I think we historically had `self.rnd` in `Dataset` but then made it local (to e.g. `get_seq_order_for_epoch`) (I don't quite remember anymore...).
I would always say that local vars are more clean - i.e. having less state in the object. So better move it to `_get_random_dataset_seq_order`.

Btw, this "random_dataset", is this really sth you use? It doesn't sound so useful to me.

MetaDataset.py
```diff
@@ -885,23 +884,140 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+        dataset = self.datasets[self.dataset_idx2key_map[dataset_idx]]
+        seq_tag = dataset.get_tag(dataset_seq_idx)
+        seq_lists[dataset_idx].append(seq_tag)
+
+      # It may be large, so better delete it early, we don't need it anymore.
+      del seq_order
+
+      # Re-initialize sequence orders of sub-datasets with created sequence list.
+      for dataset_idx, dataset_key in self.dataset_idx2key_map.items():
+        self.datasets[dataset_key].init_seq_order(epoch=epoch, seq_list=seq_lists[dataset_idx])
```
albertz, line 943: Can it happen that you init some datasets twice / multiple times here? (If the `dataset_key` is used multiple times.)

MetaDataset.py
```diff
@@ -885,23 +884,140 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+        counters[dataset_idx] += 1
+        self.dataset_sorted_seq_idx_list.append((dataset_idx, dataset_seq_idx))
+
+      # We only want to load those sequences in the sub-datasets that appear in seq_order. For this, we extract
+      # sequence lists containing the subset of sequences for each dataset from seq_order.
+      seq_lists = [[] for _ in self.datasets]
+      for seq_idx in seq_order:
+        dataset_idx, dataset_seq_idx = self._seq_idx_to_dataset_seq_idx(seq_idx)
+        dataset = self.datasets[self.dataset_idx2key_map[dataset_idx]]
+        seq_tag = dataset.get_tag(dataset_seq_idx)
```
albertz, line 935: I just noticed you call `get_tag` here, which is again not possible in general (unless you have loaded that seq). E.g. check the implementation `CachedDataset2.get_tag`.

There is `get_all_tags`,  which is for the whole corpus, i.e. using corpus seq idx.
(Unfortunately, the implementation in the base class `Dataset.get_all_tags` is pretty buggy, and has the same problem. We maybe should remove that, and also introduce sth like `have_all_tags`. But anyway, this is independent of your PR here...)

There is also `get_corpus_seq_idx`, which gets in the current epoch seq idx, and returns the corpus seq idx.

So some code like this could work:
```
seq_corpus_tags_per_ds = {}
for key, ds in self.datasets.items():
  assert ds.have_corpus_seq_idx()
  seq_corpus_tags_per_ds[key] = ds.get_all_tags()

...

for seq_idx in seq_order:
  ...
  dataset_corpus_seq_idx = dataset.get_corpus_seq_idx(dataset_seq_idx)
  seq_tag = seq_corpus_tags_per_ds[self.dataset_idx2key_map[dataset_idx]][dataset_corpus_seq_idx]
```

MetaDataset.py
albertz: Ah sorry, no, this should be unique.

albertz:
Ok, so a remaining problem with `get_tag`. But see my suggested code, which should solve this.

When this is done, maybe after your gained experience in the messy RETURNN dataset API (to be clear: I like the general design behind it, having these generic data streams, having a possibly infinite stream of data, etc; but the specific API has grown this way for historic reasons, and is somewhat strange or maybe unintuitive, and has certain buggy implementations in some cases, maybe because of that):
* We maybe should extend the documentation of the `Dataset` base class to formalize these conditions, and make that really more clear. A couple of things we stumbled upon here are not really documented well (it only becomes somewhat clear when you check specific implementations).
* Maybe we could do some cleanup. I.e. rename some functions to make their usage more clear. Or introduce other new functions which makes it simpler.
* Maybe you have other ideas?

MetaDataset.py
```diff
@@ -885,23 +883,142 @@ def init_seq_order(self, epoch=None, seq_list=None):
...
+  def _get_random_dataset_seq_order(self, total_num_seqs):
+    """
+    Choose datasets randomly but preserve order within each dataset. This sorting method is unique to CombinedDataset.
+
+    :param int total_num_seqs:
+    :param int epoch:
+    :returns: sequence order
+    :rtype: list[int]
+    """
+    full_epoch = (self.epoch - 1) // self.partition_epoch + 1 if self.epoch else 1
```
albertz, line 987: Btw, there is also `self._get_random_seed_for_epoch`, which might be better to use here.

MetaDataset.py
patrick-wilken: I've made the random seed local.

I remember having used 'random_dataset' for speech translation to mix audio and text input sequences, where it does not make sense to sort by input length, because you would basically separate the corpora. But anyway, this was the only way to shuffle before I made changes to CombinedDataset so I hope somebody was using it :D

MetaDataset.py
patrick-wilken: `Dataset.get_all_tags()` looks strange. It claims to disable partition_epoch but it doesn't. And also it returns the tags in sorted order, therefore your code snippet does not work like this. If I use `seq_tag = seq_corpus_tags_per_ds[self.dataset_idx2key_map[dataset_idx]][dataset_seq_idx]` (so without calculating `dataset_corpus_seq_idx, instead using the sorted index) it works, but I assume this is not how `get_all_tags()` is supposed to be used.

As a separate issue, we had performance issues related to loading and passing tags when using a huge number of sequences. Therefore I would like to avoid loading a list of all tags into memory...

In our code, we even did not use tags here at all, instead I extended `Dataset.init_seq_order()` to take a list of "reinit_indices" that reorder the previously initialised sequence ordering. But that code is probably not general enough...

MetaDataset.py
patrick-wilken: Ah, yes, looks better.

MetaDataset.py
albertz: > `Dataset.get_all_tags()` looks strange. It claims to disable partition_epoch but it doesn't.

Yes, that's what I meant by saying:

>> (Unfortunately, the implementation in the base class `Dataset.get_all_tags` is pretty buggy, and has the same problem. We maybe should remove that, and also introduce sth like have_all_tags. But anyway, this is independent of your PR here...)

So just ignore the `Dataset.get_all_tags` (base class) implementation for now.
(Or maybe fix it. Or maybe just throw `OptionalNotImplementError` in there.)

But still this API is valid, and as I described (corpus seq idx -> seq tag), and correctly implemented by many datasets.

> And also it returns the tags in sorted order, therefore your code snippet does not work like this.

This is just the buggy base class implementation. Ignore it.

> As a separate issue, we had performance issues related to loading and passing tags when using a huge number of sequences. Therefore I would like to avoid loading a list of all tags into memory...

It is usually anyway in memory already. But yes, this creates a new list (usually with references to already existing string objects).
Also yes, this is maybe not optimal.

However, if you don't like this API, then it means you also need to create a new API for this.
 
> In our code, we even did not use tags here at all, instead I extended `Dataset.init_seq_order()` to take a list of "reinit_indices" that reorder the previously initialised sequence ordering. But that code is probably not general enough...

This sounds very ugly, hacky, and not generic. We definitely should not do that.

Unless you maybe mean something different: The current `seq_list` argument is a list of seq tags. We could also allow it to be a list of corpus seq idx. This would be valid and clean and reasonable (for those datasets where corpus seq idx makes sense; basically where `have_corpus_seq_idx()` would return True).

However, we need to make sure that all datasets handle this correctly, or at least catch it and throw an error if not supported. You need to go through all existing dataset implementations and check that.

patrick-wilken:
I now just used your suggest method and made it work for TranslationDataset by implementing the `get_all_tags` method there, hoping that we can finish this PR. I will work a more efficient way.

In the base class I don't think it is possible to provide a working default behaviour, except for the case of `partition_epoch=1` and `seq_ordering="default"`, then using `get_tag()` on all sorted indices would work. We could raise NotImplemented only otherwise.

albertz:
> I now just used your suggest method and made it work for TranslationDataset by implementing the `get_all_tags` method there, hoping that we can finish this PR. I will still work on a more efficient way.

Yes, it will make it really much simpler if we keep stuff separate. It's very ineffective to always check through all the changes again and again each time I'm reviewing this.

> In the base class I don't think it is possible to provide a working default behaviour, except for the case of `partition_epoch=1` and `seq_ordering="default"`, then using `get_tag()` on all sorted indices would work. We could raise NotImplemented only otherwise.

Even then, it has the same problem I raised here, i.e. it uses `get_tag()`, which is invalid in the general case (can lead to loading the whole dataset into memory). So I think `OptionalNotImplementedError` is really the better default behavior.

Looks mostly good now. (I don't like too much that you introduced another attrib for `seq_corpus_tags_per_ds`, so again more state and code more spread around, but I guess you did this for efficiency...)

Maybe let's wait for the tests before we merge.

PR #291, opened 2020-05-12
Allow passing np array directly for init

vieting:
Modified `get_initializer()` in `TFUtil.py` such that it is possible to directly pass a numpy array for initialization, e.g. using `{'class': 'conv',..., 'forward_weights_init': np.array([1, 2, 3])}`. The shape of the array has to be compatible with the shape of the initialized tensor.

### DIFF
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -3814,7 +3814,7 @@ def load_txt_file_initializer(filename, dtype=tf.float32):
 
 def get_initializer(s, seed=None, eval_local_ns=None, dtype=tf.float32):
   """
-  :param str|dict[str]|float s: e.g. "glorot_uniform" or "truncated_normal" or "orthogonal",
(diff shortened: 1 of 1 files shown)


TFUtil.py
```diff
@@ -3869,6 +3869,8 @@ def error():
         f = ns[s + "_initializer"]()
       elif s in ns:
         f = ns[s]()
+    elif isinstance(s, numpy.ndarray):
+      f = ns["Constant"].from_config({'value': s, 'verify_shape': True})
```
albertz, line 3873: Why so complicated and indirect?
Just in the beginning of this function, after `if isinstance(s, (float, int))`, just add this case, and directly use `tf.constant_initializer`.
Then you also don't need other changes in this function.

vieting: Right. From the online documentation I got the impression that the option `verify_shape` wouldn't be available this way but sure it is and that's nicer.

PR #294, opened 2020-05-22
Dataset API discussion

JackTemaki:
This is not really a Pull Request, but a discussion place for the Dataset API. I will push all my comments or additional documentation in a single commit, and they are all marked with TODO for better visibility.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -131,8 +131,9 @@ class Dataset(object):
     self.weights = {}
     self.nbytes = 0
     self.num_running_chars = 0  # CTC running chars.
-    self._num_timesteps = 0
+    self._num_timesteps = 0 # TODO: this is only used by the HDFDataset, nowhere else
     self._num_codesteps = None  # type: typing.Optional[int]  # Num output frames, could be different from input, seq2seq, ctc.  # nopep8
+    # TODO: this variable is never set
     self._num_seqs = 0
     self._estimated_num_seqs = estimated_num_seqs
     self.min_chunk_size = min_chunk_size
@@ -227,6 +228,8 @@ class Dataset(object):
 
   def sliding_window(self, xr):
     """
+    TODO: This function is only used inside datasets -> could be protected
+
     :type xr: numpy.ndarray
     :rtype: numpy.ndarray
     """
@@ -242,6 +245,10 @@ class Dataset(object):
   # noinspection PyMethodMayBeStatic
   def preprocess(self, seq):
     """
+    TODO: This seems to be only called in a function of CachedDataset, which
+      is then only called in HDFDataset. Seems to be completely useless,
+      as there is no override at any point.
+
     :type seq: numpy.ndarray
     :rtype: numpy.ndarray
     """
@@ -249,6 +256,8 @@ class Dataset(object):
 
   def is_cached(self, start, end):
     """
+    TODO: Is only called within the Dataset itself, can be protected or even private
+
     :param int start: like in load_seqs(), sorted seq idx
     :param int end: like in load_seqs(), sorted seq idx
     :rtype: bool
@@ -261,6 +270,9 @@ class Dataset(object):
 
   def get_seq_length_nd(self, sorted_seq_idx):
     """
+    TODO: This function is only replaced in CachedDataset, and used only in get_seq_lenght of the
+      Dataset itself, so this code should never be executed.
+
     :type sorted_seq_idx: int
     :rtype: numpy.ndarray
     :returns the len of the input features and the len of each target sequence.
@@ -277,11 +289,13 @@ class Dataset(object):
 
   def get_seq_length(self, seq_idx):
     """
+    TODO: This function seems to be replaced in all existing subclasses, so the body might be removed here.
+
     :param int seq_idx:
     :rtype: NumbersDict
     :returns the len of the input features and the len of the target sequence.
     """
-    assert self.__class__.get_seq_length_nd is not Dataset.get_seq_length_nd, "Override get_seq_length."
+    assert self.__class__.get_seq_length_nd is not Dataset.get_seq_length_nd, "Override get_seq_length_nd in a subclass"
     input_len, output_len = self.get_seq_length_nd(seq_idx)
     d = {"data": input_len}
     d.update({k: output_len for k in self.get_target_list()})
@@ -291,11 +305,13 @@ class Dataset(object):
     """
     :rtype: int
     """
-    assert self._num_timesteps > 0
+    assert self._num_timesteps > 0 # TODO: this can only be valid for HDFDataset
     return self._num_timesteps
 
   def get_num_codesteps(self):
     """
+    TODO: this function has no usages
+
     :rtype: int|list[int]
     """
     if self._num_codesteps is None:
@@ -346,6 +362,9 @@ class Dataset(object):
 
   def _load_seqs(self, start, end):
     """
+    TODO: the existance of _load_seqs and load_seqs with the same name is not obvious to me,
+      maybe one should be renamed
+
     Load data sequences.
     :param int start: inclusive seq idx start
     :param int end: exclusive seq idx end. can be more than num_seqs
@@ -376,6 +395,9 @@ class Dataset(object):
     Returns the order of the given epoch.
     This is mostly a static method, except that is depends on the configured type of ordering,
     such as 'default' (= as-is), 'sorted' or 'random'. 'sorted' also uses the sequence length.
+
+    TODO: This function as to be used in the subclasses explicitely, so it is possible to write a dataset
+      that does not support the seq_ordering parameter and it will never cause a warning or assert
 
     :param int epoch: for 'random', this determines the random seed
     :param int num_seqs:
@@ -527,6 +549,9 @@ class Dataset(object):
 
   def _get_random_seed_for_epoch(self, epoch):
     """
+    Makes sure that each epoch is shuffled differently,
+    but fixed for repeated training runs.
+
     :param int|None epoch:
     :rtype: int
     """
@@ -545,6 +570,8 @@ class Dataset(object):
 
     This is called when we start a new epoch, or at initialization.
     Call this when you reset the seq list.
+
+    TODO: it does not seem to be obvious what the sub-classes do/should do when this function is caleld
     """
     self.epoch = epoch
     self.rnd_seq_drop = Random(self._get_random_seed_for_epoch(epoch=epoch))
@@ -555,11 +582,15 @@ class Dataset(object):
     This would get called at the end of the epoch (currently optional only).
     After this, further calls to :func:`get_data` or :func:`load_seqs` are invalid,
     until a new call to :func:`init_seq_order` follows.
+
+    TODO: currently this seems to be important only for the SprintDataset
     """
     self.epoch = None
 
   def get_current_seq_order(self):
     """
+    TODO: This function is only used by the MetaDataset, so this might be flagged as protected
+
     :return: many datasets use self.get_seq_order_for_epoch. this function would return the current seq order
       for the current epoch, after self.init_seq_order was called.
       Not all datasets implement this.
@@ -568,6 +599,10 @@ class Dataset(object):
     raise OptionalNotImplementedError
 
   def _base_init(self):
+    """
+    TODO: seems to be only used in initialize and one location SprintDataset, might be moved
+    :return:
+    """
     self.nbytes = 0
     self.zpad = None
     # We expect that the following attributes are already set elsewhere, by a derived class.
@@ -588,6 +623,8 @@ class Dataset(object):
 
   def initialize(self):
     """
+    TODO: Pycharm gives some weird usage places here...
+
     Does the main initialization before it can be used.
     This needs to be called before self.load_seqs() can be used.
     """
@@ -596,6 +633,8 @@ class Dataset(object):
 
   def get_times(self, sorted_seq_idx):
     """
+    TODO: is used only in eval mode, and implemented only by the CachedDataset
+
     :param int sorted_seq_idx:
     """
     raise OptionalNotImplementedError
@@ -606,6 +645,10 @@ class Dataset(object):
     :param str key: data-key, e.g. "data" or "classes"
     :rtype: numpy.ndarray
     :returns features or targets: format 2d (time,feature) (float)
+
+    TODO: the dependencies between the different get_X_data functions is not clear,
+      e.g. HDFDatasets calls this function from the other two
+
     """
     # Fallback implementation for old-style subclasses.
     if key == "data":
@@ -634,6 +677,7 @@ class Dataset(object):
   def get_ctc_targets(self, sorted_seq_idx):
     """
     Warning: This is deprecated/obsolete.
+    TODO: only used in the theano part
 
     :param int sorted_seq_idx:
     :rtype: numpy.ndarray|None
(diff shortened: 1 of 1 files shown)


JackTemaki:
I couldnt find you in the reviewer list, this is also for @patrick-wilken.

JackTemaki:
Also in relation to #290

albertz:
I don't quite understand. What exactly is this about? What do you want to discuss? Why is this a pull request for a discussion?
Do you want to discuss the documentation? Unclear parts, or incorrect parts, or incomplete/missing parts?
Or what else do you want to discuss?

JackTemaki:
> Why is this a pull request for a discussion?

I thought the discussion is easier when I can directly update the comments, and in a pull request the changed code is marked, in contrast to an issue.

> Do you want to discuss the documentation? Unclear parts, or incorrect parts, or incomplete/missing parts?

Yes, that should be a part of it

> Or what else do you want to discuss?

 - Which functions of the Dataset class are relevant in which subclasses (Datasets)
 - Which functions are not needed any more, or can be moved to subclasses as they are only
 used for specific cases
 - which functions should be marked as public or protected, so it is visible if they should be overridden in subclasses, or actually define access to the dataset from outside.
 
In general, I just wanted a place to discuss how to proceed with the confusions about how to use the Dataset (As there are many things I do not understand or easily see from the code), and when things are clear, update the docstrings. And eventually removing/moving/adding code where necessary.

albertz:
For most of these questions, I don't know (or don't remember), you would just have to check the code.
But you could ask more specific questions. But also then, in most cases, I don't know (or don't remember), we would have to check the code.
I will go through your TODOs. Maybe you can also try to answer them yourself (here via a comment) - because as I said, I often don't remember all the small details.

Dataset.py
```diff
@@ -131,8 +131,9 @@ def __init__(self, name=None,
     self.weights = {}
     self.nbytes = 0
     self.num_running_chars = 0  # CTC running chars.
-    self._num_timesteps = 0
+    self._num_timesteps = 0 # TODO: this is only used by the HDFDataset, nowhere else
```
albertz, line 134: (Syntactic note: Space missing before `#`.)
When I search for `_num_timesteps`, I see lots of usages. So your comment is wrong? Or you did not found them? Or what exactly is the question?

Dataset.py
```diff
@@ -131,8 +131,9 @@ def __init__(self, name=None,
...
     self._num_codesteps = None  # type: typing.Optional[int]  # Num output frames, could be different from input, seq2seq, ctc.  # nopep8
+    # TODO: this variable is never set
```
albertz, line 136: You mean `_num_seqs`? But it is set in the next line, right below this comment?

Dataset.py
```diff
@@ -227,6 +228,8 @@ def _load_seq_list_file(filename, use_cache_manager=False, expect_list=True):
 
   def sliding_window(self, xr):
     """
+    TODO: This function is only used inside datasets -> could be protected
```
albertz, line 231: Yes sure.
Why do you add this as a TODO? Why don't you simply do it? (I mean is pretty trivial to check where it is used, and then I think this is a case which is pretty clear. Or if not clear, you can write a short Slack message and ask about it.)

Dataset.py
```diff
@@ -242,13 +245,19 @@ def sliding_window(self, xr):
   # noinspection PyMethodMayBeStatic
   def preprocess(self, seq):
     """
+    TODO: This seems to be only called in a function of CachedDataset, which
+      is then only called in HDFDataset. Seems to be completely useless,
+      as there is no override at any point.
```
albertz, line 250: I don't remember anymore.
Maybe there were some other usages in the past? You could check the history.
But if there are none right now, sure, there is no reason to not remove it.

Dataset.py
```diff
@@ -242,13 +245,19 @@ def sliding_window(self, xr):
...
+
     :type seq: numpy.ndarray
     :rtype: numpy.ndarray
     """
     return seq
 
   def is_cached(self, start, end):
     """
+    TODO: Is only called within the Dataset itself, can be protected or even private
```
albertz, line 259: Well, I see potential usages also from outside for this. Although this might not be correctly implemented by all datasets. So maybe prefix with `_` (the convention for protected in Python; there is nothing for private).

Dataset.py
```diff
@@ -261,6 +270,9 @@ def is_cached(self, start, end):
 
   def get_seq_length_nd(self, sorted_seq_idx):
     """
+    TODO: This function is only replaced in CachedDataset, and used only in get_seq_lenght of the
+      Dataset itself, so this code should never be executed.
```
albertz, line 274: It is executed all the time. See the usages. This is called in `get_seq_length`.
I agree that this function (or API) is there for historical reasons. But you cannot just remove it. First you would need to fix all usages. Basically you would need to implement `CachedDataset.get_seq_length` instead. (If I'm not missing something now. This has to be tested well, as this is very critical code.)

Dataset.py
```diff
@@ -277,11 +289,13 @@ def get_seq_length_nd(self, sorted_seq_idx):
 
   def get_seq_length(self, seq_idx):
     """
+    TODO: This function seems to be replaced in all existing subclasses, so the body might be removed here.
```
albertz, line 292: Are you sure? I see only 6 overwrites. But there are far more subclasses.

Dataset.py
```diff
@@ -291,11 +305,13 @@ def get_num_timesteps(self):
     """
     :rtype: int
     """
-    assert self._num_timesteps > 0
+    assert self._num_timesteps > 0 # TODO: this can only be valid for HDFDataset
     return self._num_timesteps
 
   def get_num_codesteps(self):
     """
+    TODO: this function has no usages
```
albertz, line 313: I remember that I have seen usages in some user scripts (for dumping HDF or so - external, not part of RETURNN). But not sure. Maybe I remember wrongly. Maybe ask @doetsch or @pavelgolik . Or maybe remove and see if someone complains.

Dataset.py
```diff
@@ -346,6 +362,9 @@ def _shuffle_frames_in_seqs(self, start, end):
 
   def _load_seqs(self, start, end):
     """
+    TODO: the existance of _load_seqs and load_seqs with the same name is not obvious to me,
+      maybe one should be renamed
```
albertz, line 366: Well, this is somewhat common in Python (and also in general, in any language), that you see some public function, and some internal function (e.g. just recently I saw `GetNext` and `GetNextInternal` in the C++ API of the TF dataset iterator). The public function usually has some simple logic around, and you are supposed to overwrite the internal function.

This specific example mostly follows this common convention. I'm not sure if another name helps. Maybe `_load_seqs_internal`, but you could argue that the `_` prefix already says the same thing.

Of course, I'm always open to extend the documentation to make this more clear.

albertz:
Btw, as a general comment about the documentation of how to write subclasses: This should be in the docstring of the class itself (as an overview, describing the general logic, maybe pointing to the most important functions), and then more specific in the docstring of the individual functions. So I agree that the docstring of `Dataset` can be extended.

Note that I recommend to subclass from `CachedDataset2` instead, which simplifies many things. And that class already provides such a (brief) documentation.

Dataset.py
```diff
@@ -377,6 +396,9 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
     This is mostly a static method, except that is depends on the configured type of ordering,
     such as 'default' (= as-is), 'sorted' or 'random'. 'sorted' also uses the sequence length.
 
+    TODO: This function as to be used in the subclasses explicitely, so it is possible to write a dataset
+      that does not support the seq_ordering parameter and it will never cause a warning or assert
```
albertz, line 400: It is possible to write all sorts of buggy or strange code. You cannot avoid that.

A good behaving subclass should of course check for `seq_ordering` (in `init_seq_order`), and do the corresponding thing, and if it cannot handle some case, it should throw an exception. I think most datasets do this correctly, or not?

Dataset.py
JackTemaki: It is overwritten in CachedDataset, CachedDataset2, GeneratingDataset, MetaDataset and SprintBaseDataset, which covers all Datasets that I found so far. And only in CachedDataset there is a reference to `self.get_seq_length_nd`, which is why I thought this is not needed in the Dataset anymore

Dataset.py
```diff
@@ -545,6 +570,8 @@ def init_seq_order(self, epoch=None, seq_list=None):
 
     This is called when we start a new epoch, or at initialization.
     Call this when you reset the seq list.
+
+    TODO: it does not seem to be obvious what the sub-classes do/should do when this function is caleld
```
albertz, line 574: Well, then extend the documentation if you think it's unclear?

Dataset.py
```diff
@@ -555,11 +582,15 @@ def finish_epoch(self):
     This would get called at the end of the epoch (currently optional only).
     After this, further calls to :func:`get_data` or :func:`load_seqs` are invalid,
     until a new call to :func:`init_seq_order` follows.
+
+    TODO: currently this seems to be important only for the SprintDataset
```
albertz, line 586: Yes, so? This could potentially be important for all kinds of datasets. This potentially allows for some early cleanup, or better error handling, etc.

Dataset.py
```diff
@@ -555,11 +582,15 @@ def finish_epoch(self):
...
     """
     self.epoch = None
 
   def get_current_seq_order(self):
     """
+    TODO: This function is only used by the MetaDataset, so this might be flagged as protected
```
albertz, line 592: No, this is not just used via `self`, but also for other datasets (externally). Protected functions/attributes are mostly just used via `self`. This is an example which I would make public.

Dataset.py
JackTemaki: This is still unclear for me, so I am not sure if I can do it.

Dataset.py
```diff
@@ -568,6 +599,10 @@ def get_current_seq_order(self):
     raise OptionalNotImplementedError
 
   def _base_init(self):
+    """
+    TODO: seems to be only used in initialize and one location SprintDataset, might be moved
```
albertz, line 603: I think I put it there, so that other datasets potentially could also use it. I think it covers some more exotic cases, so normally it does not matter whether this is called or not. And it might also be too specific for `SprintDataset` for historical reasons but not sure. I don't remember the details anymore. But when I look at the code, I guess that most other datasets handle `window` wrongly (or not at all), and also `nbytes` or `zpad` are probably not correctly handled (but also that might not really be relevant in most cases).

Dataset.py
```diff
@@ -588,6 +623,8 @@ def _base_init(self):
 
   def initialize(self):
     """
+    TODO: Pycharm gives some weird usage places here...
```
albertz, line 626: What is this TODO about?
(Probably you speak about cases where PyCharm cannot correctly infer the type of some object. Then it cannot know whether this is this `initialize`, or from some other class, so it shows it (under "unclassified usages" or sth like that). This is the best thing it can do. You could go through this list and check why PyCharm could not correctly infer the type. But that's other code, not related to here in `Dataset.py`. There is no problem here in `Dataset.py`.)

Dataset.py
```diff
@@ -596,6 +633,8 @@ def initialize(self):
 
   def get_times(self, sorted_seq_idx):
     """
+    TODO: is used only in eval mode, and implemented only by the CachedDataset
```
albertz, line 636: Yes, maybe. This could be for historical reasons. I don't remember anymore.
But what is your TODO about? What do you want to do about it?

Dataset.py
```diff
@@ -606,6 +645,10 @@ def get_data(self, seq_idx, key):
     :param str key: data-key, e.g. "data" or "classes"
     :rtype: numpy.ndarray
     :returns features or targets: format 2d (time,feature) (float)
+
+    TODO: the dependencies between the different get_X_data functions is not clear,
+      e.g. HDFDatasets calls this function from the other two
```
albertz, line 650: They are there for historical reasons.
In an optimal world, you would only overwrite `get_data`, and also only use `get_data`.

Dataset.py
```diff
@@ -634,6 +677,7 @@ def get_targets(self, target, sorted_seq_idx):
   def get_ctc_targets(self, sorted_seq_idx):
     """
     Warning: This is deprecated/obsolete.
+    TODO: only used in the theano part
```
albertz, line 680: What is the TODO about?

Dataset.py
```diff
@@ -642,6 +686,8 @@ def get_ctc_targets(self, sorted_seq_idx):
 
   def get_data_slice(self, seq_idx, key, start_frame, end_frame):
     """
+    TODO: only used in the theano part
```
albertz, line 689: What is the TODO about?

Dataset.py
```diff
@@ -655,6 +701,9 @@ def get_data_slice(self, seq_idx, key, start_frame, end_frame):
     return data[start_frame:end_frame]
 
   def _get_data_slice_sparse(self, seq_idx, key, start_frame, end_frame):
+    """
+    TODO: This function has no usages
```
albertz, line 705: That's wrong. It's used in `get_data_slice`.

Dataset.py
```diff
@@ -694,6 +743,13 @@ def get_all_tags(self):
 
   def get_total_num_seqs(self):
     """
+    TODO: this function is only used in DataSet and MetaDataset, so can be protected
```
albertz, line 746: No, as explained, this is used via external usage, not via `self`, so it should be public.

Dataset.py
```diff
@@ -694,6 +743,13 @@ def get_all_tags(self):
...
+      This function should either be removed, or given a useful purpose
+      In Dataset is only used for an assert when self.seq_tags_filter is not None
```
albertz, line 748: This is very much the standard case, that `seq_tags_filter` is `None`.

Dataset.py
albertz: Also, it's used in other places as well.

albertz: Also, it's used in more places.

Dataset.py
```diff
@@ -694,6 +743,13 @@ def get_all_tags(self):
...
+      In MetaDataset, it is surrounded with a try that catches for NotImplemented,
+      and just continues without any exception.
+      Meaning, this function is completely useless right now, and not used at any relevant point
```
albertz, line 751: This function is actually very important for any kind of meta dataset. I don't really understand why you think it's not important?

Dataset.py
albertz: There are about 20-30 datasets.

Dataset.py
albertz: I don't exactly understand. What is unclear?

albertz: This is called when we start a new epoch, or at initialization. The subclasses can do whatever they want here.
Usually they would initialize the seq order, maybe do some other preparation for the epoch (whatever they want).

Dataset.py
JackTemaki: Yes, and all are based on the 6 Mentioned. So the function from Dataset itself is never called, also not via super()

JackTemaki:
I will split this into smaller pull requests for the part where it is necessary, as it might not be beneficial to discuss this in a larger PR

Dataset.py
albertz: Well, if you are sure (I did not check, but I assume you know how to check this), you can remove this dummy code, or rather move to `CachedDataset`, and put a `raise NotImplementedError` here.

PR #295, opened 2020-05-25
removed unused get_seq_length/get_seq_length_nd bodies

JackTemaki:
As discussed in #294, I think the function bodies can be removed/replaced by a NotImplementedException

I ran the dataset tests locally, but made a PR to make sure all tests are fine.

### DIFF
--- a/Dataset.py
+++ b/Dataset.py
@@ -251,33 +251,13 @@ class Dataset(object):
     assert start < end
     return False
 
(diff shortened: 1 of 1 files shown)


albertz:
No, we discussed that this can be moved to `CachedDataset`. It's used in many places there. I wonder why no test case covers this.

PR #298, opened 2020-06-08
UTF-8 compatibility for bliss-to-ogg-zip tool

JackTemaki:
I switched recently from my own custom tool for ogg zip creation to the official one, and found that it crashes when I have non-ASCII characters in the "orth" segments. I could fix the problem by adding a custom parser that runs with UTF-8 encoding. I am just not sure if this creates any unwanted side-effects.


albertz:
I also don't know. But I'm just merging this for now.

Issue #304, opened 2020-06-15
Targeting Beam as Training Target Causes Dimension Error


npwynands:
This issue is related to the following StackOverflow question: [Teacher-Student System: Training Student With k Target Sequences for Each Input Sequence](https://stackoverflow.com/questions/62344553/teacher-student-system-training-student-with-k-target-sequences-for-each-input/62359741#62359741)

I want to configure a teacher-student system, where a teacher seq2seq model generates a top-*k* list of hypotheses, which are used to train a student seq2seq model.

I select the top-*k* hypotheses list from the teacher’s `ChoiceLayer` (or output layer) by:
```python
"teacher_hypotheses": {
    "class": "copy", "from": ["extra.search:teacher_output"],
    "register_as_extern_data": "teacher_hypotheses_data"
}
```
The output `Data` of that layer has a batch axis length `batch_size`=*k*=4 times the length of the input `Data`’s batch axis length (cf. doc and code of: `Data.copy_extend_with_beam`, `SearchChoices.translate_to_common_search_beam`).

`teacher_hypotheses_data` is selected as the student’s training target. But this leads to the following error:
```shell
TensorFlow exception: assertion failed: [shape[0]:] [92] [!=] [dim:] [23]
     [[node student_output/rec/subnet_base/check_seq_len_batch_size/check_input_dim/assert_equal_1/Assert/Assert (defined at home/philipp/Documents/bachelor-thesis/returnn/returnn-venv/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py:1748) ]]
```
Which is, I assume, due to the issue that the target data of the student, the hypotheses list, has a batch axis length *k*=4 times longer than the one of the student’s input data/encoder state data.

According to Albert’s comment, this might be a bug. I attached you a simplified version of my config, which causes the same error. In this case it uses dummy data, so no datasets are needed.

**Environment**:
- OS: Ubuntu 18 (local machine, not the cluster)
- Python 3.7.5
- TensorFlow 1.15.2
- RETURNN (commit 8595f10b03067b0a2253d0ee65d7dd7efd62ce76; 27th May, 15:40)

**Example Config**:
[beam-as-target.py.txt](https://github.com/rwth-i6/returnn/files/4779979/beam-as-target.py.txt)

albertz:
This should be fixed now (via PR #310). Please test.

albertz:
Btw, if you want to see how to work on a fix like this:

[...]

PR #310, opened 2020-06-21
Fix target beam #304

albertz:


albertz:
This should fix #304.

PR #312, opened 2020-06-22
ActivationLayer output dtype

vieting:
I added `get_out_data_from_opts()` for the `ActivationLayer` and added the case of the abs activation on complex values, in which case the output `dtype` should change from `complex64` to `float32`.


vieting:
There was a failing test case (`test_Loss_NCHW` in `test_TFNetworkLayer.py`) because the `ActivationLayer` was called directly without setting `output` explicitly. I adapted the call according to how e.g. the `SliceLayer` is called in `test_SliceLayer_NCHW`, i.e., setting output using `ActivationLayer.get_out_data_from_opts()`. Not sure though if we should change the test this way...

vieting:
Just saw that this is related to #240 which can be closed now I suppose.

PR #313, opened 2020-06-23
Support for attention energies without encoder dependencies

JackTemaki:
This PR adds an additional test case to enable the construction of specific attention methods that do not require an explicit encoder context. This added test case uses an explicit DimensionTag for correct template construction and an `initial_output` for the weight-feedback based on the encoder for correct tensor shapes.

Two fixes were needed for this to run:
- the deepcopy function does not try to copy tensors and dimension tags
- the `declare_same_as` function to adopt to external DimensionTags is now aware of a possible beam search and adjusts the batch axis accordingly

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -4812,7 +4812,7 @@ class ConvLayer(_ConcatInputLayer):
             filter_size=filter_size[i], stride=strides[i], dilation_rate=dilation_rate[i], padding=padding)
     feature_dim_axis = NotSpecified
     # Swap the dims if the input dim order doesn't fit the flag auto_use_channel_first.
-    if (TFUtil.is_gpu_available() and auto_use_channel_first) or data.is_batch_feature_major:
+    if TFUtil.is_gpu_available() and (auto_use_channel_first or data.is_batch_feature_major):
       feature_dim_axis = 1
       shape = shape[-1:] + shape[:-1]
     return {
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -252,6 +252,12 @@ class DimensionTag(object):
           self.src_data.size_placeholder = {}
         self.src_data.size_placeholder[
           self.src_data.get_batch_axis_excluding_batch(self.src_axis)] = self.same_as.dyn_size
+        # if the tag is used in a recurrent layer during search, the placeholder has to be expanded by the beam size
+        if self.src_data.beam and (not self.same_as.src_data or not self.same_as.src_data.beam):
+          for i, v in sorted(self.src_data.size_placeholder.items()):
+            with same_control_flow_ctx(v):
+              self.src_data.size_placeholder[i] = tile_transposed(v, axis=0, multiples=self.src_data.beam.beam_size)
+            self.set_tag_on_size_tensor(self.src_data.size_placeholder[i])
     # If others dyn_size is None but we have a dyn_size, maybe update others dyn_size.
     if self.dyn_size is not None and self.same_as.dyn_size is not self.dyn_size:
       # Could be unset if it comes from the config, or from prev graph creation.
@@ -2125,7 +2131,8 @@ class Data(object):
     axis_wo_batch = self.get_batch_axis_excluding_batch(axis)
     if axis_wo_batch in self.size_placeholder:
       return True  # not quite the same as get_dynamic_axes
-    assert isinstance(self.batch_shape[axis], int)
+    assert isinstance(self.batch_shape[axis], int), (
+      "%s: the requested axis has neither a size_placeholder entry nor a fixed size" % self)
     return False
 
   def get_dynamic_size(self, axis):
--- a/Util.py
+++ b/Util.py
@@ -2183,17 +2183,26 @@ def unicode_to_str(s):
   return s
 
 
-def deepcopy(x):
+def deepcopy(x, stop_types=None):
   """
   Simpler variant of copy.deepcopy().
   Should handle some edge cases as well, like copying module references.
 
   :param T x: an arbitrary object
+  :param list[type]|None stop_types: objects of these types will not be deep-copied, only the reference is passed
   :rtype: T
   """
   # See also class Pickler from TaskSystem.
   # Or: https://mail.python.org/pipermail/python-ideas/2013-July/021959.html
   from TaskSystem import Pickler, Unpickler
+
+  persistent_memo = {}  # id -> obj
+
+  def persistent_id(obj):
+    if stop_types and isinstance(obj, tuple(stop_types)):
+      persistent_memo[id(obj)] = obj
+      return id(obj)
+    return None
 
   def pickle_dumps(obj):
     """
@@ -2202,6 +2211,7 @@ def deepcopy(x):
     """
     sio = BytesIO()
     p = Pickler(sio)
+    p.persistent_id = persistent_id
     p.dump(obj)
     return sio.getvalue()
 
@@ -2212,6 +2222,7 @@ def deepcopy(x):
     :rtype: object
     """
     p = Unpickler(BytesIO(s))
+    p.persistent_load = persistent_memo.__getitem__
     return p.load()
 
   s = pickle_dumps(x)
--- a/tests/test_TFEngine.py
+++ b/tests/test_TFEngine.py
@@ -1078,6 +1078,87 @@ def test_attention_train_then_search():
   engine.finalize()
 
 
+def test_attention_no_encoder_dependency():
+  from GeneratingDataset import DummyDataset
+  seq_len = 5
+  n_data_dim = 2
+  n_classes_dim = 7
+  train_data = DummyDataset(input_dim=n_data_dim, output_dim=n_classes_dim, num_seqs=2, seq_len=seq_len)
+  train_data.init_seq_order(epoch=1)
+  dev_data = DummyDataset(input_dim=n_data_dim, output_dim=n_classes_dim, num_seqs=2, seq_len=seq_len)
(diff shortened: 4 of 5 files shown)


TFUtil.py
```diff
@@ -252,6 +252,12 @@ def declare_same_as(self, other):
           self.src_data.size_placeholder = {}
         self.src_data.size_placeholder[
           self.src_data.get_batch_axis_excluding_batch(self.src_axis)] = self.same_as.dyn_size
+        # if the ag is used in a recurrent layer during search, the placeholder has to be expanded by the beam size
```
albertz, line 255: Small typo...

Util.py
```diff
@@ -2183,25 +2183,35 @@ def unicode_to_str(s):
...
 
 
-def deepcopy(x):
+def deepcopy(x, stop_types=None):
   """
   Simpler variant of copy.deepcopy().
   Should handle some edge cases as well, like copying module references.
 
   :param T x: an arbitrary object
+  :param list[type] stop_types: objects of these types will not be copied
```
albertz, line 2192: Wrong type. Should be `list[type]|None`.
Also wrong or misleading documentation. They will be copied, just not deep but shallow (i.e. just the ref).

Util.py
```diff
@@ -2183,25 +2183,35 @@ def unicode_to_str(s):
...
   :rtype: T
   """
   # See also class Pickler from TaskSystem.
   # Or: https://mail.python.org/pipermail/python-ideas/2013-July/021959.html
   from TaskSystem import Pickler, Unpickler
 
+  persistent_memo = {}
+
+  def persistent_id(obj):
+    if type(obj) in stop_types:
```
albertz, line 2202: `stop_types` can also be `None`.
Also, maybe `isinstance` is better? It's not exactly the same, but I think `isinstance` would make more sense.

tests/test_TFEngine.py
```diff
@@ -1078,6 +1078,86 @@ def test_attention_train_then_search():
...
+  n_classes_dim = 7
+  train_data = DummyDataset(input_dim=n_data_dim, output_dim=n_classes_dim, num_seqs=2, seq_len=seq_len)
+  train_data.init_seq_order(epoch=1)
+  dev_data = DummyDataset(input_dim=n_data_dim, output_dim=n_classes_dim, num_seqs=2, seq_len=seq_len)
+  dev_data.init_seq_order(epoch=1)
+
+  from TFUtil import DimensionTag
+  enc_time = DimensionTag(kind=DimensionTag.Types.Spatial, description="enc time")
+
+  config = Config()
```
albertz, line 1094: You should include `debug_print_layer_output_template` (and maybe other things) in the config. Like:
```suggestion
  config = Config({"debug_print_layer_output_template": True})
```

albertz: Ah sorry, did not see that you call `update` then. (You could skip that by a single call...)

albertz:
Also check the failing tests.

Util.py
JackTemaki: If we also want to check for possible derived classes, yes. I am not sure if this exists.

albertz: `isinstance` does exactly that. And usually it is recommend to use `isinstance` instead of `type(...) == ...` (PyCharm should also warn you about that). So I think it would follow that general convention (unless there are very good/explicit reasons to not do that).

JackTemaki: As the relation is "in", there is no warning, but I will change this to isinstance.

TFNetworkLayer.py
```diff
@@ -4697,7 +4697,7 @@ def __init__(self, n_out, filter_size, padding, strides=1, dilation_rate=1,
     if input_add_feature_dim:
       # Add a feature dimension; any other static dims will be used as dynamic dims below.
       input_data = input_data.copy_add_feature_dim()
-    if self.output.is_batch_feature_major:
+    if self.output.is_batch_feature_major and TFUtil.is_gpu_available():
```
albertz, line 4700: This is wrong. The output *must* be as defined. The output should be correctly defined in `get_out_type_from_opts`.

TFUtil.py
```diff
@@ -2125,7 +2131,9 @@ def is_axis_dynamic(self, axis):
     axis_wo_batch = self.get_batch_axis_excluding_batch(axis)
     if axis_wo_batch in self.size_placeholder:
       return True  # not quite the same as get_dynamic_axes
-    assert isinstance(self.batch_shape[axis], int)
+    assert isinstance(self.batch_shape[axis], int), (
+      "the requested axis has neither a size_placeholder entry nor a fixed size"
+    )
```
albertz, line 2136: The closing bracket should be on the same line as the text. (See all other code. Keep code consistent.)
Also, you should add `"%s: the requested ..." % self)` for a better debug message.

Util.py
```diff
@@ -2183,25 +2183,35 @@ def unicode_to_str(s):
...
   :rtype: T
   """
   # See also class Pickler from TaskSystem.
   # Or: https://mail.python.org/pipermail/python-ideas/2013-July/021959.html
   from TaskSystem import Pickler, Unpickler
 
+  persistent_memo = {}
+
+  def persistent_id(obj):
+    if any([isinstance(obj, stop_type) for stop_type in (stop_types or [])]):
```
albertz, line 2202: This is way too complicated and hard to read.
Just do
```suggestion
    if stop_types and isinstance(obj, stop_types):
```

albertz:
Btw, it would be good to cleanup the commits (not keeping separate commits for fixing the typo, introduced in a previous commit; just merge them together).

JackTemaki:
> Btw, it would be good to cleanup the commits (not keeping separate commits for fixing the typo, introduced in a previous commit; just merge them together).

albertz:
Why did you close this?

JackTemaki:
> Why did you close this?

This was not intended

Util.py
JackTemaki: It was not clear to me that isinstance takes lists, I will update this

JackTemaki:
> Btw, it would be good to cleanup the commits (not keeping separate commits for fixing the typo, introduced in a previous commit; just merge them together).

Do you mean rebasing teach fix into the original commit introducing the code, or squashing the minor fixes into one commit?

Util.py
albertz: It takes tuples. Maybe we should enforce the type to be a tuple (then you must fix the usage), or convert to tuple here.
```suggestion
    if stop_types and isinstance(obj, tuple(stop_types)):
```

albertz:
> > Btw, it would be good to cleanup the commits (not keeping separate commits for fixing the typo, introduced in a previous commit; just merge them together).
> 
> Do you mean rebasing each fix into the original commit introducing the code, or squashing the minor fixes into one commit?

The commit history should be clean. There is no point in having a separate (single or multiple) "cleanup" commit, which cleans up previous commits.

Util.py
JackTemaki: `isinstance` does not accept lists, but tuples work

TFUtil.py
```diff
@@ -252,6 +252,12 @@ def declare_same_as(self, other):
           self.src_data.size_placeholder = {}
         self.src_data.size_placeholder[
           self.src_data.get_batch_axis_excluding_batch(self.src_axis)] = self.same_as.dyn_size
+        # if the Tag is used in a recurrent layer during search, the placeholder has to be expanded by the beam size
```
albertz, line 255: "tag" should be lower case.

Util.py
```diff
@@ -2183,25 +2183,35 @@ def unicode_to_str(s):
...
 
   :param T x: an arbitrary object
+  :param list[type]|None stop_types: objects of these types will not be deep-copied, only the reference is passed
   :rtype: T
   """
   # See also class Pickler from TaskSystem.
   # Or: https://mail.python.org/pipermail/python-ideas/2013-July/021959.html
   from TaskSystem import Pickler, Unpickler
 
+  persistent_memo = {}
```
albertz, line 2199: I think it would be good to add a comment of the content:
```suggestion
  persistent_memo = {}  # id -> obj
```

JackTemaki: Do you want things like this as user hint only, or as PyCharm type hint, such as `persistent_memo = {}  # type: dict[id, object]`?

albertz: (There should be two spaces.)
This is not a valid PyCharm type hint. The valid PyCharm type hint would be `typing.Dict[int,object]`.
However, even more important here is the semantic information, `id -> obj`. The type is probably already correctly automatically derived by PyCharm. I leave it to you if you want to have both here or just the semantic info.

JackTemaki: For me this was working in PyCharm, but you are right that typing should be used. I will just use short info.

albertz: This definitely does not work. It probably ignored it (but even with a warning that it is incorrect), and just used the auto-derived type.

JackTemaki: It definitely did change the function autocompletion suggestion when I edited the types in the hint.

albertz:
Btw, please click on "resolve" when you resolved some of the comments.
There is still an outstanding problem.

Util.py
JackTemaki: Please try it for yourself in an empty file, for me there is no warning, and correct autocompletion hints.

TFNetworkLayer.py
JackTemaki: Yes, you are right that this is redundant code, as the output will be set correctly anyway. I will remove it.

albertz:
You added an empty file `xyz.py`. Please remove this.

PR #315, opened 2020-06-24
TFNetworkLayer: add DctLayer

vieting:
I added a `DctLayer` which wraps `tf.signal.dct`. So far, that's all it does, so no specific handling of axes is done. `tf.signal.dct` always computes the DCT over the last dimension.

### DIFF
--- a/TFNetworkLayer.py
+++ b/TFNetworkLayer.py
@@ -4963,6 +4963,45 @@ class PoolLayer(_ConcatInputLayer):
       beam=data.beam)
 
 
+class DctLayer(_ConcatInputLayer):
+  """
+  Layer to perform DCT
+  Wraps :func:`tf.signal.dct`. For further documentation on the input arguments, refer to
+  https://www.tensorflow.org/api_docs/python/tf/signal/dct
+  """
+
+  layer_class = "dct"
+  recurrent = True  # we should not shuffle in the time-dimension
+
+  def __init__(self, type=2, n=None, norm=None, **kwargs):
+    """
+    :param int type: DCT type to perform. Must be 1, 2, 3, or 4
+    :param int|None n: length of the transform
+    :param str|None norm: normalization to apply. Must be None or "ortho"
+    """
+    assert n is None, 'Not implemented yet for n other than None'
+    super(DctLayer, self).__init__(**kwargs)
+    input_data = self.input_data.copy_as_batch_spatial_major()
+    x = input_data.placeholder
+    y = TFCompat.v1.spectral.dct(x, type=type, n=n, norm=norm)
+    self.output.placeholder = y
+    if n is None:
+      self.output.size_placeholder = self.input_data.size_placeholder.copy()
+    else:
+      raise NotImplementedError
+
+  @classmethod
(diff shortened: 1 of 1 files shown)


TFNetworkLayer.py
```diff
@@ -4963,6 +4963,45 @@ def get_out_data_from_opts(cls, name, pool_size, strides=None, dilation_rate=1,
...
+  Wraps :func:`tf.signal.dct`. For further documentation on the input arguments, refer to
+  https://www.tensorflow.org/api_docs/python/tf/signal/dct
+  """
+
+  layer_class = "dct"
+  recurrent = True  # we should not shuffle in the time-dimension
+
+  def __init__(self, type=2, n=None, norm=None, **kwargs):
+    """
+    :param type: DCT type to perform
```
albertz, line 4978: Type missing. I guess `int`?

TFNetworkLayer.py
```diff
@@ -4963,6 +4963,45 @@ def get_out_data_from_opts(cls, name, pool_size, strides=None, dilation_rate=1,
...
+    :param n: length of the transform
```
albertz, line 4979: Type missing. I guess `int`?

TFNetworkLayer.py
```diff
@@ -4963,6 +4963,45 @@ def get_out_data_from_opts(cls, name, pool_size, strides=None, dilation_rate=1,
...
+    :param norm: normalization to apply
```
albertz, line 4980: Type missing? I have no idea, maybe `str|None`?

TFNetworkLayer.py
```diff
@@ -4963,6 +4963,45 @@ def get_out_data_from_opts(cls, name, pool_size, strides=None, dilation_rate=1,
...
+    """
+    assert n is None, 'Not implemented yet for n other than None'
+    super(DctLayer, self).__init__(**kwargs)
+    input_data = self.input_data.copy_as_batch_spatial_major()
+    x = input_data.placeholder
+    y = tf.spectral.dct(x, type=type, n=n, norm=norm)
```
albertz, line 4986: Better use `tf.signal.dct`. I think `tf.spectral` is deprecated since a long time.

vieting: `tf.signal` only works with tf versions >1.12. I am mostly using 1.8, that's why I used `tf.spectral`. Should we add some workaround for older versions?

albertz: But I think this code does not work on TF 2 then. And I would say that supporting a recent TF version has a higher priority than a very outdated TF version. Also see the contributing guidelines about this. There is the rule that new code must work with a recent TF version.

You could maybe use `TFCompat.v1.spectral.dct`.
Or you could simply update your TF version, which I anyway would recommend.

vieting: I agree of course to the priorities.

> You could maybe use `TFCompat.v1.spectral.dct`

Do you mean as kind of a fallback if `tf.signal` is not working?

albertz: The code should also be simple, so avoid having two code paths there.

vieting: I'll add the compatibility version then for now.

PR #316, opened 2020-06-29
TFUtil: np array initialization

vieting:
I added another case in `get_initializer()` that allows passing strings which result in a numpy array when passed to `eval()`. Example for a possible usage: `'forward_weights_init': 'numpy.hanning(400).reshape((400, 1, 1, 1))'`

### DIFF
--- a/TFUtil.py
+++ b/TFUtil.py
@@ -3924,6 +3924,8 @@ def get_initializer(s, seed=None, eval_local_ns=None, dtype=tf.float32):
       if f == 1:
         return TFCompat.v1.ones_initializer(dtype=dtype)
       return TFCompat.v1.constant_initializer(f, dtype=dtype)
+    elif isinstance(f, numpy.ndarray):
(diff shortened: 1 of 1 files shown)


albertz:
Why is this needed? You could simply directly use `'forward_weights_init': numpy.hanning(400).reshape((400, 1, 1, 1))`. This looks cleaner, and also should be faster, right?

vieting:
That's nicer, sure. But if the config should be created with sisyphus, it's straight forward to write a string there but not that much to create it like you wrote it.

albertz:
Well, I would argue, this should better be fixed on Sisyphus then, as sth like this should be straight-forward. Also, there are many other cases (maybe not here but elsewhere) where it can be more complex, e.g. if you pass a function/lambda or so (e.g. for EvalLayer), etc.

But anyway, ok, it doesn't hurt too much to merge this in.

PR #319, opened 2020-07-06
ExtractAudioFeatures: enable multichannel raw audio

vieting:
Allow using multiple channels for raw audio in `ExtractAudioFeatures` and therefore in `OggZipDataset`. If `num_channels` is set, the data will be of shape `(T, num_channels, num_feature_filters)`, otherwise just `(T, num_feature_filters)` as before.

### DIFF
--- a/GeneratingDataset.py
+++ b/GeneratingDataset.py
@@ -1004,7 +1004,7 @@ class ExtractAudioFeatures:
                norm_mean=None, norm_std_dev=None,
                features="mfcc", feature_options=None, random_permute=None, random_state=None, raw_ogg_opts=None,
                pre_process=None, post_process=None,
-               sample_rate=None,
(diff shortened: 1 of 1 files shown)


GeneratingDataset.py
```diff
@@ -3019,6 +3048,12 @@ def get_total_num_seqs(self):
     """
     return len(self._data)
 
+  def get_data_shape(self, key):
```
albertz, line 3051: Documentation missing.

albertz:
I just merged now. Add the missing doc directly in master, or any other further improvements.

vieting:
I tested it with my new multichannel `OggZipDataset` and with an older setup which uses single channel. But not for all options that are available and only for raw features, not MFCC etc.

PR #325, opened 2020-08-06
BinaryCrossEntropyLoss: added weight for positive labels

patrick-wilken:
I currently need weighting of positive vs. negative labels when doing a binary decision. Probably a nice to have.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -6645,6 +6645,13 @@ class BinaryCrossEntropyLoss(Loss):
   """
   class_name = "bin_ce"
 
+  def __init__(self, pos_weight=None, **kwargs):
+    """
+    :param float|None pos_weight: weight of positive labels, see tf.nn.weighted_cross_entropy_with_logits.
+    """
+    super(BinaryCrossEntropyLoss, self).__init__(**kwargs)
+    self._pos_weight = pos_weight
+
   def _check_init(self):
     assert self.target
     assert self.target.batch_ndim == self.output.batch_ndim, (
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -6645,6 +6645,13 @@ class BinaryCrossEntropyLoss(Loss):
   """
   class_name = "bin_ce"
 
+  def __init__(self, pos_weight=None, **kwargs):
+    """
+    :param float pos_weight: weight of positive labels, see tf.nn.weighted_cross_entropy_with_logits.
```
albertz, line 6650: `float|None` is the type.

returnn/tf/layers/basic.py
```diff
@@ -6655,8 +6662,12 @@ def get_value(self):
     :rtype: tf.Tensor
     """
     with tf.name_scope("loss_bin_ce"):
-      out = tf.nn.sigmoid_cross_entropy_with_logits(
-        logits=self.output_flat, labels=tf.cast(self.target_flat, self.output_flat.dtype))
+      target_flat = tf.cast(self.target_flat, self.output_flat.dtype)
+      if self._pos_weight is None:
+        out = tf.nn.sigmoid_cross_entropy_with_logits(logits=self.output_flat, labels=target_flat)
+      else:
```
albertz, line 6668: Do you need this branch? Wouldn't the other code just always work?

JackTemaki: The implementation is not exactly identical, so I think it is safer to keep it separated.

albertz: Ah sorry, I didn't realized this is a different function.

patrick-wilken:
Fixed the type. And yes, I didn't want to always use the weighted loss because the formulas in the implementations look quite different, maybe there are numerical differences, just to be sure.

PR #329, opened 2020-08-07
StaticDataset: sequence ordering and fixes

patrick-wilken:
`get_data_dtype()` was returning `numpy.dtype` instead of `str` (which caused an assertion error when creating `Data` instances for me).

 `Dataset.is_data_sparse()` does not work in case of dense scalars, i.e. single float values as data (with or without time dimension). `Dataset.num_outputs`, which is of the form `(num-classes, len(shape)`, would be `[1, 1]` or `[1, 0]` in those cases, I guess. But data is assumed to be dense if `len(shape) <= 1`. (Even if we required a feature dimension, in this case of size 1, the case without time dimension would still fail.) I get that this is a best guess in the base class, but not many Datasets override this, e.g. `CachedDataset2.is_data_sparse()` makes the same assumption, so this is a more general problem.
Anyway, I just inferred sparseness from the data type now in StaticDataset.

And I added sequence ordering for StaticDataset. Which might go a bit against the name of it... But as StaticDataset is the easiest way to create a dataset on the fly when using RETURNN as a framework, this feature was necessary for me.

### DIFF
--- a/returnn/datasets/generating.py
+++ b/returnn/datasets/generating.py
@@ -870,10 +870,11 @@ class StaticDataset(GeneratingDataset):
       data=data, target_list=dataset.get_target_list(),
       output_dim=dataset.num_outputs, input_dim=dataset.num_inputs)
 
-  def __init__(self, data, target_list=None, output_dim=None, input_dim=None, **kwargs):
+  def __init__(self, data, target_list=None, output_dim=None, input_dim=None, sort_data_key=None, **kwargs):
     """
     :param list[dict[str,numpy.ndarray]] data: list of seqs, each provide the data for each data-key
     :param int|None input_dim:
+    :param str|None sort_data_key: key of data stream used for sorting by length, defaults to "data" if present
     :param int|dict[str,(int,int)|list[int]] output_dim:
     """
     assert len(data) > 0
@@ -881,6 +882,7 @@ class StaticDataset(GeneratingDataset):
     num_seqs = len(data)
     first_data = data[0]
     self.data_keys = sorted(first_data.keys())
+    self.sort_data_key = sort_data_key or ("data" if "data" in self.data_keys else False) or self.data_keys[0]
     if target_list is not None:
       for key in target_list:
         assert key in self.data_keys
@@ -909,14 +911,34 @@ class StaticDataset(GeneratingDataset):
         assert output_dim[key][0] == first_data_output.shape[-1]
     assert sorted(output_dim.keys()) == self.data_keys, "output_dim does not match the given data"
 
+    self.seq_order = None
+
     super(StaticDataset, self).__init__(input_dim=input_dim, output_dim=output_dim, num_seqs=num_seqs, **kwargs)
+
+  def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
+    """
+    :param int|None epoch:
+    :param list[str]|None seq_list:
+    :param list[int]|None seq_order:
+    :rtype: bool
+    """
+    super(StaticDataset, self).init_seq_order(epoch=epoch, seq_list=None, seq_order=None)
+    if seq_order:
+      self.seq_order = seq_order
+    elif seq_list:
+      self.seq_order = [int(tag[len("seq-"):]) for tag in seq_list]
+    else:
+      self.seq_order = numpy.array(self.get_seq_order_for_epoch(
+        epoch=epoch, num_seqs=len(self.data), get_seq_len=lambda i: len(self.data[i][self.sort_data_key])))
+    self._num_seqs = len(self.seq_order)
+    return True
 
   def generate_seq(self, seq_idx):
     """
     :param int seq_idx:
     :rtype: DatasetSeq
     """
-    data = self.data[seq_idx]
+    data = self.data[self.seq_order[seq_idx]]
     return DatasetSeq(seq_idx=seq_idx, features={key: data[key] for key in self.data_keys})
 
   def get_data_keys(self):
--- a/returnn/datasets/generating.py
+++ b/returnn/datasets/generating.py
@@ -938,6 +938,14 @@ class StaticDataset(GeneratingDataset):
     """
     return str(self.data[0][key].dtype)
 
+  def is_data_sparse(self, key):
+    """
+    :param str key:
+    :rtype: bool
+    """
+    data_type = self.get_data_dtype(key)
+    return data_type.startswith("int")
+
 
 class CopyTaskDataset(GeneratingDataset):
   """
--- a/returnn/datasets/generating.py
+++ b/returnn/datasets/generating.py
@@ -936,7 +936,7 @@ class StaticDataset(GeneratingDataset):
     :param str key:
     :rtype: str
     """
-    return self.data[0][key].dtype
+    return str(self.data[0][key].dtype)
 
 
 class CopyTaskDataset(GeneratingDataset):


returnn/datasets/generating.py
```diff
@@ -938,6 +938,14 @@ def get_data_dtype(self, key):
     """
     return str(self.data[0][key].dtype)
 
+  def is_data_sparse(self, key):
+    """
+    :param str key:
+    :rtype: bool
+    """
+    data_type = self.get_data_dtype(key)
+    return data_type.startswith("int")
```
patrick-wilken, line 969: I guess no need to check for "uint" etc., right?

albertz: Why not?

albertz:
> `Dataset.is_data_sparse()` does not work in case of dense scalars, i.e. single float values as data (with or without time dimension).

Why don't you fix `Dataset.is_data_sparse` then?

> `Dataset.num_outputs`, which is of the form `(num-classes, len(shape)`, would be `[1, 1]` or `[1, 0]` in those cases, I guess. But data is assumed to be dense if `len(shape) <= 1`. (Even if we required a feature dimension, in this case of size 1, the case without time dimension would still fail.) I get that this is a best guess in the base class, but not many Datasets override this, e.g. `CachedDataset2.is_data_sparse()` makes the same assumption, so this is a more general problem.
> Anyway, I just inferred sparseness from the data type now in StaticDataset.

The problem is also basically `num_outputs` itself, and its design. It is too specific to really describe all possible cases. That is why I would anyway try to move away from it more and more. I don't think a lot of code still uses it. Only some old code, or some of the default implementations of the base class would use it as a fallback.

I would prefer a more generic format, like we have with `extern_data`. I sometimes already started using so (e.g. see `LibriSpeechCorpus` with `raw`). But this is not really fully supported yet.

Your "solution" is basically to replace one somewhat incorrect assumption with another somewhat incorrect assumption. So no the base class is inconsistent to `StaticDataset`. I'm not sure if this is a good idea.

Just changing that `Dataset.is_data_sparse` returns always False if it has float/complex dtype, would fix the problem, right? However, that only make sense if `Dataset.get_data_dtype` is also specialized...

> And I added sequence ordering for StaticDataset. Which might go a bit against the name of it... 

`StaticDataset` derives from `GeneratingDataset`. `GeneratingDataset` is intended for datasets which generate data on-the-fly, where it does not make sense to have a predefined order. It also can be of infinite size.

You are now trying to work around this. But I think the actual problem is that this is probably not the right base class. Better directly make it derived from `Dataset` instead.

> But as StaticDataset is the easiest way to create a dataset on the fly when using RETURNN as a framework, this feature was necessary for me.

I don't quite understand. `StaticDataset` should really only be useful for debugging and to write tests. I don't see how this could be useful otherwise. It requires that you load your whole data into memory. This is almost never a good idea. Or even if it works now for you, this solution will not scale. And also, it probably means that you currently have some work-flow like this:

```
data = load_my_data_somehow()
dataset = StaticDataset(data=data, ...)
```

This is very inefficient, suboptimal, and not very flexible.

Instead, a much better way would be to just implement your own dataset. Derive it from `CachedDataset2`. Then mostly you need to specify `_collect_single_seq`. And maybe some other functions if you want more custom stuff.
(If you think this is too complicated, maybe we could create a higher-level base class, based on `CachedDataset2`, or so...)

Now, you can either directly use your dataset. Which is already faster than the workflow you had before.

But also, you could use other existing tools. E.g. like `hdf_dump`, to convert it to HDF.

patrick-wilken:
Yes, the new extern_data format is much nicer :)
I could look into using it everywhere, but that's going to be some work, obviously.

> Why don't you fix Dataset.is_data_sparse then?
Well, using `num_outputs` it's not possible and there is this comment that using `Dataset.get_data_dtype` would lead to a recursion, so I did not do that :D But maybe we can make it work if we check that we are not calling the base class method...

So you want to support using an integer datatype for dense data? What would be a use case for this? But I agree that setting `sparse=True` explicitly somehow would be better than inferring it from the type.

StaticDataset is also used for `Engine.search_single_seq()`, which we use for a server setting, where we definitely don't want to write incoming requests to disk. In general, when you have a trained model and want to load it in an external application and do simple (possibly single) inferences, StaticDataset is currently the simplest way (maybe the only reasonable one?). It's like a slightly more complicated version of `tf.data.Dataset.from_tensor_slices()`. And I think we should offer such a simple interface. But yes, GeneratingDatatset is not the right base class...
You could argue that sequence ordering is not needed in those cases. But for example we could sort the sequences to run into out of memory right at the start like it's done in normal search. And actually I use StaticDataset for training in one scenario now, where it is run from within another tool, data consist of only a couple of scalars and the number of sequences is not too large. I know that it does not scale, but it doesn't have to, so I rather plug in the data directly rather than creating an HDF file.

albertz:
> Yes, the new extern_data format is much nicer :)
> I could look into using it everywhere, but that's going to be some work, obviously.

Yes. Also should be a separate pull request.
We could check where exactly `num_inputs`/`num_outputs` is really used. As I said, I think it's not too often. So maybe not so hard after all.
But the question about this design goes a bit further. `ExternData` and `Data` is currently specific to TF. Although not really too much. When you ignore the placeholder itself, it is generic. Or we could make a much stripped down version of those for the dataset (`GenericExternData`, `GenericData` or so). Not sure...

> > Why don't you fix Dataset.is_data_sparse then?
> 
> Well, using `num_outputs` it's not possible and there is this comment that using `Dataset.get_data_dtype` would lead to a recursion, so I did not do that :D But maybe we can make it work if we check that we are not calling the base class method...

Yes, that's what I meant. Although maybe some other classes overload it and also call `is_data_sparse`. So you would still need to check through all existing implementations. Or simply check for a recursive call.
But that might all be too complicated. I'm not really sure about the best solution.

> So you want to support using an integer datatype for dense data? What would be a use case for this?

There can be many use cases. I often use it to encode sequence length, or also to encode offsets.

> But I agree that setting `sparse=True` explicitly somehow would be better than inferring it from the type.

The default logic in `Data` is somewhat clever about this. I think it actually follows your logic. The default is inferred from the dtype, I think.

> StaticDataset is also used for `Engine.search_single_seq()`, which we use for a server setting, where we definitely don't want to write incoming requests to disk.

That is for a single sequence, where seq ordering anyway is not needed. That's exactly the use case of `StaticDataset`, despite tests.

Also, this function is really a bad API, more for testing as well. In general you would not want to search a single seq. That would be inefficient.

> In general, when you have a trained model and want to load it in an external application and do simple (possibly single) inferences, StaticDataset is currently the simplest way (maybe the only reasonable one?). It's like a slightly more complicated version of `tf.data.Dataset.from_tensor_slices()`. And I think we should offer such a simple interface.

But for these things, there are already the helper functions like `search_single_seq`?
Also, again, you would not want to have any sequence ordering logic for `StaticDataset` in this case.

> You could argue that sequence ordering is not needed in those cases. But for example we could sort the sequences to run into out of memory right at the start like it's done in normal search.

I don't understand. How do you sort a single sequence?

> And actually I use StaticDataset for training in one scenario now, where it is run from within another tool, data consist of only a couple of scalars and the number of sequences is not too large. I know that it does not scale, but it doesn't have to, so I rather plug in the data directly rather than creating an HDF file.

Why HDF file? As I said, you simply would derive from `CachedDataset2`, and directly define it there. That is simpler, more flexible, and more efficient, and scales.

But in any case, if you really want that, then you better should rewrite `StaticDataset` and directly derive from `Dataset`. Everything else would be hacky.

patrick-wilken:
We also process batches with `search_single_seq`/`StaticDataset` in the server, e.g. when translating whole files.

Ok, I see that writing a custom dataset is an option. This is then a similar API as e.g. pytorch datasets. Nevertheless, feeding data into RETURNN via a basic data structure would be very handy, I think. Let me try to implement this more generic StaticDataset. In the meantime I guess we can close this PR.

> But for these things, there are already the helper functions like search_single_seq?

A bit unrelated, but `search_single_seq` is not flexible enough for us anymore (I think it was originally added for our MT server). We now want to have multiple inputs and outputs (translation factors...) and also fetch things like position wise word probabilities. I hacked that in for us, but I was thinking anyway about writing a more generic helper function. E.g. it could take a data list like StaticDataset but would do the vocabulary lookup for all string data like `search_single_string_to_string_seq`. Otherwise we would also have to manually build a dataset outside of RETURNN.
 
By the way, what would be the exact definition of sparse data? :D When it is used to index something, right? So when you say you have sequence length as dense integer then it is used in some float operation and not e.g. for an embedding lookup?

albertz:
> Ok, I see that writing a custom dataset is an option. This is then a similar API as e.g. pytorch datasets. Nevertheless, feeding data into RETURNN via a basic data structure would be very handy, I think.

But isn't a custom dataset exactly that as well, just differently?

> A bit unrelated, but `search_single_seq` is not flexible enough for us anymore (I think it was originally added for our MT server). We now want to have multiple inputs and outputs (translation factors...) and also fetch things like position wise word probabilities. ...

I think no-one uses `search_single_seq` and `search_single_string_to_string_seq`, except you. I don't see any tool, test or demo using this. So feel free to change this as you want. But do it in a clean, general and flexible way, of course.

Actually, if this becomes too complex and also too specific, instead of making `search_single_seq` too complicated, just directly call `search_single`, or even `run_single`, and then you can fetch and feed whatever you want. Or directly call `session.run`.

See the code of `run_single`. It's extremely simple. It sounds a bit like you add some arbitrary complexity around `StaticDataset` to somehow be able to use `search_single_seq`, while it could be much simpler.

> By the way, what would be the exact definition of sparse data? :D When it is used to index something, right? So when you say you have sequence length as dense integer then it is used in some float operation and not e.g. for an embedding lookup?

I did not came up with this terminology and concept. That was Patrick, and comes from the very early time of RETURNN. And then it has just grown historically. So there never really was a formal exact definition.

I would probably define it like you said: It means that it represents a class index. `dim` defines the number of classes. `vocab` more specifically would define the classes (when those are a vocabulary).

It also means there is no dense feature dimension. This is just a matter of interpretation / terminology. You can still have a tensor of shape [B, T, D], but the last axis D here would not be referred to as a feature dimension axis, but just as another spatial axis.

There are cases where the value of `data.sparse` does not really matter at all. I'm also undecided for bool tensors. E.g. the `CompareLayer` can be configured to return the data as sparse or not. Some other layers might expect it in a certain way. The `LengthLayer` also has such a flag. Or `RangeLayer`. Or `RangeInAxisLayer`.

* When you concat the input from multiple layers, this is disallowed when `data.sparse is True`. (Basically, concat would concat in the feature axis, but there is none in this case.)
* `sparse is True` implies that `feature_dim_axis is None` (although `shape` can be anything).
* `sparse is True` also implies that the dtype is int, uint or bool.
* `sparse is True` expects that you explicitly define `dim` (initially that must have been an integer, but at some point I also allowed `None`). You can also define the `vocab` in this case.
* For a `LinearLayer`, if the input is sparse, this will be a lookup. So there is a real behavior difference between sparse and not sparse. This behavior originates from our Theano code. I think this was one of the main original ideas behind this concept / flag. Note that `dim` is still important in any case. The matrix will be of shape (input.dim, output.dim), no matter if input is sparse or not.

patrick-wilken:
Thanks for the clarification!

So just for reference, I have now implemented my use case without the need for the changes of this PR and changes to the Engine helper functions. For that I created a custom dataset like this:

```
class SimpleDataset(CachedDataset2):
    def __init__(self, data_list, output_dim, sort_data_key=None, **kwargs):
        super().__init__(**kwargs)
        self._data_list = data_list
        self.num_outputs = output_dim
        self._data_keys = list(sorted(output_dim.keys()))
        self._seq_order = None
        self._sort_data_key = sort_data_key or ("data" if "data" in self.data_keys else False) or self._data_keys[0]

    def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
        super().init_seq_order(epoch=epoch, seq_list=None, seq_order=None)
        self._seq_order = np.array(
            self.get_seq_order_for_epoch(
                epoch=epoch, num_seqs=len(self._data_list),
                get_seq_len=lambda i: len(self._data_list[i][self._sort_data_key])))
        self._num_seqs = len(self._seq_order)
        return True

    def _collect_single_seq(self, seq_idx):
        data = self._data_list[self._seq_order[seq_idx]]

        return DatasetSeq(seq_idx=seq_idx, features={key: data[key] for key in self._data_keys})

    def is_data_sparse(self, key):
        return False

    def get_data_dim(self, key):
        return self.num_outputs[key][0]

...

dataset = SimpleDataset(data_list=data, output_dim=num_outputs, seq_ordering=seq_ordering, sort_data_key="feature_0")
```

`is_data_sparse` is obviously specific to my data. Before, I only had the last line using StaticDataset. So we could say that this is just what the user needs to implement, or we could put a generalized version of this into the code. I would prefer the latter, because it is a simpler interface and also because the user does not need to use internals like `DatasetSeq`, which we cannot change in the future if we regard them as part of the API.

I also call `self._engine.tf_session.run` directly now like this:
```
output_dict = {"my_layer": self._engine.network.layers["my_layer"].output.get_placeholder_as_time_major(), ...}
feed_dict = self._engine.get_specific_feed_dict(dataset=dataset, seq_idx=-1)
outputs = self._engine.tf_session.run(output_dict, feed_dict=feed_dict)
```

Which is pretty brief, as you predicted, but also not super nice, again because we have to access internals. Also, in case of sparse data with vocabulary lookups this would become more complex.

Before, I had
```
output_layers = ["my_layer", "my_layer_2"]
outputs = self._engine.forward_single(dataset, seq_idx=-1, output_layer_name=output_layers)
```
for which I implemented support for multiple output layers for `Engine.forward_single()`.


Yeah, so should we use something like the Dataset above to replace StaticDataset to have a generic "from python list"-dataset?

albertz:
> class SimpleDataset(CachedDataset2):

Why do you call it `SimpleDataset` now? I thought this is supposed to be a reimplementation of `StaticDataset`? Or is just a demo for your config now?

>     def __init__(self, data_list, output_dim, sort_data_key=None, **kwargs):
>         super().__init__(**kwargs)

Note that this is not Python 2 compatible. I think you need `super(SimpleDataset, self)`.
(Did PyCharm generated this code for you? Normally it should do that correctly, when you instruct it to insert the super call.)

Note that as I said, I advise to not use the dataset like that (i.e. to preload the data into memory). (This is exactly the same problem in your dataset. That is the reason why I also would not recommend `StaticDataset`.)

Rather, you should implement a dataset which covers the data loading on-the-fly (in the `_collect_single_seq` function).


> So we could say that this is just what the user needs to implement, 

I don't understand. The user should definitely not implement it like that. This has the same problem like `StaticDataset`. This doesn't make sense.

> or we could put a generalized version of this into the code.

I don't quite understand. You mean to fix/extend `StaticDataset`? Yes, as I said, you can do that.
But also as I said, this is more for tests, or maybe very simple demos, but not recommended for any serious stuff.

> I would prefer the latter, because it is a simpler interface and also because the user does not need to use internals like `DatasetSeq`, which we cannot change in the future if we regard them as part of the API.

If you say it is too difficult currently for the user to implement an own dataset, this is another topic. This is also what I explained already. We can make a even higher-level abstraction based on `CachedDataset2` or so. See also how this is usually done in e.g. PyTorch or other frameworks.

> I also call `self._engine.tf_session.run` directly now like this:
> 
> ```
> output_dict = {"my_layer": self._engine.network.layers["my_layer"].output.get_placeholder_as_time_major(), ...}
> feed_dict = self._engine.get_specific_feed_dict(dataset=dataset, seq_idx=-1)
> outputs = self._engine.tf_session.run(output_dict, feed_dict=feed_dict)
> ```
> 
> Which is pretty brief, as you predicted, but also not super nice, again because we have to access internals.

What internals?
Also, this is not how I meant it. I meant to generate the `feed_dict` yourself, and not use the RETURNN dataset at all (if you anyway have your own data loading and custom loop). You would not generate any RETURNN dataset for that.

> Also, in case of sparse data with vocabulary lookups this would become more complex.

Why?

> Before, I had
> 
> ```
> output_layers = ["my_layer", "my_layer_2"]
> outputs = self._engine.forward_single(dataset, seq_idx=-1, output_layer_name=output_layers)
> ```

I thought you are speaking about training? You wrote:

> And actually I use StaticDataset for training in one scenario now, ...

I thought this is the use case we discuss here.

For `forward_single` or `search_single_seq`, there was no issue about `StaticDataset`? (And other stuff you are mentioning are unrelated to `StaticDataset`, or not?)

JackTemaki:
I think that Patricks situation is the following (lets keep training and the additional inputs aside):
 - RETURNN is used as framework
 - The data is already in the memory (this is what was not clear I think)
 - The data should be sortable with available functions of RETURNN

I am also using `StaticDataset` for the same scenario, except for the sorting, as this is the most straight-forward solution. I agree with Patrick that we definitely need a dataset which operates on data already in the memory, but provides the following functions:
 - support of all possible data shapes as definable in `extern_data` 
 - support of all implemented sequence ordering methods.

This is how I understood it.

A side question:
> Note that this is not Python 2 compatible

How is this relevant? User code does not need to be Python 2 compatible.

patrick-wilken:
I guess I mixed up training and inference, yes. I need both.

> Why do you call it SimpleDataset now? I thought this is supposed to be a reimplementation of StaticDataset? Or is just a demo for your config now?

In this case I'm using RETURNN as a framework and this is _my_ dataset in _my_ application :D It is the minimal dataset that gets my code running, so that's what I will use. I don't need data loading on the fly (or Python 2 support).
I just wanted to point out that this is how the current interface looks like. I don't think it is nice that everyone has to implement this for the simple use case of just getting data as-is from external python code into RETURNN (also for training!).

> You mean to fix/extend StaticDataset? Yes, as I said, you can do that.
But also as I said, this is more for tests, or maybe very simple demos, but not recommended for any serious stuff.

Why not provide a simpler interface which does not require writing a dataset? What I imagine for this is StaticDataset (derived from CachedDataset2) + sequence ordering + option to load data on the fly, e.g. by passing the data via a generator or callable. This could be used for tests but also for training/inference. Not all "serious stuff" requires huge amounts of data and a super efficient dataset. Using a callable this should even be no issue.

>  We can make a even higher-level abstraction based on CachedDataset2.

What do you mean exactly? What would be different from CachedDataset2 and would it simplify my code above? I'm not talking about the design of `Dataset` in general here, only about what a user needs to do to get the data in. My custom dataset is pretty basic, but I'm saying we could avoid the need to write it for many use cases.


For inference, my only problem was there was no suitable helper function which is as I said somewhat unrelated.

> I meant to generate the feed_dict yourself.

Ah, ok. But also here I only wanted to show what is necessary for me at the moment. With a more general forward single functions this would be cleaner and higher level.

But in the end it's "only" a design decision, my application now runs without changes to RETURNN, so we could leave it like that. I just find it inconvenient how it is at the moment. So does "StaticDataset (derived from CachedDataset2) + sequence ordering + option to load data on the fly" sound good?

albertz:
> I am also using `StaticDataset` for the same scenario, except for the sorting, as this is the most straight-forward solution.

Which scenario exactly? To train? As I explained, this is not a good idea.

If this is currently the most straight-forward solution, this should be changed, rather than pushing even more a bad straight-forward solution. Rather, we should make the good/right solution more straight-forward to use.

> I agree with Patrick that we definitely need a dataset which operates on data already in the memory, but provides the following functions:
> 
> * support of all possible data shapes as definable in `extern_data`
> * support of all implemented sequence ordering methods.

As I said, yes, we can extend `StaticDataset`, there is no real discussion about that at all here.

I think Patricks speaks about multiple situations, not just a single. As I understood it:
* There is one situation where he needs sorting, for training.
* Then there are other situations, where he doesn't need sorting, and also no training, and actually only a single seq in the dataset. This is actually all fine already with the current `StaticDataset`, and I think most of the discussion gets really off-topic here, and is mostly about the interface of `forward_single` and co (not at all about `StaticDataset`).

So, I keep only speaking about the first situation now. Please read my earlier explanations why it is a bad idea to use `StaticDataset` for that, except maybe for small tests.

Anyway, even under these conditions that this is a bad idea, surely we can fix/extend `StaticDataset` to be more general/flexible.

> > Note that this is not Python 2 compatible
> 
> How is this relevant? User code does not need to be Python 2 compatible.

I thought this was the suggestion for the new implementation of `StaticDataset`, to be part of RETURNN, not in the config. (As this is what this PR is about.)

albertz:
> I just wanted to point out that this is how the current interface looks like. I don't think it is nice that everyone has to implement this for the simple use case of just getting data as-is from external python code into RETURNN (also for training!).

As I explained, no-one should do it like that anyway (for training). The data loading should be part of the dataset, not external.

If you say implementing this is too complicated, that is also what I mentioned already. This can be simplified as much as we want. But this gets off-topic here. If you want to discuss how to make that simpler, open a separate issue about that, and let's discuss there. (Let's keep discussion more focused.)

> Why not provide a simpler interface which does not require writing a dataset? What I imagine for this is StaticDataset (derived from CachedDataset2) + sequence ordering + option to load data on the fly, e.g. by passing the data via a generator or callable. This could be used for tests but also for training/inference. Not all "serious stuff" requires huge amounts of data and a super efficient dataset. Using a callable this should even be no issue.

What you say is exactly writing a dataset. So kind of a contradiction to your first sentence?
This is exactly what I already mentioned a couple of times, to simplify it to write a dataset.
But this is not about `StaticDataset`.

> > We can make a even higher-level abstraction based on CachedDataset2.
> 
> What do you mean exactly? What would be different from CachedDataset2 and would it simplify my code above? I'm not talking about the design of `Dataset` in general here, only about what a user needs to do to get the data in. My custom dataset is pretty basic, but I'm saying we could avoid the need to write it for many use cases.

One sentence before, you gave a description how a simple interface could look like.

> > I meant to generate the feed_dict yourself.
> 
> Ah, ok. But also here I only wanted to show what is necessary for me at the moment. With a more general forward single functions this would be cleaner and higher level.

But that is what I explained will not be the case. This will be more complicated, and more indirect. Because you first construct a (dummy) dataset. It will be simpler to skip all that. No matter how simple it is to create the dataset, it will be simpler to not create a dataset at all. You cannot really have `forward_single` generic/general enough for all possible use cases. It will just become more and more complicated.
 
> But in the end it's "only" a design decision, my application now runs without changes to RETURNN, so we could leave it like that. I just find it inconvenient how it is at the moment. So does "StaticDataset (derived from CachedDataset2) + sequence ordering + option to load data on the fly" sound good?

Now you speak again about the training situation, right? But this is not a `StaticDataset`. This is a simple way how the user would implement an own dataset.

JackTemaki:
> This is a simple way how the user would implement an own dataset.

The user should not be required to do that (or at least not for simple applications). The user will always go for the most simple solution from his perspective. With his above example Patrick showed exactly what the simplest solution is to solve his problem (data in memory, sorting required) is right now **without** touching RETURNN code and only writing user code.

You can not answer that the user should not implement it like that. If it solves the users problem, he will do it how it is most convenient with respect to to time and effort. If the user has the option of misusing existing datasets versus writing a lot of code, he will always go for the first option.

I think we should start to discuss about real life situations and how we can improve RETURNN to fit the requirements. This PR was intended to improve of RETURNN w.r.t. to one of these real life situations, but I agree that the solution might be suboptimal.

So we should discuss what are the needs, and what could the optimal solutions be without having to change too much of the RETURNN code.

albertz:
> > This is a simple way how the user would implement an own dataset.
> 
> The user should not be required to do that (or at least not for simple applications). The user will always go for the most simple solution from his perspective.

Maybe you misunderstand what I mean. I say, the simplest solution should be to define/write/implement an own dataset, e.g. in the way Patrick briefly described ("e.g. by passing the data via a generator or callable").

`StaticDataset` is not a good solution. If that is currently the simplest solution, we should try to make other (better) solutions more simple.

I'm very much arguing to make it more simple. I'm just saying, we should not make bad solutions (like `StaticDataset`) more simple, but rather good solutions more simple.

> With his above example Patrick showed exactly what the simplest solution is to solve his problem (data in memory, sorting required) is right now **without** touching RETURNN code and only writing user code. (In contrast, the PR was about the simplest solution **with** touching RETURNN code.)

The PR is about fixing `StaticDataset`. All what we are discussing here is not really about `StaticDataset`. I think we are mixing up things.

> You can not answer that the user should not implement it like that. If it solves the users problem, he will do it how it is most convenient with respect to time and effort. If the user has the option of misusing existing datasets versus writing a lot of code, he will always go for the first option.

Exactly that is my point. It is not good if the most convenient way (using `StaticDataset`) is sub-optimal. We should change that. The most convenient way should be a good solution (like defining an own dataset by a generator or whatever).

> I think we should start to discuss about real life situations and how we can improve RETURNN to fit the requirements. This PR was intended to improve of RETURNN w.r.t. to one of these real life situations, but I agree that the solution might be suboptimal.

As I said multiple times, I have no objections about fixing/extending `StaticDataset` (what this PR is about).

I'm just saying, it is not a good idea to use `StaticDataset` like that, and if this is the simplest solution, we should think about how to make better solutions more simple, such that better solutions become more simple over using `StaticDataset`.

patrick-wilken:
> The PR is about fixing StaticDataset

I agreed earlier that the commits of this PR are not want we want for my use case, so I'm not sure there is anything to fix about StaticDataset, given the use cases you mention for it.

What I would like to have is a Dataset class defined _inside RETURNN_ which simply takes a list/generator/callable as data, without any major restriction of use case. I don't mind whether we call this a new good solution or an extension of StaticDataset :D

> The most convenient way should be a good solution (like defining an own dataset by a generator or whatever).

So what you still suggest is that the user should implement this kind of dataset class outside of RETURNN him/herself? Possible, but how could this ever become simpler than the one-liner defining a StaticDataset (instance)?

albertz:
> > The PR is about fixing StaticDataset
> 
> I agreed earlier that the commits of this PR are not want we want for my use case, so I'm not sure there is anything to fix about StaticDataset, given the use cases you mention for it.

I thought that you still wanted to generalize/extend `StaticDataset`. Or not? In any case, this is what this PR is about: Extending/generalizing `StaticDataset`.

If you don't plan to do that anymore (which I'm totally fine with, as I don't think there is really a need for it, and as I explained, it would be a bad idea to use it like you intend to do), then we can close this PR.

Anything else, like a discussion about simplifying how to get some data source into RETURNN (in whatever way) should be a separate issue, esp if this is not about `StaticDataset` (as it should not be about `StaticDataset`, as a I explained).
 
> What I would like to have is a Dataset class defined _inside RETURNN_ which simply takes a list/generator/callable as data, without any major restriction of use case.

Yes, this is what I already said in my very first comment here.

> I don't mind whether we call this a new good solution or an extension of StaticDataset :D

We should not invent a completely new way, which would be very different from `StaticDataset`, and then implement that and call it `StaticDataset`, when it is something completely different. `StaticDataset` is pretty clear in what it is. It should not be changed to become sth else.

> > The most convenient way should be a good solution (like defining an own dataset by a generator or whatever).
> 
> So what you still suggest is that the user should implement this kind of dataset class outside of RETURNN him/herself? Possible, but how could this ever become simpler than the one-liner defining a StaticDataset (instance)?

Another one-liner? You suggested just the same? You said:

> [user implements dataset] which simply takes a list/generator/callable ...

Sth like `GeneratorDataset(generator)` or whatever.
This is also what I briefly explained in my very first comment here:

> ... specify `_collect_single_seq`. (If you think this is too complicated, maybe we could create a higher-level base class, based on `CachedDataset2`, or so [like `GeneratorDataset`] ...)

Anyway, let's discuss that somewhere else. We are mixing up stuff here.

patrick-wilken:
> [user implements dataset] which simply takes a list/generator/callable ...

How did you translate "Dataset class defined _inside RETURNN_" into "[user implements dataset]"? :D No, that's exactly the point, avoiding that each user has to implement something that could be done in a general way.

Let's continue here: #353

PR #333, opened 2020-08-11
Added details to the doc of CompareLayer

mattiadg:
I found the docstring for CompareLayer not very clear, so I added some details.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4555,7 +4555,11 @@ class RemoveLayer(LayerBase):
 
 class CombineLayer(LayerBase):
   """
-  Applies some binary operation on all sources, such as addition.
+  Applies a binary operation, such as addition, to all sources while accumulating the partial results.
+  In the first step, the binary operation is performed on the first two sources.
+  After the first step, the previous results is always the left-hand operator.
+
+  Its basic working is similar to the `reduce` function used in functional programming.
   Also see :class:`ActivationLayer`, or :class:`CompareLayer`.
   """
   layer_class = "combine"
@@ -4565,10 +4569,10 @@ class CombineLayer(LayerBase):
                eval=None, eval_locals=None, eval_for_output_loss=False,
                **kwargs):
     """
-    :param str kind: e.g. "average" or "add", or "eval"
+    :param str kind: currently accepted values are `average`, `add`, `sub`, `mul`, or `eval`
     :param list[LayerBase] sources:
     :param str|None activation: if provided, activation function to apply, e.g. "tanh" or "relu"
-    :param bool with_bias: if given, will add a bias
+    :param bool with_bias: if given, will add a trainable bias tensor
     :param str|callable eval: for kind="eval", will eval this string. or function. see :func:`_op_kind_eval`
     :param dict[str]|None eval_locals: locals for eval
     :param bool eval_for_output_loss: will do the same eval on layer.output_loss
@@ -4769,39 +4773,53 @@ class EvalLayer(CombineLayer):
 
 class CompareLayer(LayerBase):
   """
-  This layer performs an elementwise comparison of all provided sources, and returnns a boolean vector
-  of the same shape.
-  If the `value` parameter is provided, all elements in the provided sources are also compared to this value.
+  Compares element-wise the tokens of all input sequences among themselves and/or with a specified given value.
+  The comparisons are performed in a chain according to the order in which they are listed.
 
-  If more than two sources are provided, each additional source will be compared against the first source,
-  and the results will be combined by a logical conjunction.
+  Example::
+
+      {"class": "compare", "from": ["i1", "i2"], "value": val, "kind": "less"}
+
+  computes i1 < i2 < val and it is true only if the whole chain of operations is true.
+  The final result is the logical "and" of all comparisons. Note that `value` is the last element to be compared to.
+
+  A common example usage is the `end` layer in a rec subnetwork to specify the stopping criterion,
+  e.g. the last generated token is equal to the end-of-sentence token::
+
+      "output": {"class": "rec", "from": [], "unit": {
+          .
+          .
+          .
+          "end": {"class": "compare", "from": "output", "value": end_of_sentence_id}
+      }, "target": "classes0"}
+
   """
   layer_class = "compare"
 
   def __init__(self, kind="equal", value=None, **kwargs):
     """
-    :param str kind: Which comparison operation to use, e.g. "equal" or other supported tf comparison ops
+    :param str kind: which comparison operation to use, e.g. "equal", "greater", "less"
+      or other supported TF comparison ops
     :param float|int|None value: if specified, will also compare to this
     """
     super(CompareLayer, self).__init__(**kwargs)
     assert len(self.sources) >= 1
     if value is None:
-      assert len(self.sources) >= 2
+      assert len(self.sources) >= 2, "{} requires at least two elements to compare".format(self)
     op = getattr(tf, kind)  # e.g. tf.equal
-    from returnn.tf.util.basic import swapaxes
-    x = self.sources[0].output.placeholder
+    from returnn.tf.util.basic import swapaxes, opt_logical_and
+    common_data = Data.get_common_data([s.output for s in self.sources])
+    x = self.sources[0].output.copy_compatible_to(common_data).placeholder
     batch_axis = self.sources[0].output.batch_dim_axis
-    r_last = None
-    if value is not None:
-      r_last = op(x, value)
+    r_last = True
     for source in self.sources[1:]:
-      x2 = source.output.placeholder
+      x2 = source.output.copy_compatible_to(common_data).placeholder
       if source.output.batch_dim_axis != batch_axis:
         x2 = swapaxes(x2, batch_axis, source.output.batch_dim_axis)
-      r = op(x, x2)
-      if r_last is not None:
-        r = tf.logical_and(r_last, r)
-      r_last = r
+      r_last = opt_logical_and(r_last, op(x, x2))
+      x = x2
+    if value is not None:
+      r_last = opt_logical_and(r_last, op(x, value))
     self.output.placeholder = r_last
 
   @classmethod
@@ -4812,16 +4830,31 @@ class CompareLayer(LayerBase):
     :param list[LayerBase] sources:
     :rtype: Data
     """
-    if n_out is NotSpecified and not out_type:
-      if not sources[0] or sources[0].output.undefined:
-        return Data.create_undefined(name="%s_output" % kwargs["name"])
-      out_type = sources[0].output.get_kwargs()
-      out_type["name"] = "%s_output" % kwargs["name"]
-      if out_type.get("sparse", False):
-        out_type["dim"] = 2  # True or False
-      out_type["dtype"] = "bool"
-      out_type["vocab"] = None
-    return super(CompareLayer, cls).get_out_data_from_opts(n_out=n_out, out_type=out_type, sources=sources, **kwargs)
+    out_type_ = {}
+    sources_ = [s for s in sources if s and not s.output.undefined]
+    if sources and not sources_:
+      if n_out is NotSpecified and not out_type:
+        from returnn.tf.network import CannotHandleUndefinedSourcesException
+        raise CannotHandleUndefinedSourcesException(
+          layer_name=kwargs["name"], layer_desc=dict(
+            n_out=n_out, out_type=out_type, sources=sources, **kwargs))
+    sources = sources_
+    if sources:
+      out_type_.update(Data.get_common_data([s.output for s in sources], warnings_out=log.v4).get_kwargs())
+    if n_out is not NotSpecified:
+      out_type_["dim"] = n_out
+    elif out_type_.get("sparse", False):
+      out_type_["dim"] = 2
+    out_type_["dtype"] = "bool"
+    out_type_["name"] = "%s_output" % kwargs["name"]
+    if out_type:
+      if isinstance(out_type, dict):
+        out_type_.update(out_type)
+      elif callable(out_type):
+        out_type_ = out_type  # just overwrite
+      else:
+        raise TypeError("unexpected type of out_type %r" % (out_type,))
+    return super(CompareLayer, cls).get_out_data_from_opts(n_out=n_out, out_type=out_type_, sources=sources, **kwargs)
 
 
 class SwitchLayer(LayerBase):


albertz:
Quite a coincidence that just a few minutes ago (in a735faddb3af8d6a0ad1ac4f79ef0), @JackTemaki also improved exactly this documentation. Although I think it is still not optimal and somewhat misleading now.

That's also why your commit conflicts with the master now.

Maybe you can update the PR, by basing it on current master, and also take my comment (in a735faddb3af8d6a0ad1ac4f79ef0, [here](https://github.com/rwth-i6/returnn/commit/a735faddb3af8d6a0ad1ac4f79ef0261143aa51a#r41361887)) into account.

returnn/tf/layers/basic.py
```diff
@@ -4792,6 +4806,12 @@ def __init__(self, kind="equal", value=None, **kwargs):
       if r_last is not None:
         r = tf.logical_and(r_last, r)
       r_last = r
+    if value is not None:
+      r = op(x, value)
+      if r_last is not None:
+        r = tf.logical_and(r_last, r)
+      else:
+        r_last = r
```
albertz, line 4814: You want to change the behavior that `value` is checked last?
But does this make sense? Is this how you would expect it? (@JackTemaki What do you think?)
Do you actually have multiple sources?
I am thinking whether we should maybe only allow multiple sources + value if `kind=="equal"`, or otherwise only a single source if value is given as well.

returnn/tf/layers/basic.py
```diff
@@ -4764,13 +4764,29 @@ def __init__(self, eval, **kwargs):
...
-  Compares (e.g. equality check) all the sources element-wise.
+  Compares element-wise the tokens of all input sequences among themselves and/or with a specified given value.
+  The comparisons are performed in a chain according to the order in which they are listed.
+  The final result is the logical "and" of all comparisons.
+
+  Example:
+    {"class": "compare", "from": ["i1", "i2"], "value": val, "kind": "less"}
+    computes i1 < i2 < val and it is true only if the whole chain of operations is true.
+
+  This layer is used in particular as the `end` node in an `output` subnetwork to specify the stopping criterion for generation,
```
albertz, line 4775: Note: This is only a common example where it is used. But:
* The `end` layer doesn't need to be a `CompareLayer`. It could be anything else which determines the end.
* This is not the only use of the `CompareLayer`. It is always useful when you want to compare things.

returnn/tf/layers/basic.py
mattiadg: I replied in the other thread, this is just to propose a unified behavior, if we can find it. 
For sure it is not finished as we need to assign x2 to x after each comparison to achieve that behavior. So far it is still buggy. But we need to agree on the behavior first.

returnn/tf/layers/basic.py
albertz: This comment is mostly about the formulation. I would maybe just reformulate it, like:
```suggestion
  A common example usage is the `end` layer in a rec subnetwork to specify the stopping criterion,
```

returnn/tf/layers/basic.py
JackTemaki: > I am thinking whether we should maybe only allow multiple sources + value if `kind=="equal"`, or otherwise only a single source if value is given as well.

This should be good.

JackTemaki:
Please check that you do not have merge commits in your pull request, so always update with `git pull --rebase`.

returnn/tf/layers/basic.py
albertz: I realized now that even for a single source, this is kind of arbitrary. See the discussion on the commit comment.
Esp consider this:
`{"class": "compare", "kind": "greater", "from": "x", "value": 0}`
vs:
`{"class": "compare", "kind": "greater", "value": 0, "from": "x"}`

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4769,53 @@ def __init__(self, eval, **kwargs):
...
+      x2 = source.output.copy_compatible_to(common_data).placeholder
       if source.output.batch_dim_axis != batch_axis:
         x2 = swapaxes(x2, batch_axis, source.output.batch_dim_axis)
       r = op(x, x2)
       if r_last is not None:
         r = tf.logical_and(r_last, r)
       r_last = r
+      x = x2
+    if value is not None:
+      r_last = tf.logical_and(r_last, op(x, value))
```
JackTemaki, line 4818: `r_last` can still be `None`. Is `tf.logical_and` capable of handling that?

mattiadg: Fixed this with my last commit

mattiadg: > 
> 
> The `out_type` variable is still based on the first input:
> 
> `out_type = sources[0].output.get_kwargs()`
> 
> I think this needs to be changed as well, you can have a look at the code in `CombineLayer`.

By looking at the code for `get_out_data_from_opts` from `CombineLayer` I noticed some differences that I don't know how to solve.

1. In case of error, `CombineLayer` raises a `CannotHandleUndefinedSourcesException` while `CompareLayer` returns a `Data.create_undefined`. 
2. `CombineLayer` checks if `out_type` is a dict while `CompareLayer` assumes that it is.
3. actully the point discussed, one uses the kwargs from all sources while the other only from one element. Is it related to the fact that the output is a boolean or we need to fix this?

@albertz please, give us your opinion

returnn/tf/layers/basic.py
```diff
@@ -4565,10 +4569,10 @@ def __init__(self, kind, sources, activation=None, with_bias=False,
                eval=None, eval_locals=None, eval_for_output_loss=False,
                **kwargs):
     """
-    :param str kind: e.g. "average" or "add", or "eval"
+    :param str kind: accepted values are `average`, `add`, `sub`, `mul`, or `eval`
```
albertz, line 4572: There are much more accepted values. That's why I wrote "e.g.".

albertz: Ah, sorry, not in current implementation, I think. But this sounds again like it would be restricted to that. But sth like `kind="div"` should also be possible, it's just not implemented yet.

mattiadg: Actually, you set an assert for those values `assert kind in ["average", "add", "sub", "mul", "eval"]`

returnn/tf/layers/basic.py
```diff
@@ -4565,10 +4569,10 @@ def __init__(self, kind, sources, activation=None, with_bias=False,
...
     :param list[LayerBase] sources:
     :param str|None activation: if provided, activation function to apply, e.g. "tanh" or "relu"
-    :param bool with_bias: if given, will add a bias
+    :param bool with_bias: if given, will add a bias tensor
```
albertz, line 4575: I think crucial is also that this is a trainable parameter.

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,57 @@ def __init__(self, eval, **kwargs):
...
   """
-  This layer performs an elementwise comparison of all provided sources, and returnns a boolean vector
-  of the same shape.
-  If the `value` parameter is provided, all elements in the provided sources are also compared to this value.
+  Compares element-wise the tokens of all input sequences among themselves and/or with a specified given value.
+  The comparisons are performed in a chain according to the order in which they are listed. 
 
-  If more than two sources are provided, each additional source will be compared against the first source,
-  and the results will be combined by a logical conjunction.
+  Example:
```
albertz, line 4779: Note: You should follow [rst syntax](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) here, because we use this for the documentation.

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,57 @@ def __init__(self, eval, **kwargs):
...
+        "end": {"class": "compare", "from": "output", "value": end_of_sentence_id}
+    }, "target": "classes0"},
   """
   layer_class = "compare"
 
   def __init__(self, kind="equal", value=None, **kwargs):
     """
-    :param str kind: Which comparison operation to use, e.g. "equal" or other supported tf comparison ops
+    :param str kind: Which comparison operation to use, e.g. "equal", "greater", "less"
+    or other supported tf comparison ops
```
albertz, line 4799: Make an indent for this line (by 2 spaces).
Also, if you make this a proper sentence (i.e. you start with upper case "Which"), then I would also add a "." at the end.

albertz: Also, use upper case "TF".

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,57 @@ def __init__(self, eval, **kwargs):
...
     :param float|int|None value: if specified, will also compare to this
     """
     super(CompareLayer, self).__init__(**kwargs)
     assert len(self.sources) >= 1
     if value is None:
-      assert len(self.sources) >= 2
+      assert len(self.sources) >= 2, "{} requires at least two elements to compare.".format(self.__name__)
```
albertz, line 4805: Use just `self`, which adds more/nicer/further debug information.

albertz: Also, no "." here.

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,57 @@ def __init__(self, eval, **kwargs):
...
     op = getattr(tf, kind)  # e.g. tf.equal
     from returnn.tf.util.basic import swapaxes
-    x = self.sources[0].output.placeholder
+    common_data = Data.get_common_data([s.output for s in self.sources])
```
albertz, line 4808: Note that when you do that here, you should also fix `get_out_data_from_opts`. See `CombineLayer.get_out_data_from_opts`, which you can basically copy here.

returnn/tf/layers/basic.py
albertz: Yea because it's not implemented otherwise. Maybe I should have raised `NotImplementedError` instead. I was just too lazy to type that.

returnn/tf/layers/basic.py
albertz: `CannotHandleUndefinedSourcesException` vs `Data.create_undefined`, doesn't really matter.
`CompareLayer` does not assume that `out_type` is a dict (documentation of the type is slightly wrong for `out_type`).
The code of `CompareLayer` (including `get_out_data_from_opts`) probably is outdated, and incomplete, as you see from the whole discussion here. It never was used for these more complex cases so far, so it didn't matter.

albertz:
I think your commits are messed up. Rebase on current master. And then also cleanup. (Use PyCharm interactive rebase, or `git rebase -i`.)

albertz:
Btw, also check the failing tests, esp the code style check (e.g. [here](https://travis-ci.org/github/rwth-i6/returnn/jobs/716954114)). Specifically:
```
returnn/tf/layers/basic.py:4777: WEAK WARNING Pep8CodeStyle: W291: trailing whitespace
returnn/tf/layers/basic.py:4783: WEAK WARNING Pep8CodeStyle: W293: blank line contains whitespace
```

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,56 @@ def __init__(self, eval, **kwargs):
...
-  This layer performs an elementwise comparison of all provided sources, and returnns a boolean vector
-  of the same shape.
-  If the `value` parameter is provided, all elements in the provided sources are also compared to this value.
-
-  If more than two sources are provided, each additional source will be compared against the first source,
-  and the results will be combined by a logical conjunction.
+  Compares element-wise the tokens of all input sequences among themselves and/or with a specified given value.
+  The comparisons are performed in a chain according to the order in which they are listed. 
+
+  Example::
```
albertz, line 4779: I think there must be an empty line after `::` (right @JackTemaki ?).

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,56 @@ def __init__(self, eval, **kwargs):
...
+    {"class": "compare", "from": ["i1", "i2"], "value": val, "kind": "less"}
+  computes i1 < i2 < val and it is true only if the whole chain of operations is true.
+  The final result is the logical "and" of all comparisons. Note that `value` is the last element to be compared to.
+  
+  A common example usage is the `end` layer in a rec subnetwork to specify the stopping criterion,
+  e.g. the last generated token is equal to the end-of-sentence token::
```
albertz, line 4785: Here as well, one empty line after `::`.

returnn/tf/layers/basic.py
```diff
@@ -4812,16 +4833,30 @@ def get_out_data_from_opts(cls, n_out=NotSpecified, out_type=None, sources=(), *
...
+        raise CannotHandleUndefinedSourcesException(
+          layer_name=kwargs["name"], layer_desc=dict(
+            n_out=n_out, out_type=out_type, sources=sources, **kwargs))
+    sources = sources_
+    if sources:
+      out_type_.update(Data.get_common_data([s.output for s in sources], warnings_out=log.v4).get_kwargs())
+    if n_out is not NotSpecified:
+      out_type_["dim"] = n_out
+    out_type_["name"] = "%s_output" % kwargs["name"]
+    out_type_["dtype"] = bool
```
albertz, line 4851: This is wrong. Be careful. This should be a string, i.e. `"bool"`.

returnn/tf/layers/basic.py
```diff
@@ -4812,16 +4833,30 @@ def get_out_data_from_opts(cls, n_out=NotSpecified, out_type=None, sources=(), *
...
+      if n_out is NotSpecified and not out_type:
+        from returnn.tf.network import CannotHandleUndefinedSourcesException
+        raise CannotHandleUndefinedSourcesException(
+          layer_name=kwargs["name"], layer_desc=dict(
+            n_out=n_out, out_type=out_type, sources=sources, **kwargs))
+    sources = sources_
+    if sources:
+      out_type_.update(Data.get_common_data([s.output for s in sources], warnings_out=log.v4).get_kwargs())
+    if n_out is not NotSpecified:
+      out_type_["dim"] = n_out
```
albertz, line 4849: In addition, I think you should copy the old behavior (`elif out_type.get("sparse", False): ...`).

returnn/tf/layers/basic.py
albertz:
```suggestion
    out_type_["dtype"] = "bool"
```

returnn/tf/layers/basic.py
albertz, line 4849:
```suggestion
    if n_out is not NotSpecified:
      out_type_["dim"] = n_out
    elif out_type.get("sparse", False):
      out_type_["dim"] = 2    
```

returnn/tf/layers/basic.py
JackTemaki: Yes, always use an empty line after `::` and also after the code example.

albertz:
Why do you have the "verbose docstring for CompareLayer" (by @JackTemaki ) commit in this PR? But actually it looks like this commit contains none of the changes from Nick, but only your changes? You should clean up the commits.

mattiadg:
Is it better now?

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4769,55 @@ def __init__(self, eval, **kwargs):
...
+      or other supported TF comparison ops
     :param float|int|None value: if specified, will also compare to this
     """
     super(CompareLayer, self).__init__(**kwargs)
     assert len(self.sources) >= 1
     if value is None:
-      assert len(self.sources) >= 2
+      assert len(self.sources) >= 2, "{} requires at least two elements to compare".format(self)
     op = getattr(tf, kind)  # e.g. tf.equal
     from returnn.tf.util.basic import swapaxes
```
albertz, line 4809:
```suggestion
    from returnn.tf.util.basic import swapaxes, opt_logical_and
```

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4769,55 @@ def __init__(self, eval, **kwargs):
...
-    x = self.sources[0].output.placeholder
+    common_data = Data.get_common_data([s.output for s in self.sources])
+    x = self.sources[0].output.copy_compatible_to(common_data).placeholder
     batch_axis = self.sources[0].output.batch_dim_axis
     r_last = None
```
albertz, line 4813:
```suggestion
    r_last = True
```

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4769,55 @@ def __init__(self, eval, **kwargs):
...
-      r_last = op(x, value)
     for source in self.sources[1:]:
-      x2 = source.output.placeholder
+      x2 = source.output.copy_compatible_to(common_data).placeholder
       if source.output.batch_dim_axis != batch_axis:
         x2 = swapaxes(x2, batch_axis, source.output.batch_dim_axis)
       r = op(x, x2)
       if r_last is not None:
         r = tf.logical_and(r_last, r)
       r_last = r
```
albertz, line 4821:
```suggestion
      r_last = opt_logical_and(r_last, op(x, x2))
```

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4769,55 @@ def __init__(self, eval, **kwargs):
...
+      x = x2
+    if value is not None:
+      r_last = tf.logical_and(r_last, op(x, value))
```
albertz:
```suggestion
      r_last = opt_logical_and(r_last, op(x, value))
```

albertz: There was still the problem that `r_last` could have been `None` here.
But I added a few suggestions to use `opt_logical_and`, which simplifies the code anyway.

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,52 @@ def __init__(self, eval, **kwargs):
...
   """
-  This layer performs an elementwise comparison of all provided sources, and returnns a boolean vector
-  of the same shape.
-  If the `value` parameter is provided, all elements in the provided sources are also compared to this value.
+  Compares element-wise the tokens of all input sequences among themselves and/or with a specified given value.
+  The comparisons are performed in a chain according to the order in which they are listed.
+
+  Example::
+
+    {"class": "compare", "from": ["i1", "i2"], "value": val, "kind": "less"}
```
albertz, line 4781: Also newline after the code example.

albertz: I think the code example needs to be indented by 4 spaces (after the `::`) (right @JackTemaki ?).

returnn/tf/layers/basic.py
```diff
@@ -4769,39 +4773,52 @@ def __init__(self, eval, **kwargs):
...
+  computes i1 < i2 < val and it is true only if the whole chain of operations is true.
+  The final result is the logical "and" of all comparisons. Note that `value` is the last element to be compared to.
+
+  A common example usage is the `end` layer in a rec subnetwork to specify the stopping criterion,
+  e.g. the last generated token is equal to the end-of-sentence token::
+
+    "output": {"class": "rec", "from": [], "unit": {
```
albertz, line 4788: This needs to be indented (I think by 4 spaces).

returnn/tf/layers/basic.py
mattiadg: Should I use tab or spaces for indenting?

albertz: Always spaces.

albertz: Please follow the PyCharm settings as described in the wiki.

returnn/tf/layers/basic.py
```diff
@@ -4812,16 +4830,31 @@ def get_out_data_from_opts(cls, n_out=NotSpecified, out_type=None, sources=(), *
...
+        from returnn.tf.network import CannotHandleUndefinedSourcesException
+        raise CannotHandleUndefinedSourcesException(
+          layer_name=kwargs["name"], layer_desc=dict(
+            n_out=n_out, out_type=out_type, sources=sources, **kwargs))
+    sources = sources_
+    if sources:
+      out_type_.update(Data.get_common_data([s.output for s in sources], warnings_out=log.v4).get_kwargs())
+    if n_out is not NotSpecified:
+      out_type_["dim"] = n_out
+    elif out_type.get("sparse", False):
```
albertz, line 4846: Sorry, this is wrong. It should be:
```suggestion
    elif out_type_.get("sparse", False):
```

albertz: Please run some of the tests yourself. You would have seen that directly.

returnn/tf/layers/basic.py
JackTemaki: > I think the code example needs to be indented by 4 spaces (after the `::`) (right @JackTemaki ?).

I think the depth of the indentation is not relevant, I checked the build and it compiles as it is now.

albertz: Maybe because it is already 4 spaces from the very left? Maybe it was also for PyCharm (which also parses the docstring and formats it). I don't remember anymore. Or maybe I just remember wrong. But shouldn't really matter, as long as it is indented.

mattiadg:
Anything more is needed?

JackTemaki:
`from returnn.tf.network import CannotHandleUndefinedSourcesException` might be changed to a global import, because it is used multiple times now, but the rest seems fine.

albertz:
> `from returnn.tf.network import CannotHandleUndefinedSourcesException` might be changed to a global import, because it is used multiple times now, but the rest seems fine.

I actually prefer to not clutter the global imports too much with somewhat specific stuff. I like to only have stuff there which is broadly used by most of the code, and central for the functionality. Other stuff, used only once, twice, or so, should be local.
But this is maybe personal preference. Also I'm not really sure where to draw the line.

albertz:
@mattiadg Note that the commit message still is not good. The first line is supposed to be a summary. Then there should be an empty line. Then there can be additional details. (But merged now, so remember that for next time.)

Also note that the mail you use for the commit (Apptek) is not registered in your GitHub account. You might want to do that if you want that this commit can be associated with your account.

mattiadg:
Sorry, I'll do it better by next time

PR #334, opened 2020-08-11
Added no-conversion option in bliss-to-ogg-zip.py

JackTemaki:
For some reason the FFMPEG calls had very faulty results (mostly missing 3/4 of the audio content) when the input was already in .ogg format, and the file duration was already correct. I was not able to find the source of the error, so for me the safest solution was to add the same option (not doing any conversion or as I already had in my old custom Job for creating ogg-zip-files.

An alternative would be to create a separate job, that would just pass through the audio and also support different formats than .ogg (because the OggZipDataset supports basically all formats of the soundfile/libsndfile library).

This is open for discussion.

### DIFF
--- a/tools/bliss-to-ogg-zip.py
+++ b/tools/bliss-to-ogg-zip.py
@@ -255,6 +255,8 @@ def main():
   arg_parser.add_argument("bliss_filename")
   arg_parser.add_argument("--subset_segment_file")
   arg_parser.add_argument("--no_ogg", help="skip generating ogg files", action="store_true")
+  arg_parser.add_argument(
+    "--no_conversion", help="skip ffmpeg call, assume audio is correct already", action="store_true")
   arg_parser.add_argument("--no_cleanup", help="don't delete our temp files", action="store_true")
   arg_parser.add_argument("--sprint_cache", help="filename of feature cache for synchronization")
   arg_parser.add_argument("--raw_sample_rate", help="sample rate of audio input", type=int, default=8000)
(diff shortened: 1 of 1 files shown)


tools/bliss-to-ogg-zip.py
```diff
@@ -366,19 +368,25 @@ def main():
...
-      if limit_duration:
-        cmd += ["-t", str(duration)]
-      cmd += [dest_filename, "-loglevel", args.ffmpeg_loglevel]
-      print("$ %s" % " ".join(cmd))
-      check_call(cmd)
+      if args.no_conversion:
+        print("skip ffmpeg, copy instead (%s -> %s)" % (
+        os.path.basename(rec_filename), dest_filename[len(dest_dirname) + 1:]))
+        shutil.copy(src=source_filename,
+                    dst=dest_filename)
```
albertz, line 375: Note: It is a bad idea to format code like this.
Better (if it fits in one line, which is the case here I think):
```suggestion
        shutil.copy(src=source_filename, dst=dest_filename)
```
Or:
```suggestion
        shutil.copy(
          src=source_filename, dst=dest_filename)
```
Or:
```suggestion
        shutil.copy(
          src=source_filename,
          dst=dest_filename)
```

tools/bliss-to-ogg-zip.py
```diff
@@ -255,6 +255,8 @@ def main():
   arg_parser.add_argument("bliss_filename")
   arg_parser.add_argument("--subset_segment_file")
   arg_parser.add_argument("--no_ogg", help="skip generating ogg files", action="store_true")
+  arg_parser.add_argument("--no_conversion", help="skip ffmpeg call, assume audio is correct already",
+                          action="store_true")
```
albertz, line 259: Same formatting comment as below.

tools/bliss-to-ogg-zip.py
```diff
@@ -366,19 +368,25 @@ def main():
...
-      if start_time:
-        cmd += ["-ss", str(start_time)]
-      if limit_duration:
-        cmd += ["-t", str(duration)]
-      cmd += [dest_filename, "-loglevel", args.ffmpeg_loglevel]
-      print("$ %s" % " ".join(cmd))
-      check_call(cmd)
+      if args.no_conversion:
+        print("skip ffmpeg, copy instead (%s -> %s)" % (
+        os.path.basename(rec_filename), dest_filename[len(dest_dirname) + 1:]))
```
albertz, line 373: This line needs an indent.

tools/bliss-to-ogg-zip.py
```diff
@@ -366,19 +368,25 @@ def main():
...
-      if args.number_of_channels > 0:
-        cmd += ["-ac", str(args.number_of_channels)]
-      if start_time:
-        cmd += ["-ss", str(start_time)]
-      if limit_duration:
-        cmd += ["-t", str(duration)]
-      cmd += [dest_filename, "-loglevel", args.ffmpeg_loglevel]
-      print("$ %s" % " ".join(cmd))
-      check_call(cmd)
+      if args.no_conversion:
```
albertz, line 371: Shouldn't there be a check like
`assert rec_filename.endswith(".ogg")`?

tools/bliss-to-ogg-zip.py
albertz: Note, for the reasoning of this: Now when you change the name of `arg_parser`, this will mess up the formatting. And if you fix the formatting afterwards, the commit diff becomes more convoluted.
Also, this style can lead to cases where stuff is extremely right indented, which is bad.
I think you can find some articles which go into more detail. Maybe it's also a matter of taste in the end, but we should stay consistent in our code style.

albertz:
In general, I don't have objections against this option. (See minor comments.)

However, I think it would be good to better understand what goes wrong here. Maybe this also happens in other cases when ffmpeg is used. This would be very bad. But the question is, what exactly goes wrong there with ffmpeg. Maybe this is also just a hickup with NFS and not related to ffmpeg? Maybe this is some specific ffmpeg version? If we better understand the real problem, we could maybe better catch and solve that.

tools/bliss-to-ogg-zip.py
JackTemaki: I found the option in PyCharm which enables the correct automatic line breaking when exceeding 120 characters, I will add this to the Wiki page.

tools/bliss-to-ogg-zip.py
JackTemaki: We could also support any file ending, but I will go with the assert for now.

albertz:
Ah nice, the code style checker works in Travis. :)
See [here](https://travis-ci.org/github/rwth-i6/returnn/jobs/716866435):
```
tools/bliss-to-ogg-zip.py:373: WEAK WARNING Pep8CodeStyle: E122: continuation line missing indentation or outdented
```
Didn't you see that directly in PyCharm?

JackTemaki:
> Didn't you see that directly in PyCharm?

To be honest, I did this change directly in vim for a quick fix of my setup and forgot to check it in PyCharm again for the PR, so no.

PR #335, opened 2020-08-11
signal_processing.py: fix code style issues

vieting:
Fix code style issues in `signal_processing.py`. Mainly long lines, variables with uppercase etc. Also usages of `tf.contrib` are changed for TF 2 compatibility.

### DIFF
--- a/returnn/tf/layers/signal_processing.py
+++ b/returnn/tf/layers/signal_processing.py
@@ -1,5 +1,9 @@
+"""
+Defines multiple signal processing-related layers.
+"""
 
 from __future__ import print_function
+from __future__ import division
 
 import tensorflow as tf
 import returnn.tf.compat as tf_compat
@@ -9,17 +13,13 @@ from returnn.tf.util.basic import Data
 
 class AlternatingRealToComplexLayer(_ConcatInputLayer):
   """
-  This layer converts a real valued input tensor into a complex valued output
-  tensor.
-  For this even and odd features are considered the real and imaginary part of
-  one complex number, respectively
+  This layer converts a real valued input tensor into a complex valued output tensor.
+  For this even and odd features are considered the real and imaginary part of one complex number, respectively
   """
 
   layer_class = "alternating_real_to_complex"
 
   def __init__(self, **kwargs):
-    """
-    """
     super(AlternatingRealToComplexLayer, self).__init__(**kwargs)
 
     input_placeholder = self.input_data.get_placeholder_as_batch_major()
@@ -31,21 +31,29 @@ class AlternatingRealToComplexLayer(_ConcatInputLayer):
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, n_out=None, **kwargs):
-    return super(AlternatingRealToComplexLayer, cls).get_out_data_from_opts(name=name, sources=sources, out_type={"dim": n_out, "dtype": "complex64", "batch_dim_axis": 0, "time_dim_axis": 1}, **kwargs)
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int|None|returnn.util.basic.NotSpecified n_out:
+    :rtype: Data
+    """
+    return super(AlternatingRealToComplexLayer, cls).get_out_data_from_opts(
+      name=name, sources=sources,
+      out_type={"dim": n_out, "dtype": "complex64", "batch_dim_axis": 0, "time_dim_axis": 1}, **kwargs)
 
 
 class BatchMedianPoolingLayer(_ConcatInputLayer):
   """
   This layer is used to pool together batches by taking their medium value.
-  Thus the batch size is divided by pool_size. The stride is hard coded to be
-  equal to the pool size
+  Thus the batch size is divided by pool_size.
+  The stride is hard coded to be equal to the pool size.
   """
 
   layer_class = "batch_median_pooling"
 
   def __init__(self, pool_size=1, **kwargs):
     """
-    :param pool_size int: size of the pool to take median of (is also used as stride size)
+    :param int pool_size: size of the pool to take median of (is also used as stride size)
     """
     super(BatchMedianPoolingLayer, self).__init__(**kwargs)
 
@@ -73,6 +81,13 @@ class BatchMedianPoolingLayer(_ConcatInputLayer):
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, pool_size, n_out=None, **kwargs):
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int pool_size:
+    :param int|None|returnn.util.basic.NotSpecified n_out:
+    :rtype: Data
+    """
     input_data = get_concat_sources_data_template(sources)
     assert not input_data.sparse
     return Data(
@@ -90,11 +105,21 @@ class BatchMedianPoolingLayer(_ConcatInputLayer):
 
 
 class ComplexLinearProjectionLayer(_ConcatInputLayer):
+  """
+  Complex linear projection layer
+  For the original idea, see:
+  Variani, Ehsan, et al. "Complex linear projection (CLP): A discriminative approach to joint feature extraction and
+  acoustic modeling." (2016).
+  """
   layer_class = "complex_linear_projection"
 
   def __init__(self, nr_of_filters, clp_weights_init="glorot_uniform", **kwargs):
+    """
+    :param int nr_of_filters:
+    :param str|dict[str]|float|numpy.ndarray clp_weights_init:
+    """
     if 'n_out' in kwargs and kwargs['n_out'] != nr_of_filters:
-        raise Exception('argument n_out of layer MelFilterbankLayer can not be different from nr_of_filters')
+      raise Exception('argument n_out of layer MelFilterbankLayer can not be different from nr_of_filters')
     kwargs['n_out'] = nr_of_filters
     self._nr_of_filters = nr_of_filters
     super(ComplexLinearProjectionLayer, self).__init__(**kwargs)
@@ -110,7 +135,8 @@ class ComplexLinearProjectionLayer(_ConcatInputLayer):
       clp_weights_initializer = get_initializer(
         clp_weights_init, seed=self.network.random.randint(2 ** 31), eval_local_ns={"layer": self})
       clp_kernel = self.add_param(tf_compat.v1.get_variable(
-        name="clp_kernel", shape=(2, kernel_width, kernel_height), dtype=tf.float32, initializer=clp_weights_initializer))
+        name="clp_kernel", shape=(2, kernel_width, kernel_height), dtype=tf.float32,
+        initializer=clp_weights_initializer))
     return clp_kernel
 
   def _build_clp_multiplication(self, clp_kernel):
@@ -130,6 +156,10 @@ class ComplexLinearProjectionLayer(_ConcatInputLayer):
 
   @classmethod
   def get_out_data_from_opts(cls, nr_of_filters, **kwargs):
+    """
+    :param int nr_of_filters:
+    :rtype: Data
+    """
     if 'n_out' not in kwargs:
       kwargs['n_out'] = nr_of_filters
     return super(ComplexLinearProjectionLayer, cls).get_out_data_from_opts(**kwargs)
@@ -137,40 +167,36 @@ class ComplexLinearProjectionLayer(_ConcatInputLayer):
 
 class ComplexToAlternatingRealLayer(_ConcatInputLayer):
   """
-  This layer converts a complex valued input tensor into a real valued output
-  tensor.
-  For this the even and odd parts of the output are considered the real and imaginary part of
-  one complex number, respectively
+  This layer converts a complex valued input tensor into a real valued output tensor.
+  For this the even and odd parts of the output are considered the real and imaginary part of one complex number,
+  respectively.
   """
 
   layer_class = "complex_to_alternating_real"
 
   def __init__(self, **kwargs):
-    """
-    """
-    def _interleaveVectors(vec1, vec2):
-        vec1 = tf.expand_dims(vec1, 3)
-        vec2 = tf.expand_dims(vec2, 3)
-        interleaved = tf.concat([vec1, vec2], 3)
-        interleaved = tf.reshape(interleaved, (tf.shape(vec1)[0], tf.shape(vec1)[1], tf.shape(vec1)[2] * 2))
-        return interleaved
+    def _interleave_vectors(vec1, vec2):
+      vec1 = tf.expand_dims(vec1, 3)
+      vec2 = tf.expand_dims(vec2, 3)
+      interleaved = tf.concat([vec1, vec2], 3)
+      interleaved = tf.reshape(interleaved, (tf.shape(vec1)[0], tf.shape(vec1)[1], tf.shape(vec1)[2] * 2))
+      return interleaved
+
     super(ComplexToAlternatingRealLayer, self).__init__(**kwargs)
 
     input_placeholder = self.input_data.get_placeholder_as_batch_major()
 
     real_value = tf_compat.v1.real(input_placeholder)
     imag_value = tf_compat.v1.imag(input_placeholder)
-    self.output.placeholder = _interleaveVectors(real_value, imag_value)
+    self.output.placeholder = _interleave_vectors(real_value, imag_value)
     self.output.size_placeholder = {0: self.input_data.size_placeholder[self.input_data.time_dim_axis_excluding_batch]}
 
 
 class MaskBasedGevBeamformingLayer(LayerBase):
   """
-  This layer applies GEV beamforming to a multichannel signal. The different
-  channels are assumed to be concatenated to the
-  input feature vector. The first source to the layer must contain the complex
-  spectrograms of the single channels and the
-  second source must contain the noise and speech masks
+  This layer applies GEV beamforming to a multichannel signal. The different channels are assumed to be concatenated to
+  the input feature vector. The first source to the layer must contain the complex spectrograms of the single channels
+  and the second source must contain the noise and speech masks
   """
 
   layer_class = "mask_based_gevbeamforming"
@@ -186,22 +212,38 @@ class MaskBasedGevBeamformingLayer(LayerBase):
     """
     super(MaskBasedGevBeamformingLayer, self).__init__(**kwargs)
     assert len(self.sources) == 2
-
+    # noinspection PyUnresolvedReferences
     from tfSi6Proc.audioProcessing.enhancement.beamforming import TfMaskBasedGevBeamformer
 
-    complexSpectrogram = self.sources[0].output.get_placeholder_as_batch_major()
-    complexSpectrogram = tf.transpose(tf.reshape(complexSpectrogram, (tf.shape(complexSpectrogram)[0], tf.shape(complexSpectrogram)[1], nr_of_channels, tf.shape(complexSpectrogram)[2] // nr_of_channels)), [0, 1, 3, 2])
-    masks = tf.transpose(self.sources[1].output.placeholder, [self.sources[1].output.batch_dim_axis, self.sources[1].output.time_dim_axis, self.sources[1].output.feature_dim_axis])
-    masks = tf.transpose(tf.reshape(masks, (tf.shape(masks)[0], tf.shape(masks)[1], nr_of_channels, tf.shape(masks)[2] / nr_of_channels)), [0, 1, 3, 2])
-    noiseMasks = masks[:, :, :(tf.shape(masks)[2] // 2), :]
-    speechMasks = masks[:, :, (tf.shape(masks)[2] // 2):, :]
+    complex_spectrogram = self.sources[0].output.get_placeholder_as_batch_major()
+    complex_spectrogram = tf.transpose(tf.reshape(complex_spectrogram, (
+      tf.shape(complex_spectrogram)[0], tf.shape(complex_spectrogram)[1], nr_of_channels,
+      tf.shape(complex_spectrogram)[2] // nr_of_channels)), [0, 1, 3, 2])
+    masks = tf.transpose(
+      self.sources[1].output.placeholder,
+      [self.sources[1].output.batch_dim_axis, self.sources[1].output.time_dim_axis,
+       self.sources[1].output.feature_dim_axis])
+    masks = tf.transpose(
+      tf.reshape(masks, (tf.shape(masks)[0], tf.shape(masks)[1], nr_of_channels, tf.shape(masks)[2] // nr_of_channels)),
+      [0, 1, 3, 2])
+    noise_masks = masks[:, :, :(tf.shape(masks)[2] // 2), :]
+    speech_masks = masks[:, :, (tf.shape(masks)[2] // 2):, :]
 
-    gevBf = TfMaskBasedGevBeamformer(flag_inputHasBatch=1, tfFreqDomInput=complexSpectrogram, tfNoiseMask=noiseMasks, tfSpeechMask=speechMasks, postFilterId=postfilter_id, qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
-    bfOut = gevBf.getFrequencyDomainOutputSignal()
-    self.output.placeholder = bfOut
+    gev_bf = TfMaskBasedGevBeamformer(
+      flag_inputHasBatch=1, tfFreqDomInput=complex_spectrogram, tfNoiseMask=noise_masks, tfSpeechMask=speech_masks,
+      postFilterId=postfilter_id, qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
+    bf_out = gev_bf.getFrequencyDomainOutputSignal()
+    self.output.placeholder = bf_out
 
   @classmethod
-  def get_out_data_from_opts(cls, out_type={}, n_out=None, **kwargs):
+  def get_out_data_from_opts(cls, out_type=None, n_out=None, **kwargs):
+    """
+    :param dict[str]|None out_type:
+    :param int|None|returnn.util.basic.NotSpecified n_out:
+    :rtype: Data
+    """
+    if out_type is None:
+      out_type = {}
     out_type.setdefault("dim", n_out)
     out_type["batch_dim_axis"] = 0
     out_type["time_dim_axis"] = 1
@@ -210,11 +252,9 @@ class MaskBasedGevBeamformingLayer(LayerBase):
 
 class MaskBasedMvdrBeamformingWithDiagLoadingLayer(LayerBase):
   """
-  This layer applies GEV beamforming to a multichannel signal. The different
-  channels are assumed to be concatenated to the
-  input feature vector. The first source to the layer must contain the complex
-  spectrograms of the single channels and the
-  second source must contain the noise and speech masks
+  This layer applies GEV beamforming to a multichannel signal. The different channels are assumed to be concatenated to
+  the input feature vector. The first source to the layer must contain the complex spectrograms of the single channels
+  and the second source must contain the noise and speech masks.
   """
 
   layer_class = "mask_based_mvdrbeamforming"
@@ -229,19 +269,35 @@ class MaskBasedMvdrBeamformingWithDiagLoadingLayer(LayerBase):
     super(MaskBasedMvdrBeamformingWithDiagLoadingLayer, self).__init__(**kwargs)
     assert len(self.sources) == 2
 
+    # noinspection PyUnresolvedReferences
     from tfSi6Proc.audioProcessing.enhancement.beamforming import TfMaskBasedMvdrBeamformer
 
-    complexSpectrogramWithConcatChannels = self.sources[0].output.get_placeholder_as_batch_major()
-    complexSpectrogram = tf.transpose(tf.reshape(complexSpectrogramWithConcatChannels, (tf.shape(complexSpectrogramWithConcatChannels)[0], tf.shape(complexSpectrogramWithConcatChannels)[1], nr_of_channels, tf.shape(complexSpectrogramWithConcatChannels)[2] // nr_of_channels)), [0, 1, 3, 2])
-    noiseMasks = self.sources[1].output.get_placeholder_as_batch_major()
-    noiseMasks = tf.transpose(tf.reshape(noiseMasks, (tf.shape(noiseMasks)[0], tf.shape(noiseMasks)[1], nr_of_channels, tf.shape(noiseMasks)[2] // nr_of_channels)), [0, 1, 3, 2])
+    complex_spectrogram_with_concat_channels = self.sources[0].output.get_placeholder_as_batch_major()
+    complex_spectrogram = tf.transpose(tf.reshape(complex_spectrogram_with_concat_channels, (
+      tf.shape(complex_spectrogram_with_concat_channels)[0], tf.shape(complex_spectrogram_with_concat_channels)[1],
+      nr_of_channels, tf.shape(complex_spectrogram_with_concat_channels)[2] // nr_of_channels)), [0, 1, 3, 2])
+    noise_masks = self.sources[1].output.get_placeholder_as_batch_major()
+    noise_masks = tf.transpose(tf.reshape(
+      noise_masks,
+      (tf.shape(noise_masks)[0], tf.shape(noise_masks)[1], nr_of_channels, tf.shape(noise_masks)[2] // nr_of_channels)),
+      [0, 1, 3, 2])
 
-    mvdrBf = TfMaskBasedMvdrBeamformer(flag_inputHasBatch=1, tfFreqDomInput=complexSpectrogram, tfNoiseMask=noiseMasks, tfDiagLoadingCoeff=tf.constant(diag_loading_coeff, dtype=tf.float32), qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
-    bfOut = mvdrBf.getFrequencyDomainOutputSignal()
-    self.output.placeholder = bfOut
+    mvdr_bf = TfMaskBasedMvdrBeamformer(
+      flag_inputHasBatch=1, tfFreqDomInput=complex_spectrogram, tfNoiseMask=noise_masks,
+      tfDiagLoadingCoeff=tf.constant(diag_loading_coeff, dtype=tf.float32), qrAlgorithmSteps=qralgorithm_steps,
+      outputNanFilter=output_nan_filter)
+    bf_out = mvdr_bf.getFrequencyDomainOutputSignal()
+    self.output.placeholder = bf_out
 
   @classmethod
-  def get_out_data_from_opts(cls, out_type={}, n_out=None, **kwargs):
+  def get_out_data_from_opts(cls, out_type=None, n_out=None, **kwargs):
+    """
+    :param dict[str]|None out_type:
+    :param int|None|returnn.util.basic.NotSpecified n_out:
+    :rtype: Data
+    """
+    if out_type is None:
+      out_type = {}
     out_type.setdefault("dim", n_out)
     out_type["batch_dim_axis"] = 0
     out_type["time_dim_axis"] = 1
@@ -250,18 +306,20 @@ class MaskBasedMvdrBeamformingWithDiagLoadingLayer(LayerBase):
 
 class MelFilterbankLayer(_ConcatInputLayer):
   """
-  This layer applies the log Mel filterbank to the input
+  This layer applies the Mel filterbank to the input.
   """
 
   layer_class = "mel_filterbank"
 
   def __init__(self, sampling_rate=16000, fft_size=1024, nr_of_filters=80, **kwargs):
     """
-    :param sampling_rate int: sampling rate of the signal which the input originates from
-    :param fft_size int: fft_size with which the time signal was transformed into the intput
-    :param nr_of_filters int: number of output filter bins
+    :param int sampling_rate: sampling rate of the signal which the input originates from
+    :param int fft_size: fft_size with which the time signal was transformed into the intput
+    :param int nr_of_filters: number of output filter bins
     """
-    def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
+
+    # noinspection PyShadowingNames
+    def tf_mel_filter_bank(f_min, f_max, sampling_rate, fft_size, nr_of_filters):
       """
       Returns the filter matrix which yields the mel filter bank features, when applied to the spectrum as
       tf.matmul(freqDom, filterMatrix), where freqDom has dimension (time, frequency)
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/signal_processing.py
```diff
@@ -426,32 +459,46 @@ def _get_nr_of_channels_from_input_placeholder(self):
...
-            import scipy.signal
-            window = tf.constant(scipy.signal.blackman(frame_size), dtype=tf.float32)
-        if self._window == "None" or self._window == "ones":
+          try:
+            from tensorflow.signal import hann_window
+          except ImportError:
+            from tensorflow.contrib.signal import hann_window
+          window = hann_window(window_length, dtype=dtype)
+        elif self._window == "blackman":
+          tf_compat.v1.assert_equal(frame_size, window_length)
```
albertz, line 478: This assert does not have any effect.
You should use `with tf.control_dependencies([...])` instead.

returnn/tf/layers/signal_processing.py
```diff
@@ -464,7 +511,9 @@ def _padTimeSignal(input_placeholder, frame_size):
         concat_feature_dim = channel_wise_stft.shape[2] * channel_wise_stft.shape[3]
         channel_concatenated_stft = tf.reshape(channel_wise_stft, (batch_dim, time_dim, concat_feature_dim))
         if channel_wise_stft_res_list:
-          channel_concatenated_stft = _cropStftOutputToReferenceFrameSizeLength(channel_concatenated_stft, get_shape(channel_wise_stft_res_list[0])[1])
+          channel_concatenated_stft = \
+            _crop_stft_output_to_reference_frame_size_length(channel_concatenated_stft,
+                                                             get_shape(channel_wise_stft_res_list[0])[1])
```
albertz, line 516: Do not use `\`. Instead, add brackets.
Also, do not indent like that. We don't want to have such wide hanging indents.
```suggestion
          channel_concatenated_stft = (
            _crop_stft_output_to_reference_frame_size_length(
              channel_concatenated_stft,
              get_shape(channel_wise_stft_res_list[0])[1]))
```

returnn/tf/layers/signal_processing.py
```diff
@@ -524,7 +574,8 @@ def __init__(self, nr_of_frames, **kwargs):
     super(NoiseEstimationByFirstTFramesLayer, self).__init__(**kwargs)
     self._nr_of_frames = nr_of_frames
     noise_vector = self._get_noise_vector()
-    self.output.placeholder = tf.tile(noise_vector, (1, tf.shape(self.input_data.get_placeholder_as_batch_major())[1], 1))
+    self.output.placeholder = tf.tile(noise_vector,
+                                      (1, tf.shape(self.input_data.get_placeholder_as_batch_major())[1], 1))
```
albertz, line 578:
```suggestion
    self.output.placeholder = tf.tile(
      noise_vector, (1, tf.shape(self.input_data.get_placeholder_as_batch_major())[1], 1))
```

returnn/tf/layers/signal_processing.py
```diff
@@ -561,48 +613,51 @@ def __init__(self, noise_power_spectrum_tensor):
       def from_layer(cls, layer):
         return cls(layer.output.get_placeholder_as_batch_major())
 
-      def getNoisePowerSpectrum(self):
+      def get_noise_power_spectrum(self):
         return self._noise_power_spectrum_tensor
 
-    def _getParametersFromConstructorInputs(parameters, l_overwrite, p_overwrite, q_overwrite, average_parameters):
+    def _get_parameters_from_constructor_inputs(parameters_, l_overwrite_, p_overwrite_, q_overwrite_,
```
albertz, line 619: In the case that you need to rename arguments of the local function, I would rather prefer to not do that. So the `# noinspection ...` is a cleaner here, I think.

returnn/tf/layers/signal_processing.py
```diff
@@ -561,48 +613,51 @@ def __init__(self, noise_power_spectrum_tensor):
...
-        q = tf.expand_dims(parameter_vector[:, :, 2], axis=-1)
-      return l, p, q
+        q_ = tf.expand_dims(parameter_vector[:, :, 2], axis=-1)
+      return l_, p_, q_
 
     filter_input_placeholder = filter_input.output.get_placeholder_as_batch_major()
     if filter_input_placeholder.dtype != tf.complex64:
       filter_input_placeholder = tf.cast(filter_input_placeholder, dtype=tf.complex64)
-    tf_compat.v1.assert_equal(noise_estimation.output.get_placeholder_as_batch_major().shape[-1], filter_input_placeholder.shape[-1])
+    tf_compat.v1.assert_equal(noise_estimation.output.get_placeholder_as_batch_major().shape[-1],
```
albertz, line 647: This assert has no effect.

returnn/tf/layers/signal_processing.py
```diff
@@ -681,10 +737,15 @@ def __init__(self, nr_of_channels=1, **kwargs):
 
     input_placeholder = self.input_data.get_placeholder_as_batch_major()
 
-    output = tf.reshape(input_placeholder, [tf.shape(input_placeholder)[0], tf.shape(input_placeholder)[1], nr_of_channels, tf.shape(input_placeholder)[2] / nr_of_channels])
-    self.output.placeholder = tf.transpose(tf.reshape(tf.transpose(output, [1, 3, 0, 2]), (tf.shape(output)[1], tf.shape(output)[3], tf.shape(output)[0] * tf.shape(output)[2])), [2, 0, 1])
+    output = tf.reshape(input_placeholder,
+                        [tf.shape(input_placeholder)[0], tf.shape(input_placeholder)[1], nr_of_channels,
+                         tf.shape(input_placeholder)[2] / nr_of_channels])
```
albertz, line 742:
```suggestion
    output = tf.reshape(
      input_placeholder,
      [tf.shape(input_placeholder)[0], tf.shape(input_placeholder)[1],
       nr_of_channels,
       tf.shape(input_placeholder)[2] // nr_of_channels])
```
Note also the fix for `//` (important!).

returnn/tf/layers/signal_processing.py
```diff
@@ -681,10 +737,15 @@ def __init__(self, nr_of_channels=1, **kwargs):
...
+    self.output.placeholder = tf.transpose(tf.reshape(tf.transpose(output, [1, 3, 0, 2]), (
+      tf.shape(output)[1], tf.shape(output)[3], tf.shape(output)[0] * tf.shape(output)[2])), [2, 0, 1])
```
albertz, line 744: I would prefer such code to be multiple separate statements, where you also comment on the shape after each transformation. This is too complicated to read otherwise.

returnn/tf/layers/signal_processing.py
```diff
@@ -681,10 +737,15 @@ def __init__(self, nr_of_channels=1, **kwargs):
...
     # work around to obtain result like numpy.repeat(size_placeholder, nr_of_channels)
-    self.output.size_placeholder = {self.output.time_dim_axis_excluding_batch: tf.reshape(tf.tile(tf.reshape(self.input_data.size_placeholder[self.input_data.time_dim_axis_excluding_batch], [-1, 1]), [1, nr_of_channels]), [-1])}
+    self.output.size_placeholder = {self.output.time_dim_axis_excluding_batch: tf.reshape(
+      tf.tile(tf.reshape(self.input_data.size_placeholder[self.input_data.time_dim_axis_excluding_batch], [-1, 1]),
+              [1, nr_of_channels]), [-1])}
```
albertz, line 748:
```suggestion
    self.output.size_placeholder = {
      self.output.time_dim_axis_excluding_batch: ...}
```
This is also way too complicated. Isn't this just the same as `tile_transposed`?

returnn/tf/layers/signal_processing.py
vieting: What exactly do you mean here? I didn't change anything in this line but can still update it of course.

returnn/tf/layers/signal_processing.py
```diff
@@ -722,7 +783,8 @@ def get_out_data_from_opts(cls, name, sources, repetitions, n_out=None, **kwargs
     assert not input_data.sparse
     return Data(
       name="%s_output" % name,
-      shape=[input_data.get_placeholder_as_batch_major().shape[1].value, input_data.get_placeholder_as_batch_major().shape[2].value * repetitions],
+      shape=[input_data.get_placeholder_as_batch_major().shape[1].value,
+             input_data.get_placeholder_as_batch_major().shape[2].value * repetitions],
```
albertz, line 787: This code doesn't make sense. Why do you get the tensor? `input_data` already has that information. It's just `input_data.shape[0]` and `input_data.shape[1]`, right?

returnn/tf/layers/signal_processing.py
albertz: Yea, I know, it was wrong before. But I thought, when you already change this, you should maybe also fix such obvious bugs. Or just remove the assert if you want to have the old behavior.

returnn/tf/layers/signal_processing.py
vieting: That goes for all functions, I guess?

albertz: What do you mean?

vieting: Then I'll revert all changes that include renaming function arguments and use `# noinspection...` instead.

But renaming is ok, so I leave that?

albertz: I don't understand what you ask about renaming the function. The code should pass the code style check. You can name the function whatever way you like as long as it passed the code style check.

albertz:
Btw, check the Travis test. There are still a lot of warnings. E.g. see [here](https://travis-ci.org/github/rwth-i6/returnn/jobs/716935448) for the latest test.
```
returnn/tf/layers/signal_processing.py:36: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
returnn/tf/layers/signal_processing.py:80: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
returnn/tf/layers/signal_processing.py:97: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
...
```
Maybe you have not enabled the respective warning.

@JackTemaki this could be added to the wiki as well.
https://github.com/rwth-i6/returnn/wiki/PyCharm-Configuration

JackTemaki:
> @JackTemaki this could be added to the wiki as well.
> https://github.com/rwth-i6/returnn/wiki/PyCharm-Configuration

Done. We could also think about mentioning "Missing type hinting for function definition", which is also disabled by default.

returnn/tf/layers/signal_processing.py
vieting: Probably yes. But it might not be batch major, right?

returnn/tf/layers/signal_processing.py
vieting: I fully agree. However, I don't have any setup which uses this code so I cannot test it. If I change stuff here, I would have to assume how it was used (e.g. input shapes) and hope that it does not affect the behaviour.

returnn/tf/layers/signal_processing.py
albertz: `data.shape` does not include the batch-dim, so time-dim vs batch-dim doesn't matter.

returnn/tf/layers/signal_processing.py
albertz: This doesn't need testing. This would just be a straight-forward code transformation. This is not different to what you do already now.
About the comments, also this doesn't need testing. You can simply read that from the code. I just mean to write it like [B,T,D], or whatever there is.

returnn/tf/layers/signal_processing.py
vieting: I removed it now.

returnn/tf/layers/signal_processing.py
vieting: You did not want the arguments to be renamed. Therefore, I was wondering if you want the function to be renamed. From your last comment I assume that is alright. 

I now undid the renaming of arguments and used `# noinspection` instead, but kept the updated function name.

returnn/tf/layers/signal_processing.py
vieting: I removed it now.

returnn/tf/layers/signal_processing.py
vieting: I split this into several lines to make it more readable.

returnn/tf/layers/signal_processing.py
vieting: Okay, I just use `input_data.shape[0]` and `input_data.shape[1]` now.

JackTemaki:
Can you try to delete `signal_processing.py` from `tests/lint_common.py`? I think then the correct test should run right away for the Pull request.

returnn/tf/layers/signal_processing.py
albertz: Ok for now. But I think further work on this should simplify this. If this is just `tile_transposed`, you should just use this function.

returnn/tf/layers/signal_processing.py
```diff
@@ -270,111 +325,106 @@ def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
...
 
-        :type melVal: float
-        :param melVal: value in mel domain
+        :param float melVal: value in mel domain
         :rtype: float
         """
         return 700.0 * (np.exp(float(melVal) / 1125) - 1)
 
-      def filterCenter(filterId, fMin, fMax, samplingRate, fftSize, nrOfFilters):
+      # noinspection PyPep8Naming,PyShadowingNames
```
albertz, line 357: Why `PyPep8Naming`? You should only ignore this warning for very good reasons. I don't understand why it is needed here. The arguments should be renamed to match the PEP8 style.

returnn/tf/layers/signal_processing.py
```diff
@@ -387,33 +437,34 @@ def __init__(self, frame_shift, frame_sizes, fft_sizes, window="hanning", use_rf
...
+    :param bool pad_last_frame: padding of last frame with zeros or discarding of last frame
     """
+
     def _compute_size_placeholder():
       size_placeholder_dict = {}
       nr_of_full_frames = (self.input_data.size_placeholder[0] - self._reference_frame_size) // self._frame_shift + 1
       nf_of_paded_frames = 0
-      if self._pad_last_frame and ((self.input_data.size_placeholder[0] - self._reference_frame_size) - (nr_of_full_frames - 1) * self._frame_shift > 0):
+      if self._pad_last_frame and ((self.input_data.size_placeholder[0] - self._reference_frame_size) -
+                                   (nr_of_full_frames - 1) * self._frame_shift > 0):
```
albertz, line 451: You should never indent like that (such wide to the right). Introduce further brackets, and then you can easily indent right after `if (`.

returnn/tf/layers/signal_processing.py
```diff
@@ -426,32 +477,52 @@ def _get_nr_of_channels_from_input_placeholder(self):
 
   def _apply_stft_to_input(self):
     from returnn.tf.util.basic import get_shape
-    def _cropStftOutputToReferenceFrameSizeLength(channel_concatenated_stft, crop_size):
-      return tf.slice(channel_concatenated_stft, [0, 0, 0], [get_shape(channel_concatenated_stft)[0], crop_size, get_shape(channel_concatenated_stft)[2]])
+
+    # noinspection PyShadowingNames
+    def _crop_stft_output_to_reference_frame_size_length(channel_concatenated_stft, crop_size):
+      return tf.slice(channel_concatenated_stft, [0, 0, 0], [get_shape(channel_concatenated_stft)[0], crop_size,
+                                                             get_shape(channel_concatenated_stft)[2]])
```
albertz, line 484: You should never indent like that. Rather, add newline after `tf.slice(` and then indent properly.

returnn/tf/layers/signal_processing.py
```diff
@@ -426,32 +477,52 @@ def _get_nr_of_channels_from_input_placeholder(self):
...
         return window
 
-      def _padTimeSignal(input_placeholder, frame_size):
+      # noinspection PyShadowingNames
+      def _pad_time_signal(input_placeholder, frame_size):
         if frame_size > self._reference_frame_size:
-          return tf.concat([input_signal, tf.ones([get_shape(input_signal)[0], frame_size-self._reference_frame_size, get_shape(input_signal)[2]])*1e-7], axis=1)
+          return tf.concat([input_signal, tf.ones(
+            [get_shape(input_signal)[0], frame_size - self._reference_frame_size, get_shape(input_signal)[2]]) * 1e-7],
+                           axis=1)
```
albertz, line 513: You should never indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -473,12 +547,18 @@ def _padTimeSignal(input_placeholder, frame_size):
...
     if use_rfft:
-        n_out = fft_size // 2 + 1
+      n_out = fft_size // 2 + 1
     n_out *= nr_of_channels
     return n_out
 
   @classmethod
   def get_out_data_from_opts(cls, fft_sizes, use_rfft=True, nr_of_channels=1, **kwargs):
+    """
+    :param list(int) fft_sizes:
```
albertz, line 557:
```suggestion
    :param list[int] fft_sizes:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -31,21 +30,29 @@ def __init__(self, **kwargs):
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, n_out=None, **kwargs):
-    return super(AlternatingRealToComplexLayer, cls).get_out_data_from_opts(name=name, sources=sources, out_type={"dim": n_out, "dtype": "complex64", "batch_dim_axis": 0, "time_dim_axis": 1}, **kwargs)
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int|None|NotSpecified n_out:
```
albertz, line 36: `NotSpecified` is not imported here, so you cannot refer to it directly like that. (You will see that PyCharm cannot correctly infer the type, when you move the mouse over it, and try to ctrl+click on it .)
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -73,6 +80,13 @@ def __init__(self, pool_size=1, **kwargs):
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, pool_size, n_out=None, **kwargs):
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int pool_size:
+    :param int|None|NotSpecified n_out:
```
albertz, line 87:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -186,22 +211,37 @@ def __init__(self, nr_of_channels=1, postfilter_id=0, qralgorithm_steps=None, ou
...
-    gevBf = TfMaskBasedGevBeamformer(flag_inputHasBatch=1, tfFreqDomInput=complexSpectrogram, tfNoiseMask=noiseMasks, tfSpeechMask=speechMasks, postFilterId=postfilter_id, qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
-    bfOut = gevBf.getFrequencyDomainOutputSignal()
-    self.output.placeholder = bfOut
+    complex_spectrogram = self.sources[0].output.get_placeholder_as_batch_major()
+    complex_spectrogram = tf.transpose(tf.reshape(complex_spectrogram, (
+      tf.shape(complex_spectrogram)[0], tf.shape(complex_spectrogram)[1], nr_of_channels,
+      tf.shape(complex_spectrogram)[2] // nr_of_channels)), [0, 1, 3, 2])
+    masks = tf.transpose(self.sources[1].output.placeholder,
+                         [self.sources[1].output.batch_dim_axis, self.sources[1].output.time_dim_axis,
+                          self.sources[1].output.feature_dim_axis])
```
albertz, line 223: Never indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -186,22 +211,37 @@ def __init__(self, nr_of_channels=1, postfilter_id=0, qralgorithm_steps=None, ou
...
+    masks = tf.transpose(
+      tf.reshape(masks, (tf.shape(masks)[0], tf.shape(masks)[1], nr_of_channels, tf.shape(masks)[2] / nr_of_channels)),
+      [0, 1, 3, 2])
+    noise_masks = masks[:, :, :(tf.shape(masks)[2] // 2), :]
+    speech_masks = masks[:, :, (tf.shape(masks)[2] // 2):, :]
+
+    gev_bf = TfMaskBasedGevBeamformer(flag_inputHasBatch=1, tfFreqDomInput=complex_spectrogram, tfNoiseMask=noise_masks,
+                                      tfSpeechMask=speech_masks, postFilterId=postfilter_id,
+                                      qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
```
albertz, line 232: Never indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -186,22 +211,37 @@ def __init__(self, nr_of_channels=1, postfilter_id=0, qralgorithm_steps=None, ou
...
+    bf_out = gev_bf.getFrequencyDomainOutputSignal()
+    self.output.placeholder = bf_out
 
   @classmethod
-  def get_out_data_from_opts(cls, out_type={}, n_out=None, **kwargs):
+  def get_out_data_from_opts(cls, out_type=None, n_out=None, **kwargs):
+    """
+    :param dict[str]|None out_type:
+    :param int|None|NotSpecified n_out:
```
albertz, line 240: Same.

returnn/tf/layers/signal_processing.py
```diff
@@ -229,19 +267,34 @@ def __init__(self, nr_of_channels=1, diag_loading_coeff=0, qralgorithm_steps=Non
...
-    bfOut = mvdrBf.getFrequencyDomainOutputSignal()
-    self.output.placeholder = bfOut
+    complex_spectrogram_with_concat_channels = self.sources[0].output.get_placeholder_as_batch_major()
+    complex_spectrogram = tf.transpose(tf.reshape(complex_spectrogram_with_concat_channels, (
+      tf.shape(complex_spectrogram_with_concat_channels)[0], tf.shape(complex_spectrogram_with_concat_channels)[1],
+      nr_of_channels, tf.shape(complex_spectrogram_with_concat_channels)[2] // nr_of_channels)), [0, 1, 3, 2])
+    noise_masks = self.sources[1].output.get_placeholder_as_batch_major()
+    noise_masks = tf.transpose(tf.reshape(noise_masks, (
+      tf.shape(noise_masks)[0], tf.shape(noise_masks)[1], nr_of_channels, tf.shape(noise_masks)[2] // nr_of_channels)),
+                               [0, 1, 3, 2])
```
albertz, line 280: Do not indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -229,19 +267,34 @@ def __init__(self, nr_of_channels=1, diag_loading_coeff=0, qralgorithm_steps=Non
...
+
+    mvdr_bf = TfMaskBasedMvdrBeamformer(flag_inputHasBatch=1, tfFreqDomInput=complex_spectrogram,
+                                        tfNoiseMask=noise_masks,
+                                        tfDiagLoadingCoeff=tf.constant(diag_loading_coeff, dtype=tf.float32),
+                                        qrAlgorithmSteps=qralgorithm_steps, outputNanFilter=output_nan_filter)
```
albertz, line 285: Never indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -229,19 +267,34 @@ def __init__(self, nr_of_channels=1, diag_loading_coeff=0, qralgorithm_steps=Non
...
+    bf_out = mvdr_bf.getFrequencyDomainOutputSignal()
+    self.output.placeholder = bf_out
 
   @classmethod
-  def get_out_data_from_opts(cls, out_type={}, n_out=None, **kwargs):
+  def get_out_data_from_opts(cls, out_type=None, n_out=None, **kwargs):
+    """
+    :param dict[str]|None out_type:
+    :param int|None|NotSpecified n_out:
```
albertz, line 293:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
albertz:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -250,18 +303,20 @@ def get_out_data_from_opts(cls, out_type={}, n_out=None, **kwargs):
...
-    :param sampling_rate int: sampling rate of the signal which the input originates from
-    :param fft_size int: fft_size with which the time signal was transformed into the intput
-    :param nr_of_filters int: number of output filter bins
+    :param int sampling_rate: sampling rate of the signal which the input originates from
+    :param int fft_size: fft_size with which the time signal was transformed into the intput
+    :param int nr_of_filters: number of output filter bins
     """
-    def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
+
+    # noinspection PyPep8Naming,PyShadowingNames
```
albertz, line 318: Why `PyPep8Naming`? You should never use that (unless there are very good reasons why you really need it).

returnn/tf/layers/signal_processing.py
```diff
@@ -270,111 +325,106 @@ def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
...
 
-        :type freq: float
-        :param freq: frequency value to transform onto mel scale
+        :param float freq: frequency value to transform onto mel scale
         :rtype: float
         """
         return 1125.0 * np.log(1 + float(freq) / 700)
 
-      def invMelScale(melVal):
+      # noinspection PyPep8Naming
```
albertz, line 347: Why?

returnn/tf/layers/signal_processing.py
```diff
@@ -270,111 +325,106 @@ def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
...
     super(MelFilterbankLayer, self).__init__(**kwargs)
 
     input_placeholder = self.input_data.get_placeholder_as_batch_major()
 
-    mel_fbank_mat = tfMelFilterBank(0, sampling_rate / 2.0, sampling_rate, fft_size, nr_of_filters)
-    self.output.placeholder = tf.einsum('btf,bfc->btc', input_placeholder, tf.tile(tf.expand_dims(mel_fbank_mat, axis=0), [tf.shape(input_placeholder)[0], 1, 1]))
+    mel_fbank_mat = tf_mel_filter_bank(0, sampling_rate / 2.0, sampling_rate, fft_size, nr_of_filters)
+    self.output.placeholder = tf.einsum('btf,bfc->btc', input_placeholder,
+                                        tf.tile(tf.expand_dims(mel_fbank_mat, axis=0),
+                                                [tf.shape(input_placeholder)[0], 1, 1]))
```
albertz, line 401: Do not indent like this.

returnn/tf/layers/signal_processing.py
```diff
@@ -270,111 +325,106 @@ def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
...
     self.output.size_placeholder = {0: self.input_data.size_placeholder[self.input_data.time_dim_axis_excluding_batch]}
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, n_out=None, **kwargs):
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int|None|NotSpecified n_out:
```
albertz, line 409:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -270,111 +325,106 @@ def tfMelFilterBank(fMin, fMax, samplingRate, fftSize, nrOfFilters):
...
     return super(MelFilterbankLayer, cls).get_out_data_from_opts(
       name=name, sources=sources, out_type={"dim": n_out, "batch_dim_axis": 0, "time_dim_axis": 1}, **kwargs)
 
 
 class MultiChannelMultiResolutionStftLayer(_ConcatInputLayer):
   """
   The layer applys a STFT to every channel separately and concatenates the frequency domain vectors for every frame.
-  The STFT is applied with multiple different frame- and fft-sizes
-  and the resulting multi-channel stfts are concatenated.
+  The STFT is applied with multiple different frame- and fft-sizes and the resulting multi-channel stfts are
```
albertz, line 419: Keep consistent. Why lower case "stfts"?

albertz: Also why lower case "fft"?

returnn/tf/layers/signal_processing.py
```diff
@@ -490,12 +570,13 @@ def get_out_data_from_opts(cls, fft_sizes, use_rfft=True, nr_of_channels=1, **kw
...
   """
-  The layer applys a STFT to every channel separately and concatenates the frequency domain vectors for every frame
+  The layer applys a STFT to every channel separately and concatenates the frequency domain vectors for every frame.
   """
   recurrent = True
   layer_class = "multichannel_stft_layer"
 
-  def __init__(self, frame_shift, frame_size, fft_size, window="hanning", use_rfft=True, nr_of_channels=1, pad_last_frame=False, **kwargs):
+  def __init__(self, frame_shift, frame_size, fft_size, window="hanning", use_rfft=True, nr_of_channels=1,
+               pad_last_frame=False, **kwargs):
```
albertz, line 579: Note: For the function arguments in a function definition, I think the indentation is ok. But only for that.

returnn/tf/layers/signal_processing.py
```diff
@@ -681,13 +802,30 @@ def __init__(self, nr_of_channels=1, **kwargs):
 
     input_placeholder = self.input_data.get_placeholder_as_batch_major()
 
-    output = tf.reshape(input_placeholder, [tf.shape(input_placeholder)[0], tf.shape(input_placeholder)[1], nr_of_channels, tf.shape(input_placeholder)[2] / nr_of_channels])
-    self.output.placeholder = tf.transpose(tf.reshape(tf.transpose(output, [1, 3, 0, 2]), (tf.shape(output)[1], tf.shape(output)[3], tf.shape(output)[0] * tf.shape(output)[2])), [2, 0, 1])
+    # For shape comments: C = nr_of_channels
+    output = tf.reshape(input_placeholder, [
+      tf.shape(input_placeholder)[0], tf.shape(input_placeholder)[1], nr_of_channels,
+      tf.shape(input_placeholder)[2] / nr_of_channels])  # shape B x T x C x (F / C)
```
albertz, line 808:
```suggestion
      tf.shape(input_placeholder)[2] // nr_of_channels])  # shape B x T x C x (F / C)
```

returnn/tf/layers/signal_processing.py
```diff
@@ -681,13 +802,30 @@ def __init__(self, nr_of_channels=1, **kwargs):
...
+    size_placeholder_np_repeat = tf.reshape(size_placeholder_np_repeat, [-1])  # shape (T * C)
+    self.output.size_placeholder = {self.output.time_dim_axis_excluding_batch: size_placeholder_np_repeat}
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, nr_of_channels, n_out=None, **kwargs):
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int nr_of_channels:
+    :param int|None|NotSpecified n_out:
```
albertz, line 826:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

returnn/tf/layers/signal_processing.py
```diff
@@ -718,11 +856,18 @@ def __init__(self, repetitions=1, **kwargs):
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources, repetitions, n_out=None, **kwargs):
+    """
+    :param str name:
+    :param list[LayerBase] sources:
+    :param int repetitions:
+    :param int|None|NotSpecified n_out:
```
albertz, line 863:
```suggestion
    :param int|None|returnn.util.basic.NotSpecified n_out:
```

albertz:
Btw, this `tfSi6Proc`, is that public? We should put it along the other external modules. I.e. make a Git submodule in `returnn/extern/`. And then also import it that way, i.e. like `from returnn.extern import tfSi6Proc`.

albertz:
If the TF 2 tests for signal processing run through now, you should also enable the test.

returnn/tf/layers/signal_processing.py
vieting: I did rename the arguments previously but you commented:
> In the case that you need to rename arguments of the local function, I would rather prefer to not do that. So the `# noinspection ...` is a cleaner here, I think.

Therefore, I undid the renaming and used `# noinspection PyPep8Naming` instead

albertz: Sorry, this was about `PyShadowingNames`. You do not necessarily need to (or should not) rename arguments to avoid `PyShadowingNames`.

You almost always should avoid `PyPep8Naming`.

vieting: Okay, then I'll rename the arguments again. 

Is there some guideline or way of determining these priorities?

albertz: Mostly just common sense. :)

vieting:
Btw, I still get a PyCharm warning
> package containing module 'tensorflow' is not listed in project requirements
It is also present in other files. Do you disable this warning in PyCharm? If so, we should add it to the wiki
https://github.com/rwth-i6/returnn/wiki/PyCharm-Configuration

returnn/tf/layers/signal_processing.py
vieting: Sure. As I said, I renamed them in the first place and just undid it because of your comment.

vieting: If I rename them anyway, should we then also avoid `PyShadowingNames` warnings or still use `noinspection` there?

albertz:
> Btw, I still get a PyCharm warning
> 
> > package containing module 'tensorflow' is not listed in project requirements
> > It is also present in other files. Do you disable this warning in PyCharm? If so, we should add it to the wiki
> > https://github.com/rwth-i6/returnn/wiki/PyCharm-Configuration

The PyCharm settings we use for the GA CI tests are [here](https://github.com/rwth-i6/returnn/blob/master/tests/PyCharm.idea/inspectionProfiles/Project_Default.xml).
I'm not really decided whether it is the best solution to just ignore them all, as you see there.
But I also want to have it explicit, i.e. that we indeed get a warning when there is some random new package imported.
Feel free to edit/extend the wiki. (@JackTemaki maybe check as well.)

returnn/tf/layers/signal_processing.py
albertz: As I said (that is why I meant with my initial comment), I would usually not rename the arguments just for `PyShadowingNames`.

vieting:
> 
> 
> If the TF 2 tests for signal processing run through now, you should also enable the test. I.e. add a line in `.github/workflows/main.yml` (after [here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L22)), and then also remove the TF 1.15 specific test ([here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L75)).

They don't, but I guess that should be fixed in a separate PR, right?

albertz:
> > If the TF 2 tests for signal processing run through now, you should also enable the test. I.e. add a line in `.github/workflows/main.yml` (after [here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L22)), and then also remove the TF 1.15 specific test ([here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L75)).
> 
> They don't, but I guess that should be fixed in a separate PR, right?

I don't understand. I thought this is what this PR is about? You wrote:

> Also usages of tf.contrib are changed for TF 2 compatibility.

What else is missing?

returnn/tf/layers/signal_processing.py
```diff
@@ -186,22 +211,38 @@ def __init__(self, nr_of_channels=1, postfilter_id=0, qralgorithm_steps=None, ou
...
+    complex_spectrogram = self.sources[0].output.get_placeholder_as_batch_major()
+    complex_spectrogram = tf.transpose(tf.reshape(complex_spectrogram, (
+      tf.shape(complex_spectrogram)[0], tf.shape(complex_spectrogram)[1], nr_of_channels,
+      tf.shape(complex_spectrogram)[2] // nr_of_channels)), [0, 1, 3, 2])
+    masks = tf.transpose(
+      self.sources[1].output.placeholder,
+      [self.sources[1].output.batch_dim_axis, self.sources[1].output.time_dim_axis,
+       self.sources[1].output.feature_dim_axis])
+    masks = tf.transpose(
+      tf.reshape(masks, (tf.shape(masks)[0], tf.shape(masks)[1], nr_of_channels, tf.shape(masks)[2] / nr_of_channels)),
```
albertz, line 226:
```suggestion
      tf.reshape(masks, (tf.shape(masks)[0], tf.shape(masks)[1], nr_of_channels, tf.shape(masks)[2] // nr_of_channels)),
```

returnn/tf/layers/signal_processing.py
```diff
@@ -1,3 +1,6 @@
+"""
+Defines multiple signal processing-related layers.
+"""
 
 from __future__ import print_function
```
albertz, line 5:
```suggestion
from __future__ import print_function
from __future__ import division
```

albertz:
What about `tfSi6Proc`?

vieting:
> 
> 
> > > If the TF 2 tests for signal processing run through now, you should also enable the test. I.e. add a line in `.github/workflows/main.yml` (after [here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L22)), and then also remove the TF 1.15 specific test ([here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L75)).
> > 
> > 
> > They don't, but I guess that should be fixed in a separate PR, right?
> 
> I don't understand. I thought this is what you also wanted to fix in this PR? You wrote:
> 
> > Also usages of tf.contrib are changed for TF 2 compatibility.
> 
> What else is missing?

I have to check more closely. At least it was not as trivial as the changes done here before. But if you like we can also add it to this PR.

albertz:
> > > > If the TF 2 tests for signal processing run through now, you should also enable the test. I.e. add a line in `.github/workflows/main.yml` (after [here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L22)), and then also remove the TF 1.15 specific test ([here](https://github.com/rwth-i6/returnn/blob/194ddfaa20f0867f1e4634b60ade452bc47a4f47/.github/workflows/main.yml#L75)).
> > > 
> > > 
> > > They don't, but I guess that should be fixed in a separate PR, right?
> > 
> > 
> > I don't understand. I thought this is what you also wanted to fix in this PR? You wrote:
> > > Also usages of tf.contrib are changed for TF 2 compatibility.
> > 
> > 
> > What else is missing?
> 
> I have to check more closely. At least it was not as trivial as the changes done here before. But if you like we can also add it to this PR.

No, it's ok. It's already too big, so better let's merge it in now, and make it a separate PR then for further TF 2 compat fixes (whatever is missing).

Also, as I said, we should resolve the questions about tfSi6Proc. See my comments above.

albertz:
Note that the commits should be cleaned up. Or I will just use squash here (it's ok to have them all in one single commit).

vieting:
> 
> 
> What about `tfSi6Proc`?

Tobi had it as a private repository and transferred it to `rwth-i6` before he left. I think he would agree to making it public but didn't want to do it without discussion at the chair. Might be something to discuss "offline" ;)

vieting:
> 
> 
> Note that the commits should be cleaned up. Or I will just use squash here (it's ok to have them all in one single commit).

Done

albertz:
@JackTemaki can you also check again?

Issue #327, opened 2020-08-07
RelativePositionalEncodingLayer Does Not Pass Sanity Check in Recent Commits


npwynands:
I recently updated my RETURNN/master from commit fa0a7ad8 to e1a9974c (or a728507a). Now I have issues with the [`RelativePositionalEncodingLayer`](https://returnn.readthedocs.io/en/latest/api/tf.layers.rec.html?highlight=relative%20positional%20encoding#returnn.tf.layers.rec.RelativePositionalEncodingLayer) I have been using in my Transformer models:

```shell
...
Exception creating layer root/'teacherMT_enc_01_rel_pos' of class RelativePositionalEncodingLayer with opts:
{'forward_weights_init': "variance_scaling_initializer(mode='fan_in', "
                         "distribution='uniform', scale=0.78)",
 'n_out': 64,
 'name': 'teacherMT_enc_01_rel_pos',
 'network': <TFNetwork 'root' extra_nets={'extra.search': <TFNetwork 'extra.search' train=<tf.Tensor 'globals/train_flag:0' shape=() dtype=bool> search>} train=<tf.Tensor 'globals/train_flag:0' shape=() dtype=bool>>,
 'output': Data(name='teacherMT_enc_01_rel_pos_output', shape=(None, None, 64), batch_dim_axis=None, time_dim_axis=0, available_for_inference=False, batch_shape_meta=[T|'time:var:extern_data:source_text','time:var:extern_data:source_text',F|64]),
 'sources': [<LayerNormLayer 'teacherMT_enc_01_self_att_laynorm' out_type=Data(shape=(None, 512), available_for_inference=False, batch_shape_meta=[B,T|'time:var:extern_data:source_text',F|512])>],
 'trainable': False}
Traceback (most recent call last):
  File "/home/philipp/Documents/bachelor-thesis/returnn/repository/returnn/__main__.py", line 642, in main
    execute_main_task()
  ...
  File "/home/philipp/Documents/bachelor-thesis/returnn/repository/returnn/tf/network.py", line 776, in add_layer
    layer = self._create_layer(name=name, layer_class=layer_class, **layer_desc)
  File "/home/philipp/Documents/bachelor-thesis/returnn/repository/returnn/tf/network.py", line 723, in _create_layer
    output_template.sanity_check(ignore_placeholder=True)  # placeholder might be overwritten later
[...]

PR #336, opened 2020-08-12
ADDED: proper test case for RelativePositionalEncodingLayer

npwynands:
Related to issue #327

### DIFF
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -5573,6 +5573,54 @@ def test_PositionalEncodingLayer_offset_in_rec():
     print(out)  # random...
 
 
+def test_RelativePositionalEncodingLayer():
+  # Test `RelativePositionalEncodingLayer` with an example similar to its example usage.
+  rnd = numpy.random.RandomState(42)
+  n_batch, n_out, n_time = 3, 5, 7
+  EncKeyTotalDim = 9
+  EncValueTotalDim = 18
+  AttNumHeads = 3
+  net_dict = {
+    'rel_pos': {
+      "class": "relative_positional_encoding",
+      "from": "data",
+      "n_out": EncKeyTotalDim // AttNumHeads
+    },
+    'output': {
+      "class": "self_attention",
+      "num_heads": AttNumHeads,
+      "total_key_dim": EncKeyTotalDim,
+      "n_out": EncValueTotalDim, "from": "data",
+      "attention_left_only": False,
+      "key_shift": 'rel_pos'
+    }
+  }
+  config = Config()
+  config.update({
+    "extern_data": {"data": {"dim": n_out, "sparse": False}},
+    "debug_print_layer_output_template": True
+  })
+  with make_scope() as session:
+    network = TFNetwork(config=config, train_flag=True)
+    pprint(network.extern_data.data)
+    network.construct_from_dict(net_dict)
+    fetches = network.get_fetches_dict()
+    data_input = network.extern_data.data["data"]
+    assert data_input.batch_shape == (None, None, n_out)
+    train_out = network.get_layer("output").output
+    session.run(tf_compat.v1.variables_initializer(tf_compat.v1.global_variables() + [network.global_train_step]))
+    rand_data = rnd.rand(n_batch, n_time, n_out)
+    outputs = [train_out.placeholder]
+    info, out = session.run(
+      (fetches, outputs),
+      feed_dict={
+        data_input.placeholder: rand_data,
+        data_input.size_placeholder[0]: numpy.array([n_time] * n_batch, dtype="float32"),
+      })
+    print(info)
+    print(out)  # random...
+
+
 if __name__ == "__main__":
   try:
     better_exchook.install()


returnn/tf/layers/rec.py
```diff
@@ -8123,6 +8123,7 @@ def get_out_data_from_opts(cls, name, sources, n_out, **kwargs):
     data = data.copy_template().copy_as_batch_major()
     data.batch_dim_axis = None
     data.feature_dim_axis = NotSpecified
+    data.dim = n_out
```
albertz, line 8126: This was already committed, right?
Please rebase on current master.

albertz:
You reformatted the whole file. Don't do that. Please just add your test, without changing anything else. (So delete all the existing commits, and then do a new commit which just adds the test.)

albertz:
Why are there multiple commits? This is just a single test case. Just use a single commit.

albertz:
Your formatting is wrong. We use two space indents. Please always stay consistent to existing code! See the contribution guidelines. Also see PyCharm settings in wiki.

npwynands:
Which commit message is too long?

albertz:
> Which commit message is too long?

You see that here on GitHub. Just click on commits. It will show you which is too long by removing the trail and adding "...".

npwynands:
Ah, you mean the "REMOVED: [...]"? Okay.

npwynands:
Okay, I changed the space indent and squashed my commits. Is it fine now?

albertz:
The commit message is still somewhat inconsistent to other commit messages. We never have sth like "ADDED" as a prefix. Please keep it consistent. In this case, I would just call it "test_RelativePositionalEncodingLayer".

albertz:
Otherwise looks fine now. So this test failed before and runs fine now?

JackTemaki:
> Ah, you mean the "REMOVED: [...]"? Okay.

If you write your commit messages with git commit using vim, it will show you the allowed commit message length in yellow, and exceeding characters in white.

albertz:
> > Ah, you mean the "REMOVED: [...]"? Okay.
> 
> If you write your commit messages with git commit using vim, it will show you the allowed commit message length in yellow, and exceeding characters in white.

With PyCharm as well, when you configure it correctly.

JackTemaki:
> With PyCharm as well, when you configure it correctly.

I did not know that the configuration option existed. I will add this as well. Is 72 the correct maximum?

albertz:
> > With PyCharm as well, when you configure it correctly.
> 
> I did not know that the configuration option existed.

Well, this was kind of obvious that it does, or not?
I just was a bit surprised that it is not enabled by default. Although I think it was in some earlier version.

> I will add this as well.

I already added it.

> Is 72 the correct maximum?

I don't really know. I think so.

npwynands:
> The commit message is still somewhat inconsistent to other commit messages. We never have sth like "ADDED" as a prefix. Please keep it consistent. In this case, I would just call it "test_RelativePositionalEncodingLayer".

I don’t see any pattern/convention in your commit messages and you already merged one of my commits where I’ve been using such a prefix (8ce5f9494d9b379bc4713aa4c5387f13846421cf). I will remember that for future commits, but I don’t think that matters here...

JackTemaki:
> > Is 72 the correct maximum?
> 
> I don't really know. I think so.

It is 50 for the subject and 72 for the body, I changed it accordingly

npwynands:
> > Ah, you mean the "REMOVED: [...]"? Okay.
> 
> If you write your commit messages with git commit using vim, it will show you the allowed commit message length in yellow, and exceeding characters in white.

Ah okay, I simply write my commit messages into the terminal. There are no such assisting tools.

npwynands:
> Otherwise looks fine now. So this test failed before and runs fine now?

Yes, it does fail due to the sanity check.

albertz:
> > > Is 72 the correct maximum?
> > 
> > 
> > I don't really know. I think so.
> 
> It is 50 for the subject and 72 for the body, I changed it accordingly

I just checked. No, 50 is wrong. It is 72 for both the subject and body. At least on GitHub.
But this is also what you find similarly for the Linux kernel, which you can kind of see as the gold standard.
[Here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664):
> For these reasons, the ``summary`` must be no more than 70-75 characters

JackTemaki:
> Ah okay, I simply write my commit messages into the terminal. There are no such assisting tools.

you can add:
```
[core]
    editor = vim
```
in your `.gitconfig` file in your home folder. If you then call `git commit` it will open vim with a pre-defined template and correct colouring.

albertz:
> Ah okay, I simply write my commit messages into the terminal. There are no such assisting tools.

Never do that (i.e. never use `-m`). Always use some editor.

JackTemaki:
> > > > Is 72 the correct maximum?
> > > 
> > > 
> > > I don't really know. I think so.
> > 
> > 
> > It is 50 for the subject and 72 for the body, I changed it accordingly
> 
> I just checked. No, 50 is wrong. It is 72 for both the subject and body. At least on GitHub.
> But this is also what you find similarly for the Linux kernel, which you can kind of see as the gold standard.
> [Here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664):
> 
> > For these reasons, the `summary` must be no more than 70-75 characters

But Github does give a warning when you specify summaries greater than 50 characters, somebody posted a screenshot of this here: https://github.com/microsoft/vscode/issues/60300

albertz:
> > Ah okay, I simply write my commit messages into the terminal. There are no such assisting tools.
> 
> you can add:
> 
> ```
> [core]
>     editor = vim
> ```
> 
> in your `.gitconfig` file in your home folder. If you then call `git commit` it will open vim with a pre-defined template and correct colouring.

This is the default already, or not? Or rather, it will check for `$EDITOR`, which is even better. You better should not overwrite this in the `.gitconfig`. E.g. some environments will overwrite `$EDITOR`, and then Git should follow this, and not be overwritten.

albertz:
> > I just checked. No, 50 is wrong. It is 72 for both the subject and body. At least on GitHub.
> > But this is also what you find similarly for the Linux kernel, which you can kind of see as the gold standard.
> > [Here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664):
> > > For these reasons, the `summary` must be no more than 70-75 characters
> 
> But Github does give a warning when you specify summaries greater than 50 characters, somebody posted a screenshot of this here: [microsoft/vscode#60300](https://github.com/microsoft/vscode/issues/60300)
> 
> But we can go with 72, I also see this as more convenient.

It's not really a warning. It is a "pro tip". This is also the same recommendation by Linus and for Linux.

> So shoot for 50 characters, but consider 72 the hard limit.

This is what I would also recommend. It doesn't necessarily have to be exactly 50 though.

albertz:
Merging now. @npwynands Next time, please make the Git message consistent to our normal style.

albertz:
@npwynands Note that the mail you use for the commit (RWTH) is not registered in your GitHub account. You might want to add that if you want that this commit can be associated with your account.

PR #340, opened 2020-08-13
GitHub actions workflow to replace Travis

### LINKED ISSUE #308: GitHub CI instead of Travis?
Should we switch over to [GitHub CI](https://help.github.com/en/actions/building-and-testing-code-with-continuous-integration/about-continuous-integration) / [GitHub actions](https://help.github.com/en/actions/building-and-testing-code-with-continuous-integration/setting-up-continuous-integration-using-github-actions) instead of [Travis](https://travis-ci.org/github/rwth-i6/returnn/)? Or you a mixture of both?

Has anyone looked at GitHub CI and can share some experience? Advantages/disadvantages vs Travis?

I don't like that the Travis tests take so long to complete. But that is maybe just because we are on the free plan. Otherwise I'm mostly happy with Travis.

albertz:
This would fix #308.


JackTemaki:
So did you come to the conclusion that Github CI might be better than Travis? I saw #308, but I did not look into that topic yet.

albertz:
Note: This is work-in-progress. Do not review yet. I'm just testing. (I'm also using draft PR, as I'm not sure if this will correctly trigger the action.)

I have also not really decided. I just want to test.

I'm a bit annoyed at Travis because they just support 5 concurrent jobs (for the free plan), and it takes quite long to complete.
(Btw, I asked them, how much would it cost for more, e.g. 100 concurrent jobs. That would be 4700$/month. So I guess not an option.)

GitHub actions provides 20 concurrent jobs (for the free plan), so much better already.
Also, it has very nice support for caching across builds, which will enable us to speedup many of the things a bit (at least the installation/setup stuff). (Not sure if maybe Travis also supports that, but I haven't seen it...)

albertz:
Ok, I think this looks good already.
@JackTemaki any comments? Should we already adapt? It seems to work, and it is faster than Travis.

JackTemaki:
> Ok, I think this looks good already.
> @JackTemaki any comments? Should we already adapt? It seems to work, and it is faster than Travis.

I think this looks much nicer than Travis in nearly every aspect. It has a test overview directly in the PR window, you can use the "back" function of the browser (which did not work when opening the Travis page for me), and the logs are structured. 

Is there a reason that you did not enable testing for Python 2.7? Maybe it would be sufficient to test `SprintInterface` with Python 2.7 only.

albertz:
> Is there a reason that you did not enable testing for Python 2.7? Maybe it would be sufficient to test SprintInterface with Python 2.7 only.

I enabled only some of them for Python 2.7, e.g. TFUtil and TFNativeOp. This is different to before, where basically all tests were also run with Python 2.7. I can change that as well, if we want. Or select a couple of specific tests for Python 2.7. Suggestions?

But otherwise, it also sounds like you prefer it. Then I will merge it in soon.

jtrmal:
I have to say this looks super-cool

jtrmal:
> I'm a bit annoyed at Travis because they just support 5 concurrent jobs (for the free plan), and it takes quite long to complete.
> (Btw, I asked them, how much would it cost for more, e.g. 100 concurrent jobs. That would be 4700$/month. So I guess not an option.)
> 
> GitHub actions provides 20 concurrent jobs (for the free plan), so much better already.
> Also, it has very nice support for caching across builds, which will enable us to speedup many of the things a bit (at least the installation/setup stuff). (Not sure if maybe Travis also supports that, but I haven't seen it...)

Just fyi... 
a) for kaldi we have the smallest commercial plan and works quite OK -- I think there is also a higher priority for the commercial plan jobs...
b) we do caching via ccache in Travis (and Travis does support storing the ccache for the "next time". Kyrill would know details

PR #343, opened 2020-08-13
Fixing demos/demo-hyper-param-tuning.config

mattiadg:
This small edit fixes the error in #345 by assigning the "training" task to each run's config.

### DIFF
--- a/returnn/tf/hyper_param_tuning.py
+++ b/returnn/tf/hyper_param_tuning.py
@@ -635,6 +635,7 @@ class _IndividualTrainer:
     for p in self.optim.hyper_params:
       print(" %s -> %s" % (p.description(), hyper_param_mapping[p]), file=log.v2)
     config = self.optim.create_config_instance(hyper_param_mapping, gpu_ids=self.gpu_ids)
+    # init_train_from_config expects to have the train task
     config.set("task", "train")
     engine = Engine(config=config)
     train_data = StaticDataset.copy_from_dataset(self.optim.train_data)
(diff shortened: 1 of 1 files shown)


returnn/tf/hyper_param_tuning.py
```diff
@@ -635,6 +635,7 @@ def run(self):
     for p in self.optim.hyper_params:
       print(" %s -> %s" % (p.description(), hyper_param_mapping[p]), file=log.v2)
     config = self.optim.create_config_instance(hyper_param_mapping, gpu_ids=self.gpu_ids)
+    config.set("task", "train")
```
albertz, line 638: Why is that needed? That is already the default.

mattiadg: here it is set to `"hyper_param_tuning"`

albertz: But why is that relevant?

albertz:
Why is your stacktrace not correct? It misses the local variables. Do you have changed sth? Why is there `pdb`? Please always report the standard stacktrace.

albertz:
Also, I don't quite understand. Your change is not related to the error you report? So you change/fix one thing, but then report the error of another problem? Please do not mix things up. What exactly are you fixing here? Which error?

mattiadg:
I've added the error without pdb and created an issue with the original error. Now, the problem is an inconsistency between the dtype of the Data object and the dtype of the data it contains. I hope that it is limited to the `hyper_param_tuning` task.

returnn/tf/hyper_param_tuning.py
albertz: You should add a comment to this why it is relevant.
Sth like:
```suggestion
    # init_train_from_config expects to have the train task
    config.set("task", "train")
```

albertz:
> I've added the error without pdb and created an issue with the original error.

It would have been easier to just post that here, as this is what this PR is about.

> Now, the problem is an inconsistency between the dtype of the Data object and the dtype of the data it contains.

This is a separate thing. Neither this PR is about this, nor the issue you posted is about this.
You could fill in another issue, or directly fix that. But don't mix up things, and keep it simple.

PR #344, opened 2020-08-13
move non-log stdout print messages to log.v3 

JackTemaki:
I want to use RETURNN in a pipeable script (with log-verbosity:1), and right now these messages might get in my `stdout`. I am not sure how to find other messages that can still appear.

I assumed `stderr` is fine, because I did not want to add an additional dependency on `Log`.

### DIFF
--- a/returnn/tf/native_op.py
+++ b/returnn/tf/native_op.py
@@ -7,6 +7,7 @@ Wrappers for most relevant NativeOp ops.
 from __future__ import print_function
 
 import os
+import sys
 import tensorflow as tf
 try:
   from tensorflow.python.ops.nn import rnn_cell
@@ -74,6 +75,7 @@ class OpMaker(object):
   global_lock = RLock()
   mod_cache = {}  # cache_key -> mod
   op_cache = {}  # cache_key -> op
+  log_stream = sys.stdout  # type: typing.TextIO
 
   def __init__(self, description, compiler_opts=None,
                search_for_runtime_blas=True, search_for_numpy_blas=True, search_for_system_blas=True,
@@ -529,6 +531,7 @@ class OpMaker(object):
       include_deps=[self.support_native_op_cpp_filename],
       ld_flags=ld_flags,
       use_cuda_if_available=self.with_cuda,
+      log_stream=self.log_stream,
       **dict(self.compiler_opts))
     mod = comp.load_tf_module()
     mod._op_compiler = comp
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -919,7 +919,7 @@ class _DeviceAttributes:
 _list_local_devices = None
 
 
-def get_tf_list_local_devices(tf_session_opts=None):
+def get_tf_list_local_devices(tf_session_opts=None, file=None):
   """
   This uses tensorflow.device_lib.list_local_devices().
   Note that a call to this will trigger the internal TF thread pool inits,
@@ -929,13 +929,14 @@ def get_tf_list_local_devices(tf_session_opts=None):
   You can get the list available in a given TF session by :func:`tf.compat.v1.Session.list_devices`.
 
   :param dict[str]|None tf_session_opts: if given, will init a temp tf.compat.v1.Session with these opts
+  :param typing.TextIO|None file: log_stream stream for print statements, defaults to sys.stdout
   :rtype: list[tensorflow.core.framework.device_attributes_pb2.DeviceAttributes|_DeviceAttributes]
   """
   check_initial_tf_thread_pool_init(tf_session_opts=tf_session_opts)
   global _list_local_devices
   if _list_local_devices is not None:
     return _list_local_devices
-  print("Collecting TensorFlow device list...")
+  print("Collecting TensorFlow device list...", file=file)
   if tf_session_opts and tf_session_opts.get("gpu_options", {}):
     # On any gpu_options, e.g. gpu_options.per_process_gpu_memory_fraction:
     # The first usage of a device will use the provided options.
@@ -990,7 +991,7 @@ def print_available_devices(tf_session_opts=None, file=None):
   so you should call :func:`setup_tf_thread_pools` first.
 
   :param dict[str]|None tf_session_opts: if given, will init a temp Session with these opts
-  :param io.FileIO file:
+  :param typing.TextIO|None file: file stream for print statements, defaults to sys.stdout
   """
   if file is None:
     file = sys.stdout
@@ -1003,7 +1004,7 @@ def print_available_devices(tf_session_opts=None, file=None):
   if tf_session_opts and tf_session_opts.get("gpu_options", {}).get("visible_device_list", None):
     print("TF session gpu_options.visible_device_list is set to %r." % (
       tf_session_opts["gpu_options"]["visible_device_list"],), file=file)
-  devs = get_tf_list_local_devices(tf_session_opts=tf_session_opts)
+  devs = get_tf_list_local_devices(tf_session_opts=tf_session_opts, file=file)
   print("Local devices available to TensorFlow:", file=file)
   for i, dev in enumerate(devs):
     print("  %i/%i: %s" % (i + 1, len(devs), "\n       ".join(str(dev).splitlines())), file=file)
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3164,7 +3164,7 @@ class NativeCodeCompiler(object):
                is_cpp=True, c_macro_defines=None, ld_flags=None,
                include_paths=(), include_deps=None,
                static_version_name=None, should_cleanup_old_all=True, should_cleanup_old_mydir=False,
-               use_cxx11_abi=False, verbose=False):
+               use_cxx11_abi=False, log_stream=None, verbose=False):
     """
     :param str base_name: base name for the module, e.g. "zero_out"
     :param int|tuple[int] code_version: check for the cache whether to reuse
@@ -3180,6 +3180,7 @@ class NativeCodeCompiler(object):
     :param bool should_cleanup_old_all: whether we should look in the cache dir
       and check all ops if we can delete some old ones which are older than some limit (self._cleanup_time_limit_days)
     :param bool should_cleanup_old_mydir: whether we should delete our op dir before we compile there.
+    :param typing.TextIO|None log_stream: file stream for print statements
     :param bool verbose: be slightly more verbose
     """
     if self.CollectedCompilers is not None:
@@ -3203,8 +3204,9 @@ class NativeCodeCompiler(object):
       self._cleanup_old()
     self._should_cleanup_old_mydir = should_cleanup_old_mydir
     self.use_cxx11_abi = use_cxx11_abi
+    self._log_stream = log_stream
     if self.verbose:
-      print("%s: %r" % (self.__class__.__name__, self))
+      print("%s: %r" % (self.__class__.__name__, self), file=log_stream)
 
   def __repr__(self):
     return "<%s %r in %r>" % (self.__class__.__name__, self.base_name, self._mod_path)
@@ -3410,7 +3412,7 @@ class NativeCodeCompiler(object):
     cmd_bin = self._get_compiler_bin()
     cmd_args = [cmd_bin] + opts
     from subprocess import Popen, PIPE, STDOUT, CalledProcessError
-    print("%s call: %s" % (self.__class__.__name__, " ".join(cmd_args)))
+    print("%s call: %s" % (self.__class__.__name__, " ".join(cmd_args)), file=self._log_stream)
     proc = Popen(cmd_args, cwd=self._mod_path, stdout=PIPE, stderr=STDOUT)
     stdout, stderr = proc.communicate()
     assert stderr is None  # should only have stdout
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -357,6 +357,8 @@ def init_backend_engine():
     setup_tf_thread_pools(log_file=log.v3, tf_session_opts=tf_session_opts)
     # Print available devices. Also make sure that get_tf_list_local_devices uses the correct TF session opts.
     print_available_devices(tf_session_opts=tf_session_opts, file=log.v2)
+    from returnn.tf.native_op import OpMaker
+    OpMaker.log_stream = log.v3
     debug_register_better_repr()
     if config.is_true("distributed_tf"):
       import returnn.tf.distributed


albertz:
We should not start to forward random statements to stderr, this is not a solution.

albertz:
Usually we would pass a `file` (or `stream`) argument to the function, and then pass `file=log.v3` or so.

JackTemaki:
> Usually we would pass a `file` (or `stream`) argument to the function, and then pass `file=log.v3` or so.

This was exactly my question, so ok, I will also add `from returnn.log import log` there. I thought this was maybe missing on purpose.

albertz:
> > Usually we would pass a `file` (or `stream`) argument to the function, and then pass `file=log.v3` or so.
> 
> This was exactly my question, so ok, I will also add `from returnn.log import log` there. I thought this was maybe missing on purpose.

No, you missunderstood. You should not add the dependency.
You would add an argument (`file`) to the function (`get_tf_list_local_devices` or whatever) such that you do not need the depedency.

JackTemaki:
> You would add an argument (`file`) to the function (`get_tf_list_local_devices` or whatever) such that you do not need the depedency.

But I would have to pass this through many functions, I would have to touch about 10-15 function headers.

albertz:
`get_tf_list_local_devices` is used in 10-15 different places?
But anyway, I think we should not add the dependency on `log`. This part should stay totally independent.

JackTemaki:
> `get_tf_list_local_devices` is used in 10-15 different places?
> But anyway, I think we should not add the dependency on `log`. This part should stay totally independent.

Sorry, I meant for `_maybe_compile_inner(self):`, For `get_tf_list_local_devices` it is less (about 5 or so)

albertz:
> > `get_tf_list_local_devices` is used in 10-15 different places?
> > But anyway, I think we should not add the dependency on `log`. This part should stay totally independent.
> 
> Sorry, I meant for `_maybe_compile_inner(self)`, for `get_tf_list_local_devices` it is less (about 5 or so)

I don't understand. That is a hidden function. You would not call that.
Why don't you just pass it to `NativeCodeCompiler`?

In any case, I think we should not add the dependency on `log`. This part should stay totally independent.

JackTemaki:
> > > `get_tf_list_local_devices` is used in 10-15 different places?
> > > But anyway, I think we should not add the dependency on `log`. This part should stay totally independent.
> > 
> > 
> > Sorry, I meant for `_maybe_compile_inner(self)`, for `get_tf_list_local_devices` it is less (about 5 or so)
> 
> I don't understand. That is a hidden function. You would not call that.
> Why don't you just pass it to `NativeCodeCompiler`?
> 
> In any case, I think we should not add the dependency on `log`. This part should stay totally independent.

This is just too much to edit, and does not make sense at all. `get_tf_list_local_devices` is called indirectly by `is_gpu_available()` which as a total of 48 usages, from which 26 are in core Returnn code. I do not want to search by hand for all cases where the function could be called for the first time, and thus the print statement might occur.

We can do a local import, so that there is no additional global dependency, I do not see any other solution right now.

albertz:
`get_tf_list_local_devices` is cached for that reason. You don't really need that.
This is called via `print_available_devices`, which is called very early in RETURNN. So you should not have this problem.

JackTemaki:
> `get_tf_list_local_devices` is cached for that reason. You don't really need that.
> This is called via `print_available_devices`, which is called very early in RETURNN. So you should not have this problem.

Ok if this is reliable, then this part is not an issue. What about the `NativeCodeCompiler` then? I can add `file` as a parameter, but  I would likely need to pass it for all `OpCodeCompiler` (subclass) as well, right? Do you just want to avoid having the log dependency in the `returnn.util` package, and the rest is ok? Or also not in the `returnn.tf.util` package?

albertz:
> What about the `NativeCodeCompiler` then? I can add `file` as a parameter, but I would likely need to pass it for all `OpCodeCompiler` (subclass) as well, right? Do you just want to avoid having the log dependency in the `returnn.util` package, and the rest is ok? Or also not in the `returnn.tf.util` package?

In none of them. You should never add further dependencies if there is not a really good reason to do so. (This is in general, for any type of code, any project.)

Btw, for `NativeCodeCompiler`, I would not call the parameter `file`. That would be very misleading/confusing. Better sth like `log_output_stream` or so. And default should be `sys.stdout`.

JackTemaki:
> In none of them. You should never add further dependencies if there is not a really good reason to do so.

Well, I think this is a very vague definition. Logging is a very basic functionality that should be accessible in most if not all parts of the code. If I add the log dependency now to `returnn.tf.native_op.py` or `returnn.tf.util.basic.py` does not make a difference to me, but doing the later would reduce the code changes by a lot.

albertz:
> > In none of them. You should never add further dependencies if there is not a really good reason to do so.
> 
> Well, I think this is a very vague definition. Logging is a very basic functionality that should be accessible in most if not all parts of the code. If I add the log dependency now to `returnn.tf.native_op.py` or `returnn.tf.util.basic.py` does not make a difference to me, but doing the later would reduce the code changes by a lot.

It does for everyone who wants to use RETURNN as a framework and wants to use some parts from utils, TF utils, or TF native op. It means that they cannot do anymore without further setup-code because it expects that the RETURNN logging system is setup (`log.initialize`). It fails otherwise. Also, they might not want to adapt this logging system.

But in general, really, any type of dependencies are bad. You should keep the dependencies always as minimal as possible. This is important to be able to maintain code, and to find bugs. When everything depends on everything, you have an unmaintainable mess, and debugging is extremely hard.

Also, this case now really only requires maybe 1 or 2 lines of code change, not more. And it makes it more clean as well. I don't see why you need a lot of code changes here.

JackTemaki:
> Also, this case now really only requires maybe 1 or 2 lines of code change, not more. And it makes it more clean as well. I don't see why you need a lot of code changes here.

I do not see a solution without bloating everything, maybe there is something I clearly do not understand.

I put an `assert False` at the position where I need the log stream (next to the `print` statement) and this is my stacktrace through all modules that do not contain the `log` module:

```python
  File "/home/nick/github/returnn/returnn/tf/layers/rec.py", line 635, in _get_cell
    line: cell = rnn_cell_class(
            n_hidden=n_hidden, n_input_dim=self.input_data.dim,
            input_is_sparse=self.input_data.sparse,
            step=self._direction, **unit_opts)
    locals:
      cell = <not found>
      rnn_cell_class = <local> <class 'returnn.tf.native_op.NativeLstmCell'>
      n_hidden = <local> 256
      n_input_dim = <not found>
      self = <local> <RecLayer 'lstm1_fw' out_type=Data(shape=(None, 256), batch_dim_axis=1, batch_shape_meta=[T|'time:var:extern_data:data',B,F|256])>
      self.input_data = <local> Data(name='embed_output', shape=(None, 128), batch_shape_meta=[B,T|'time:var:extern_data:data',F|128])
      self.input_data.dim = <local> 128
      input_is_sparse = <not found>
      self.input_data.sparse = <local> False
      step = <not found>
      self._direction = <local> 1
      unit_opts = <local> {}
  File "/home/nick/github/returnn/returnn/tf/native_op.py", line 708, in __init__
    line: self.op = make_lstm_op()
    locals:
      self = <local> <returnn.tf.native_op.NativeLstmCell object at 0x7f44f835a750>
      self.op = <local> !AttributeError: 'NativeLstmCell' object has no attribute 'op'
      make_lstm_op = <global> <function make_lstm_op at 0x7f44f7019680>
  File "/home/nick/github/returnn/returnn/tf/native_op.py", line 650, in make_lstm_op
    line: return make_op(native_op.LstmGenericBase, **kwargs)
    locals:
      make_op = <global> <function make_op at 0x7f44f70195f0>
      native_op = <global> <module 'returnn.native_op' from '/home/nick/github/returnn/returnn/native_op.py'>
      native_op.LstmGenericBase = <global> <class 'returnn.native_op.LstmGenericBase'>
      kwargs = <local> {}
  File "/home/nick/github/returnn/returnn/tf/native_op.py", line 640, in make_op
    line: return maker.make_op()
    locals:
      maker = <local> <returnn.tf.native_op.OpMaker object at 0x7f44f83642d0>
      maker.make_op = <local> <bound method OpMaker.make_op of <returnn.tf.native_op.OpMaker object at 0x7f44f83642d0>>
  File "/home/nick/github/returnn/returnn/tf/native_op.py", line 546, in make_op
    line: mod = self._make_mod()
    locals:
      mod = <not found>
      self = <local> <returnn.tf.native_op.OpMaker object at 0x7f44f83642d0>
      self._make_mod = <local> <bound method OpMaker._make_mod of <returnn.tf.native_op.OpMaker object at 0x7f44f83642d0>>
  File "/home/nick/github/returnn/returnn/tf/native_op.py", line 533, in _make_mod
    line: mod = comp.load_tf_module()
    locals:
      mod = <not found>
      comp = <local> <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>
      comp.load_tf_module = <local> <bound method OpCodeCompiler.load_tf_module of <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>>
  File "/home/nick/github/returnn/returnn/tf/util/basic.py", line 2533, in load_tf_module
    line: self._maybe_compile()
    locals:
      self = <local> <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>
      self._maybe_compile = <local> <bound method NativeCodeCompiler._maybe_compile of <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>>
  File "/home/nick/github/returnn/returnn/util/basic.py", line 3362, in _maybe_compile
    line: self._maybe_compile_inner()
    locals:
      self = <local> <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>
      self._maybe_compile_inner = <local> <bound method NativeCodeCompiler._maybe_compile_inner of <OpCodeCompiler 'LstmGenericBase' in '/tmp/nick/returnn_tf_cache/ops/LstmGenericBase/fd388345ec'>>
  File "/home/nick/github/returnn/returnn/util/basic.py", line 3416, in _maybe_compile_inner
    line: assert False
    locals:
       no locals
```

So from what I see, I would need to pass a stream variable here:
```python
 cell = rnn_cell_class(
            n_hidden=n_hidden, n_input_dim=self.input_data.dim,
            input_is_sparse=self.input_data.sparse,
            step=self._direction, **unit_opts)
```
Where it would be part of the `**kwargs` until here: `
``` python
def make_op(cls, **kwargs):
  """
  :param type[returnn.native_op.NativeOpGenBase] cls:
  :param kwargs: passed to OpMaker
  :return: op
  :rtype: (tf.Tensor) -> tuple[tf.Tensor]
  """
  maker = OpMaker(OpDescription.from_gen_base(cls), **kwargs)
  return maker.make_op()
```
Either passed in the constructor or the `make_op` function.
And inside the `OpMaker` it would then need to be passed to the `OpCodeCompiler`:
```python
    comp = tf_util.OpCodeCompiler(
      base_name=self.name, code_version=self.description.code_version,
      code=self._make_code(),
      include_deps=[self.support_native_op_cpp_filename],
      ld_flags=ld_flags,
      use_cuda_if_available=self.with_cuda,
      **dict(self.compiler_opts))
```

And this is only one case, and there are also additional calls such as:
```python
      if self.description.is_grad_defined:
        assert not grad_func
        grad_description = self.description.grad()
        grad_op_maker = OpMaker(description=grad_description, compiler_opts=self.compiler_opts,
                                search_for_numpy_blas=self.search_for_numpy_blas,
                                blas_lib=self.blas_lib)
        grad_op = grad_op_maker.make_op()
```
In the `OpMaker` itself.

So what am I missing here? How can I easily add the stream?

albertz:
There are really countless of options. One random for example:

Add a global attribute to `OpMaker`, like:
```
class OpMaker:
  log_output_stream = sys.stdout
  ...
```
In `__main__`, you can then init that somewhere (e.g. after TF init).

For `NativeCodeCompiler`, you can make it an argument to `__init__`, that should be fine.

And then that's it?

Btw, why do you want to do that at all? Piping the output of RETURNN sounds like a very fragile solution which will easily break again and again. Why don't you just use RETURNN as a framework? That should be much simpler, cleaner, and more stable.

JackTemaki:
> There are really countless of options. One random for example:
> 
> Add a global attribute to `OpMaker`, like:
> 
> ```
> class OpMaker:
>   log_output_stream = sys.stdout
>   ...
> ```
> 
> In `__main__`, you can then init that somewhere (e.g. after TF init).
> 
> For `NativeCodeCompiler`, you can make it an argument to `__init__`, that should be fine.
> 
> And then that's it?

Hmm okay, I am not really in favour of global variables like this, but I can do that.

> Btw, why do you want to do that at all? Piping the output of RETURNN sounds like a very fragile solution which will easily break again and again. Why don't you just use RETURNN as a framework? That should be much simpler, cleaner, and more stable.

I am using RETURNN as a framework, this is why I said pipeable script (a script that uses RETURNN as framework). But this does not work when there are "hidden" prints of RETURNN to stdout. A framework should never do that (unless explicitly requested).

albertz:
> I am using RETURNN as a framework, this is why I said pipeable script (a script that uses RETURNN as framework). But this does not work when there are "hidden" prints of RETURNN to stdout. A framework should never do that (unless explicitly requested).

I don't quite understand. You mean your custom script, which uses RETURNN as a framework, is intended to be used for piping (or in general to parse stdout)? This will just shift the problem somewhere else. This is in general a fragile solution. Why don't you make your custom script a Python module and import it?

But yes, I agree, it would be nice to have some control over what ends up on stdout. But you are aware that TF itself also prints lots of stuff on stdout, right? Also, there are many other libs which potentially could do the same, and sometimes do (maybe rarely, or only in strange situations; maybe just not on your OS or in your current env).

JackTemaki:
> Why don't you make your custom script a Python module and import it?

I do not understand, how will that help with the stdout prints? I want to have an executable python binary/script that takes stdin as input and writes only the same lines (which are also edited by some RETURNN operation during the process) to stdout.

> But you are aware that TF itself also prints lots of stuff on stdout, right?

This did not happen for me yet, all TF messages were directed to stderr.

albertz:
> > Why don't you make your custom script a Python module and import it?
> 
> I do not understand, how will that help with the stdout prints? I want to have an executable python binary/script that takes stdin as input and writes only the same lines (which are also edited by some RETURNN operation during the process) to stdout.

I meant to not use pipes at all. It should be simpler (and more stable) to just do `import your_script` and do whatever you want to do.

JackTemaki:
> > > Why don't you make your custom script a Python module and import it?
> > 
> > 
> > I do not understand, how will that help with the stdout prints? I want to have an executable python binary/script that takes stdin as input and writes only the same lines (which are also edited by some RETURNN operation during the process) to stdout.
> 
> I meant to not use pipes at all. It should be simpler (and more stable) to just do `import your_script` and do whatever you want to do.

This is not an option, the only alternative would be to use actual files (or pipes as file descriptor on linux), but anyway, this should not be discussed here.

JackTemaki:
Is this what you Imagined? I still would prefer the `stderr` solution for the compile commands (especially if you want people to import only the compiled ops without loading returnn).

returnn/tf/native_op.py
```diff
@@ -74,6 +75,7 @@ class OpMaker(object):
   global_lock = RLock()
   mod_cache = {}  # cache_key -> mod
   op_cache = {}  # cache_key -> op
+  log_stream = sys.stdout  # io.TextIOBase
```
albertz, line 78: Is this supposed to be a type? Then you need `# type: io.TextIOBase`.

returnn/tf/util/basic.py
```diff
@@ -929,13 +929,14 @@ def get_tf_list_local_devices(tf_session_opts=None):
   You can get the list available in a given TF session by :func:`tf.compat.v1.Session.list_devices`.
 
   :param dict[str]|None tf_session_opts: if given, will init a temp tf.compat.v1.Session with these opts
+  :param io.FileIO file: file stream for print statements, defaults to sys.stdout
```
albertz, line 932: I think the type is wrong. It should be `io.TextIOBase`. Or maybe even more general, like `io.TextIOBase|io.StringIO|typing.TextIO|None`.

returnn/util/basic.py
```diff
@@ -3180,6 +3180,7 @@ def __init__(self, base_name, code_version, code,
     :param bool should_cleanup_old_all: whether we should look in the cache dir
       and check all ops if we can delete some old ones which are older than some limit (self._cleanup_time_limit_days)
     :param bool should_cleanup_old_mydir: whether we should delete our op dir before we compile there.
+    :param io.TextIOBase log_stream: file stream for print statements, defaults to sys.stdout
```
albertz, line 3183: Type is wrong. At least `...|None`. Or maybe `io.TextIOBase|io.StringIO|typing.TextIO|None`.

albertz:
Yes, this would be one solution.
Some types are wrong (don't you get type warnings).
But otherwise I think it's fine.

returnn/tf/util/basic.py
JackTemaki: I assumed the existing type had a reason (in `print_available_devices`). I will also correct that one as well.

albertz: It's a bit inconsistent and also wrong in other places. You will not always get warnings, but you will get warnings when you use different types. E.g. what should work (without warning) is `sys.stdout`, result from `open()`, and also `StringIO()`.

returnn/util/basic.py
JackTemaki: `io.StringIO` and `typing.TextIO` are not needed, as they are inheriting `io.TextIOBase`.

returnn/tf/native_op.py
```diff
@@ -74,6 +75,7 @@ class OpMaker(object):
   global_lock = RLock()
   mod_cache = {}  # cache_key -> mod
   op_cache = {}  # cache_key -> op
+  log_stream = sys.stdout  # type: io.TextIOBase
```
albertz, line 78: Better `typing.TextIO`. Also fixes the warning.
```suggestion
  log_stream = sys.stdout  # type: typing.TextIO
```
Or just remove the type at all. It is clear already just from the code.
```suggestion
  log_stream = sys.stdout
```

returnn/util/basic.py
albertz: I think you should better use `typing.TextIO`, as this is more generic (the inheritance is not correct / not relevant I think).
I think `io.TextIOBase` is missing some functions as well, so you might get warnings elsewhere.

returnn/tf/native_op.py
JackTemaki: This is an Issue with the type checking. All cases, (`sys.stdout`, `open`, and `StringIO`) are based on `io.TextIOBase`. For some reason, PyCharm does not complain when I use `_io.TextIOBase`, but for `io.TextIOBase`. Maybe `typing.TextIO` solves the Issue.

returnn/util/basic.py
```diff
@@ -3180,6 +3180,7 @@ def __init__(self, base_name, code_version, code,
     :param bool should_cleanup_old_all: whether we should look in the cache dir
       and check all ops if we can delete some old ones which are older than some limit (self._cleanup_time_limit_days)
     :param bool should_cleanup_old_mydir: whether we should delete our op dir before we compile there.
+    :param typing.TextIO log_stream: file stream for print statements, defaults to sys.stdout
```
albertz, line 3183:
```suggestion
    :param typing.TextIO|None log_stream: file stream for print statements
```

PR #348, opened 2020-08-14
SelfAttention: Reduce memory footprint when in loop

Zettelkasten:
When inside a recurrent loop, store a single `prev_k` and `prev_v` instead of a concatenation `prev_kv`. This saves a split-operation of the `prev_kv`-tensor with length (t-1) at time step t into seperate `prev_k` and `prev_v`, which in total had quadratic memory consumption in t. We still have quadratic memory usage even with this change, but it is much better than before.

As a reference: I managed to reduce the peak memory (measured by tfprof) of a 6 Layer Trafo for MT with mostly recurrent decoder layers (besides embeddings) with Batch Size 2350 from around 9000 MB to 6800 MB.
This is still not ideal, as both the in-loop version could have the same memory footprint as the optimized one, however that would be a major change to implement (the fully parallel model only requires 4800 MB).

Note that `test_reclayer_optimize_out_selfatt_left` tests this already.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -5982,8 +5982,9 @@ class SelfAttentionLayer(_ConcatInputLayer):
     if data.time_dim_axis is None:
       # Assume inside RecLayer. See get_rec_initial_extra_outputs.
       total_value_dim = n_out
-      return {"k_left": tf.TensorShape((None, num_heads, None, total_key_dim // num_heads)),
-              "v_left": tf.TensorShape((None, num_heads, None, total_value_dim // num_heads))}
+      return {
+        "k_left": tf.TensorShape((None, num_heads, None, total_key_dim // num_heads)),
+        "v_left": tf.TensorShape((None, num_heads, None, total_value_dim // num_heads))}
     return {}
 
   def post_process_final_rec_vars_outputs(self, rec_vars_outputs, seq_len):
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -5744,18 +5744,26 @@ class SelfAttentionLayer(_ConcatInputLayer):
       if self._rec_previous_layer:
         assert self.input_data.time_dim_axis is None
         assert attention_left_only
-        # (batch,heads,time,kv-dim//heads)
-        prev_kv_left = self._rec_previous_layer.rec_vars_outputs["kv_left"]
+        # (batch,heads,time,k-dim//heads)
+        prev_k_left = self._rec_previous_layer.rec_vars_outputs["k_left"]
+        # (batch,heads,time,v-dim//heads)
+        prev_v_left = self._rec_previous_layer.rec_vars_outputs["v_left"]
       else:
         assert self.input_data.time_dim_axis is not None
         batch_dim = self.input_data.get_batch_dim()
-        prev_kv_left = (
-          RnnCellLayer.get_rec_initial_state_inner(
+        if initial_state is not None:
+          prev_k_left = RnnCellLayer.get_rec_initial_state_inner(
             initial_state=initial_state, name=self.name, rec_layer=self,
-            state_key="kv_left",
-            initial_shape=(batch_dim, num_heads, 0, (total_key_dim + total_value_dim) // num_heads),
-            shape_invariant=(None, num_heads, None, (total_key_dim + total_value_dim) // num_heads))
-          if initial_state is not None else None)
+            state_key="k_left",
+            initial_shape=(batch_dim, num_heads, 0, total_key_dim // num_heads),
+            shape_invariant=(None, num_heads, None, total_key_dim // num_heads))
+          prev_v_left = RnnCellLayer.get_rec_initial_state_inner(
+            initial_state=initial_state, name=self.name, rec_layer=self,
+            state_key="v_left",
+            initial_shape=(batch_dim, num_heads, 0, total_value_dim // num_heads),
+            shape_invariant=(None, num_heads, None, total_value_dim // num_heads))
+        else:
+          prev_k_left, prev_v_left = None, None
     x = self.input_data.placeholder
     if self.input_data.sparse:
       x = tf.nn.embedding_lookup(mat, to_int32_64(x))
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/rec.py
```diff
@@ -5960,7 +5982,8 @@ def get_rec_initial_extra_outputs_shape_invariants(cls, num_heads, total_key_dim
     if data.time_dim_axis is None:
       # Assume inside RecLayer. See get_rec_initial_extra_outputs.
       total_value_dim = n_out
-      return {"kv_left": tf.TensorShape((None, num_heads, None, (total_key_dim + total_value_dim) // num_heads))}
+      return {"k_left": tf.TensorShape((None, num_heads, None, total_key_dim // num_heads)),
+              "v_left": tf.TensorShape((None, num_heads, None, total_value_dim // num_heads))}
```
albertz, line 5986: Do not format like this. Make newline after the `{` and then properly indent.

albertz:
> When inside a recurrent loop, store a single prev_k and prev_v instead of a concatenation prev_kv. This saves a split-operation of the prev_kv-tensor with length (t-1) at time step t into seperate prev_k and prev_v, which in total had quadratic memory consumption in t. We still have quadratic memory usage even with this change

You will always have quadratic memory consumption. I don't understand how this is relevant here?

> but it is much better than before.

I don't quite understand why this should save memory. Can you explain? I would expect that this actually increases the memory (very slightly, because of fragmentation, maybe not relevant).

Also, what is your condition exactly? Are you speaking about training? Then prev_kv should not be used at all, as this should be optimized out of the loop. I.e. this will have no effect in training. So you speak about inference?

Also, despite memory consumption, runtime is also very important (or maybe even more important). Did you measure that?

I think I copied this behavior from Tensor2Tensor originally. Not exactly sure anymore. Anyway, did you check current Tensor2Tensor, and maybe Lingvo code, and maybe Huggingface or so, how they do it?

Zettelkasten:
Sorry, I could've made that more clear. I am talking about training of a modified config, where the self attention cannot be optimized out of the loop. In my case this happened because I added recurrence to the encoder-decoder attention (e.g. by concatenate the attention weights using the attention weights of the previous target time steps). This means encoder-decoder attention of layer 1-6 cannot be optimized out anymore, and thus also self attention of layers 2-6 cannot be optimized out anymore.
Because of this, memory consumption almost doubled, forcing us to decrease the batch size by half, thus slowing down training  by a lot.

Both the current implementation and the one I propose use a [batch,heads,1,total_key_dim]-tensor `k` and [batch,heads,1,total_key_dim] `v`.

The current implementation takes a stored [batch,heads,t-1,total_key_dim+total_value_dim]-tensor `prev_kv` for every time step t. To append the t-th key/value pair, it takes one `tf.concat`-op along feature dim to create a vector `kv` of dim [batch,heads,1,total_key_dim+total_value_dim]. This is concatenated along the time axis resulting in a [batch,heads,t,total_key_dim+total_value_dim]-tensor, and then split again in the feature axis, resulting in a [batch,heads,t,total_key_dim]-tensor and a [batch,heads,t,total_value_dim]-tensor.

My proposal takes a stored [batch,heads,t-1,total_key_dim]-tensor `prev_k` and [batch,heads,t-1,total_value_dim]-tensor `prev_v`, to which we can directly concat the new `k` and `v`, giving tensors of dimension [batch,heads,t,total_key_dim] and [batch,heads,t,total_key_dim].

The difference is that my proposed version saves the memory for the `kv`-concat along the feature-dimension, and - more importantly - the split of `prev_kv` along the feature-dimension. For time step t, my change should save
`batch*heads*(total_key_dim+total_value_dim + (t-1) * (total_key_dim + total_value_dim)`.

Note all those tensors (their hidden states + gradients) have to be stored simultaneously for all time steps when we do a backward pass.

I didn't look at any other Trafo implementations - but as this is not a problem for the vanilla Trafo training where it is parallel anyway, the authors probably didn't think of this.

Also, I did not measure training time, but I can of course do that. But I do not see why this should be slower, in the end, there are less ops.

albertz:
> I am talking about training of a modified config, where the self attention cannot be optimized out of the loop ... because I added recurrence ...

Ah ok, I see.

> The difference is that my proposed version saves the memory for the `kv`-concat along the feature-dimension, and - more importantly - the split of `prev_kv` along the feature-dimension. For time step t, my change should save
> `batch*heads*(total_key_dim+total_value_dim + (t-1) * (total_key_dim + total_value_dim)`.
> Note all those tensors (their hidden states + gradients) have to be stored simultaneously for all time steps when we do a backward pass.

Are you sure this is the case here? I understand that without any optimization, this is true, but I thought that some simple optimization is used here, which could have easily eliminated this. To be sure, you could take a look at the computation graph. If it is stored, you will find some `TensorStack` for these specific values (stored per frame).

It is also possible to do that explicitly via `tf.Defun`, to avoid that intermediate values are stored. I read somewhere that cleverly using `tf.Defun` in a recurrent loop can safe you quite some memory, while sacrificing a bit runtime because you effectively need to recalculate stuff. But in your/this specific case, the runtime overhead of a saved concat or split is probably really irrelevant.

> I didn't look at any other Trafo implementations - but as this is not a problem for the vanilla Trafo training where it is parallel anyway, the authors probably didn't think of this.

Yes, probably true. But anyway, maybe still do that. You often find some interesting stuff nonetheless. More modern implementations, like Hugging Face, or Lingvo (I'm not sure how much Tensor2Tensor is still updated).

> Also, I did not measure training time, but I can of course do that. But I do not see why this should be slower, in the end, there are less ops.

This is not so clear. I agree, I assume that this will probably not have a big effect, or even maybe not a measurable effect. But better measure it anyway, to be sure.

albertz:
@JackTemaki @papar22 @talkhouli @patrick-wilken @curufinwe @ringoreality You are all using Transformer models. Not all of you need to review here, but some of you at least should take a look. Also, maybe some of you have seen other Transformer implementations, and know a bit how it is done elsewhere?

albertz:
@JackTemaki @papar22 @talkhouli @patrick-wilken @curufinwe @ringoreality someone?

JackTemaki:
I have no experience at all with self-attention/transformer models. Just from a quick look at the code it seems more reasonable to keep `k` and `v` separate, as there seems to be no computation on the concatenation of both. So I do not think that this change will make the computation slower, but definitely reduces the number of operations and memory.

albertz:
Ok, then at least someone else with some experience with self-attention / transformers should take a look at this.
@papar22 @talkhouli @patrick-wilken @curufinwe @ringoreality someone?

patrick-wilken:
I also don't have any particular expertise here, but the code looks good to me. Same conclusion as Nick 😬

albertz:
Ok, but at least someone with some experience with self-attention / transformers should take a look at this.
@papar22 @talkhouli @curufinwe @ringoreality?

albertz:
@papar22 @talkhouli @curufinwe @ringoreality Ping? Someone can take a look?

talkhouli (commented):
Unfortunately I also don't have experience with transformer code. The code looks good to me.

albertz:
@Zettelkasten did you change something?

Zettelkasten:
I accidentally pushed the commits from my other PR to this one, but shortly after dropped those changes here again. So no, i didn’t change anything here :D

albertz:
@curufinwe No feedback yet? So I assume ok?
I will merge this in now.

PR #354, opened 2020-08-24
PostfixInTime: Allow postfix to be another layer

Zettelkasten:
This allows to pass a name of a layer as `postfix` of the `postfix_in_time`-Layer.
Currently, this requires the postfix-Layer to not have a time dimension itself. However to implement that, it might be nicer to reuse the code from `PrefixInTimeLayer`. Maybe then it's nicer to add a `ConcatInTime`-Layer that concats all `from`-Arguments or something like that.
Also, this PR adds a test case for `PrevTimeLayer` that also tests its old behaviour.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3909,15 +3909,21 @@ class PostfixInTimeLayer(_ConcatInputLayer):
 
   def __init__(self, postfix=0.0, repeat=1, **kwargs):
     """
-    :param float|int postfix: constant
+    :param float|int|LayerBase postfix: constant or other layer without time axis to use as postfix
     :param int repeat: how often to repeat the postfix
     """
     from returnn.tf.util.basic import DimensionTag
     super(PostfixInTimeLayer, self).__init__(**kwargs)
     self.output = self.input_data.copy(name="%s_output" % self.name)
     assert self.output.time_dim_axis is not None
-    assert isinstance(postfix, (float, int)), "other layer src not yet supported"
-    c = tf.constant(postfix, dtype=self.output.dtype)
+    assert isinstance(postfix, (float, int, LayerBase))
+    if isinstance(postfix, LayerBase):
+      assert not postfix.output.have_time_axis(), 'Postfix layer with time axis not implemented yet'
+      postfix = postfix.output.copy_compatible_to(self.output)
+      assert self.output.time_dim_axis_excluding_batch not in postfix.size_placeholder
+      c = postfix.placeholder
+    else:
+      c = tf.constant(postfix, dtype=self.output.dtype)
     added_shape = [
       ((self.output.batch_shape[i] or tf.shape(self.output.placeholder)[i])
        if (i != self.output.time_dim_axis)
@@ -3950,6 +3956,19 @@ class PostfixInTimeLayer(_ConcatInputLayer):
     """
     # Note: Time seq len is not correct...
     return get_concat_sources_data_template(sources, name="%s_output" % name)
+
+  @classmethod
+  def transform_config_dict(cls, d, network, get_layer):
+    """
+    :param dict[str] d:
+    :param returnn.tf.network.TFNetwork network:
+    :param get_layer:
+    """
+    super(PostfixInTimeLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+    if d.get("postfix", None):
+      postfix = d["postfix"]
+      if isinstance(postfix, str):
+        d["postfix"] = get_layer(postfix)
 
 
 class TimeChunkingLayer(_ConcatInputLayer):
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -2154,6 +2154,41 @@ def test_ResizeLayer_fill_dropout():
       assert_equal([s for s in out[i] if s != fill_value], src_seqs[i])
 
 
+def test_PostfixInTimeLayer():
+  with make_scope() as session:
+    import numpy as np
+    net = TFNetwork(extern_data=ExternData())
+    src = InternalLayer(name="src", network=net, out_type={"dim": 2, "dtype": "int32"})
+    src_seqs = np.array([[[1, 1], [2, 2], [3, 3], [4, 4], [5, 5]], [[6, 6], [7, 7], [8, 8], [0, 0], [0, 0]]])
+    src_seq_lens = [5, 3]
+    src.output.placeholder = tf.constant(src_seqs, dtype=tf.int32)
+    src.output.size_placeholder = {0: tf.constant(src_seq_lens, dtype=tf.int32)}
+
+    static_postfix = -7
+    layer_postfix = InternalLayer(
+      name="postfix", network=net, out_type={"dim": 2, "time_dim_axis": None, "dtype": "int32"})
+    layer_postfix.output.placeholder = tf.constant([[-7, -8], [-9, -10]], dtype=tf.int32)
+
+    for postfix in [static_postfix, layer_postfix]:
+      for repeat in (1, 3):
+        layer = PostfixInTimeLayer(
+          name="postfix_in_time", network=net,
+          sources=[src], postfix=postfix, repeat=repeat,
+          output=PostfixInTimeLayer.get_out_data_from_opts(
+            name="postfix_in_time", network=net, sources=[src], postfix=postfix, repeat=repeat))
+        out, seq_lens = session.run([layer.output.placeholder, layer.output.size_placeholder[0]])
+        print(out)
+        print(seq_lens)
+        assert isinstance(out, numpy.ndarray)
+        assert isinstance(seq_lens, numpy.ndarray)
+        assert out.shape == (2, 5 + repeat, 2)
+        assert all(new_len == src_len + repeat for new_len, src_len in zip(seq_lens, src_seq_lens))
+        assert out[0, src_seq_lens[0] - 1, 0] == src_seqs[0, src_seq_lens[0] - 1, 0]
+        assert out[1, src_seq_lens[1] - 1, 0] == src_seqs[1, src_seq_lens[1] - 1, 0]
+        assert out[0, src_seq_lens[0], 0] == -7
+        assert out[0, src_seq_lens[0] + repeat - 1, 0] == -7
+
+
 def test_DotLayer():
   with make_scope() as session:
     B = 2


returnn/tf/layers/basic.py
```diff
@@ -3909,15 +3909,21 @@ class PostfixInTimeLayer(_ConcatInputLayer):
...
     super(PostfixInTimeLayer, self).__init__(**kwargs)
     self.output = self.input_data.copy(name="%s_output" % self.name)
     assert self.output.time_dim_axis is not None
-    assert isinstance(postfix, (float, int)), "other layer src not yet supported"
-    c = tf.constant(postfix, dtype=self.output.dtype)
+    assert isinstance(postfix, (float, int, LayerBase))
+    if isinstance(postfix, LayerBase):
+      assert not postfix.output.have_time_axis(), 'Postfix layer with time axis not implemented yet'
+      postfix = postfix.output.copy_compatible_to(self.output)
+      assert postfix.size_placeholder.get(self.output.time_dim_axis_excluding_batch, 1) == 1
```
albertz, line 3923: This check is wrong (if it is a tensor, `==` does not work like this). Just check instead:
```suggestion
      assert self.output.time_dim_axis_excluding_batch not in postfix.size_placeholder
```
Or maybe no need to check this at all.

Zettelkasten:
@albertz Thanks. I fixed up the old commit with your change.

albertz (approved):

Zettelkasten:
I think I didn't implement `get_out_data_from_opts` correctly, as I got an error when I tried to do search.
Adding this fixed it for me:


```python
  @classmethod
  def get_out_data_from_opts(cls, name, sources, postfix=0.0, **kwargs):
    """
    :param str name:
    :param int|float|LayerBase postfix:
    :rtype: Data
    """
    # Note: Time seq len is not correct...
    sources_output = [s.output for s in sources]
    if isinstance(postfix, LayerBase):
      sources_output += [postfix.output]
    out_type = Data.get_common_data(sources_output, warnings_out=log.v4).get_kwargs()
    out_type["name"] = "%s_output" % name
    return super(PostfixInTimeLayer, cls).get_out_data_from_opts(
      out_type=out_type, name=name, sources=sources, postfix=postfix, **kwargs)
```

This is inspired by the `CombineLayer` code, but I'm not 100% sure if this is the way it should be. Maybe also set the time seq length correctly?
The entire Data / template construction part of Returnn still is black magic to me ^^

Also: Should I add the postfix to `get_dep_layers()` if it is a layer? I saw that `GenericAttentionLayer` does this with `weights`, but it worked fine without for me.

albertz:
> I think I didn't implement `get_out_data_from_opts` correctly, as I got an error when I tried to do search.

What is the situation exactly? The layer is still outside the recurrent loop, right? Or operates on another time axis. Or on what time axis exactly?

What is the error?

> Adding this fixed it for me:
> 
> ```python
>   @classmethod
>   def get_out_data_from_opts(cls, name, sources, postfix=0.0, **kwargs):
>     """
>     :param str name:
>     :param list[LayerBase] sources:
>     :param int|float|LayerBase postfix:
>     :rtype: Data
>     """
>     # Note: Time seq len is not correct...
>     sources_output = [s.output for s in sources]
>     if isinstance(postfix, LayerBase):
>       sources_output += [postfix.output]
>     out_type = Data.get_common_data(sources_output, warnings_out=log.v4).get_kwargs()
>     out_type["name"] = "%s_output" % name
>     return super(PostfixInTimeLayer, cls).get_out_data_from_opts(
>       out_type=out_type, name=name, sources=sources, postfix=postfix, **kwargs)
> ```
> 
> This is inspired by the `CombineLayer` code, but I'm not 100% sure if this is the way it should be for `PostfixInTimeLayer`. Maybe also set the time seq length correctly?

This is wrong. It has to use `get_concat_sources_data_template`, as this is what the layer does (concatenates multiple inputs).

In your current implementation, you assume that the `postfix` layer has a subset of dimensions of the input. This assumption is fine. But then the old `get_out_data_from_opts` should also already be fine. So I wonder what the error is.

> Also: Should I add the postfix to `get_dep_layers()` if it is a layer? I saw that `GenericAttentionLayer` does this with `weights`, but it worked fine without for me.

Yes, this is important.

Anyway, this should not be discussed on this closed PR. You should open a new PR (e.g. with failing test, or maybe also suggested fix) or issue (describing the problem, and error you get).

---

> The entire Data / template construction part of Returnn still is black magic to me ^^

What exactly is unclear?
Maybe make another separate issue where you ask more specifically which parts of the documentation and/or code are unclear, i.e. where we can improve things.

Zettelkasten:
> What is the situation exactly? The layer is still outside the recurrent loop, right? Or operates on another time axis. Or on what time axis exactly?

I am using the layer inside a rec-layer, but operate on another time-axis.

> What is the error?

The template layer of a postfix_in_time layer did not have a beam:

```
  File "/u/petrick/software/returnn/personal/returnn/tf/layers/rec.py", line 1514, in get_layer
    line: assert (layer.output.beam == layer_template.output.beam and
                  layer_choices.beam_size == layer.output.beam.beam_size == layer_template.output.beam.beam_size), (
            "Layer %r has buggy search choices resolution." % layer,
            self.net.debug_search_choices(layer) or "see search choices debug output")
    locals:
      layer = <local> <PostfixInTimeLayer output/'dec_01_att_value' out_type=Data(shape=(8, None, 64), time_dim_axis=2, beam=SearchBeam(name='output/prev:output', beam_size=12), batch_shape_meta=[B,8,T|'time-with-postfix:output/dec_01_att_value',F|64])>
      layer.output = <local> Data(name='dec_01_att_value_output', shape=(8, None, 64), time_dim_axis=2, beam=SearchBeam(name='output/prev:output', beam_size=12), batch_shape_meta=[B,8,T|'time-with-postfix:output/dec_01_att_value',F|64])
      layer.output.beam = <local> SearchBeam(name='output/prev:output', beam_size=12)
      layer_template = <local> <_TemplateLayer(PostfixInTimeLayer)(:template:postfix_in_time) output/'dec_01_att_value' out_type=Data(shape=(8, None, 64), time_dim_axis=2, batch_shape_meta=[B,8,T|'time:var:extern_data:data',F|64]) (construction stack 'dec_01_att0')>
      layer_template.output = <local> Data(name='dec_01_att_value_output', shape=(8, None, 64), time_dim_axis=2, batch_shape_meta=[B,8,T|'time:var:extern_data:data',F|64])
      layer_template.output.beam = <local> None
      layer_choices = <local> <SearchChoices owner='prev:output' beam_size=12 beam_scores=shaped:(None,None)>
      layer_choices.beam_size = <local> 12
      layer.output.beam.beam_size = <local> 12
      layer_template.output.beam.beam_size = <local> !AttributeError: 'NoneType' object has no attribute 'beam_size'
      self = <local> <_SubnetworkRecCell of <RecLayer 'output' out_type=Data(shape=(None,), dtype='int32', sparse=True, dim=19094, batch_dim_axis=1, beam=SearchBeam(name='output/output', beam_size=12, dependency=SearchBeam(name='output/prev:output', beam_size=12)), batch_shape_meta=[T|?,B])>>
      self.net = <local> <TFNetwork 'root/output:rec-subnet' parent_layer=<RecLayer 'output' out_type=Data(shape=(None,), dtype='int32', sparse=True, dim=19094, batch_dim_axis=1, beam=SearchBeam(name='output/output', beam_size=12, dependency=SearchBeam(name='output/prev:output', beam_size=12)), batch_shape_meta=[T|?,B])>...
      self.net.debug_search_choices = <local> <bound method TFNetwork.debug_search_choices of <TFNetwork 'root/output:rec-subnet' parent_layer=<RecLayer 'output' out_type=Data(shape=(None,), dtype='int32', sparse=True, dim=19094, batch_dim_axis=1, beam=SearchBeam(name='output/output', beam_size=12, dependency=SearchBeam(name='output/prev:out...
AssertionError: ("Layer <PostfixInTimeLayer output/'dec_01_att_value' out_type=Data(shape=(8, None, 64), time_dim_axis=2, beam=SearchBeam(name='output/prev:output', beam_size=12), batch_shape_meta=[B,8,T|'time-with-postfix:output/dec_01_att_value',F|64])> has buggy search choices resolution.", 'see search choices debug output')

```

And this is the layer:
```
 'dec_01_att_value': {   'class': 'postfix_in_time',
  'from': ['base:dec_01_att_value_encoder'],
  'postfix': 'prev:dec_01_att0'},
```

> This is wrong. It has to use `get_concat_sources_data_template`, as this is what the layer does (concatenates multiple inputs).
> 
> In your current implementation, you assume that the `postfix` layer has a subset of dimensions of the input. This assumption is fine. But then the old `get_out_data_from_opts` should also already be fine. So I wonder what the error is.
> 
> > Also: Should I add the postfix to `get_dep_layers()` if it is a layer? I saw that `GenericAttentionLayer` does this with `weights`, but it worked fine without for me.
> 
> Yes, this is important.

Okay, I see. I wil try to change that, test it with my concrete config, and then create a PR if that works for me.
If not, I will try to nail down the problem further and work on a test case. We can move further discussion to that PR then.

> What exactly is unclear?
> Maybe make another separate issue where you ask more specifically which parts of the documentation and/or code are unclear, i.e. where we can improve things.

Yes, that sounds good. We are having a deadline soon so I do not have that much time currently, but I will collect a list an then create an issue.

albertz:
> > What is the situation exactly? The layer is still outside the recurrent loop, right? Or operates on another time axis. Or on what time axis exactly?
> 
> I am using the layer inside a rec-layer, but operate on another time-axis.

Be careful with that! The layer operates on the time axis. But which is the time axis in each case? It might be different depending on whether it is inside the rec loop or outside the loop. You will see that in the stdout (what the time axis is set to). If it is different in each case, you will get incorrect behavior (but without an error!).

> > What is the error?
> 
> The template layer of a postfix_in_time layer did not have a beam: ...

Ah, then the error is clear.
Sth like this should work:
```
  @classmethod
  def get_out_data_from_opts(cls, name, sources, postfix=0.0, **kwargs):
    """
    :param str name:
    :param float|int|LayerBase postfix: constant or other layer without time axis to use as postfix
    :rtype: Data
    """
    # Note: Time seq len is not correct...
    out = get_concat_sources_data_template(sources, name="%s_output" % name)
    if isinstance(postfix, LayerBase):
      out.beam = SearchBeam.get_combined_beam(out.beam, postfix.output.beam)
    return out
```

Zettelkasten:
Thanks, your proposed change works for both training and search.
I also tried to do it with `get_concat_sources_data_template`, but then got a dependency loop.

I will create a PR (also with `get_dep_layers()`).

PR #355, opened 2020-08-25
tf util: add GammatoneFilterbankInitializer

vieting:
I added a `GammatoneFilterbankInitializer` class that can be used to initialize a convolutional layer with the parameters of a gammatone filterbank.

Main references are the paper 
>Schlueter, Ralf, et al. "Gammatone features and feature combination for large vocabulary speech recognition." ICASSP 2007

and the RASR code as indicated in the comments.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -1354,12 +1354,15 @@ class GammatoneFilterbankInitializer(init_ops.Initializer):
   def __call__(self, shape, dtype=None, partition_info=None):
     """
     :param tuple[int] shape:
-    :param tf.DType dtype:
+    :param tf.DType|None dtype:
     :param partition_info:
     :rtype: tf.Tensor
     """
+    import numpy
     fbank = self.gammatone_filterbank.get_gammatone_filterbank()
-    fbank = numpy.reshape(numpy.transpose(fbank), shape).astype(dtype.as_numpy_dtype)
+    fbank = numpy.reshape(fbank, shape)
+    if dtype is not None:
+      fbank = fbank.astype(dtype.as_numpy_dtype)
     return tf.convert_to_tensor(fbank, dtype=dtype)
 
 
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -1351,18 +1351,26 @@ class GammatoneFilterbankInitializer(init_ops.Initializer):
     from returnn.util.sig_proc import GammatoneFilterbank
     self.gammatone_filterbank = GammatoneFilterbank(**kwargs)
 
-  def __call__(self, shape, dtype=None, partition_info=None):
+  def __call__(self, shape, dtype=tf.float32, partition_info=None):
     """
     :param tuple[int] shape:
-    :param tf.DType|None dtype:
+    :param tf.DType dtype:
     :param partition_info:
     :rtype: tf.Tensor
     """
     import numpy
     fbank = self.gammatone_filterbank.get_gammatone_filterbank()
+    # Check whether shape and specified parameters for gammatone filterbank match. Currently, this is just implemented
+    # for the case that the expected shape is (sample_rate * length, num_channels) with possibly additional 1-dim axes
+    # in between.
+    gammatone_shape = (
+      int(self.gammatone_filterbank.sample_rate * self.gammatone_filterbank.length),
+      self.gammatone_filterbank.num_channels
+    )
+    assert shape[0] == gammatone_shape[0], "Expected shape does not match given gammatone filterbank paramerers"
+    assert shape[-1] == gammatone_shape[-1], "Expected shape does not match given gammatone filterbank paramerers"
     fbank = numpy.reshape(fbank, shape)
-    if dtype is not None:
-      fbank = fbank.astype(dtype.as_numpy_dtype)
+    fbank = fbank.astype(dtype.as_numpy_dtype)
     return tf.convert_to_tensor(fbank, dtype=dtype)
 
 
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -9,6 +9,7 @@ import tensorflow as tf
 from tensorflow.python.client import device_lib
 from tensorflow.python.ops import init_ops
 import contextlib
+import numpy as np
 import os
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


returnn/tf/util/basic.py
```diff
@@ -9,6 +9,7 @@
 from tensorflow.python.client import device_lib
 from tensorflow.python.ops import init_ops
 import contextlib
+import numpy as np
```
albertz, line 12: Do not import Numpy like that. Keep consistent to our other code. Use just `import numpy`.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+  The default values are taken from RASR, see https://github.com/rwth-i6/rasr/blob/master/src/Signal/GammaTone.cc
+  :param float x:
+  :param float scaling_constant: A in [1]
+  :param float constant_of_integration: k in [1]
+  :param float slope: a in [1]
+  :return: frequency corresponding to given fractional length
+  :rtype: float
+  """
+  assert 0 <= x <= 1, "fractional length has to be between 0 and 1"
+  return scaling_constant * (np.power(10.0, slope * x) - constant_of_integration)
```
albertz, line 1358: This looks like Numpy only. Then this should not be in `tf/util` but just in `util`.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
   return LoadTxtFileInitializer()
 
 
+def greenwood_function(x, scaling_constant=165.4, constant_of_integration=0.88, slope=2.1):
+  """
+  Greenwood function, convert fractional length to frequency, see
+  Schlueter, Ralf, et al. "Gammatone features and feature combination for large vocabulary speech recognition."
+  ICASSP 2007 and also
+  https://en.wikipedia.org/wiki/Greenwood_function
+  The default values are taken from RASR, see https://github.com/rwth-i6/rasr/blob/master/src/Signal/GammaTone.cc
```
albertz, line 1349: Add a new line after your description, before the first param.

About the default values, is there any reason they are like this? (Also in RASR?)

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+  return scaling_constant * (np.power(10.0, slope * x) - constant_of_integration)
+
+
+def inv_greenwood_function(freq, scaling_constant=165.4, constant_of_integration=0.88, slope=2.1):
+  """
+  Inverse greenwood function, convert frequency to fractional length, see
+  Schlueter, Ralf, et al. "Gammatone features and feature combination for large vocabulary speech recognition."
+  ICASSP 2007 and also
+  https://en.wikipedia.org/wiki/Greenwood_function
+  The default values are taken from RASR, see https://github.com/rwth-i6/rasr/blob/master/src/Signal/GammaTone.cc
```
albertz, line 1367: Another newline here again, and again maybe explain default values.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+  :param float freq:
+  :param float scaling_constant: A in [1]
+  :param float constant_of_integration: k in [1]
+  :param float slope: a in [1]
+  :return: fractional length corresponding to given frequency
+  :rtype: float
+  """
+  return np.log10(freq / scaling_constant + constant_of_integration) / slope
```
albertz, line 1375: This looks like Numpy-only again. Then add this to `util`, not `tf/util`.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+
+
+class GammatoneFilterbankInitializer(init_ops.Initializer):
+  """
+  Initializer for a gammatone filterbank, e.g., to initialize weights of a convolutional layer.
+  Based on
+  [1] Schlueter, Ralf, et al. "Gammatone features and feature combination for large vocabulary speech recognition."
+  ICASSP 2007
+  """
```
albertz, line 1384: Add newline here before `__init__`.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    fbank = np.vstack(fbank)
+    if self.normalization:
+      fbank = self.normalize_filters(fbank)
+    fbank = np.reshape(np.transpose(fbank), shape).astype(dtype.as_numpy_dtype)
+    return tf.convert_to_tensor(fbank, dtype=dtype)
+
+  @staticmethod
+  def center_frequencies(num_channels, freq_max, freq_min):
+    """
+    Determine center frequencies for gammatone filterbank
```
albertz, line 1421: Add newline after the description.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    x_max = inv_greenwood_function(freq_max)
+    x_linear = np.linspace(x_min, x_max, num_channels)
+    freq_center = np.vectorize(greenwood_function)(x_linear)
+    return freq_center
+
+  @staticmethod
+  def bandwidth_by_center_frequency(freq, l_=24.7, q_=9.264491981582191):
+    """
+    Get bandwidth (named B in [1]) by center frequency
+    The default values are taken from RASR, see https://github.com/rwth-i6/rasr/blob/master/src/Signal/GammaTone.cc
```
albertz, line 1438: Newline.
Also reason for these defaults?

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    :param float l_:
+    :param float q_: filter quality factor
+    :return: bandwidth
+    :rtype: float
+    """
+    return l_ * (1 / (l_ * q_) * freq + 1.0)
+
+  def gammatone_impulse_response(self, f_center, length, sample_rate, k=1., n=4, phi=0.):
+    """
+    Compute gammatone impulse response based on [1]
```
albertz, line 1449: Newline.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    :param float f_center: center frequency
+    :param int length: length of finite impulse response in seconds
+    :param int sample_rate: sample rate of audio signal in Hz
+    :param float k: output gain
```
albertz, line 1453: You should not name a variable or argument like this, if the description is short enough such that you can use it directly as the name.
```suggestion
    :param float output_gain:
```

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    :param int n: order of filter
+    :param float phi: phase shift
```
albertz, line 1455:
```suggestion
    :param float phase_shift:
```

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    """
+    center_freqs = self.center_frequencies(self.num_channels, self.freq_max, self.freq_min)
+    fbank = []
+    for freq in center_freqs:
+      fbank.append(self.gammatone_impulse_response(freq, self.length, self.sample_rate))
+    fbank = np.vstack(fbank)
+    if self.normalization:
+      fbank = self.normalize_filters(fbank)
+    fbank = np.reshape(np.transpose(fbank), shape).astype(dtype.as_numpy_dtype)
+    return tf.convert_to_tensor(fbank, dtype=dtype)
```
albertz, line 1416: It looks like most of the initializer is pure Numpy code, and the only relevant TF call happens in this line.

You should move everything to `util`. (Btw, maybe you can create a new file there, `util/numpy.py` or so. Or maybe `util/sig_proc.py`.)

And then, this `GammatoneFilterbankInitializer` in here, should be very simple (like most other initializers), and mostly call into the Numpy generator.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    :rtype: np.array
+    """
+    x_min = inv_greenwood_function(freq_min)
+    x_max = inv_greenwood_function(freq_max)
+    x_linear = np.linspace(x_min, x_max, num_channels)
+    freq_center = np.vectorize(greenwood_function)(x_linear)
+    return freq_center
+
+  @staticmethod
+  def bandwidth_by_center_frequency(freq, l_=24.7, q_=9.264491981582191):
```
albertz, line 1435: Why do you use `l_` or `q_` as name? Why the postfix `_`?

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1340,142 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+    """
+    Get bandwidth (named B in [1]) by center frequency
+    The default values are taken from RASR, see https://github.com/rwth-i6/rasr/blob/master/src/Signal/GammaTone.cc
+    :param freq:
```
albertz, line 1439: What is the type?

albertz (changes requested):
See comments.

returnn/tf/util/basic.py
vieting: Done. Btw, is there a reason that we do it like this? It's quite unusual, isn't it?

returnn/tf/util/basic.py
vieting: They are apparently recommended for this model when applied to human data. Both Ralf's paper and Wikipedia cite the paper by Greenwood which I added in the comment now, but which is not publicly accessible, unfortunately.

returnn/tf/util/basic.py
vieting: I took the names from RASR and kept the postfix because PyCharm issues a warning if a variable is just named `l`. I could rename `q_` into `quality_factor`, but for `l_` I don't know a good name. I checked the paper and couldn't find a descriptive name in there.

returnn/tf/util/basic.py
albertz: I think I have seen both quite often, so i would not say this is unusual. And I think I have seen `import numpy as np` mostly in quite hacky code or scripts only.

returnn/util/sig_proc.py
```diff
@@ -0,0 +1,159 @@
...
+    :rtype: numpy.array
+    """
+    x_min = inv_greenwood_function(freq_min)
+    x_max = inv_greenwood_function(freq_max)
+    x_linear = numpy.linspace(x_min, x_max, num_channels)
+    freq_center = numpy.vectorize(greenwood_function)(x_linear)
+    return freq_center
+
+  @staticmethod
+  def bandwidth_by_center_frequency(freq, l_=24.7, q_=9.264491981582191):
```
albertz, line 110: Why `l_` and `q_` (with `_` postfix)?

returnn/util/sig_proc.py
```diff
@@ -0,0 +1,159 @@
...
+    self.sample_rate = sample_rate
+    self.freq_max = freq_max
+    self.freq_min = freq_min
+    self.normalization = normalization
+
+  def get_gammatone_filterbank(self):
+    """
+    Returns an array with the parameters of the gammatone filterbank
+
+    :return: gammatone filterbank
```
albertz, line 79: What will be the shape? Is it dependent on `num_channels` and `length`?

returnn/tf/util/basic.py
vieting: Even on their own website, I always see `import numpy as np`. Also here
https://numpy.org/doc/stable/user/absolute_beginners.html#how-to-import-numpy

But doesn't really matter, I was just curious.

returnn/util/sig_proc.py
vieting: I took the names from RASR and kept the postfix because PyCharm issues a warning if a variable is just named `l`. I could rename `q_` into `quality_factor`, but for `l_` I don't know a good name. I checked the paper and couldn't find a descriptive name in there.

returnn/util/sig_proc.py
vieting: Yes, it's of shape `(self.length * self.sample_rate, self.num_channels)`. I'll add a comment with the shape

returnn/util/sig_proc.py
albertz: Yes, because PyCharm is right, you should not use `l` and `q`. And using `_` is not really a solution. Name it `a` and `b` or `x` and `y` if you really think there are no good names for these. Or `x` and `quality_factor`. But maybe you have a better idea what this `x` is about.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1339,33 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+  def __call__(self, shape, dtype=None, partition_info=None):
+    """
+    :param tuple[int] shape:
+    :param tf.DType|None dtype:
+    :param partition_info:
+    :rtype: tf.Tensor
+    """
+    import numpy
+    fbank = self.gammatone_filterbank.get_gammatone_filterbank()
+    fbank = numpy.reshape(fbank, shape)
```
albertz, line 1363: Normally, an initializer should be generic for any shape. But in your case, it looks like any other shape than (sample_rate * length, num_channels) this is not possible, right? Then this should rather be a check `assert shape == (...)` instead of doing a reshape here.

returnn/util/sig_proc.py
```diff
@@ -0,0 +1,159 @@
...
+    :param int filter_order: order of filter, named n in [1]
+    :param float phase_shift: phase shift, named phi in [1]
+    :return: gammatone impulse response
+    :rtype: numpy.array
+    """
+    bandwidth = self.bandwidth_by_center_frequency(f_center)  # bandwidth (duration of impulse response)
+    num_samples = int(numpy.floor(sample_rate * length))
+    t = numpy.linspace(1. / sample_rate, length, num_samples)
+    return output_gain * numpy.power(t, filter_order - 1) * numpy.exp(-2 * numpy.pi * bandwidth * t) * (
+      numpy.cos(2 * numpy.pi * f_center * t + phase_shift))
```
albertz, line 144: You should not indent one part of it and not the other. Either all or nothing.
```suggestion
    return (
      output_gain *
      numpy.power(t, filter_order - 1) *
      numpy.exp(-2 * numpy.pi * bandwidth * t) *
      numpy.cos(2 * numpy.pi * f_center * t + phase_shift))
```

returnn/tf/util/basic.py
vieting: Basically that's right. However, in my case e.g. the desired shape has another axis of dimension one which is then added by this reshape. I can still do an assertion but I'm not sure if `sample_rate * length` is always the first and `num_channels` is always the last axis.

returnn/tf/util/basic.py
```diff
@@ -1339,6 +1339,33 @@ def __call__(self, shape, dtype=None, partition_info=None):
...
+  """
+
+  def __init__(self, **kwargs):
+    """
+    :param kwargs: kwargs for GammatoneFilterbank
+    """
+    from returnn.util.sig_proc import GammatoneFilterbank
+    self.gammatone_filterbank = GammatoneFilterbank(**kwargs)
+
+  def __call__(self, shape, dtype=None, partition_info=None):
```
albertz, line 1354: Is this the right default for `dtype`?
What should be returned for this default? I guess a tensor of type float32, right? This would be wrong in your case.

returnn/tf/util/basic.py
albertz: Note that if you reorder it, you would get very wrong results then, and would not even notice.
Adding a 1-dim axis somewhere is always fine, though.
Not sure sure if you want to get clever about the check, or maybe the reshape is fine.

returnn/tf/util/basic.py
vieting: I adapted it from the default dtype in `LoadTxtFileInitializer` but I guess you are right and `dtype=tf.float32` should be used by default.

returnn/tf/util/basic.py
vieting: Yes, reordering would indeed be a problem. I'll add the assertion with first and last axis. Once this becomes an issue we can check it. Better this way then not noticing possible problems.

albertz (approved):

PR #356, opened 2020-08-27
tf engine: added min_seq_length

JackTemaki:


returnn/tf/engine.py
```diff
@@ -1520,6 +1526,7 @@ def train_epoch(self):
         batch_size=self.batch_size,
         max_seqs=self.max_seqs,
         max_seq_length=self.max_seq_length,
+        min_seq_length=self.min_seq_length,
```
albertz, line 1529: It's not used?

JackTemaki: what do you mean?

albertz: Ah sorry, I didn't realized it is already implemented.

PR #360, opened 2020-09-02
Added NormLayer class

vieting:
I added a class `NormLayer` which normalizes over specified axes, e.g. time and/or feature axis.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -625,24 +625,16 @@ class NormLayer(_ConcatInputLayer):
     super(NormLayer, self).__init__(**kwargs)
     assert not self.input_data.sparse
     x = self.input_data.placeholder
-    if scale or bias:
-      if isinstance(param_shape, int):
-        param_shape = [param_shape]
-      elif ((isinstance(param_shape, list) or isinstance(param_shape, tuple)) and
-            all(isinstance(x, int) for x in param_shape)):
-        param_shape = param_shape
-      elif (isinstance(param_shape, str) or
-            ((isinstance(param_shape, list) or isinstance(param_shape, tuple)) and
-            all(isinstance(x, str) for x in param_shape))):
-        param_shape = self.input_data.get_bc_shape(opts={ax: -1 for ax in param_shape})
-      assert len(param_shape) == len(self.input_data.batch_shape), (
-        "Dimension does not match input. Maybe broadcast dimensions are not included explicitly?")
-      assert all(isinstance(x, int) for x in param_shape)
+    assert isinstance(param_shape, str)  # not implemented otherwise yet
+    param_axes = sorted(self.input_data.get_axes_from_description(param_shape))
+    param_shape = [self.input_data.batch_shape[axis] for axis in param_axes]
+    assert all(isinstance(dim, int) for dim in param_shape), "%s: only static param shape allowed" % self
+    param_bc_shape = [dim if axis in param_axes else 1 for (axis, dim) in enumerate(self.input_data.batch_shape)]
     axes = self.input_data.get_axes_from_description(axes)
 
     with self.var_creation_scope():
-      scale_params = self.add_param(tf_compat.v1.get_variable("scale", param_shape, initializer=tf.ones_initializer()))
-      bias_params = self.add_param(tf_compat.v1.get_variable("bias", param_shape, initializer=tf.zeros_initializer()))
+      scale_params = self.add_param(tf_compat.v1.get_variable("scale", param_bc_shape, initializer=tf.ones_initializer()))
+      bias_params = self.add_param(tf_compat.v1.get_variable("bias", param_bc_shape, initializer=tf.zeros_initializer()))
     mean = tf.reduce_mean(x, axis=axes, keepdims=True, name="mean")
     variance = tf.reduce_mean(tf.square(x - mean), axis=axes, keepdims=True, name="variance")
     with tf.name_scope("normalized"):
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -607,6 +607,59 @@ class LayerNormLayer(_ConcatInputLayer):
     return get_concat_sources_data_template(sources, name="%s_output" % name)
 
 
+class NormLayer(_ConcatInputLayer):
+  """
+  Normalize over specified axes, e.g. time and/or feature axis.
+  In case of just feature, this corresponds to layer norm.
+  """
+  layer_class = "norm"
+
+  def __init__(self, axes, param_shape="F", scale=True, bias=True, epsilon=1e-6, **kwargs):
+    """
+    :param str|list[str] axes: axes over which the mean and variance are computed, e.g. "F" or "TF"
+    :param str|list[str]|int|list[int] param_shape: shape of the scale and bias parameters
+    :param bool scale: add trainable scale parameters
+    :param bool bias: add trainable bias parameters
+    :param float epsilon: epsilon for numerical stability
+    """
+    super(NormLayer, self).__init__(**kwargs)
+    assert not self.input_data.sparse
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -607,6 +607,59 @@ def get_out_data_from_opts(cls, sources, name, **kwargs):
...
+        dim = [self.input_data.batch_shape[ax] for ax in self.input_data.get_axes_from_description(param_shape)]
+      assert len(dim) > 0
+      assert all(isinstance(x, int) for x in dim)
+    axes = self.input_data.get_axes_from_description(axes)
+
+    with self.var_creation_scope():
+      scale = 1 if not scale else (
+        self.add_param(tf_compat.v1.get_variable("scale", dim, initializer=tf.ones_initializer())))
+      bias = 0 if not bias else(
+        self.add_param(tf_compat.v1.get_variable("bias", dim, initializer=tf.zeros_initializer())))
```
albertz, line 645: Don't do that. Do not change the type (from `bool` now to `int|tf.Variable`). Create separate Python variables.
But also, do not do this at all. It would be a wasted computation. Just conditionally execute this (scale and bias).

returnn/tf/layers/basic.py
```diff
@@ -607,6 +607,59 @@ def get_out_data_from_opts(cls, sources, name, **kwargs):
...
+    :param float epsilon: epsilon for numerical stability
+    """
+    super(NormLayer, self).__init__(**kwargs)
+    assert not self.input_data.sparse
+    x = self.input_data.placeholder
+    dim = []
+    if scale or bias:
+      if isinstance(param_shape, int):
+        dim = [param_shape]
+      elif isinstance(param_shape, list) and all(isinstance(x, int) for x in param_shape):
```
albertz, line 632: I would also allow tuple.

returnn/tf/layers/basic.py
```diff
@@ -607,6 +607,59 @@ def get_out_data_from_opts(cls, sources, name, **kwargs):
...
+        dim = param_shape
+      elif isinstance(param_shape, str) or (
+           isinstance(param_shape, list) and all(isinstance(x, int) for x in param_shape)):
```
albertz, line 635: The check is wrong. You want that all x are str.

albertz: Also, indentation is wrong.

returnn/tf/layers/basic.py
```diff
@@ -607,6 +607,59 @@ def get_out_data_from_opts(cls, sources, name, **kwargs):
...
+    :param str|list[str] axes: axes over which the mean and variance are computed, e.g. "F" or "TF"
+    :param str|list[str]|int|list[int] param_shape: shape of the scale and bias parameters
+    :param bool scale: add trainable scale parameters
+    :param bool bias: add trainable bias parameters
+    :param float epsilon: epsilon for numerical stability
+    """
+    super(NormLayer, self).__init__(**kwargs)
+    assert not self.input_data.sparse
+    x = self.input_data.placeholder
+    dim = []
```
albertz, line 628: `dim` is a confusing name. Call it `param_shape`. I.e. keep using `param_shape` and just adapt it.

returnn/tf/layers/basic.py
vieting: >You want that all x are str.
Yes, definitely.

>Also, indentation is wrong.
What level of indentation would you prefer? 2 spaces compared to elif is the same as the next line, which is not nice. 4 spaces triggers a PyCharm warning regarding PEP8.

returnn/tf/layers/basic.py
albertz: Also, it's a bit more complicated (for the case that the feature dim is not the last axis, or you have a more complex param shape). You should explicitly add broadcast dimensions (such that number of axes is same as for the input). Sth like this:
```
scale = tf.reshape(scale, [dim if dim in param_shape for i, dim in enumerate(self.input_data.batch_shape)])
```
(This is not 100% correct... To be correct, you would need some more complex code above for `param_shape`.)

returnn/tf/layers/basic.py
albertz: Not really sure on indentation. Even two more (because last line ended with `(`). Or maybe three more (such that it is a multiple of 2 in total).
Or maybe directly use `elif (` and then newline, with 4 or 6 spaces indent.

returnn/tf/layers/basic.py
vieting: Maybe we could also use `self.input_data.get_bc_shape()` here? Something like `self.input_data.get_bc_shape(opts={ax: -1 for ax in param_shape})` but then it would not be possible to specify the shape manually, e.g. as a list of ints.

albertz: I suggested to use `param_shape` as the argument, but in the code, this should better be `param_axes` if it holds the axes. First we should check the user input (of `param_shape`), and then define `param_axes` based on that, and then derive `param_shape` based on that. But this code doesn't need to be too complex. You don't really need to cover all possible cases. Just common cases, or what you actually would use, and other cases can be implemented later.
E.g. you can just write:
```
assert isinstance(param_shape, str)  # not implemented otherwise yet
param_axes = sorted(self.input_data.get_axes_from_description(param_shape))
param_shape = [self.input_data.batch_shape[axis] for axis in param_axes]
param_bc_shape = [dim if axis in param_axes else 1 for (axis, dim) in enumerate(self.input_data.batch_shape)]
```

vieting: I used `self.input_data.get_bc_shape(opts={ax: -1 for ax in param_shape})` now and explicitly specify the shape with broadcast dimension before creating the variable.

vieting: Oh sorry, your comment didn't show up yet. I can do it like that as well of course.

returnn/tf/layers/basic.py
```diff
@@ -625,24 +625,16 @@ def __init__(self, axes, param_shape="F", scale=True, bias=True, epsilon=1e-6, *
...
+    param_shape = [self.input_data.batch_shape[axis] for axis in param_axes]
+    assert all(isinstance(dim, int) for dim in param_shape), "%s: only static param shape allowed" % self
+    param_bc_shape = [dim if axis in param_axes else 1 for (axis, dim) in enumerate(self.input_data.batch_shape)]
     axes = self.input_data.get_axes_from_description(axes)
 
     with self.var_creation_scope():
-      scale_params = self.add_param(tf_compat.v1.get_variable("scale", param_shape, initializer=tf.ones_initializer()))
-      bias_params = self.add_param(tf_compat.v1.get_variable("bias", param_shape, initializer=tf.zeros_initializer()))
+      scale_params = self.add_param(tf_compat.v1.get_variable("scale", param_bc_shape, initializer=tf.ones_initializer()))
+      bias_params = self.add_param(tf_compat.v1.get_variable("bias", param_bc_shape, initializer=tf.zeros_initializer()))
```
albertz, line 637: The param itself should not have the broadcasted shape. It should just use `param_shape`. You add the broadcast dims afterwards (via `tf.reshape`).

albertz: Also singular, `scale_param`, `bias_param`.
Also, this should be conditional, only if `scale`, and only if `bias` (two separate `if`s).

albertz: Actually, you already have the `if` below. Just move this code over there.

albertz:
Now PEP8 warning:
```
  returnn/tf/layers/basic.py:636: WEAK WARNING Pep8CodeStyle: E501: line too long (122 > 120 characters)
  returnn/tf/layers/basic.py:637: WEAK WARNING Pep8CodeStyle: E501: line too long (121 > 120 characters)
```

albertz (approved):

PR #363, opened 2020-09-16
TransposedConvLayer: fix output shape

vieting:
When doing a transposed convolution with strides, the output shape is not configured correctly (note that by default, strides are set to filter size). 

The proposed change adapts the shape of the non-dynamic spatial axes according to the stride. Note that this is only valid for `padding="SAME"`, which is always used here because it is the default in `tf.nn.conv2d_transpose`.


albertz (approved):

albertz:
@JackTemaki no review?

PR #369, opened 2020-10-07
Generalize DotLayer to support inputs without a batch-dim

curufinwe:
Pretty straight forward extension of existing DotLayer code. Comes with a test case for the new behavior.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4162,8 +4162,8 @@ class DotLayer(LayerBase):
     """
     from returnn.tf.util.basic import prod
     super(DotLayer, self).__init__(**kwargs)
-    a_out = self.sources[0].output.copy_as_batch_major()
-    b_out = self.sources[1].output.copy_as_batch_major()
+    a_out = self.sources[0].output.copy()
+    b_out = self.sources[1].output.copy()
     a_reduce_axes = a_out.get_axes_from_description(red1)
     b_reduce_axes = b_out.get_axes_from_description(red2)
     assert a_reduce_axes and b_reduce_axes, "%s: sources %r, red1 %r, red2 %r" % (self, self.sources, red1, red2)
@@ -4275,9 +4275,9 @@ class DotLayer(LayerBase):
     """
     assert len(sources) == 2, "dot-layer %r: needs exactly two sources" % (name,)
     # See __init__.
-    a_out = sources[0].output.copy_as_batch_major()
+    a_out = sources[0].output.copy()
     a_reduce_axes = a_out.get_axes_from_description(red1)
-    b_out = sources[1].output.copy_as_batch_major()
+    b_out = sources[1].output.copy()
     assert not a_out.beam or not b_out.beam or a_out.beam == b_out.beam
     b_reduce_axes = b_out.get_axes_from_description(red2)
     assert a_reduce_axes and b_reduce_axes, "%s: sources %r, red1 %r, red2 %r" % (name, sources, red1, red2)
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -4292,18 +4292,24 @@ def get_out_data_from_opts(cls, name, sources, red1=-1, red2=-2, var1=-2, var2=-
...
-      time_dim_axis = cls._axis2_to_output(
-        b_out.time_dim_axis, b_rem_axes=b_rem_axes, a_var_axes=a_var_axes, b_var_axes=b_var_axes)
-    if time_dim_axis is None and (a_out.time_dim_axis is not None or b_out.time_dim_axis is not None):
-      # We had some time dim axis before and reduced it now.
-      # But maybe there are others, so let's automatically figure out.
-      time_dim_axis = NotSpecified
+
+    def find_axis(a_axis, b_axis):
+      if a_axis is not None:
+        return cls._axis1_to_output(a_axis, a_rem_axes=a_rem_axes, a_var_axes=a_var_axes)
```
albertz, line 4298: The logic slightly changed here. It should not directly return, right? It should check if the output from `_axis1_to_output` is not None, and only return in that case. Otherwise, it should go on with the further code.

returnn/tf/layers/basic.py
```diff
@@ -4292,18 +4292,24 @@ def get_out_data_from_opts(cls, name, sources, red1=-1, red2=-2, var1=-2, var2=-
...
+      if b_axis is not None:
+        return cls._axis2_to_output(b_axis, b_rem_axes=b_rem_axes, a_var_axes=a_var_axes, b_var_axes=b_var_axes)
+      if a_axis is not None or b_axis is not None:
```
albertz, line 4301: This can never be True currently with your change.

returnn/tf/layers/basic.py
```diff
@@ -4292,18 +4292,24 @@ def get_out_data_from_opts(cls, name, sources, red1=-1, red2=-2, var1=-2, var2=-
...
+        # We had some time dim axis before and reduced it now.
+        # But maybe there are others, so let's automatically figure out.
+        return NotSpecified
```
albertz, line 4304: `NotSpecified` would be wrong for the batch-dim-axis. This is only valid for the time-dim-axis.
For batch-dim-axis, if you had it before, you must have it afterwards, so you would need to raise an exception here.

albertz (changes requested):
See comments.
Also, see failing test cases (I assume because of this).

returnn/tf/layers/basic.py
```diff
@@ -4292,18 +4292,24 @@ def get_out_data_from_opts(cls, name, sources, red1=-1, red2=-2, var1=-2, var2=-
...
-      time_dim_axis = cls._axis1_to_output(a_out.time_dim_axis, a_rem_axes=a_rem_axes, a_var_axes=a_var_axes)
-    if time_dim_axis is None and b_out.time_dim_axis is not None:
-      time_dim_axis = cls._axis2_to_output(
-        b_out.time_dim_axis, b_rem_axes=b_rem_axes, a_var_axes=a_var_axes, b_var_axes=b_var_axes)
-    if time_dim_axis is None and (a_out.time_dim_axis is not None or b_out.time_dim_axis is not None):
-      # We had some time dim axis before and reduced it now.
-      # But maybe there are others, so let's automatically figure out.
-      time_dim_axis = NotSpecified
+
+    def find_axis(a_axis, b_axis):
```
albertz, line 4296: This is missing a docstring which defines the types and rtype (keep consistent to other code) (also see failing style test because of that).

albertz (approved):

PR #371, opened 2020-10-09
CombinedDataset: fix crash at end of epoch if sampling_size > num_seq

patrick-wilken:
This is something I overlooked when we decided to use `_expand_dataset_sec_idxs()` for both known and unknown num_seqs. `self.used_num_seqs_per_subset` in the case of known num_seqs is not just the total number of sequences in the subsets, but we have to account for sequence ordering, sampling, etc.
This lead to running into the exception at the bottom of `_expand_dataset_sec_idxs()` at the end of an epoch, however only in case we artificially increase the number of sequences, e.g. by repeat_epoch or sampling more sequences than the total number.

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -922,8 +922,6 @@ class CombinedDataset(CachedDataset2):
       total_num_seqs = None
 
     if total_num_seqs is not None:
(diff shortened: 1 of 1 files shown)


albertz (approved):
Looks ok from a short glance...

PR #373, opened 2020-10-14
tf.network: remove dead code

JackTemaki:
When browsing through the `TFNetwork` code, I found that there two functions are called at no position in the code. I also did a complete search for the function names, which led to no results.


albertz:
Hm, we definitely had that supported at some point. This was the standard/original way for the Theano backend, to define the network like `network = hidden,hidden,hidden`. I wonder where/when this got lost for TF. But as it is lost apparently now, i.e. currently not supported, and no-one was ever complaining, you are right, we can remove this.

Issue #372, opened 2020-10-12
Data is accessed despite using "available_for_inference": False


JackTemaki:
In my current setup, I am using a window function to convert the `data` input into a `padded_data_target`, which is used by the 'output' as target:

```python 
  'windowed_data': {'class': 'window', 'from': ['data'], 'stride': 3, 'window_right': 2, 'window_size': 3},
  'padded_data_target': { 'axes': ['T', 'static:0'],
                          'class': 'merge_dims',
                          'from': ['windowed_data'],
                          'n_out': 80,
                          'register_as_extern_data': 'padded_data_target'},
  'output': { 'class': 'combine',
              'from': ['dec_output', 'post_conv_tf'],
              'kind': 'add',
              'loss': 'mean_l1',
              'loss_scale': 0.25,
              'n_out': 80,
              'target': 'padded_data_target',
               },
```

`data` and `padded_data_target` are correctly labeled with `available_for_inference=False` according to the log:

```python
data: Data(shape=(None, 80), available_for_inference=False, batch_shape_meta=[...]
padded_data_target: Data(shape=(None, 80), available_for_inference=False, batch_shape_meta=[...]
```

Nevertheless, RETURNN adds `data` to the used data keys (somewhere), although is not available.
(This fails at some point in `check_matched_dataset` with some unrelated error.)

When removing the `register_as_extern_data` parameter, everything runs fine.

albertz:
How do you define `extern_data` in your config? (But that seems fine here, `data` in `extern_data` seems correct.)

Why is `data` defined in your dataset, and why does it have dimension 0? (`<MetaDataset 'eval' epoch=1> num_outputs {'data': [0, 2], ...`) (This seems wrong.)

This error looks different to what you described to me. This seems to be some problem in `check_matched_dataset` and your dataset but not (directly) about `register_as_extern_data`. Maybe we should just ignore data keys with `available_for_inference=False` at inference in `check_matched_dataset`.

I assumed that you have an error during the construction of the layers for `register_as_extern_data`.

JackTemaki:
No, I said that it tries to access the data which does not exist (This is why the dimension is 0), which only happens when using `register_as_extern_data`, I never said that there is anything with the layer construction.

The question is why `data` appears in the `network.used_data_keys` variable when having a layer with registered output based on `data` despite being marked as `'available_for_inference': False`
```python
extern_data = {'classes': {'dim': 77, 'shape': (None,), 'sparse': True},
[...]

PR #374, opened 2020-10-14
tf.network: check for data availabibility

JackTemaki:
- Adds a DataNotAvailable Exception
 - Aborts construction stack when the output is flagged
   as available_for_inference=False

This is related to issue #372, and approaches the problem by introducing an Exception that is caught when calling the network construction from "register_as_extern_data". In all other cases the Exception will be raised globally.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -80,7 +80,7 @@ class LayerBase(object):
 
     :param str name:
     :param returnn.tf.network.TFNetwork network:
-    :param Data output:
+    :param Data|None output: Set a specific output instead of using :func:`get_out_data_from_opts`
     :param NotSpecified|None|int n_out: output dim
     :param dict[str] out_type: kwargs for Data class. more explicit than n_out.
     :param list[LayerBase] sources: via self.transform_config_dict()
@@ -1341,6 +1341,16 @@ class InternalLayer(LayerBase):
   """
   This is not supposed to be used by the user.
   It is used by some code to construct a wrapper layer or so.
+  """
+
+
+class DataNotAvailableLayer(InternalLayer):
+  """
+  This is a dummy layer that is created when the output template is flagged "not available for inference".
+  The output template should be passed to the constructor to correctly forward the information
+  in case any dependent output is exported with "register_as_extern_data".
+
+  See :func:`returnn.tf.network._create_layer`
   """
 
 
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -695,6 +695,9 @@ class TFNetwork(object):
           layer_class.__name__, name, layer_desc)
         output_template.sanity_check(ignore_placeholder=True)  # placeholder might be overwritten later
         output_template_special_axes = output_template.get_special_axes_dict()
+        if not output_template.available_for_inference and not self.eval_flag:
+          from returnn.tf.layers.base import DataNotAvailableLayer
+          return DataNotAvailableLayer(name=layer_desc['name'], network=layer_desc['network'], output=output_template)
         layer = layer_class(**layer_desc)
         layer.post_init(layer_desc)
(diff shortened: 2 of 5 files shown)


albertz:
But as I explained in #372, this approach has the problem that there is no real way that you can make `register_as_extern_data` work.

I think the approach via `_TemplateLayer` or `UnavailableDataLayer` is better, because then `register_as_extern_data` works naturally.

JackTemaki:
I did not really get your point in the issue discussion, but maybe I do now. I think what you mean to say is that if you use any kind of dummy layer which only returns the templated output, the templated output will be added to the extern_data and thus being able to give other layers that depend on the registered data the information that the data is not available.

Also, I did not see that you can use the "output" parameter of the `LayerBase` directly to create an empty layer with the specified output.

returnn/tf/network.py
```diff
@@ -695,6 +695,9 @@ def _create_layer(self, name, layer_class, **layer_desc):
           layer_class.__name__, name, layer_desc)
         output_template.sanity_check(ignore_placeholder=True)  # placeholder might be overwritten later
         output_template_special_axes = output_template.get_special_axes_dict()
+        if output_template.available_for_inference is False and not self.eval_flag:
```
albertz, line 698:
```suggestion
        if not output_template.available_for_inference and not self.eval_flag:
```

returnn/tf/network.py
```diff
@@ -695,6 +695,9 @@ def _create_layer(self, name, layer_class, **layer_desc):
...
+          from returnn.tf.layers.base import DataNotAvailableLayer
+          return DataNotAvailableLayer(layer_desc['name'], layer_desc['network'], output_template)
```
albertz, line 700: Always use kwargs.

returnn/tf/layers/base.py
```diff
@@ -1344,6 +1344,14 @@ class InternalLayer(LayerBase):
   """
 
 
+class DataNotAvailableLayer(LayerBase):
```
albertz, line 1347: It should derive from `InternalLayer`.

returnn/tf/layers/base.py
```diff
@@ -1344,6 +1344,14 @@ class InternalLayer(LayerBase):
   """
 
 
+class DataNotAvailableLayer(InternalLayer):
+  """
+  This is dummy layer that is created when the output template is "not available for inference"
+  The output
+
+  """
```
albertz, line 1352: Incomplete documentation?

albertz:
Is this ready? Does it work?

JackTemaki:
Ok if this is the question, then I better add a specific test case.

returnn/tf/layers/base.py
JackTemaki: Ah yes, it seems I stashed a part of it away.

albertz:
I mean, if you verified / tested it for your specific case, we don't necessarily need a test case. But I would not add this without any testing at all.

JackTemaki:
It seems it was a good decision to build a test case anyway. While it does work for my TTS system ("forward"-task), it does not work yet for the test I wrote that is using search.

JackTemaki:
I was not able to get my test case running so far, and I did not even include the "register_as_extern_data" part so far. It seems there is either a bug or I am misunderstanding something in how search should be performed in RETURNN.

albertz:
So probably `classes` ends up (incorrectly) in `used_data_keys`, as this is checked in `check_matched_dataset`. So the question is, why does it end up there. It should not be used (with search).

JackTemaki:
So the `used_data_keys` were correct, but the problem was in the `extra_fetches`. I just blocked by the relevant code parts by checking for `do_eval` now, but we could also add a different parameter. This will at least keep the default behaviour in all cases that do not crash.

albertz (approved):
Looks ok.

PR #375, opened 2020-10-21
custom iterate_seqs for RNN-T chunked training

Spotlight0xff:
Hey,

I've added a test for a custom `Dataset.iterate_seqs` for RNN-T chunked training.
Also, to use this during training in a config, there is an option `custom_iterate_seqs`
which simply monkey-patches the `__class__` of the train_data to use a user-specified `iterate_seqs`.
The patch is only applied to the training dataset, because for the other cases, we don't want chunking anyways.

I know that monkey-patching should be discouraged (and hacky), but since this feature is probably very seldom used,
this may be ok. If anyone has a better suggestion where to implement this, please let me know :)

Also the custom `iterate_seqs` for RNN-T is not very efficient and could very well be improved,
but this should only show how it can be done.

Cheers, André

### DIFF
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -1006,6 +1006,10 @@ class Engine(EngineBase):
     if config.has("eval_datasets"):
       for dataset_name, dataset_opts in config.typed_value("eval_datasets", {}).items():
         self.eval_datasets[dataset_name] = init_dataset(dataset_opts, default_kwargs={"name": dataset_name})
+    if config.has("custom_iterate_seqs"):  # monkey-patch `iterate_seqs`
+      custom_iterate_seqs = config.typed_value("custom_iterate_seqs")
+      assert callable(custom_iterate_seqs)
+      self.train_data.__class__.iterate_seqs = custom_iterate_seqs
     self.start_epoch, self.start_batch = self.get_train_start_epoch_batch(config)
     self.batch_size = config.typed_value('batch_size', 1)
     self.shuffle_batches = config.bool('shuffle_batches', False)
--- a/tests/test_Dataset.py
+++ b/tests/test_Dataset.py
@@ -61,6 +61,215 @@ def test_iterate_seqs_chunking_1():
   assert_equal(seqs[5], (1, 10, 11))
 
 
+def iterate_seqs_rnnt(dataset, chunk_size=None, chunk_step=None, used_data_keys=None,
+                      alignment_key="alignment", classes_key="classes", data_key="data",
+                      blank_index=0):
+  """
+  Takes chunking into consideration, special case for RNN-T.
+  Because RNN-T does not operate time-synchronous
+  (but rather alignment-synchronously), we have to take this into
+  consideration when doing chunking.
+  This implementation currently chunks along the data-dimension,
+  that means that the chunks are constant for "data" (except the last chunk).
+  Also because our alignment is computed on the encoder-axis,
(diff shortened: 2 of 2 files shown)


albertz (changes requested):
Don't monkey patch the class. That will have lots of unintended side effects. You could just set this for this instance:
```
self.train_data.iterate_seqs = lambda *args, **kwargs: custom_iterate_seqs(self.train_data, *args, **kwargs)
```

But even that is hacky and ugly, and will break in some cases, e.g. when you define `custom_iterate_seqs` in some epochs, and later not anymore. Then it would not correctly reset this.

I was originally thinking about just a new argument to `Dataset`, i.e. `custom_iterate_seqs`. And then inside the original `Dataset.iterate_seqs`, you would check for `custom_iterate_seqs` and call that. Although that probably still needs some extra handling in `_maybe_update_config`.

I'm not sure about the cleanest way... I anyway don't like too much that we currently have coupled the dataset + batch preparation logic. In the future, we might (or should) decouple that, but then this might make it difficult to keep old configs compatible in all cases.

In that sense, it is maybe really better to have it globally defined in config, and not as option for `Dataset`. However, it also should be clear that this is for train only, i.e. `train_custom_iterate_seqs`. But then the implementation somehow should be cleaner.

JackTemaki:
I definitely agree that this should not be part of `Dataset`. There is nearly no access to "self" in `Dataset._generate_batches` and `Dataset.iterate_seqs`, except for using sampling with weights. The only dataset that overrides `_generate_batches` is the `ClusteringDataset`. It seems that `iterate_seqs` is not replaced in any dataset. And maybe we can also just pass the weights as "reference" parameter to the new independent `_generate_batches`.

As far as I understand, `tf/engine/batch.py` is only used for the `legacy` data pipeline, and not for the `tf.data` pipeline, so maybe we can move the code here.

We could then give `dataset.generate_batches` an optional parameter to replace the generator, in the situations where this is appropriate.

And for the `ClusteringDataset` we could just override the `generate_batches` function to use a different default generator.

albertz:
@JackTemaki Mostly agree. But we should move the discussion about reorganizing this code part (decoupling dataset + batching) elsewhere (own separate issue). (Btw, you say "legacy", but so far, this is still the default pipeline, and the other is still incomplete, or at least it doesn't support our standard behavior of batching, and also no chunking.)

The question is, can we reasonably add support for `train_custom_iterate_seqs` now, which would also work when we reorganize this code later? And how? Or should we first wait for the reorganization? (Although it's not nice that we cannot have such a simple functionality now...)

For the new `tf.data` pipeline, it's even more tricky to support `train_custom_iterate_seqs`, although also this can work by some wrappers and `tf.py_func`.

Btw, this logic of chunking, or also `train_custom_iterate_seqs`, could in principle be another (optional) step in the dataset pipeline. It could also be implemented as a `Dataset` (like most `tf.data` functions are also mapping `tf.Dataset` -> `tf.Dataset`).

albertz:
I added #376. Let's move the discussion about reorganization the code over there.

albertz:
> I was originally thinking about just a new argument to `Dataset`, i.e. `custom_iterate_seqs`. And then inside the original `Dataset.iterate_seqs`, you would check for `custom_iterate_seqs` and call that. Although that probably still needs some extra handling in `_maybe_update_config`.

This is exactly what I did now in #1075.

I think we can do further cleanup in a separate step. This is anyway not totally clear yet, how we really want it, and needs further discussion in #376.

PR #379, opened 2020-10-24
Warp-RNA TF wrapper

Spotlight0xff:
Hi,

because my custom TF code for the RNA loss implementation is too slow for training,
I wrote a Tensorflow binding to the [warp-rna](https://github.com/1ytic/warp-rna) code (which is for PyTorch).
The binding itself is in my [fork](https://github.com/Spotlight0xff/warp-rna).

Other than that, this is fairly standard and very similar to the `HawkAaronWarpTransducer`.

Cheers, André


albertz (approved):
Looks good.

Issue #380, opened 2020-10-26
Error in RecLayer construction when input is [B|F|T]


JackTemaki:
The network is not constructed correctly when using `"auto_use_channel_first": True` in the network.

The test case is here: https://github.com/rwth-i6/returnn/compare/master...JackTemaki:axes_optim_test
with the network:
```python3
  net_dict = {
    'conv0': {'L2': 1e-07, 'activation': 'relu', 'class': 'conv', 'filter_size': (5,), 'from': ['data'],
                    'n_out': 9, 'padding': 'same', 'auto_use_channel_first': True},
    'lstm0': {'class': 'rec',
                 'direction': 1,
                 'from': ['conv0'],
                 'n_out': 2,
                 'unit': 'nativelstm2',
                 },
    'output': {'class': 'copy', 'from': 'lstm0', 'target': 'data', 'loss': 'mse'}
  }
```

The error log can be found here: https://gist.github.com/JackTemaki/d0f407233bfd6149d41bc79431bca170,
in short:
```python3
File "/home/nick/github/returnn/returnn/tf/util/data.py", line 606, in sanity_check
    line: assert axis != self.batch_dim_axis, "%s: axis %s (%i) must be different from batch_dim_axis (%i)" % (
            self, axis_name, axis, self.batch_dim_axis)
    locals:
      axis = <local> 1
      self = <local> Data(name='lstm0_output', shape=(2, None), batch_dim_axis=1, feature_dim_axis=1, batch_shape_meta=[T|2,B|F,'time:var:extern_data:data'])
      self.batch_dim_axis = <local> 1
      axis_name = <local> 'feature_dim_axis', len = 16
AssertionError: Data(name='lstm0_output', shape=(2, None), batch_dim_axis=1, feature_dim_axis=1, batch_shape_meta=[T|2,B|F,'time:var:extern_data:data']): axis feature_dim_axis (1) must be different from batch_dim_axis (1)
```

I am assuming that there is something wrong in the template construction when the data input is in the shape `[B|F|T]` as returned from the `ConvLayer` in this test case. I will investigate later.

albertz:
If the problem is the input shape [B,F,T], we can construct a simpler test case, where you just feed in such a tensor to a reclayer.

JackTemaki:
The test case was originally intended to search for a different bug. I will add a different test case when submitting the PR with the fix.

JackTemaki:
Fixed with #381

PR #381, opened 2020-10-27
Fix SwapAxesLayer and [B,F,T] input for RecLayer

JackTemaki:
This fixes the problems described in #380 and additional problem in the template output of the `SwapAxesLayer` when swapping the feature axis away from the last position.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -362,23 +362,18 @@ class RecLayer(_ConcatInputLayer):
     """
     from tensorflow.python.util import nest
     source_data = get_concat_sources_data_template(sources) if sources else None
-    if source_data and not source_data.have_time_axis():
-      # We expect to be inside another RecLayer, and should do a single step (like RnnCellLayer).
-      out_time_dim_axis = None
-      out_batch_dim_axis = 0
-    else:
-      out_time_dim_axis = 0
-      out_batch_dim_axis = 1
     n_out = kwargs.get("n_out", NotSpecified)
     out_type = kwargs.get("out_type", None)
     loss = kwargs.get("loss", None)
     deps = list(sources)  # type: typing.List[LayerBase]
     deps += [layer for layer in nest.flatten(initial_state) if isinstance(layer, LayerBase)]
     if out_type or n_out is not NotSpecified or loss:
-      if out_type:
-        assert out_type.get("time_dim_axis", out_time_dim_axis) == out_time_dim_axis
-        assert out_type.get("batch_dim_axis", out_batch_dim_axis) == out_batch_dim_axis
       out = super(RecLayer, cls).get_out_data_from_opts(network=network, sources=sources, **kwargs)
+      if source_data and not source_data.have_time_axis():
+        # We expect to be inside another RecLayer, and should do a single step (like RnnCellLayer).
+        out = out.copy_as_batch_major()  # The output is then [B,F]
+      else:
+        out = out.copy_as_time_batch_major()  # Otherwise the output is always [T,B,F]
     else:
       out = None
     if isinstance(unit, dict):  # subnetwork
@@ -390,15 +385,15 @@ class RecLayer(_ConcatInputLayer):
         assert sub_out.dim == out.dim
         assert sub_out.shape == out.shape
       out = sub_out
-      out_batch_dim_axis = out.batch_dim_axis  # maybe the subnet does not have a batch-axis (rare but valid)
       deps += subnet.get_parent_deps()
     assert out
-    out.time_dim_axis = out_time_dim_axis
-    out.batch_dim_axis = out_batch_dim_axis
     cls._post_init_output(output=out, sources=sources, network=network, **kwargs)
     for dep in deps:
       if dep:
         out.beam = SearchBeam.get_combined_beam(out.beam, dep.output.beam)
+    if out_type:
+      assert out_type.get("time_dim_axis", out.time_dim_axis) == out.time_dim_axis
+      assert out_type.get("batch_dim_axis", out.batch_dim_axis) == out.batch_dim_axis
     return out
 
   def get_absolute_name_scope_prefix(self):
@@ -531,11 +526,7 @@ class RecLayer(_ConcatInputLayer):
     """
     assert self.input_data
     if self.input_data.have_time_axis():
-      x = self.input_data.placeholder  # (batch,time,dim) or (time,batch,dim)
-      if not self.input_data.is_time_major:
-        assert self.input_data.batch_dim_axis == 0
-        assert self.input_data.time_dim_axis == 1
-        x = self.input_data.get_placeholder_as_time_major()  # (time,batch,[dim])
+      x = self.input_data.copy_as_time_batch_major().placeholder
       seq_len = self.input_data.get_sequence_lengths()
       return x, seq_len
     else:  # no time-dim-axis, expect to be inside another RecLayer
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -3604,6 +3604,32 @@ def test_reclayer_enc_time_dim_eval():
     session.run(output_layer.output.placeholder, feed_dict=feed_dict)
 
 
+def test_reclayer_batch_feature_input():
+  """
+  Test if the RecLayer is capable of handling [B,F,T] input
+  """
+  with make_scope() as session:
+    config = Config()
+    config.update({
+      "debug_print_layer_output_template": True,
+      "debug_print_layer_output_shape": True,
+      "extern_data": {
+        "data": {'dim': 5, 'shape': (5, None), 'time_dim_axis': 2, 'feature_dim_axis': 1},
+      },
+      "network": {
+        'output': {
+          'class': 'rec', 'direction': 1, 'from': 'data', 'n_out': 2, 'unit': 'nativelstm2'},
+        }
+      })
+    network = TFNetwork(config=config, train_flag=True)
+    network.construct_from_dict(config.typed_dict["network"])
+    session.run(tf_compat.v1.global_variables_initializer())
+    output_layer = network.get_default_output_layer(must_exist=True)
+    from test_TFNetworkLayer import make_feed_dict
+    feed_dict = make_feed_dict(list(network.extern_data.data.values()))
+    session.run(output_layer.output.placeholder, feed_dict=feed_dict)
+
+
 class TransformerNetwork:
 
   def __init__(self):
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -900,6 +900,17 @@ class Data(object):
     data = data.copy_with_feature_dim_axis(1)
     return data
 
+  def copy_as_time_batch_major(self):
+    """
+    :return: copy of self with batch_dim_axis == 1 and time_dim_axis == 0
+    :rtype: Data
+    """
+    assert self.have_batch_axis() and self.have_time_axis()
+    data = self.copy_as_bt_or_tb_major()
+    if data.time_dim_axis == 1:
+      data = data.copy_move_axis(0, 1)
+    return data
+
   def copy_as_batch_spatial_major(self):
     """
     :return: copy with batch_dim_axis == 0, then all dynamic axes, then any other spatial axes, last feature axis


returnn/tf/layers/rec.py
```diff
@@ -524,10 +527,13 @@ def _get_input(self):
     """
     assert self.input_data
     if self.input_data.have_time_axis():
-      x = self.input_data.placeholder  # (batch,time,dim) or (time,batch,dim)
+      x = self.input_data.placeholder  # (batch,time,dim) or (time,batch,dim) or (batch,dim,time)
```
albertz, line 530: Why do you make it so complicated? Just introduce a function `copy_as_time_batch_major` or so. `get_placeholder_as_time_major` currently assumes (just as this comment here suggests) that it is (B,T,D) or (T,B,D), which is incorrect. But if you just have a function like `copy_as_time_batch_major`, it would always be correct. You should not check all the axes in here and make the code so complicated here. Such a function `copy_as_time_batch_major` should do that logic.

returnn/tf/layers/rec.py
```diff
@@ -393,6 +393,9 @@ def get_out_data_from_opts(cls, network, unit, sources=(), initial_state=None, *
       out_batch_dim_axis = out.batch_dim_axis  # maybe the subnet does not have a batch-axis (rare but valid)
       deps += subnet.get_parent_deps()
     assert out
+    if out.is_batch_feature_major and out.time_dim_axis is not None:
+      # this correctly reorders the axes template when the input is (batch, feature, time)
+      out = out.copy_as_time_major()
```
albertz, line 398: Why is this needed? We anyway overwrite time-dim-axis and batch-dim-axis below. Maybe just also reset feature-dim-axis, like:
```
out.feature_dim_axis = NotSpecified
```

JackTemaki: If I do not do this, I will end up having `batch_dim_axis=1` and `feature_dim_axis=1`. Why can I set the feature dim axis to be `NotSpecified`? What will happen then?

returnn/tf/layers/rec.py
JackTemaki: I did not make anything complicated here, I just extended the assert here. Yes, the whole code could be removed completely if there is a `copy_as_time_batch_major`.

albertz:
Ah, actually, now I see from the test case in `test_cnn_building_block` that it was the intended behavior, that if we swap static axes, the last static axes becomes the new feature dim. (Which was the logical behavior also at the time where feature-dim-axis was always implicitly the last one, and not possible to change.)

There is also code specifically checking `if out.feature_dim_axis_or_unspecified is not NotSpecified`. So it means, if it is not explicitly specified (by default), we keep that behavior, i.e. the default feature dim axis, which is the last static dim (or slightly more complicated, see `_default_feature_dim_axis()`).

I agree, this behavior is maybe un-intuitive. Or should be better documented.

I'm not sure though it is a good idea to change the behavior now. Also, the question is, why would you even need the `SwapAxesLayer` at all then? We emphasize always that the order of the axes should never matter, and RETURNN should always automatically select the optimal order. So that implies that `SwapAxesLayer` should never be needed (with your new behavior). At least for the common cases. It's actually not exactly correct. In cases when you do `MergeDimsLayer`, the order does matter (obviously). Although maybe I would argue that it would be better that the user specifies then at the call to `MergeDimsLayer`, in what order it should be merged.

I.e. the only real use case of it is (and was) to change the tensor such that the feature dim is changed. This was at a time where there was no way to change the feature dim axis. Since we had the possibility to change the feature dim axis, the swap-layer mostly became obsolete.

What is your use case? Why do you need the `SwapAxesLayer`?

Btw, for the test case, I would suggest to simply directly input the data in the tensor format you want to test, and don't make it too complicated (using the swap-axes-layer).

returnn/tf/layers/rec.py
albertz: If it is not specified (which is the default), it will use the default feature dim axis, which is -1 for this case. (See `_default_feature_dim_axis` for the slightly more complicated behavior.)

JackTemaki:
But using the `SwapAxesLayer` was, assuming that it does what seemed intuitive, the most simple solution. I don't know how I can change the input data format by hand (without implementing a new Dataset at least), so I will just add a reinterpret_data layer here for my test case and update the documentation for the `SwapAxesLayer`.

albertz:
> But using the `SwapAxesLayer` was, assuming that it does what seemed intuitive, the most simple solution. I don't know how I can change the input data format by hand (without implementing a new Dataset at least), so I will just add a reinterpret_data layer here for my test case and update the documentation for the `SwapAxesLayer`.

I don't understand. Why do you need a dataset at all? You just define the input (in `extern_data`) to be of that format, and then provide the data in that format for the test case. A test case should ideally be directly testing what it is supposed to test, and not doing it somehow indirectly. You just want to test the behavior of `RecLayer` with some specific input format. Don't make it too complicated by having other stuff in there.

`reinterpret_data` doesn't make any sense here. This will not change the data at all, but just reinterpret it.

JackTemaki:
I found a much better test that exactly does what I need only, `test_reclayer_enc_time_dim_eval`. So I will derive my test case from this. We really should start adding a documentation page for how to write tests, and where to find good examples for which scenarios.

returnn/tf/layers/rec.py
JackTemaki: I do not see any way to avoid this, as the `batch_shape` needs to be corrected as well. I could remove the first part of the if though, as we just need to check for the time axis. Just setting the `feature_dim_axis` is definitely not working.

albertz:
> I found a much better test that exactly does what I need only, `test_reclayer_enc_time_dim_eval`. So I will derive my test case from this.

I don't understand. This looks way more complicated, and also totally unrelated?
The test really should directly test what you described, i.e. RecLayer with the tensor as you described (B,F,T), nothing else. It's just 2 lines or so.

> We really should start adding a documentation page for how to write tests, and where to find good examples for which scenarios.

I don't quite understand. This is not so much RETURNN specific, but just how unit tests should be written in general, or not?

JackTemaki:
> > I found a much better test that exactly does what I need only, `test_reclayer_enc_time_dim_eval`. So I will derive my test case from this.
> 
> I don't understand. This looks way more complicated, and also totally unrelated?
> The test really should just directly test what you described, i.e. RecLayer with the input tensor as you described (B,F,T), nothing else. It's just 2 lines or so.
> **Edit** Ah I assume you just meant the part of creating the input. But this is obvious, or not?
> 
> > We really should start adding a documentation page for how to write tests, and where to find good examples for which scenarios.
> 
> I don't quite understand. This is not so much RETURNN specific, but just how unit tests should be written in general, or not? Also, what is not obvious about this? I would not even know what exactly you want to document here. In general, you should just think, what would be the logical thing to do (really independent from what is there already, what other tests in RETURNN). And when this is clear (and this is almost always quite simple and clear), you can maybe ask specific questions, like "how to easily create some random data for some tensor format" or whatever. If sth is missing for whatever logic you need, we can also add it. But do not really rely too much on what is there already. That should not matter. Just do it in the logical and clean way, and then you can check (ask) if that is already there or not.

But this is difficult to follow in practice. I have rather basic knowledge about RETURNN, and would not be able to write test cases without having reference test cases. Also finding the logical and clean way is often not too easy. It is also a matter of time, the choice is between "doing it in the currently existing way" or "doing it not at all".

JackTemaki:
So how should I proceed with this? Is the test still too much? As for the `out = out.copy_as_time_major()`, I think the whole `get_out_data_from_opts` can be simplified, but for me this is way too risky to do, as there can be mistakes that will not be detected by test cases.

returnn/tf/layers/rec.py
albertz: Then add whatever is missing.
Using `copy_as_time_major` is anyway not really right, and confusing. Time-major just says that time should be first (i.e. time major), it doesn't say anything about the other dims. E.g. it could still be (T,F,B) or whatever else.
In any case, we know exactly how the output should look like, no matter the input. We should maybe not derive from the input at all (except for the check whether it has a time-dim, which is relevant), and definitely not do fancy things like `copy_as_time_major`, which is really even more ambiguous and unclear. Think about the future reader of this code. If the future reader wants to know, how the output looks like. With this code, it is very unclear, the reader need to think about many different possible cases. But this is totally misleading, because it is very clear, and no matter the input, it will always be the same. There are simply only two possible cases:
(T,B,F) if there is a time-dim, otherwise (B,F). Nothing else matters from the input. The code should make this very clear when looking at it.

tests/test_TFNetworkRecLayer.py
```diff
@@ -3604,6 +3604,31 @@ def test_reclayer_enc_time_dim_eval():
     session.run(output_layer.output.placeholder, feed_dict=feed_dict)
 
 
+def test_reclayer_batch_feature_input():
+  """
+  """
```
albertz, line 3609: Why an empty docstring? Either don't add a docstring at all, or add some content.

albertz:
>> Just do it in the logical and clean way, and then you can check (ask) if that is already there or not.
> 
> But this is difficult to follow in practice. I have rather basic knowledge about RETURNN, and would not be able to write test cases without having reference test cases. Also finding the logical and clean way is often not too easy. It is also a matter of time, the choice is between "doing it in the currently existing way" or "doing it not at all".

This is very much the wrong attitude. If you follow this, it would mean that RETURNN will never ever improve the code quality. And also, it means you would always keep reinventing the wheel, make things complicated over and over again, if you just look at other references and copy them.

But it doesn't have to be complicated. Just think about what you actually want to test. And then think about what you need to write that test. And if you don't know how to do some aspect of this, or don't know whether some utility already exists (like `make_feed_dict`), then just ask.

> So how should I proceed with this? Is the test still too much?

No, the test is ok now.

> As for the `out = out.copy_as_time_major()`, I think the whole `get_out_data_from_opts` can be simplified, but for me this is way too risky to do, as there can be mistakes that will not be detected by test cases.

This is again very much the wrong attitude. If it can be simplified, you/we should simplify it, rather than making it even more complex, which will likely add further bugs.
See my inline comment on this.

JackTemaki:
I removed the asserts for the `out_type` now, as this will be overwritten anyway. The only question that remains if we add assertions to check for an existing source if no unit is given, and that there are never more than 3 axes.

albertz:
> I removed the asserts for the `out_type` now, as this will be overwritten anyway.

Why? So you just ignore user configuration? This is wrong. If the user specifically asks for e.g. time-dim-axis = 2 or so, this should result in an error, and not be ignored.

You can move the assert to the end, sth like `assert out_type["time_dim_axis"] == out.time_dim_axis` etc.

> The only question that remains if we add assertions to check for an existing source if no unit is given, and that there are never more than 3 axes.

If a subnet is given, I think there can be more than 3 axes.

JackTemaki:
> > I removed the asserts for the `out_type` now, as this will be overwritten anyway.
> 
> Why? So you just ignore user configuration? This is wrong. If the user specifically asks for e.g. time-dim-axis = 2 or so, this should result in an error, and not be ignored.

Ok, but to be consistent then I would need to check for all user inputs, this would be a lot of asserts. (axis, shape, dims, etc...)

> 
> You can move the assert to the end, sth like `assert out_type["time_dim_axis"] == out.time_dim_axis` etc.
> 
> > The only question that remains if we add assertions to check for an existing source if no unit is given, and that there are never more than 3 axes.
> 
> If a subnet is given, I think there can be more than 3 axes.

Ah yes, I only thought about the direct case.

albertz:
> > > I removed the asserts for the `out_type` now, as this will be overwritten anyway.
> > 
> > Why? So you just ignore user configuration? This is wrong. If the user specifically asks for e.g. time-dim-axis = 2 or so, this should result in an error, and not be ignored.
> 
> Ok, but to be consistent then I would need to check for all user inputs, this would be a lot of asserts. (axis, shape, dims, etc...)

Yea, we can (maybe should) extend it, although then this should be a common utility function, and not be specific to `RecLayer`.

Note that many other layers also have some (partial) similar logic. E.g. in many cases, `n_out` is optional (think of `ActivationLayer`, `DotLayer`, etc), but when you provide it, it will will check whether it is correct, and otherwise doesn't just ignore it, but throw an error. But this is not consistently done in all layers, which is of course not optimal. But that doesn't mean that it should be removed. That rather means it should be completed where it is missing.

Or when you provide a loss + target, it would derive `n_out` from the target. And then also error when it is not consistent to a provided `n_out`.

Or when you provide both `out_type: {"dim": ...}` and a different (non-matching) `n_out`, it should also error. This might also not be done in all cases. But it would be good to do that, and not to silently ignore user inputs.

But anyway, all of that doesn't need to be part of this PR now.

returnn/tf/layers/rec.py
```diff
@@ -362,23 +362,22 @@ def get_out_data_from_opts(cls, network, unit, sources=(), initial_state=None, *
...
     from tensorflow.python.util import nest
     source_data = get_concat_sources_data_template(sources) if sources else None
-    if source_data and not source_data.have_time_axis():
-      # We expect to be inside another RecLayer, and should do a single step (like RnnCellLayer).
-      out_time_dim_axis = None
-      out_batch_dim_axis = 0
-    else:
-      out_time_dim_axis = 0
-      out_batch_dim_axis = 1
+    has_time = False if source_data and not source_data.have_time_axis() else True
```
albertz, line 365: This is very strange to read. Just directly write:
```suggestion
    has_time = not source_data or source_data.have_time_axis()
```

returnn/tf/layers/rec.py
```diff
@@ -362,23 +362,22 @@ def get_out_data_from_opts(cls, network, unit, sources=(), initial_state=None, *
...
     out_type = kwargs.get("out_type", None)
     loss = kwargs.get("loss", None)
     deps = list(sources)  # type: typing.List[LayerBase]
     deps += [layer for layer in nest.flatten(initial_state) if isinstance(layer, LayerBase)]
     if out_type or n_out is not NotSpecified or loss:
       if out_type:
-        assert out_type.get("time_dim_axis", out_time_dim_axis) == out_time_dim_axis
-        assert out_type.get("batch_dim_axis", out_batch_dim_axis) == out_batch_dim_axis
+        assert out_type.get("time_dim_axis", 0 if has_time else None) == 0 if has_time else None
+        assert out_type.get("batch_dim_axis", 1 if has_time else 0) == 1 if has_time else 0
```
albertz, line 374: I think it would be easier to just move that to the end, and then write
```suggestion
        assert out_type.get("time_dim_axis", out.time_dim_axis) == out.time_dim_axis
        assert out_type.get("batch_dim_axis", out.batch_dim_axis) == out.batch_dim_axis
```
That's much simpler and has less assumptions.

JackTemaki: Ah yes, that is definitely better

albertz:
Hm, some strange errors in the tests now?

JackTemaki:
Yes, I accidentally flipped a condition...

albertz (approved):

PR #386, opened 2020-11-06
Wei Zhou's solution for Pickle protocol incompatibility

Marvin84:
During fast Baum-Welch training using the FST from Sprint, which currently relies on python2, there is a Pickle protocol incompatibility. This solution proposed by Wei Zhou solves this problem.


albertz:
Please clean up the commits (rebase).

returnn/sprint/control.py
```diff
@@ -403,11 +403,41 @@ def _get_loss_and_error_signal_via_sprint_callback(self, seg_name, orthography,
     return loss, error_signal
 
   def _send(self, data):
-    Pickler(self.pipe_c2p).dump(data)
+    #Pickler(self.pipe_c2p).dump(data)
```
albertz, line 406: Do not leave commented out code.

returnn/sprint/control.py
```diff
@@ -403,11 +403,41 @@ def _get_loss_and_error_signal_via_sprint_callback(self, seg_name, orthography,
...
+    while read_size < size:
+      data_raw = self.pipe_p2c.read(size - read_size)
+      if len(data_raw) == 0:
+        raise EOFError("%s: expected to read %i bytes but got EOF after %i bytes" % (self, size, read_size))
+      read_size += len(data_raw)
+      stream.write(data_raw)
+    stream.seek(0)
+    try:
+      if PY3:
+        data = Unpickler(stream, encoding="bytes").load()
```
albertz, line 435: It seems a lot of this code was copied from `ExternSprintDataset._read_next_raw`. But why did you remove the comments from there? The comments should be left in.

returnn/sprint/error_signals.py
```diff
@@ -209,12 +209,47 @@ def _build_sprint_args(self):
...
+    while read_size < size:
+      data_raw = p.read(size - read_size)
+      if len(data_raw) == 0:
+        raise EOFError("%s: expected to read %i bytes but got EOF after %i bytes" % (self, size, read_size))
+      read_size += len(data_raw)
+      stream.write(data_raw)
+    stream.seek(0)
+    try:
+      if PY3:
+        data = Unpickler(stream, encoding="bytes").load()
```
albertz, line 242: Same comment as above.

albertz:
It seems this is heavily copied from `ExternSprintDataset._read_next_raw` (but code comments stripped away?), and now you would end up with three (almost? or even exact?) duplicate copies of the same code.

Why? We should not have duplicate code. You should introduce some small helper functions (or class) that you use in these three cases.

albertz (changes requested):
See comments.

returnn/sprint/control.py
Marvin84: > It seems a lot of this code was copied from `ExternSprintDataset._read_next_raw`. But why did you remove the comments from there? The comments should be left in.

I am not familiar with this and I just applied Wei's changes as they are.

Marvin84:
> It seems this is heavily copied from `ExternSprintDataset._read_next_raw` (but code comments stripped away?), and now you would end up with three (almost? or even exact?) duplicate copies of the same code.
> 
> Why? We should not have duplicate code. You should introduce some small helper functions (or class) that you use in these three cases.

I agree with you and that was also my first thought. But, again, this is not my solution. If you think it is not suitable I can cancel the pull request. Unfortunately, I cannot work now on this due to other priorities.

returnn/sprint/control.py
albertz: > I am not familiar with this and I just applied Wei's changes as they are.

That's not really relevant here. What matters is that the code changes are done in a clean and correct way.

albertz:
@Marvin84 or @ZhouW321 (or whoever else feels responsible for this, maybe @curufinwe ), can you clean this up and make ready for merge? We should not keep ugly custom patches around for too long. You should all be able to use the master as-is, and this should be useful for others (if you think what you do is useful for others).

albertz:
Is this still needed, or outdated now? @JackTemaki ?

JackTemaki:
>  this still needed, or outdated now?
I think it is neither. It might be still needed in some cases where you do some mixing of python versions. But I am not doing that so at least for my current setup this is not needed. I still feel that there might be some errors or inconsistencies in the current code, as there is insufficient test coverage for those parts. But I am not familiar enough with this code parts to really tell.

PR #389, opened 2020-11-11
fix sparse intput for SqueezeLayer

### LINKED ISSUE #388: Providing single speaker labels via HDFDataset
This issue is somewhat related to #331, but I still want to know if what I have is a wrong approach or a bug.

I have the following dataset:

```python
{ 'class': 'MetaDataset',
  'data_map': {'classes': ('audio', 'classes'), 'data': ('audio', 'data'), 'speaker_label': ('speaker', 'data')},
  'datasets': { 'audio': {[...]},
                'speaker': { 'class': 'HDFDataset',
                             'files': [ '/u/rossenbach/experiments/librispeech_tts/work/tts/speaker_embeddings/SpeakerLabelHDFFromBliss.xmjLCiVxKiEW/output/speaker_labels.hdf']}},
```

This is my data definition:
```python3
extern_data = {[...],
  'speaker_label': {'dim': 251, 'shape': (None,), 'sparse': True}}
```
And the stored data format is `[B,T|1]`

Unfortunately there is still not a way to store the data without a time axis, so I am doing the same as I did for my speaker vectors before, but now for the single label:
```python3
'speaker_label_notime': {'axis': 'T', 'class': 'squeeze', 'from': ['data:speaker_label']},
```

This results in the following assertion fail (the rest of the trace is not relevant):
```python3
File "/work/asr4/rossenbach/sisyphus_work_folders/librispeech_tts_work/returnn/helpers/rossenbach/checkout/CheckoutRETURNN.WE8k4r4LocNz/output/crnn/b1674b45c0449dc0fef7350bd22abaa5a03036e4/returnn/tf/layers/basic.py", line 3642, in get_out_data_from_opts
    line: return ReduceLayer.get_out_data_from_opts(
    locals:
      ReduceLayer = <global> <class 'returnn.tf.layers.basic.ReduceLayer'>
      ReduceLayer.get_out_data_from_opts = <global> <bound method ReduceLayer.get_out_data_from_opts of <class 'returnn.tf.layers.basic.ReduceLayer'>>
  File "/work/asr4/rossenbach/sisyphus_work_folders/librispeech_tts_work/returnn/helpers/rossenbach/checkout/CheckoutRETURNN.WE8k4r4LocNz/output/crnn/b1674b45c0449dc0fef7350bd22abaa5a03036e4/returnn/tf/layers/basic.py", line 3485, in get_out_data_from_opts
    line: assert not x.sparse
    locals:
      x = <local> Data(name='speaker_label', shape=(None,), dtype='int32', sparse=True, dim=251, available_for_inference=False, batch_shape_meta=[B,T|'time:var:extern_data:speaker_label'])
      x.sparse = <local> True
AssertionError
```
Which is caused by the `SqueezeLayer` using the `get_out_data_from_opts` function from the `ReduceLayer which apparently does not support sparse data.

JackTemaki:
Changes the behaviour of `get_out_data_from_opts` in `ReduceLayer` for `mode=""` to fix issue with the `SqueezeLayer` and sparse data input

Fixes #388

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3637,13 +3637,14 @@ class SqueezeLayer(_ConcatInputLayer):
     input_data = get_concat_sources_data_template(sources)
     if enforce_batch_dim_axis is not None:
       input_data = input_data.copy_with_batch_dim_axis(enforce_batch_dim_axis)
-    if axis == "auto":
-      axis = cls._get_axes(axis, input_data=input_data)
+    axis_indices = cls._get_axes(axis, input_data=input_data)
     if allow_no_op:
-      if not cls._get_axes(axis, input_data=input_data):
+      if not axis_indices:
         return input_data.copy("%s_output" % kwargs["name"])
-    return ReduceLayer.get_out_data_from_opts(
-      axis=axis, keep_dims=False, enforce_batch_dim_axis=enforce_batch_dim_axis, sources=sources, **kwargs)
+    # remove the axis in reversed order
+    for axis_idx in sorted(axis_indices, reverse=True):
+      input_data = input_data.copy_template_excluding_axis(axis_idx)
+    return input_data
 
 
 class StackLayer(LayerBase):


returnn/tf/layers/basic.py
```diff
@@ -3521,18 +3521,26 @@ def get_out_data_from_opts(cls, name, sources, mode="", axes=None, axis=None, ke
       out_size = {
         i if out_batch_dim_axis is None or out_batch_dim_axis >= i else i - 1: size
         for (i, size) in out_size.items()}  # by axis without batch-dim
-    sparse_out = mode.lower().startswith("arg")
-    if sparse_out:
-      out_feature_dim_axis = None
+    if mode == "":
+      # this is for the case that get_out_data_from_opts is called by the SqueezeLayer
```
albertz, line 3525: Why? Don't make the code in `ReduceLayer` complicated for a special case by another layer.

albertz (changes requested):
Don't make `ReduceLayer` too complicated here, esp for a use case, which is not related to `ReduceLayer` at all. Why do you want to keep using that?

Wouldn't `Data.copy_template_excluding_axis` already work?

JackTemaki:
Maybe using a loop with `Data.copy_template_excluding_axis` is ok then.

returnn/tf/layers/basic.py
```diff
@@ -3637,13 +3637,14 @@ def get_out_data_from_opts(cls, axis, enforce_batch_dim_axis=None, allow_no_op=F
...
-    if axis == "auto":
-      axis = cls._get_axes(axis, input_data=input_data)
+    axis_indices = cls._get_axes(axis, input_data=input_data)
     if allow_no_op:
-      if not cls._get_axes(axis, input_data=input_data):
+      if not axis_indices:
         return input_data.copy("%s_output" % kwargs["name"])
-    return ReduceLayer.get_out_data_from_opts(
-      axis=axis, keep_dims=False, enforce_batch_dim_axis=enforce_batch_dim_axis, sources=sources, **kwargs)
+    # remove the axis in order and shift the index for each deleted axis
```
albertz, line 3644: Don't make it so complicated. Just go in reverse order over the sorted axes.

albertz (changes requested):
Can be simpler, see comment.

returnn/tf/layers/basic.py
JackTemaki: Ah yes sure, that makes it simpler.

albertz (approved):

Issue #353, opened 2020-08-18
Simpler data input when using RETURNN as framework 


patrick-wilken:
Scenario:
- training data is provided by user application as a python object
- RETURNN is used as framework
- training is run via `Engine.init_train_from_config()` and `Engine.train()`, therefore a `Dataset` instance is required

What is the best way to create a `Dataset` containing the training data? 

---

As discussed in #329, currently, this can be done by

a) using `StaticDataset`
b) a custom dataset class that the user has to implement in his application

Disadvantages of a):
- sequence ordering is not supported
- on-the-fly data loading is not supported, all data is in memory

Disadvantages of b):
- bloated user code
- more complex, requires familiarity with `Dataset` internals
- basic things like sequence ordering have to be re-implemented

---

What I suggest is to add a new generic dataset, that would read data as a dict `data_key -> numpy_array` per sequence (same format as StaticDataset). Those dicts could be provided

- as a list
- via a generator
- via a callable, which gets `corpus_seq_idx`

A callable would be the most general. A list would not allow for on-the-fly data loading. A generator would not allow for using RETURNN's sequence ordering implementations.

Random thoughts:
All possible data shapes and types should be supported.
For sorting, we could add `get_seq_length` as a second callable for cases in which data loading is more costly than calculating the length.
If the data originates from a python generator, sequence ordering is not possible, partition epoch might be complicated.
This would provide a superset of the features of StaticDataset, so why not replace it?

albertz:
> Scenario:
> 
> * training data is provided by user application as a python object

What exactly does that mean? Where does it come from? How did you load it?

> * RETURNN is used as framework
> * training is run via `Engine.init_train_from_config()` and `Engine.train()`, therefore a `Dataset` instance is required

Just a side node: Is this a real requirement? Or do you just do this because it is currently the simplest way to do it? If that is the sole reason, that's not a good reason. You should always ask: What is the best way. And then, is that possible, or if not, how can we make it possible. You should not just look at existing functions/interfaces and assume they must be used.

> What is the best way to create a `Dataset` containing the training data?
> 
> As discussed in #329, currently, this can be done by

Again, it really doesn't matter what currently can be done. Let's discuss how it rather should be, and then what's missing to get there.

> a custom dataset class that the user has to implement in his application

[...]

PR #390, opened 2020-11-13
Generic inference function

patrick-wilken:
When using RETURNN as framework we have been using `Engine.search_single_string_to_string_seq()` to run search for a given loaded model. This function is tailored for the standard MT use case where in- and output are both a single string. What we now need is support for additional inputs - strings ("sparse") as well as floats ("dense").

This PR proposes a generalized inference function that supports any kind of in- and outputs. What we wanted to keep from `search_single_string_to_string_seq()`:
- vocabulary lookup of string in- and outputs
- removal of padding, i.e. cutting outputs according to sequence length
- output of beam scores

In addition, the new function works for both "search" and "forward".
Also, it groups the batch dim into beams, such that there is really one output per input sequence.

I used @JackTemaki 's MapDataset, but I will adjust to whatever we decide on in #353.

### DIFF
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -2395,6 +2395,154 @@ class Engine(EngineBase):
         raise Exception("invalid output_file_format %r" % output_file_format)
       output_file.close()
 
+  def single_inference(self, input_data, output_layer_names=None):
+    """
+    Generic function that runs the network on the given input data and fetches outputs from the given output layers.
+    It aims to provide a one-line interface to run a model when using RETURNN as a framework.
+
+    We expect that the network has been initialized via self.init_network_from_config(). Depending on whether
+    self.use_search_flag was set, this function will be similar to self.forward() or self.search().
+
+    Input data is given as a list (i.e. batch) of sequences. The data for each sequence is given in a
+    <data_key> -> <data> dictionary. For numeric inputs, we expect <data> to be numpy arrays. For string inputs, we
+    support doing the vocabulary lookup here. This means <data> can be a numpy array of vocabulary indices, or a list of
+    strings which we will convert to vocabulary indices internally. For this, the vocabulary has to be set for the
+    corresponding data key in external_data, e.g. by setting 'extern_data = {"data": {"vocab": "vocab.pkl", ...}}' in
+    the config.
+
+    Outputs will be returned as a list of sequence data, each being a dict <output_layer_name> -> <data>, where <data>
+    is a numpy array. If a vocabulary for the target of the output layer is set in extern_data, the output is
+    interpreted as vocabulary indices and we create an additional entry in the output dicts with the key
+    "<output_layer_name>:string" (e.g. "output:string") containing a list of strings corresponding to the vocabulary
+    indices.
+    If we are doing search and the output of an output layer has a beam, <data> will be a list of numpy arrays instead,
+    corresponding to the hyps in the beam (also for the ":string" data). In addition, there will be an output
+    "<output_layer_name>:beam_scores" which contains the beam scores for this output layer as a list of floats.
+
+    The shapes of all in- and output numpy arrays correspond to what is defined as extern_data in the config.
+
+    :param list[dict]|dict input_data: Batch of data_key -> data dicts, see above.
+      Giving a single sequence (i.e. no list) is also supported. In that case, output will be a single dict too.
+    :param str|list[str]|None output_layer_names: Names of the layers to fetch outputs from. If not set, will read
+      from config "search_output_layer".
+    :return: Outputs of all output layers and possibly additional outputs ":string" and ":beam_scores" for all
+      sequences, see above.
+    :rtype: list[dict]|dict
+    """
+    from returnn.datasets.map import FromListDataset
+
+    is_single_sequence = False
+    if not isinstance(input_data, list):
+      # Support input of a single sequence without a "batch".
+      input_data = [input_data]
+      is_single_sequence = True
+
+    assert input_data and isinstance(input_data[0], dict), (
+      "Expected list of sequence data, each of which being a dict data_key->data.")
+
+    # Get input data types (shape, dim, dtype, etc.) from network.
+    input_data_keys = input_data[0].keys()
+    input_data_types = {}
+    for data_key in input_data_keys:
+      assert data_key in self.network.extern_data.data, "Data key '{}' is unknown to the network.".format(data_key)
+      input_data_types[data_key] = self.network.extern_data.data[data_key]
+
+    # Convert all string data to vocabulary indices.
+    input_data_converted = [{} for _ in input_data]
+    for seq_idx, seq_input_data in enumerate(input_data):
+      assert seq_input_data.keys() == input_data_keys, "All input sequences must provide the same data keys."
+      for data_key in input_data_keys:
+        # For convenient type checking, but also to support any other input that can be converted to an array.
+        data = numpy.array(seq_input_data[data_key])
+        if data.dtype.char in ["S", "U"]:  # string types
+          # TODO: support string arrays with 0 or more than 1 dimension
+          assert data.ndim == 1, "Currently, only 1-D string sequences are supported for automatic vocabulary lookup."
+          vocab = self.network.extern_data.data[data_key].vocab
+          assert vocab, "Got string data but found no vocabulary for {} in extern data.".format(data_key)
+          input_data_converted[seq_idx][data_key] = numpy.array(vocab.get_seq_indices(data.tolist()))
+        else:
+          input_data_converted[seq_idx][data_key] = data
+
+    output_layer_names = output_layer_names or self.config.typed_value("search_output_layer", "output")
+    if not isinstance(output_layer_names, list):
+      output_layer_names = [output_layer_names]
+
+    # Create fetches dict. We fetch the layer outputs, and also sequence lengths and beam scores if applicable.
+    fetches_dict = {}
+    for output_layer_name in output_layer_names:
+      output_layer = self.network.layers[output_layer_name]
+      output_placeholder = output_layer.output.get_placeholder_as_batch_major()
+      fetches_dict[output_layer_name] = output_placeholder
+
+      if output_layer.output.have_time_axis():
+        output_seq_lens = output_layer.output.get_sequence_lengths()
+        fetches_dict[output_layer_name + ":seq_lens"] = output_seq_lens
+
+      search_choices = output_layer.get_search_choices()
+      if search_choices:
+        output_layer_beam_scores = search_choices.beam_scores
+        fetches_dict[output_layer_name + ":beam_scores"] = output_layer_beam_scores
+
+    dataset = FromListDataset(data_list=input_data_converted, num_outputs=input_data_types)
+    dataset.init_seq_order(epoch=1)
+
+    output_batch = self.run_single(dataset=dataset, output_dict=fetches_dict, seq_idx=-1)
+
+    output_data = []
+    for seq_idx in range(len(input_data)):
+      seq_data = {}
+      for output_layer_name in output_layer_names:
+        output_layer = self.network.layers[output_layer_name]
+
+        beam_size = output_layer.output.beam.beam_size if output_layer.output.beam else 1
+        beam_first_idx = seq_idx * beam_size
+        beam_last_idx = (seq_idx + 1) * beam_size
+
+        output = output_batch[output_layer_name][beam_first_idx:beam_last_idx, :]
+
+        # Unpack the batch dim (here: the beam) into a list.
+        output = list(output)
+
+        # Remove padding from outputs by applying sequence length.
+        if output_layer.output.have_time_axis():
+          output_seq_lens = output_batch[output_layer_name + ":seq_lens"][beam_first_idx:beam_last_idx]
+          output = [output_[:length] for output_, length in zip(output, output_seq_lens)]
+          # TODO: handle multiple dynamic axes
+
+        if output_layer_name + ":beam_scores" in fetches_dict:
+          seq_data[output_layer_name + ":beam_scores"] = (
+            output_batch[output_layer_name + ":beam_scores"][seq_idx].tolist())
+
+        # Convert indices to string if vocabulary is found.
+        vocab, string_output = None, None
+        if output_layer.target and output_layer.target in self.network.extern_data.data:
+          vocab = self.network.extern_data.data[output_layer.target].vocab
+
+        if vocab:
+          if self.network.extern_data.data[output_layer.target].ndim != 1:
+            # TODO: support string arrays with 0 or more than 1 dimension
+            print("WARNING: Currently, only 1-D string sequences are supported for automatic vocabulary lookup. "
+                  "Will not create output '{}:string'".format(output_layer_name), file=log.v2)
+          else:
+            string_output = [vocab.get_seq_labels(output_).split() for output_ in output] if vocab else None
+
+        if not output_layer.output.beam:
+          # Remove beam dim if originally there was none.
+          output = output[0]
+          string_output = string_output[0] if string_output else None
+
+        seq_data[output_layer_name] = output
+        if string_output:
+          seq_data[output_layer_name + ":string"] = string_output
+
+      output_data.append(seq_data)
+
+    if is_single_sequence:
+      assert len(output_data) == 1
+      output_data = output_data[0]
+
+    return output_data
+
   def search_single(self, dataset, seq_idx, output_layer_name=None):
     """
     Performs search.
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -2429,7 +2429,7 @@ class Engine(EngineBase):
       sequences, see above.
     :rtype: list[dict]|dict
     """
-    from returnn.datasets.map import FromListDataset
+    from returnn.datasets.map import FromListDataset, MapDatasetWrapper
 
     is_single_sequence = False
     if not isinstance(input_data, list):
@@ -2483,7 +2483,7 @@ class Engine(EngineBase):
         output_layer_beam_scores = search_choices.beam_scores
         fetches_dict[output_layer_name + ":beam_scores"] = output_layer_beam_scores
 
-    dataset = FromListDataset(data_list=input_data_converted, num_outputs=input_data_types)
+    dataset = MapDatasetWrapper(FromListDataset(data_list=input_data_converted, num_outputs=input_data_types))
     dataset.init_seq_order(epoch=1)
 
     output_batch = self.run_single(dataset=dataset, output_dict=fetches_dict, seq_idx=-1)
--- a/tests/test_TFEngine.py
+++ b/tests/test_TFEngine.py
@@ -955,6 +955,183 @@ def test_engine_search_attention():
   check_engine_search_attention()
 
 
+def test_engine_single_inference():
+  import pickle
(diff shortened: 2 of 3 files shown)


returnn/tf/engine.py
```diff
@@ -2395,6 +2395,154 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+    if not isinstance(input_data, list):
+      # Support input of a single sequence without a "batch".
+      input_data = [input_data]
+      is_single_sequence = True
+
+    assert input_data and isinstance(input_data[0], dict), (
+      "Expected list of sequence data, each of which being a dict data_key->data.")
+
+    # Get input data types (shape, dim, dtype, etc.) from network.
+    input_data_keys = input_data[0].keys()
```
patrick-wilken, line 2444: Why does PyCharm think input_data[0] is a list? Because of line 2437?

tests/test_TFEngine.py
```diff
@@ -955,6 +955,183 @@ def test_engine_search_attention():
...
+  config.update({
+    "model": "%s/model" % _get_tmp_dir(),
+    "extern_data": extern_data,
+    "network": {
+      "input_embedding": {"class": "linear", "activation": None, "n_out": 2, "from": "data:string_seq_input"},
+      "sum": {
+        "class": "combine", "kind": "add", "from": ["data:float_input", "data:float_seq_input", "input_embedding"]},
+      "prob": {"class": "softmax", "from": "sum", "target": "string_seq_output"},
+      "output": {
+        "class": "eval", "eval": "tf.argmax(source(0, auto_convert=False), axis=1)", "from": "prob",
```
patrick-wilken, line 1016: We don't have a better version of a "FirstBestLayer", right?

JackTemaki (changes requested):
We should split the wrapper implementation from the dataset interface before proceeding here. I think we already agreed on that in #353. It is better if we do this as a separate pull-request first.

albertz:
Didn't we discussed this already a while ago and came to the conclusion that this is way more complicated than it needs to be, and it would be much simpler if you just do the `session.run` on your side?

albertz (changes requested):
See comment.

patrick-wilken:
I knew this would be controversial :D
But if RETURNN does not offer such a function then I guess this will just be the implementation "on our side". 
We really need a high level access to RETURNN because we will regularly try out models with new in- and outputs and don't want to adjust code dealing with the things handled in this proposed function each time.

To understand your viewpoint: you would also say that `search_single_string_to_string_seq()` is actually too specific to be part of RETURNN?
I understand that what you want to have is the functionality of frameworks like TF and PyTorch and nothing more specific. But with RETURNN we clearly often intentionally are more specific. The output of `++task search` is a text file and not a matrix of ints, for example...

> This solution is very un-intuitive, has lots of very custom decisions,

Can you point out what the custom decisions are? From what I see:
- the exact way in which the in- and output data is structured (we can discuss whether there is a better way)
- whether output vocabulary lookup should happen automatically or explicitly and whether the string output replaces the numeric one or not
- the fact that beam scores can be returned

Apart from that, this function contains code (vocabulary lookup, handling of padding, handling of beams) that pretty much every user would have to write in this way. 

>  It will only make it more complicated to understand any code using that.

When you look at the test code `dummy_data = ...`, `outputs = engine.single_inference(input_data=dummy_data,`, isn't this pretty readable?

> And it will lead to lots and lots of further extensions and modifications because it will never really suite all possible use cases.

Do you have examples? What I can think off is returning losses, so handling the "eval" case as well. But other than that, what types of data would be a problem with the current code?


(I wouldn't say we agreed on using `sess.run()`, I just admitted that it's not unreasonable for the use case I had at that discussion, but here I'm dealing with more complex inputs: strings, beams...)

albertz:
Sorry for the late answer.

> But if RETURNN does not offer such a function then I guess this will just be the implementation "on our side".

You anyway need to have custom code on the calling side, for all the special cases.

> We really need a high level access to RETURNN because we will regularly try out models with new in- and outputs and don't want to adjust code dealing with the things handled in this proposed function each time.

If there is no change to the inputs/outputs, no code (neither on the calling side, nor in RETURNN) needs to be adjusted.

If there are changes, very likely you need to change/extend/modify this function on the RETURNN side (because it's actually not general enough), as well as the code on the calling side.

> To understand your viewpoint: you would also say that `search_single_string_to_string_seq()` is actually too specific to be part of RETURNN?

Not only that it is too specific (although that is also one argument):

I think it is not really helpful. You just add another indirection / abstraction here. The code on the calling side would probably be simpler, shorter, more straight-forward, and easier to understand, if you just would do this directly, instead of having this indirect custom specific function.

> I understand that what you want to have is the functionality of frameworks like TF and PyTorch and nothing more specific. But with RETURNN we clearly often intentionally are more specific. The output of `++task search` is a text file and not a matrix of ints, for example...

No, this is not really my argument.
I'm just arguing against having unnecessary indirection/abstraction which makes things more complicated, and not easier.

> > This solution is very un-intuitive, has lots of very custom decisions,
> 
> Can you point out what the custom decisions are? From what I see:
> 
> * the exact way in which the in- and output data is structured (we can discuss whether there is a better way)
> * whether output vocabulary lookup should happen automatically or explicitly and whether the string output replaces the numeric one or not
> * the fact that beam scores can be returned

This is all way too specific. This doesn't fit in well into a single function. To even describe the input format/API of this function, you need to write lots of documentation, and it is extremely non-intuitive. That's bad design. A function should ideally be clear how to use without even looking at the documentation. It should be straight-forward. It should be simple. It should not need much documentation.

I don't think it's possible to have such a generic function at all. That's already the wrong idea. You can simplify individual points (e.g. have code which parses text into label indices). But trying to squeeze it all into a single function will always lead to very complicated and non-intuitive code (both in RETURNN, and also on the calling side).

This is also all not really related about `Engine`. The engine is there to run a single or multiple epochs over a dataset. If that is not what you want, `Engine` is the wrong object/API for you. Trying to abuse engine for sth else will just be confusing and makes things complicated.

> Apart from that, this function contains code (vocabulary lookup, handling of padding, handling of beams) that pretty much every user would have to write in this way.

This does not belong to `Engine`.

Dataset utils should be part of dataset utils, like vocab lookup.

Padding logic is also already part of dataset utils. It's currently in `FeedDictDataProvider`, which uses a dataset. So maybe that's not optimal if you don't have a dataset. We could abstract this away. Although, once you have multiple sequences, maybe it's just simpler to re-use the dataset API anyway (not reinventing stuff), and just use the existing functions.

If some stuff from the dataset utils are too complicated to use, you could introduce some helper there. E.g. maybe creating `StaticDataset` & `FeedDictDataProvider` to get out the batch is too complicated and can be wrapped into a single function. Or maybe the vocab lookup could be simpler in some way. I don't really know what you think is too complicated there.

Handling of beam, I'm not actually sure what you handle here. You just have lots of custom complicated code to put it into some custom output format. There is no handling at all. Probably on the calling side, you again have code to parse it again out of this custom format.

> > It will only make it more complicated to understand any code using that.
> 
> When you look at the test code `dummy_data = ...`, `outputs = engine.single_inference(input_data=dummy_data,`, isn't this pretty readable?

No, very much not.

- You need ~30 lines of code just to prepare the input data.
- The code to prepare the input data is extremely non-intuitive.
   - To understand how to *write* this code, you would need to carefully read the code and documentation of `single_inference`.
   - To *read* and understand this code, just the same. Just from looking at the example, I don't really understand what would happen there.
- The code in `single_inference` tries to be clever, i.e. automatic vocab lookup if it is a string. That's often bad as well.
- The code to parse out the `outputs` is again ~35 lines of code, just to read it again.
- The code to parse out `outputs` is extremely non-intuitive. Just the same as for the input data. Reading/understanding/writing this code is not possible after carefully reading the code/doc of `single_inference`.

Instead, just doing the same logic directly, would end up in much shorter, and much more straight-forward (to write and read) code. Sth like:
```
in_data = extern_data.data["data"]
in = in_data.vocab.get_seq("Hello world")
feed_dict = {
  in_data.placeholder: [in]
  in_data.size_placeholder[0]: [len(in)]}
out_data = network.layers["output"].output.copy_as_batch_major()
out, out_sizes, out_beam_scores = session.run(
  (out_data.placeholder, out_data.size_placeholder[0], out_data.beam.beam_scores), feed_dict=feed_dict)
```
That's just 8 lines of code (much shorter), and also much simpler, and more straight-forward (to write), and directly clear what it does (i.e. reading is simple, no need to check some external documentation about some custom in/out format).

This is now for one example case. But my point is, once you change the input (e.g. have dense input, or multiple inputs), or the output, anyway your calling code would also change. So you could just directly change the code above (these 8 lines). At no point, you need something complicated like `single_inference`.

If these 8 lines are still too complicated, you can even simplify them further. E.g. maybe you could have a function:
`get_single_seq_feed_dict_for_data_for_sparse_input_lookup(data: Data, text: str) -> Dict[tf.Tensor, numpy.ndarray]` (shorten the function name if you want -- I just wanted to make it clear), and then you could write:
```
feed_dict = get_single_seq_feed_dict_for_data_for_sparse_input_lookup(
  data=extern_data.data["data"], text="Hello world")
```
If you would have multiple inputs, it would look like:
```
feed_dict = {
  **get_single_seq_feed_dict_for_data_for_sparse_input_lookup(
    data=extern_data.data["data1"], text="Hello world 1"),
  **get_single_seq_feed_dict_for_data_for_sparse_input_lookup(
    data=extern_data.data["data2"], text="Hello world 2")}
```
This is very straight-forward. And there is no magic here. You directly understand what it does. Also a year later when you read it again.

> > And it will lead to lots and lots of further extensions and modifications because it will never really suite all possible use cases.
> 
> Do you have examples? What I can think off is returning losses, so handling the "eval" case as well. But other than that, what types of data would be a problem with the current code?

I'm speaking out of experience. Whenever you have such custom code, this is usually the case.
Losses is one example. Or maybe you also want to get other statistics. Maybe you want to apply some other preprocessing or postprocessing. Or maybe your input is actually of type string (dtype="string" in `Data`). Whatever...

patrick-wilken:
I agree that there are probably too many different things happening in the function. I had it split into sub-functions at one point, but having them as members of `Engine` was ugly as well, related to what you said about the function of `Engine`. I could try to move the code to different places as you suggested. But also, in this regard the function is actually quite similar to `search()` or `eval_model()`.

> To even describe the input format/API of this function, you need to write lots of documentation, and it is extremely non-intuitive. That's bad design. 

This data format might not be the simplest one in terms of the data structure itself (depth, number of possible data types), but what I am suggesting is that *semantically* it reflects exactly what is happening. Not only to my model, but pretty much every possible model. It takes information in the form of numeric values and/or labels and predicts outputs in the form of numeric values and/or labels. You run the model for one or more sequences. If your model can give you a list of n-best predictions, the output will be a list.

>The code on the calling side would probably be simpler, shorter, more straight-forward, and easier to understand, if you just would do this directly, instead of having this indirect custom specific function.

Again the example from the test:

```
dummy_data = [
    {"string_seq_input": ["input0", "input1", "input2"], "float_input": numpy.array(1.0),
     "float_seq_input": numpy.array([[0.0, 1.0], [2.0, 3.0], [4.0, 5.0]])},
    {"string_seq_input": ["input3", "input4"], "float_input": -1.0,
     "float_seq_input": [[0.0, -1.0], [-2.0, -3.0]]},
  ]
output_layer_names = ["output", "output_embedding", "mean_output_embedding"]

outputs = engine.single_inference(input_data=dummy_data, output_layer_names=output_layer_names)
```
>You need ~30 lines of code just to prepare the input data.

Do you count the definition of extern_data? Then you have to do that in your example too 😜 But this is independent of the inference function.

> The code to prepare the input data is extremely non-intuitive.

Really? data_key -> data ... Multiple times in a list.
Of course the data keys have no meaning here in this test. And I used a list of floats and a single float here just to make sure the function does not crash in those cases, it could all just be numpy arrays. But that's not what you are talking about I guess. 

> The code to parse out the outputs is again ~35 lines of code, just to read it again.
> The code to parse out outputs is extremely non-intuitive. Just the same as for the input data.

The code has a somewhat weird structure to perform checks depending on the data type of the output, that's not representative of what you would do in an application.

Again the data format is data_key -> data multiple times in a list, which is not hard to parse. Let's e.g. say you want to do LM rescoring of the beam, this would be something like:
```
rescored_output = [do_lm_rescoring(seq["output:string"], seq["output:beam_scores"]) for seq in outputs]
```

Your code snippet is for the case of a single sequence with a single, fixed length, numeric output. Yes, in this case this is all you need. If you consider a batch, sparse outputs, variable length outputs, you would end up writing the code in the inference function. At least that was the idea, to do the "pre- and postprocessing" that basically every user would have to write for those kinds of conditions. I admit that there will always be cases that are not covered by such a function, but the implemented cases are so common in my view that it's worth having them handled inside RETURNN.

But it's not only about reducing the number of lines of code the user has to write, so about having a utility function. I also view the function as an interface to RETURNN. So the idea is, the user has created the config, therefore knows about the network (data keys, layers). He inits RETURNN in his application using the config. And then he simply provides values for the keys and gets values for given layers. He does not have to know about the session, the network class, etc.

I don't insist having this in the master 😄 But I think this is a pretty useful generic, high level interface.

albertz:
> I agree that there are probably too many different things happening in the function.

This is one of the problems, yes. But also, the things happening there are also a huge part due to the unnecessary complexity you add here because of the custom format (for input/output) and custom indirection.

E.g., why not just pass a `StaticDataset` instead of this custom `dict` as input? Then the custom input format handling and parsing already goes away.

> I had it split into sub-functions at one point, but having them as members of `Engine` was ugly as well, related to what you said about the function of `Engine`. I could try to move the code to different places as you suggested.

Yes. But actually, what exactly? Almost all helper functions you need are already there.

> But also, in this regard the function is actually quite similar to `search()` or `eval_model()`...

Yes. It's only complicated because you try to handle so much different cases. Ideally, code should be in a way that it doesn't need to handle lots of different cases.

> > To even describe the input format/API of this function, you need to write lots of documentation, and it is extremely non-intuitive. That's bad design.
> 
> This data format might not be the simplest one in terms of the data structure itself ...

But why do you need the custom data format? That's my point. This is not needed. This is simply an unnecessary abstraction/indirection.

> > The code on the calling side would probably be simpler, shorter, more straight-forward, and easier to understand, if you just would do this directly, instead of having this indirect custom specific function.
> 
> Again the example from the test:
> 
> ```
> dummy_data = [
>     {"string_seq_input": ["input0", "input1", "input2"], "float_input": numpy.array(1.0),
>      "float_seq_input": numpy.array([[0.0, 1.0], [2.0, 3.0], [4.0, 5.0]])},
>     {"string_seq_input": ["input3", "input4"], "float_input": -1.0,
>      "float_seq_input": [[0.0, -1.0], [-2.0, -3.0]]},
>   ]
> output_layer_names = ["output", "output_embedding", "mean_output_embedding"]
> 
> outputs = engine.single_inference(input_data=dummy_data, output_layer_names=output_layer_names)
> ```

Yea, that's non-straightforward. It's totally unclear what the specific logic and handling would be. It's also unclear what this would return. And there is no point in that. This is already more complicated, and more unclear, than what I showed above, where you simply directly use `session.run`.

> > The code to prepare the input data is extremely non-intuitive.
> 
> Really? data_key -> data ... Multiple times in a list.

Yea, it's unclear how exactly the data should be represented, or what extra logic might be applied in what cases.

> > The code to parse out the outputs is again ~35 lines of code, just to read it again.
> > The code to parse out outputs is extremely non-intuitive. Just the same as for the input data.
> 
> The code has a somewhat weird structure to perform checks depending on the data type of the output, that's not representative of what you would do in an application.

That's not what I mean. It's a custom data format you introduce here, and thus you need custom code to read/parse that again. This is an indirection which you would not need, which you purely artificially introduced here.

> Again the data format is data_key -> data multiple times in a list, which is not hard to parse. Let's e.g. say you want to do LM rescoring of the beam, this would be something like:
> 
> ```
> rescored_output = [do_lm_rescoring(seq["output:string"], seq["output:beam_scores"]) for seq in outputs]
> ```

This is non-straightforward and custom. Why do you really need *any* code at all to parse it? This just makes it more complicated. You simply don't need that code if you simply directly get exactly what you want out of the network, and not have this indirection with some custom input/output data format.

> Your code snippet is for the case of a single sequence with a single, fixed length, numeric output.

No, it's variable length, it also gets the sequence length. But anyway, this is an example, for the case you want to get out exactly that. That's the point I'm trying to make. For whatever use case you might have, you can very easily directly get out what you want. There is no reason for a custom indirection. You can very straight-forwardly call `session.run` directly.

> Yes, in this case this is all you need.

Yes, that's the point. Why do you want to make it so complicated? It's not needed.

You simply can directly write the code to directly pass in and get out exactly what you need. This is very simple, very straight-forward, very clear.

> At least that was the idea, to do the "pre- and postprocessing" that basically every user would have to write for those kinds of conditions.

What pre-/postprocessing do you mean? There is no pre/postprocessing. (Only in your case, because you now need to prepare the custom input format, and then parse out the custom output format. In my case nothing is needed.)

> I admit that there will always be cases that are not covered by such a function, but the implemented cases are so common in my view that it's worth having them handled inside RETURNN.

The point I'm trying to make is, this will not simplify anything, but just make it more complicated.
For any possible use case, it would always be simpler to directly call `session.run`.
With such a generic inference function, it would only be more complicated, because you would need extra code to prepare the input, and parse out the output.
 
> But it's not only about reducing the number of lines of code the user has to write, so about having a utility function. I also view the function as an interface to RETURNN. So the idea is, the user has created the config, therefore knows about the network (data keys, layers). He inits RETURNN in his application using the config. And then he simply provides values for the keys and gets values for given layers. He does not have to know about the session, the network class, etc.

You are simply shifting things. Now instead he need to know about the custom data formats, and all the details of this generic inference function. This is not simpler. This is purely shifted. And this is actually more complicated, because in most cases, e.g. to debug things, the user anyway should understand the basics of TF and also how RETURNN works internally. So you just made it more complicated.

Also, with auto-completion, getting the output layer, its output, etc, whatever you actually need, is actually simpler than having to read and understand some custom data format, where auto-completion would not work.

> I don't insist on having this in the master  But I think this is a pretty useful generic, high level interface.

I still don't exactly understand your use case. Where/how do you need this? For some custom scripts, research code, prototyping, writing test cases, etc, you don't need it to be so generic. You very exactly know what your input is, what your output is. For this, it's just more complicated to use such a function. It would be much better to directly call `session.run`. Simpler, more clear, more straight-forward, can more easily be adapted to any special case you need.

patrick-wilken:
I'm closing this as we decided to put this in a private repo.

Nevertheless, I still think the code might be useful for others.

> I still don't exactly understand your use case. Where/how do you need this?

I don't know yet. Potentially for inference of any arbitrary model we train using RETURNN. That's the point.
Currently we already use it for our MT models, with an arbitrary number of sparse in-/outputs and optionally fetching things like word-level scores, and for quality estimation models, which have an arbitrary number of numeric in-/outputs.

> For some custom scripts, research code, prototyping, writing test cases, etc, you don't need it to be so generic. You very exactly know what your input is, what your output is. For this, it's just more complicated to use such a function. It would be much better to directly call session.run. Simpler, more clear, more straight-forward, can more easily be adapted to any special case you need.

Of course I could write simpler code in each specific case. But I would have to write it, every time. We would have code for vocabulary lookup, padding removal, getting the output layers, creating the fetches dict etc. in our MT tool, in our quality estimation tool, and in every future tool that calls some RETURNN model.
Now we just have to initialize the RETURNN engine and then use an interface based on pure python objects.

Of course you could make different design choices for such a function, but we need it in some form.

PR #394, opened 2020-11-25
DotLayer: Fix size_placeholder bug

Zettelkasten:
When the sources of a DotLayer have different batch dim axes, the key
in the size_placeholder has to inferred according to the new batch dim
of the DotLayer, and not the source layer this axis was originally from.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4502,7 +4502,7 @@ class DotLayer(LayerBase):
     def get_batch_axis_excluding_batch(axis):
       """
       :param int axis:
-      :return int:
+      :rtype: int
       """
       if batch_dim_axis is None:
         return axis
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -4499,19 +4499,32 @@ def find_axis(a_axis, b_axis):
...
 
+    def get_batch_axis_excluding_batch(axis):
+      """
+      :param int axis:
+      :return int:
+      """
+      if batch_dim_axis is None:
+        return axis
+      if axis == batch_dim_axis:
+        return None
```
albertz, line 4510: This is a case which should not happen here, right? (Also the docstring would be wrong then. If this is allowed, the return type would be `int|None`.)
```suggestion
      assert axis != batch_dim_axis
```

returnn/tf/layers/basic.py
```diff
@@ -4499,19 +4499,32 @@ def find_axis(a_axis, b_axis):
     if not b_var_dims and add_var2_if_empty:
       b_var_dims.append(1)
 
+    def get_batch_axis_excluding_batch(axis):
+      """
+      :param int axis:
+      :return int:
```
albertz, line 4505: You probably meant:
```suggestion
      :rtype: int
```

PR #395, opened 2020-11-25
Fix hierarchical paths to sub-layers in RecLayer

Zettelkasten:


albertz (approved):

albertz:
@patrick-wilken maybe also take a look, but I will already merge now, as all tests pass, and this looks ok.

patrick-wilken:
@Zettelkasten, for which network/layer dependencies did you need this fix? Would help me understand.
Because my gut feeling is that `construct_layer()` in line 3150 should already cover what you fixed here. Or was this to support sublayers in combination with prev: or base:?

@albertz:
While checking this I noticed an issue with 8f16741f6. It brakes a variant of the test case `test_TFEngine.test_attention_two_targets()` where you remove the dependency on "orth_embed_1" in
```
"orth_embed": {"class": "copy", "from": ["orth_embed_0", "orth_embed_1"]]},
``` 

Before 8f16741f6, root and sublayer were always either both moved out or both in the loop. If moved out, `construct_layer()` apparently was able to handle the sublayer, i.e. the "/" in the name. (Maybe because the order was always right by chance, I'm not sure yet...)
After 8f16741f6, the altered test case breaks because `output` is in the loop, but `output/out_1` is moved outside. `root_layer` in `construct_layer()` in this case is an `InternalLayer` (the one after created in beam resolution) which returns None for `get_sub_layer()` and then we run into LayerNotFound in line 631.

So do you think we need the functionality to optimize only sublayers out? The graph for root and sublayers are both created in the `__init__()` of the root layer, so is it even possible to separate them?

albertz:
> @Zettelkasten, for which network/layer dependencies did you need this fix? Would help me understand.
> Because my gut feeling is that `construct_layer()` in line 3150 should already cover what you fixed here. Or was this to support sublayers in combination with prev: or base:?

This was for some `SplitLayer` example. Maybe we should add a test case? @Zettelkasten can you maybe add one (which failed before this fix, and now works -- this should be independent from #391).

> @albertz:
> While checking this I noticed an issue with 8f16741. It brakes a variant of the test case `test_TFEngine.test_attention_two_targets()` where you remove the dependency on "orth_embed_1" in
> ...
> Before 8f16741 ...
> After 8f16741 , the altered test case breaks because `output` is in the loop, but `output/out_1` is moved outside. `root_layer` in `construct_layer()` in this case is an `InternalLayer` (the one after created in beam resolution) which returns None for `get_sub_layer()` and then we run into LayerNotFound in line 631.

I don't exactly understand. No test case is failing currently?

> So do you think we need the functionality to optimize only sublayers out? The graph for root and sublayers are both created in the `__init__()` of the root layer, so is it even possible to separate them?

Not sure, on both questions. But I haven't really followed your argumentation.

Is this a problem you actually have? Or are you just hypothesizing?
If this is an actual problem, can you add a new reduced test case which demonstrates the problem?
Add this in a new branch, for a new PR.
Then we can also discuss further (either in the PR, or open a new issue about this).

If this is not really a problem for any of our models (and not a problem of any current existing test case), maybe we can accept it as it is for now?

patrick-wilken:
Ah, this commit would also fix the case of both root and sublayer being input layers moved out (and requested via extra_output_layers) which I never had. So yes, this commit looks good! At least I see that it could fix some cases and don't see how it could introduce issues.

Zettelkasten:
@patrick-wilken As Albert already mentioned, I had this issue when accessing the output of a SplitLayer inside a RecLayer. I added a simple test case for this too (which failed before this PR), in #399.

I don't really know if my PR was the proper way of fixing this, I really mostly assumed that that would do the trick without being that familiar with it :D Feel free to change it anyway.

PR #396, opened 2020-11-25
Repeat layer

JackTemaki:
This `RepeatLayer` extends `tf.repeat` to work with independent repetitions along the batch axis. It should in theory work on any axis, but I only wrote test cases for BTF, TBF, and BFT so far, as those are the most likely application cases.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2685,6 +2685,100 @@ class ExpandDimsLayer(_ConcatInputLayer):
     return data
 
 
+class RepeatLayer(_ConcatInputLayer):
+  """
+  A wrapper around tf.repeat, but supports an additional batch axis for the durations
+  The sum of the repetitions has to be non-zero for each sequence in the batch.
+
+  This layer can only be used with Tensorflow 1.15.0 or newer.
+  """
+  layer_class = "repeat"
+
+  def __init__(self, repetitions, axis="T", **kwargs):
+    """
+    :param LayerBase repetitions: number of repetitions for each sequence and position in target axis, [B,T] or [T,B]
+    :param str axis: (dynamic) axis for repetition (currently only time axis is supported)
+    """
+    from returnn.tf.util.data import DimensionTag
+    super(RepeatLayer, self).__init__(**kwargs)
+    self.repetitions = repetitions
+    repetitions_data = self.repetitions.output
+    input_data = self.input_data
+    assert self.input_data and self.input_data.have_batch_axis()
+    input_data = input_data.copy_as_batch_major()
+    original_axis = input_data.get_axis_from_description(axis, allow_int=False)
+    input_data = input_data.copy_move_axis(original_axis, 1)
+    repetitions_data = repetitions_data.copy_as_batch_major()
+    assert repetitions_data.dtype in ["int32", "int64"]
+    assert repetitions_data.have_batch_axis() and repetitions_data.batch_ndim == 2, (
+      "RepeatLayer only accepts repetitions of the shape (B, None|S) or (None|S, B)")
+    assert input_data.shape[0] == repetitions_data.shape[0], (
+      "Axis mismatch between input (%i) and repetitions (%i)" %
+      (input_data.shape[0], repetitions_data.shape[0]))
+    # pad the target axis
+    paddings = [(0, 1) if i == 1 else (0, 0) for i in range(input_data.batch_ndim)]
+    padded_data = tf.pad(input_data.placeholder, paddings)  # [B, T+1, ...]
+    # those are the sequence lengths after expansion
+    target_seq_len = tf.reduce_sum(repetitions_data.placeholder, axis=1)  # [B], the new size_placeholder for 'T'
+    # maximum sequence length after expansion
+    max_duration = tf.reduce_max(target_seq_len)  # [] == T'
+    # the new padding is the difference of the maximum to the new target
+    target_padding_steps = tf.expand_dims(max_duration - target_seq_len, 1)  # [B, 1]
+    # add the repetitions for the artificial padding position
+    target_repetitions = tf.concat([repetitions_data.placeholder, target_padding_steps], axis=1)  # [B, T+1]
+    # flatten batch and time
+    shape = tf_util.get_shape(padded_data)
+    flat_shape = [shape[0] * shape[1]] + shape[2:]
+    reshaped_data = tf.reshape(padded_data, flat_shape)  # [B * T+1, F]
+    reshaped_repetitions = tf.reshape(target_repetitions, (shape[0] * shape[1],))  # [B * T+1]
+    # run the repetition
+    repeated_data = tf.repeat(reshaped_data, reshaped_repetitions, axis=0)  # [B * T', ...]
+    # unflatten the output
+    target_shape = [shape[0], max_duration] + shape[2:]
+    self.output.placeholder = tf.reshape(repeated_data, target_shape)  # [B, T', ...]
+    # copy (optional) additional size placeholders
+    self.output.size_placeholder = {k: v for k, v in input_data.size_placeholder.items() if k != 0}
+    # set the size_placeholder of the target axis
+    self.output.size_placeholder[0] = target_seq_len
+    tag = DimensionTag(
+      description="repeated:%s" % self.get_absolute_name(),
+      kind=DimensionTag.Types.Spatial)
+    tag.set_tag_on_size_tensor(target_seq_len)
+
+  def get_dep_layers(self):
+    """
+    :rtype: list[LayerBase]
+    """
+    deps = super(RepeatLayer, self).get_dep_layers()
+    deps.append(self.repetitions)
+    return deps
+
+  @classmethod
+  def transform_config_dict(cls, d, network, get_layer):
+    """
+    :param dict[str] d:
+    :param returnn.tf.network.TFNetwork network:
+    :param get_layer:
+    """
+    super(RepeatLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+    d["repetitions"] = get_layer(d["repetitions"])
+
+  @classmethod
+  def get_out_data_from_opts(cls, name, axis, repetitions, sources=(), **kwargs):
+    """
+    :param str name:
+    :param str axis:
+    :param LayerBase repetitions:
+    :param list[LayerBase] sources:
+    :rtype: Data
+    """
+    data = get_concat_sources_data_template(sources, name="%s_output" % name)
+    data = data.copy_as_batch_major()
+    axis = data.get_axis_from_description(axis, allow_int=False)
+    data = data.copy_move_axis(axis, data.get_batch_axis(0))
+    return data
+
+
 class CastLayer(CopyLayer):
   """
   Cast to some other dtype.
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -1351,6 +1351,68 @@ def test_ScatterNdLayer_RangeLayer():
     print(out)  # random...
 
 
+def _run_repeat_layer(session, net, input_data_layer):
+  repetitions_data_layer = InternalLayer(
+    name="repetitions", network=net,
+    out_type={'shape': (None,), 'feature_dim_axis': None, 'dtype': 'int32'})
+  repetitions_data_layer.output.placeholder = tf.constant(
+    [[1, 3, 2, 1, 3, 4, 1, 1, 2, 1],
+     [3, 2, 1, 3, 0, 1, 1, 0, 0, 0]], dtype="int32")  # [B, T] (sparse)
+  repetitions_data_layer.output.size_placeholder = {0: tf.constant([10, 7], dtype="int32")}  # [B]
+
+  opts = {
+    'network': net, 'name': 'repeat_layer_test', 'sources': [input_data_layer],
+    'repetitions': repetitions_data_layer, 'axis': 'T'}
+  out_data = RepeatLayer.get_out_data_from_opts(**opts)
+  out_data.sanity_check()
+  print(out_data)
+  repeat_layer = RepeatLayer(output=out_data, **opts)
+  print(repeat_layer.output)
+
+  output, size_placeholder = session.run([repeat_layer.output.placeholder, repeat_layer.output.size_placeholder])
+  assert numpy.all(numpy.equal(size_placeholder[0], numpy.asarray([19, 11])))
+  assert numpy.all(numpy.equal(output.shape, numpy.asarray([2, 19, 5])))
+  # the 6 last positions of the second sequence need to be padded with zeros
+  assert numpy.all(numpy.equal(output[1, 11:], 0))
+  assert out_data.shape == (None, 5)
+  assert out_data.batch_dim_axis == 0
+  assert out_data.time_dim_axis == 1
+
+
+def test_RepeatLayerBTF():
+  with make_scope() as session:
+    net = TFNetwork(extern_data=ExternData())
+    input_data_layer = InternalLayer(name="src", network=net, out_type={'shape': (None, 5), 'dim': 5})
+    input_data_layer.output.size_placeholder = {0: tf.constant([10, 7])}  # [B]
+    input_data_layer.output.placeholder = tf_compat.v1.random_uniform((2, 10, 5))  # [B, T, F]
+
+    _run_repeat_layer(session, net, input_data_layer)
+
+
+def test_RepeatLayerTBF():
+  with make_scope() as session:
+    net = TFNetwork(extern_data=ExternData())
+    input_data_layer = InternalLayer(
+      name="src", network=net,
+      out_type={'shape': (None, 5), 'dim': 5, 'batch_dim_axis': 1, 'time_dim_axis': 0})
+    input_data_layer.output.size_placeholder = {0: tf.constant([10, 7])}  # [B]
+    input_data_layer.output.placeholder = tf_compat.v1.random_uniform((10, 2, 5))  # [T, B, F]
+
+    _run_repeat_layer(session, net, input_data_layer)
+
+
+def test_RepeatLayerBFT():
+  with make_scope() as session:
+    net = TFNetwork(extern_data=ExternData())
+    input_data_layer = InternalLayer(
+      name="src", network=net,
+      out_type={'shape': (5, None), 'dim': 5, 'time_dim_axis': 2, 'feature_dim_axis': 1})
+    input_data_layer.output.size_placeholder = {1: tf.constant([10, 7])}  # [B]
+    input_data_layer.output.placeholder = tf_compat.v1.random_uniform((2, 5, 10))  # [B, F, T]
+
+    _run_repeat_layer(session, net, input_data_layer)
+
+
 def test_ScatterNdLayer_RangeLayer_RangeInAxisLayer():
   n_batch, n_time, n_ts, n_in, n_out = 2, 3, 6, 7, 11
   rnd = numpy.random.RandomState(42)


returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    :param LayerBase repetitions: number of repetitions for each sequence and position in target axis, [B,T] or [T,B]
+    :param str|int axis: (dynamic) axis for repetition (currently only time axis is supported)
+    """
+    super(RepeatLayer, self).__init__(**kwargs)
+    repetitions_data = repetitions.output
+    input_data = self.input_data
+    assert self.input_data
+    original_axis = input_data.get_axis_from_description(axis)
+    if original_axis < input_data.batch_dim_axis:
+      input_data = input_data.copy_as_batch_major().copy_move_axis(original_axis+1, 1)
```
albertz, line 2708: Spacing is wrong.
```suggestion
      input_data = input_data.copy_as_batch_major().copy_move_axis(original_axis + 1, 1)
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    else:
+      input_data = input_data.copy_as_batch_major().copy_move_axis(original_axis, 1)
```
albertz, line 2710: This is not optimal in multiple ways:

* Code with branching is always difficult to read (on the very first glance, it looks like there are multiple different cases -- which is not really the case here).
* Also you have repeated code there, which you should always avoid (`copy_as_batch_major`, `copy_move_axis`).

This would be much better:
```suggestion
    input_data = input_data.copy_as_batch_major()
    original_axis = input_data.get_axis_from_description(axis)
    input_data = input_data.copy_move_axis(original_axis, 1)
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    reshaped_data = tf.reshape(padded_data, flat_shape)  # [B * T+1, F]
+    reshaped_repetitions = tf.reshape(target_repetitions, (shape[0] * shape[1],))  # [B * T+1]
+    # run the repetition
+    repeated_data = tf.repeat(reshaped_data, reshaped_repetitions, axis=0)  # [B * T', ...]
+    # unflatten the output
+    target_shape = tf.concat([[shape[0], max_duration], shape[2:]], axis=0)
+    self.output.placeholder = tf.reshape(repeated_data, target_shape)  # [B, T', ...]
+    # set the size_placeholder
+    self.output.size_placeholder = self.input_data.size_placeholder
+    self.output.size_placeholder[0] = target_seq_len
```
albertz, line 2735: This is wrong. You are overwriting the input size placeholder here.
Also, again the code is too complicated.
This is simpler and fixes the problem:
```suggestion
    self.output.size_placeholder = {0: target_seq_len}
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    if original_axis < input_data.batch_dim_axis:
+      input_data = input_data.copy_as_batch_major().copy_move_axis(original_axis+1, 1)
+    else:
+      input_data = input_data.copy_as_batch_major().copy_move_axis(original_axis, 1)
+    repetitions_data = repetitions_data.copy_as_batch_major()
+    # pad the target axis
+    paddings = [(0, 1) if i == 1 else (0, 0) for i in range(input_data.batch_ndim)]
+    padded_data = tf.pad(input_data.placeholder, paddings)  # [B, T+1, ...]
+    # those are the sequence lengths after expansion
+    target_seq_len = tf.reduce_sum(repetitions_data.placeholder, axis=1)  # [B], the new size_placeholder for 'T'
```
albertz, line 2716: You should install a `DimensionTag` on a newly created seq len. See other code as example (e.g. `ConvLayer`).

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+  The sum of the repetitions has to be non-zero for each sequence in the batch.
+
+  This layer can only be used with Tensorflow 1.15.0 or newer.
+  """
+  layer_class = "repeat"
+
+  def __init__(self, repetitions, axis="T", **kwargs):
+    """
+    :param LayerBase repetitions: number of repetitions for each sequence and position in target axis, [B,T] or [T,B]
+    :param str|int axis: (dynamic) axis for repetition (currently only time axis is supported)
```
albertz, line 2700: We should not allow `int` for any (new) code to specify axes.

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+      data = data.copy_as_batch_major().copy_move_axis(axis+1, 1)
+    else:
+      data = data.copy_as_batch_major().copy_move_axis(axis, 1)
+    assert repetitions.output.dtype in ["int32", "int64"]
+    assert repetitions_data.have_batch_axis() and repetitions_data.batch_ndim == 2, (
+      "RepeatLayer only accepts repetitions of the shape (B, None|S) or (None|S, B)")
+    repetitions_data = repetitions_data.copy_as_batch_major()
+    assert data.shape[0] == repetitions_data.shape[0], (
+      "Axis mismatch between input (%i) and repetitions (%i)" %
+      (data.shape[axis_wo_batch], repetitions_data.shape[0]))
```
albertz, line 2761: `get_out_data_from_opts` should be as simple as possible, just enough code to create the output `Data`. It should also work on incomplete / slightly incorrect input. This is called in many different contexts.

Any checks should be done in `__init__`.

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    """
+    data = get_concat_sources_data_template(sources)
+    data.name = "%s_output" % name
+    repetitions_data = repetitions.output.copy_template()
+    axis = data.get_axis_from_description(axis)
+    axis_wo_batch = data.get_batch_axis_excluding_batch(axis)
+    if axis < data.batch_dim_axis:
+      data = data.copy_as_batch_major().copy_move_axis(axis+1, 1)
+    else:
+      data = data.copy_as_batch_major().copy_move_axis(axis, 1)
```
albertz, line 2754:
```suggestion
    data = data.copy_as_batch_major()
    axis = data.get_axis_from_description(axis, allow_int=False)
    data = data.copy_move_axis(axis, 1)
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    assert repetitions.output.dtype in ["int32", "int64"]
+    assert repetitions_data.have_batch_axis() and repetitions_data.batch_ndim == 2, (
+      "RepeatLayer only accepts repetitions of the shape (B, None|S) or (None|S, B)")
+    repetitions_data = repetitions_data.copy_as_batch_major()
+    assert data.shape[0] == repetitions_data.shape[0], (
+      "Axis mismatch between input (%i) and repetitions (%i)" %
+      (data.shape[axis_wo_batch], repetitions_data.shape[0]))
+    return data
```
albertz, line 2762: I wonder, this would not be batch-major, but use the input data as-is. This is wrong, right?
(However, with my suggested code change above, it would be correct.)

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+  A wrapper around tf.repeat, but supports an additional batch axis for the durations
+  The sum of the repetitions has to be non-zero for each sequence in the batch.
+
+  This layer can only be used with Tensorflow 1.15.0 or newer.
+  """
+  layer_class = "repeat"
+
+  def __init__(self, repetitions, axis="T", **kwargs):
+    """
+    :param LayerBase repetitions: number of repetitions for each sequence and position in target axis, [B,T] or [T,B]
```
albertz, line 2699: If you have some custom `LayerBase` input, you need to overwrite `transform_config_dict`, but I don't see that?

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    # those are the sequence lengths after expansion
+    target_seq_len = tf.reduce_sum(repetitions_data.placeholder, axis=1)  # [B], the new size_placeholder for 'T'
+    # maximum sequence length after expansion
+    max_duration = tf.reduce_max(target_seq_len)  # [] == T'
+    # the new padding is the difference of the maximum to the new target
+    target_padding_steps = tf.expand_dims(max_duration - target_seq_len, 1)  # [B, 1]
+    # add the repetitions for the artificial padding position
+    target_repetitions = tf.concat([repetitions_data.placeholder, target_padding_steps], axis=1)  # [B, T+1]
+    # flatten batch and time
+    shape = tf.shape(padded_data)
```
albertz, line 2724: Use `tf_util.get_shape` instead.
```suggestion
    shape = tf_util.get_shape(padded_data)
```

returnn/tf/layers/basic.py
JackTemaki: I thought when I copy a data object, I will get a "new" instance of the `size_placeholder` dictionary. You code is not correct when there are multiple entries in the `size_placeholder`, which should be supported (but is untested) by the RepeatLayer

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    flat_shape = tf.concat([[shape[0] * shape[1]], shape[2:]], axis=0)
```
albertz, line 2725: Don't use TF functions if not needed. (This also assumes `get_shape` above, but the statement would be true in general.)
Just write:
```suggestion
    flat_shape = [shape[0] * shape[1]] + shape[2:]
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    paddings = [(0, 1) if i == 1 else (0, 0) for i in range(input_data.batch_ndim)]
+    padded_data = tf.pad(input_data.placeholder, paddings)  # [B, T+1, ...]
+    # those are the sequence lengths after expansion
+    target_seq_len = tf.reduce_sum(repetitions_data.placeholder, axis=1)  # [B], the new size_placeholder for 'T'
+    # maximum sequence length after expansion
+    max_duration = tf.reduce_max(target_seq_len)  # [] == T'
+    # the new padding is the difference of the maximum to the new target
+    target_padding_steps = tf.expand_dims(max_duration - target_seq_len, 1)  # [B, 1]
+    # add the repetitions for the artificial padding position
+    target_repetitions = tf.concat([repetitions_data.placeholder, target_padding_steps], axis=1)  # [B, T+1]
```
albertz, line 2722: Better use `tf_util.concat_with_opt_broadcast` instead of `tf.concat`.

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,83 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    # flatten batch and time
+    shape = tf.shape(padded_data)
+    flat_shape = tf.concat([[shape[0] * shape[1]], shape[2:]], axis=0)
+    reshaped_data = tf.reshape(padded_data, flat_shape)  # [B * T+1, F]
+    reshaped_repetitions = tf.reshape(target_repetitions, (shape[0] * shape[1],))  # [B * T+1]
+    # run the repetition
+    repeated_data = tf.repeat(reshaped_data, reshaped_repetitions, axis=0)  # [B * T', ...]
+    # unflatten the output
+    target_shape = tf.concat([[shape[0], max_duration], shape[2:]], axis=0)
```
albertz, line 2731: Again, not so complicated:
```suggestion
    target_shape = [shape[0], max_duration] + shape[2:]
```

returnn/tf/layers/basic.py
JackTemaki: This did not work with the original `tf.shape` command, as it will perform an addition instead of a concatenation.

returnn/tf/layers/basic.py
albertz: But `self.input_data` is not necessarily a new instance.
Also, why do you want to modify `self.input_data` here? It's just very non-intuitive. Leave `input_data` as-is. You never should modify the `size_placeholder` of another `Data` (except the `Data` you currently create here, i.e. `self.output`). This could easily lead to bugs in other contexts.

albertz: Also, why `self.input_data`, and not `input_data`. You potentially transposed the axes in some way, so `self.input_data.size_placeholder` might be wrong.

returnn/tf/layers/basic.py
JackTemaki: I though it is exactly the other way round, that you should fail here early so that RETURNN tries a different approach during template construction.

albertz: No.

returnn/tf/layers/basic.py
albertz: Yes you should not use `tf.shape` directly anyway. It is always better to make it a list first. Or simply use `get_shape`.

returnn/tf/layers/basic.py
JackTemaki: Ah, of course my code is wrong as well, when the spatial axis was at a different position at the beginning.

returnn/tf/layers/basic.py
JackTemaki: What naming pattern should I use?  Most layers define `spatial:%i:<name>`, but some others also have e.g. `sliced-time:<name>`. So should it be `repeated:<name>` or `repeated-spatial:<name>` or something else?

albertz: What you like. We currently don't really have fixed conventions for this. `repeated:<name>` might be good.

returnn/tf/layers/basic.py
JackTemaki: Your replies were not yet loaded for some reason. So I assume I should always create a new dictionary, I can do that.

returnn/tf/layers/basic.py
JackTemaki: Ah this was not an argument, just reasoning for the old code. I will use the list version of course.

returnn/tf/layers/basic.py
JackTemaki: What would be the reason?
```
    target_repetitions = tf_util.concat_with_opt_broadcast(
      [repetitions_data.placeholder, target_padding_steps], allow_broadcast=[False, False], axis=1)  # [B, T+1
```
looks more complicated, and does exactly the same.

albertz: Two reasons:
* It makes the broadcasting explicit. (Explicit is often better than implicit.)
* This also works on earlier TF versions. (I don't exactly remember since when they support broadcasting in concat.)

Well, maybe both are not so important. Your choice.

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,101 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+  def get_out_data_from_opts(cls, name, axis, repetitions, sources=(), **kwargs):
+    """
+    :param str name:
+    :param str axis:
+    :param LayerBase repetitions:
+    :param list[LayerBase] sources:
+    :rtype: Data
+    """
+    data = get_concat_sources_data_template(sources)
+    data.name = "%s_output" % name
```
albertz, line 2776:
```suggestion
    data = get_concat_sources_data_template(sources, name="%s_output" % name)
```

returnn/tf/layers/basic.py
```diff
@@ -2685,6 +2685,101 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
...
+    data = data.copy_as_batch_major()
+    axis = data.get_axis_from_description(axis, allow_int=False)
+    data = data.copy_move_axis(axis, 1)
```
albertz, line 2779: Btw, in case the input has no batch-dim, is this supported? I think it's wrong then, right?
```suggestion
    data = data.copy_move_axis(axis, data.get_batch_axis(0))
```

albertz:
Despite my small further comments, is this considered ready? (Last time, it clearly was not, as it was broken in multiple ways. I wonder that the test cases worked?)

JackTemaki:
The test cases worked, as I only tested the layer itself, not the layer as part of a network. So having a `transform_config_dict` was not needed, the computation was already correct. The now remaining points are related to the question if I should support inputs without a batch axis, which I would not do for now.

returnn/tf/layers/basic.py
JackTemaki: As `tf_util.concat_with_opt_broadcast` will return exactly the same `tf.concat` command as before, there is no functional difference, so I would keep it as it is.

returnn/tf/layers/basic.py
JackTemaki: So should I put assertions for an existing batch axis back into `get_out_data_from_opts`?

albertz: I would just use `data.get_batch_axis(0)` as suggested here in `get_out_data_from_opts` anyway, as this should be correct in any case, also when this is implemented later.
You can do such a check in `__init__`.

returnn/tf/layers/basic.py
albertz: It's not about that this is the same (in newer TF versions). It's about:

* Having broadcasting explicit. (Explicit is often better than implicit.)
* This also works on earlier TF versions.

JackTemaki: So you just want to have it explicitly stated that there is no broadcasting allowed, in case the code is changed later?
I am not sure why you are mentioning earlier TF versions. This will work in any TF version.

Also TF might do broadcasting on its own, even if if set to false, if I understand you correctly. There is no check for that in `concat_with_opt_broadcast`, it just returns `tf.concat`:
```
  if all([not a for a in allow_broadcast]):
    return tf.concat(values, axis=axis)
```
So using the `tf_util` functions makes it even less explicit, as you can read it as "no broadcasting will be done in any way".

albertz: > So you just want to have it explicitly stated that there is no broadcasting allowed, in case the code is changed later?

No. I said it is often better to explicitly state that sth (here: broadcasting) is done/happening. Otherwise it would be implicit.

> I am not sure why you are mentioning earlier TF versions. This will work in any TF version.

No. As I said, earlier TF versions do not support broadcasting in `tf.concat`.

> Also TF might do broadcasting on its own, even if if set to false, if I understand you correctly. There is no check for that in `concat_with_opt_broadcast`, it just returns `tf.concat`:
> 
> ```
>   if all([not a for a in allow_broadcast]):
>     return tf.concat(values, axis=axis)
> ```

This is because `tf.concat` did not support broadcasting earlier.

But anyway, `allow_broadcasting` is the other way around: It is not about disallowing broadcasting. It is about allowing it (explicitly).

You could extend the code now for newer TF versions, if you want to also check the other other way around (i.e. if allow=False, it is explicitly disallowed).

> So using the `tf_util` functions makes it even less explicit, as you can read it as "no broadcasting will be done in any way".

No. It is more explicit, because you explicitly write that you want to have broadcasting.

JackTemaki: I do not understand your comments. Here is no broadcasting happening.

If there is broadcasting, it definitely makes sense to write it explicit, and I understand that older TF version do not support this.

But in this case there is no broadcasting, I would rather disallow broadcasting here explicitly, as any broadcasting of the repetitions can only lead to incorrect results, but will never be correct.

albertz: Ah sorry, I somehow thought there would be broadcasting happening here. (Probably because of the "1" and I did not read it carefully enough.)
Then you can ignore everything what I wrote. `tf.concat` is the right choice.

albertz (approved):

PR #398, opened 2020-11-25
Failing test for case where sub-layer is optimized out

patrick-wilken:
This is the issue I discovered in #395.

The test passes if one reverts 8f16741. But the TODO suggests that this should be handled in `_move_outside_loop()` instead, I will try.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -992,6 +992,10 @@ class _SubnetworkRecCell(object):
     self._construct_template()
     self._initial_outputs = None  # type: typing.Optional[typing.Dict[str,tf.Tensor]]
     self._initial_extra_outputs = None  # type: typing.Optional[typing.Dict[str,typing.Dict[str,typing.Union[tf.Tensor,typing.Tuple[tf.Tensor,...]]]]]  # nopep8
+
+    # input_layers_moved_out, output_layers_moved_out and layers_in_loop include (used) sub-layers as separate
+    # entries, this way in- and outputting them to the loop via TensorArrays will be handled just as for normal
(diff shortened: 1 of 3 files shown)


patrick-wilken:
I added commits that fix the test case that failed before (71dc7e2). (Not sure if this is the right workflow...)

They also contain an alternative to #395. #395 works, but conceptually it is not right (which isn't obvious :sweat_smile:). `input_layers_moved_out` and `output_layers_moved_out` should contain sub-layers names, so there should be no need to handle sub-layers there in a separate case. However, `output_layers_moved_out` did not contain sub-layer names before because no dependency to sub-layer templates were added in `construct_template()`. (I did not notice this before because in my case the sub-layers were always the output layers.) 7b4e1ec fixes this. We need the sub-layer names in the layer lists because we also have to create TensorArrays for the sub-layers if they are in- or outputs of the loop. This is all pretty complicated and wasn't documented before at all, I now added at least some comments.

@Zettelkasten, your test still passes, can you try out this branch with your real-life case were you could not access a sub-layer of SplitLayer before? Just to be sure, the test seems pretty simplistic.

Zettelkasten:
> @Zettelkasten, your test still passes, can you try out this branch with your real-life case were you could not access a sub-layer of SplitLayer before? Just to be sure, the test seems pretty simplistic :)

My real-life use case still works fine with this fix, this seems fine to me. Thanks a lot! :+1:

albertz (approved):
It's somewhat difficult to follow the logic here on GitHub in the diff.
From a first glance, this looks ok.
I will just merge this in now, and we can iterate further in case there are other problems.

PR #400, opened 2020-11-27
Map-style dataset base for user-side datasets

patrick-wilken:
This is @JackTemaki 's implementation as discussed in #353.
I've now split it into a general and a RETURNN Dataset specific part and changed a few minor things.

Is the goal that giving a list to MapDatasetWrapper should work? How would we handle `get_seq_tag()` in this case? Also exception handling for the other methods which are not part of the `list` API would have to be adjusted (e.g. catching AttributeError).

### DIFF
--- a/returnn/datasets/map.py
+++ b/returnn/datasets/map.py
@@ -1,0 +1,199 @@
+"""
+Provides :class:`MapDatasetBase`
+"""
+
+from returnn.datasets.basic import DatasetSeq
+from returnn.datasets.cached2 import CachedDataset2
+from returnn.util.basic import OptionalNotImplementedError
+
+
+class MapDatasetBase(object):
+  """
+  This dataset can be used as template to implement user-side Datasets, where the data can be access in arbitrary order.
+  For global sorting, the length information needs to be known beforehand, see get_seq_len.
+  """
+
+  def __init__(self, data_types=None):
+    """
+    :param dict[str,dict] data_types: data_key -> constructor parameters of Data object, for all data streams the
+      dataset provides (inputs and targets). E.g. {'data': {'dim': 1000, 'sparse': True, ...}, 'classes': ...}.
+    """
+    self.data_types = data_types or {}
+
+  def __len__(self):
+    """
+    :return: total number of sequences in the dataset
+    :rtype: int
+    """
+    raise NotImplementedError
+
+  def __getitem__(self, seq_idx):
+    """
+    This function does the actual data loading, the order can be arbitrary.
+
+    :param int seq_idx:
+    :return: The content of a single dataset entry
+    :rtype: dict[str,numpy.array]
+    """
+    raise NotImplementedError
+
+  def get_seq_len(self, seq_idx):
+    """
+    This optional function provides the sequence length for the `seq_ordering` parameter.
+    If not specified only a limited set of options is available.
+
+    :param int seq_idx:
+    :return: sequence length
+    :rtype: int
+    """
+    raise OptionalNotImplementedError
+
+  def get_seq_tag(self, seq_idx):
+    """
+    :param int seq_idx:
+    :return: tag for the sequence of the given index, default is 'seq-{seq_idx}'.
+    :rtype: str
+    """
+    return "seq-%i" % seq_idx
+
+  def get_seq_order(self, epoch=None):
+    """
+    Override to implement a dataset specific sequence order for a given epoch.
+    The number of sequences can be less than the total number.
+    This will override the effects of `partition_epoch` and `seq_ordering` when using MapDatasetWrapper.
+
+    :param int epoch:
+    :return: sequence order (list of sequence indices)
+    :rtype: list[int]
+    """
+    raise OptionalNotImplementedError
+
+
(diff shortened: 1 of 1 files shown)


returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    :rtype: int|None
+    """
+    raise OptionalNotImplementedError
+
+  def get_seq_tag(self, seq_idx):
+    """
+    Optionally return a sequence tag.
+
+    :param int seq_idx:
+    :return:
```
albertz, line 55:
```suggestion
    :param int seq_idx:
    :return: optionally a sequence tag
```

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+from returnn.util.basic import OptionalNotImplementedError
+
+
+class MapDatasetBase(object):
+  """
+  This dataset can be used as template to implement user-side Datasets, where the data can be access in arbitrary order.
+  For global sorting, the length information needs to be known beforehand.
+  """
+
+  def __init__(self, num_outputs=None):
```
albertz, line 16: What is `num_outputs`? Is this only about outputs? What is the dict content?

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    raise NotImplementedError
+
+  def get_seq_len(self, seq_idx):
+    """
+    This optional function provides the sequence length for the `seq_ordering` parameter.
+    If not specified only a limited set of options is available.
+
+    :param seq_idx:
+    :return:
```
albertz, line 45:
```suggestion
    :return: sequence length
```

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    :rtype: int|None
```
albertz, line 46: Is this correct? If implemented, it should always return int, right?
```suggestion
    :rtype: int
```

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    return "seq-%i" % seq_idx
+
+  def get_seq_order(self, epoch=None):
+    """
+    Override to implement a dataset specific sequence order for a given epoch.
+    The number of sequences can be less than the total number.
+    This will override the effects of `partition_epoch` and `seq_ordering` when using MapDatasetWrapper.
+
+    :param epoch:
+    :return: seq_order
```
albertz, line 67: What does this mean?

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    :param int|None random_seed_offset: provide a seed offset for the sequence ordering
+    :param int|None partition_epoch:
+    """
+
+    super(MapDatasetWrapper, self).__init__(
+      name=name,
+      seq_ordering=seq_ordering,
+      random_seed_offset=random_seed_offset,
+      partition_epoch=partition_epoch
+    )
```
albertz, line 91:
```suggestion
      partition_epoch=partition_epoch)
```

returnn/datasets/map.py
```diff
@@ -0,0 +1,210 @@
...
+    :rtype: list[int]
+    """
+    raise OptionalNotImplementedError
+
+
+class MapDatasetWrapper(CachedDataset2):
+  """
+  Takes a MapDataset and turns it into a returnn.datasets.Dataset by providing the required class methods.
+  """
+  def __init__(self, map_dataset, name=None, seq_ordering='default', random_seed_offset=None, partition_epoch=None):
```
albertz, line 77: Why is there no `**kwargs`? Usually you always should use `kwargs`, and pass that on to `super`.
Why do you explicitly specify `name`, `seq_ordering`, etc here?

returnn/datasets/map.py
patrick-wilken: This is called num_outputs in other places in the code, but yes, it would be good to change this misleading name here. I've now called it data_types lacking a better idea. Might be confused with data types as in float, int, etc. But alternatives would be also not very clear: data_specs, data_defs, data_config... :thinking:

returnn/datasets/map.py
patrick-wilken: Do you just mean the `:return:  seq_order` line? I've made it more verbose.

returnn/datasets/map.py
albertz: One of the main reasons for this new alternative interface `MapDatasetBase` is that this is supposed to be more straight-forward, easier, and can get rid of other confusing parts. It doesn't really makes sense to replicate the confusing parts.

Also, `num_outputs` is already deprecated. In the main config, we have `extern_data` instead, and in the `Dataset`, we should also introduce sth more generic. But anyway, that's all irrelevant, as with `MapDatasetBase`, you start completely fresh, and can design it in any way you want. Maybe similar to the PyTorch API, as some people might already know that.

albertz:
@JackTemaki Ping?

albertz:
What is the state here? @JackTemaki ?

JackTemaki (approved):
For me this would be ok

PR #404, opened 2020-12-04
fix log activation for tf2

vieting:
Small fix to allow using `ActivationLayer` with `activation="log"` in tf2 as well.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -1213,6 +1213,8 @@ def get_activation_function(s):
     return getattr(tf.nn, s)  # e.g. relu, elu, sigmoid, softmax, ...
   elif hasattr(tf, s):
     return getattr(tf, s)  # e.g. log, abs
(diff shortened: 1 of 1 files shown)


returnn/tf/util/basic.py
```diff
@@ -1213,6 +1213,8 @@ def get_activation_function(s):
     return getattr(tf.nn, s)  # e.g. relu, elu, sigmoid, softmax, ...
   elif hasattr(tf, s):
     return getattr(tf, s)  # e.g. log, abs
+  elif hasattr(tf.math, s):
```
albertz, line 1216: `tf.math` does not exist in earlier TF versions, I think.
```suggestion
  elif hasattr(tf, "math") and hasattr(tf.math, s):
```

vieting: It exists at least in tf1.15, but this way we're on the safe side for ealier version.

albertz: There are couple of people who use TF 1.8 or maybe even earlier.

Issue #406, opened 2020-12-07
DotLayer: Support shuffled remaining dimensions


Zettelkasten:
I have a config with a dot layer that takes two inputs of e.g. the shapes `[B,T,X,Y]` and `[T,B,Y,Z]`.
Then I want to compute their dot product, getting a tensor of the form `[B,T,X,Z]` (reduce the `Y` axis, have `X` and `Z` be the variable axes).

Note however that the the remaining axes (i.e. `B` and `T`) are equal but not in the same order for these two tensors.
Currently, this is not handled by `DotLayer` at all, it always assumes that all remaining axes are in the exact same order.
If they have static dimensions, than this will throw an error at template construction.
If not, the flag `debug=True` will assert that the shapes are compatible when the session is executed.
If that flag is not set, I guess TF would throw an error as well (except when it works because of broadcasting or sth like that).

Here is a small artificial example demonstrating this (with static shapes):
```
net_dict = {
  "a": {"class": "split_dims", "axis": "static:0", "dims": (2, 3, 5)},   # [B, 2, 3, 5]
  "b": {"class": "transpose", "from": ["a"], "perm": {"static:0": "static:1", "static:1": "static:0"}},  # [B, 3, 2, 5]
  "dot": {  # should produce [B, 2, 3, 1] or [B, 3, 2, 1], but throws an Exception instead
    "class": "dot", "from": ["a", "b"],
    "red1": "static:-1", "red2": "static:-1", "var1": None, "var2": None,
    "debug": True}
  # ...
}
```

To fix this, I think we need to transpose the remaining axes of one of the inputs to match each others order, which we can e.g.
determine by matching the `DimensionTag`s of the remaining axes using `DimensionTag.is_equal(..)` - similar to `get_common_data` does this as well.
But an extra `transpose` sounds inefficient (and from what I can tell from the code efficiency was considered thoroughly in the `DotLayer`) - perhaps we can integrate this better with the logic there already is in the layer?

Also: Why doesn't `DotLayer` use `tf.einsum` instead of `tf.matmul`? This would handle those index permutations by itself. (I guess performance? Or maybe backwards compatibility?)

albertz:
That should work, obviously. It also did work in the past (your first example [B,T,X,Y] and [T,B,Y,Z]; I assume T is dynamic), because we always used `copy_as_batch_major`. However, that was removed in 90fb27be502b5c15083ebf3e41e772797ab6e656, and it seems we did not consider this.

[...]

PR #407, opened 2020-12-08
GatherLayer: Position can be any layer now

Zettelkasten:
This is my implementation of a wrapper around `tf.gather`. Sorry @albertz, I kind of completely missed that you started work on this in c968dd97bf57add1aeaafb6d0326a05939c7a9eb, so I essentially built this layer from scratch.
This implementation here however supports `position` to be any other layer and thus much more flexible (also much more complicated though..). I added a special case in `transform_config_dict` that handles the case that `position` is an int by "wrapping" it in a `ConstantLayer`. We can in the future easily adapt this to also support more complex constants like entire arrays.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -913,37 +913,199 @@ class SliceNdLayer(_ConcatInputLayer):
 
 class GatherLayer(_ConcatInputLayer):
   """
-  This is basically ``x[position]`` in some axis.
-  See also :class:`SliceLayer`.
-  See also :class:`GatherNdLayer`, which works on the time-axis, but supports different positions per batch.
+  Gathers slices on a specified axis from the input layer using indices from a ``position`` layer.
+  If the input is a layer of the shape ``[B,D,F1]``, and indices of shape ``[B,F2]``, this will yield output of the
+  shape ``[B,F2,F1]`` where
 
-  (Currently `position` can just be an integer, so it's much less generic than ``tf.gather``.
-   However, we might extend this to lists, arrays, or even other layers.)
+  ``output[b,f2,f1] = input[b,indices[b,f2],f1]``
+
+  (if ``D`` is the axis to gather from).
+  In general, all shared axes of the input and the positions will be considered as batch-axes.
+
+  It's basically a wrapper around ``tf.gather``.
+  See also :class:`GatherNdLayer`.
   """
   layer_class = "gather"
 
-  def __init__(self, axis, position, **kwargs):
+  def __init__(self, position, axis, **kwargs):
     """
-    :param str axis:
-    :param int position:
+    :param LayerBase position: Layer containing the indices used to select the slices of the input from.
+    If another layer, must be of type ``int32`` or ``int64``. Can also specify a constant ``int``.
+    :param str axis: The axis into which we gather the indices into
     """
     super(GatherLayer, self).__init__(**kwargs)
-    axis = self.input_data.get_axis_from_description(axis, allow_int=False)
-    assert isinstance(position, int)  # nothing else implemented yet
-    slices = [slice(None, None)] * axis + [position]  # type: typing.List[typing.Union[slice, int]]
-    self.output.placeholder = self.input_data.placeholder[slices]
+    self.position = position
+
+    input_data = self.input_data
+    position_data = position.output
+    old_gather_axis = input_data.get_axis_from_description(axis, allow_int=False)  # might be moved later
+
+    # determine all common axes of input_data and position_data
+    axes = self._get_common_input_position_axes(input_data, position_data, old_gather_axis)
+    common_axes_input, common_axes_position, input_axes, position_axes = axes
+    batch_dims = len(common_axes_input)
+
+    # move gather axis back if necessary s.t. common axes are always in front
+    gather_axis = max(batch_dims, old_gather_axis)
+
+    # (BatchAxes.., gather_axis, InputAxes..)
+    params = tf.transpose(
+      input_data.placeholder,
+      common_axes_input + input_axes[:gather_axis-batch_dims] + [old_gather_axis] + input_axes[gather_axis-batch_dims:])
+    # (BatchAxes.., PositionAxes..)
+    indices = tf.transpose(position_data.placeholder, common_axes_position + position_axes)
+    # (BatchAxes.., PositionAxes.., InputAxes..)
+    self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
+
+    # Take size_placeholder from input_data if they exist there, otherwise from position_data
+    size_placeholder = dict()
+    for input_axis, size in input_data.size_placeholder.items():
+      input_axis = input_data.get_batch_axis(input_axis)
+      if input_axis == old_gather_axis:
+        continue
+      output_axis = self.output.get_batch_axis_excluding_batch(
+        self.translate_input_axis(input_axis, old_gather_axis, *axes))
+      size_placeholder[output_axis] = size
+    for position_axis, size in position_data.size_placeholder.items():
+      position_axis = position_data.get_batch_axis(position_axis)
+      output_axis = self.output.get_batch_axis_excluding_batch(self.translate_position_axis(position_axis, *axes))
+      size_placeholder.setdefault(output_axis, size)
+    self.output.size_placeholder = size_placeholder
 
   @classmethod
-  def get_out_data_from_opts(cls, name, sources, axis, **kwargs):
+  def _get_common_input_position_axes(cls, input_data, position_data, old_gather_axis):
+    """
+    Determine all common axes of input_data and position_data. All other axes will be specific axes then (except for
+    the gather axis of ``input_data``)
+    :param Data input_data:
+    :param Data position_data:
+    :param int old_gather_axis: gather axis of ``input_data`` counted with batch dim (before any transformation)
+    :rtype: (list[int], list[int], list[int], list[int])
+    :return: (common_axes_input, common_axes_position, specific_input_axes, specific_position_axes), all counted with
+    batch dim.
+    """
+    is_equal_opts = dict(ignore_feature_dim=True, allow_same_spatial_dim=True, broadcast_matches=True)
+    common_axes_pairs = [
+      (input_axis, position_axis)
+      for input_axis in range(input_data.batch_ndim) for position_axis in range(position_data.batch_ndim)
+      if not input_axis == old_gather_axis
+      if input_data.get_dim_tag(input_axis).is_equal(position_data.get_dim_tag(position_axis), **is_equal_opts)
+    ]
+    common_axes_input, common_axes_position = zip(*common_axes_pairs) if common_axes_pairs else ([], [])
+    common_axes_input, common_axes_position = list(common_axes_input), list(common_axes_position)
+    specific_input_axes = [
+      axis for axis in range(input_data.batch_ndim) if axis not in common_axes_input and not axis == old_gather_axis]
+    specific_position_axes = [axis for axis in range(position_data.batch_ndim) if axis not in common_axes_position]
+    return common_axes_input, common_axes_position, specific_input_axes, specific_position_axes
+
+  # noinspection PyUnusedLocal
+  @classmethod
+  def translate_input_axis(
+    cls, input_axis, old_gather_axis, common_axes_input, common_axes_position, input_axes, position_axes):
+    """
+    :param int input_axis: batch axis of input_data
+    :param int old_gather_axis: gather axis of ``input_data`` counted with batch dim
+    :param list[int] common_axes_input:
+    :param list[int] common_axes_position:
+    :param list[int] input_axes:
+    :param list[int] position_axes:
+    :rtype: int|None
+    :return: batch axis of output
+    """
+    del common_axes_position  # add this parameter s.t. we can just call translate_input_axis(ax, gather_axis, *data).
+    if input_axis in common_axes_input:
+      return common_axes_input.index(input_axis)
+    elif input_axis == old_gather_axis:
+      return None  # the gather axis will not be present in the output
+    else:
+      return len(common_axes_input) + len(position_axes) + input_axes.index(input_axis)
+
+  # noinspection PyUnusedLocal
+  @classmethod
+  def translate_position_axis(
+    cls, position_axis, common_axes_input, common_axes_position, input_axes, position_axes):
+    """
+    :param position_axis: batch axis of position_data
+    :param list[int] common_axes_input:
+    :param list[int] common_axes_position:
+    :param list[int] input_axes:
+    :param list[int] position_axes:
+    :rtype: int
+    :return: batch axis of output
+    """
+    del input_axes  # add this parameter s.t. we can just call translate_position_axis(ax, *data).
+    if position_axis in common_axes_position:
+      return common_axes_position.index(position_axis)
+    else:
+      return len(common_axes_input) + position_axes.index(position_axis)
+
+  def get_dep_layers(self):
+    """
+    :rtype: list[LayerBase]
+    """
+    return super(GatherLayer, self).get_dep_layers() + [self.position]
+
+  @classmethod
+  def get_out_data_from_opts(cls, name, sources, position, axis, **kwargs):
     """
     :param str name:
     :param list[LayerBase] sources:
-    :param str axis:
+    :param LayerBase position:
+    :param str|int axis:
     :rtype: Data
     """
-    out = get_concat_sources_data_template(sources, name="%s_output" % name)
-    axis = out.get_axis_from_description(axis, allow_int=False)
-    return out.copy_template_excluding_axis(axis)
+    input_data = get_concat_sources_data_template(sources)
+    gather_axis = input_data.get_axis_from_description(axis)
+
+    position_data = position.output.copy_template()
+    assert position_data.dtype in ['int32', 'int64']
+
+    # determine all common axes of input_data and position_data
+    axes = cls._get_common_input_position_axes(input_data, position_data, gather_axis)
+    common_axes_input, common_axes_position, input_axes, position_axes = axes
+
+    # (BatchAxes.., PositionAxes.., InputAxes..)
+    shape = (
+      [input_data.batch_shape[ax] for ax in common_axes_input if not ax == input_data.batch_dim_axis] +
+      [position_data.batch_shape[ax] for ax in position_axes if not ax == position_data.batch_dim_axis] +
+      [input_data.batch_shape[ax] for ax in input_axes if not ax == input_data.batch_dim_axis])
+    out_type = input_data.get_kwargs()
+    out_type["name"] = "%s_output" % name
+    out_type["shape"] = shape
+    out_type["beam"] = SearchBeam.get_combined_beam(input_data.beam, position_data.beam)
+
+    # Take axes from input_data if they exist there, otherwise from position_data
+    for axis_kind in ["batch_dim_axis", "time_dim_axis", "feature_dim_axis"]:
+      input_axis, position_axis = getattr(input_data, axis_kind), getattr(position_data, axis_kind)
+      if input_axis is not None and not input_axis == gather_axis:
+        out_type[axis_kind] = cls.translate_input_axis(input_axis, gather_axis, *axes)
+      elif position_axis is not None:
+        out_type[axis_kind] = cls.translate_position_axis(position_axis, *axes)
+      else:
+        out_type[axis_kind] = None
+
+    if input_data.sparse:
+      out_type["dim"] = input_data.dim
+    else:
+      out_type["dim"] = NotSpecified
+
+    return Data(**out_type)
+
+  @classmethod
+  def transform_config_dict(cls, d, network, get_layer):
+    """
+    :param dict[str] d:
+    :param returnn.tf.network.TFNetwork network:
+    :param get_layer:
+    """
+    super(GatherLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+
+    if isinstance(d["position"], int):
+      kwargs = dict(name="constant_position", network=network, sources=[], value=d["position"], dtype='int32')
+      d["position"] = ConstantLayer(output=ConstantLayer.get_out_data_from_opts(**kwargs), **kwargs)
+    else:
+      assert isinstance(d["position"], str), "other constant types not supported yet."
+      d["position"] = get_layer(d["position"])
 
 
 class GatherNdLayer(_ConcatInputLayer):
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -914,15 +914,19 @@ class SliceNdLayer(_ConcatInputLayer):
 class GatherLayer(_ConcatInputLayer):
   """
   Gathers slices on a specified axis from the input layer using indices from a ``position`` layer.
-  If the input is a layer of the shape ``[B,D,F1]``, and indices of shape ``[B,F2]``, this will yield output of the
+  If the input is a layer of the shape ``[B,D,F1]``, and position of shape ``[B,F2]``, this will yield output of the
   shape ``[B,F2,F1]`` where
 
-  ``output[b,f2,f1] = input[b,indices[b,f2],f1]``
+  ``output[b,f2,f1] = input[b,position[b,f2],f1]``
 
   (if ``D`` is the axis to gather from).
   In general, all shared axes of the input and the positions will be considered as batch-axes.
 
+  The ``position`` argument can also be an ``int``.
+  In this case, this simply gives ``input[position]`` one the specified ``axis``.
+
   It's basically a wrapper around ``tf.gather``.
+  It provides the same functionality as the deprecated ``GatherNdLayer``, but is more generic.
   See also :class:`GatherNdLayer`.
   """
   layer_class = "gather"
@@ -930,8 +934,8 @@ class GatherLayer(_ConcatInputLayer):
   def __init__(self, position, axis, **kwargs):
     """
     :param LayerBase|int position: Layer containing the indices used to select the slices of the input from.
-    If another layer, must be of type ``int32`` or ``int64``.
-    Can also specify a constant ``int``.
+      If another layer, must be of type ``int32`` or ``int64``.
+      Can also specify a constant ``int``.
     :param str axis: The axis into which we gather the indices into
     """
     super(GatherLayer, self).__init__(**kwargs)
@@ -939,9 +943,9 @@ class GatherLayer(_ConcatInputLayer):
 
     input_data = self.input_data
     if isinstance(position, int):
-      constant_kwargs = dict(name="constant_position", network=kwargs["network"], value=position, dtype='int32')
-      position = ConstantLayer(output=ConstantLayer.get_out_data_from_opts(**constant_kwargs), **constant_kwargs)
-    position_data = position.output
+      position_data = Data.from_tensor(tf.constant(position))
+    else:
+      position_data = position.output
     old_gather_axis = input_data.get_axis_from_description(axis, allow_int=False)  # might be moved later
 
     # determine all common axes of input_data and position_data
@@ -952,13 +956,13 @@ class GatherLayer(_ConcatInputLayer):
     # move gather axis back if necessary s.t. common axes are always in front
     gather_axis = max(batch_dims, old_gather_axis)
 
-    # (BatchAxes.., gather_axis, InputAxes..)
+    # (BatchAxes.., InputAxesBeforeGatherAxis, gather_axis, InputAxesAfterGatherAxis..)
     params = tf.transpose(
       input_data.placeholder,
       common_axes_input + input_axes[:gather_axis-batch_dims] + [old_gather_axis] + input_axes[gather_axis-batch_dims:])
     # (BatchAxes.., PositionAxes..)
     indices = tf.transpose(position_data.placeholder, common_axes_position + position_axes)
-    # (BatchAxes.., PositionAxes.., InputAxes..)
+    # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
     # Take size_placeholder from input_data if they exist there, otherwise from position_data
@@ -968,12 +972,12 @@ class GatherLayer(_ConcatInputLayer):
       if input_axis == old_gather_axis:
         continue
       output_axis = self.output.get_batch_axis_excluding_batch(
-        self.translate_input_axis(input_axis, old_gather_axis, common_axes_input, input_axes, position_axes))
+        self._translate_input_axis(input_axis, old_gather_axis, common_axes_input, input_axes, position_axes))
       size_placeholder[output_axis] = size
     for position_axis, size in position_data.size_placeholder.items():
       position_axis = position_data.get_batch_axis(position_axis)
-      output_axis = self.output.get_batch_axis_excluding_batch(self.translate_position_axis(
-        position_axis, common_axes_input, common_axes_position, position_axes))
+      output_axis = self.output.get_batch_axis_excluding_batch(self._translate_position_axis(
+        position_axis, old_gather_axis, common_axes_position, position_axes))
       size_placeholder.setdefault(output_axis, size)
     self.output.size_placeholder = size_placeholder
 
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+  See also :class:`GatherNdLayer`.
   """
   layer_class = "gather"
 
-  def __init__(self, axis, position, **kwargs):
+  def __init__(self, position, axis, **kwargs):
     """
-    :param str axis:
-    :param int position:
+    :param LayerBase position: Layer containing the indices used to select the slices of the input from.
```
albertz, line 932:
```suggestion
    :param int|LayerBase position: Layer containing the indices used to select the slices of the input from.
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      position_axis = position_data.get_batch_axis(position_axis)
+      output_axis = self.output.get_batch_axis_excluding_batch(self.translate_position_axis(position_axis, *axes))
+      size_placeholder.setdefault(output_axis, size)
+    self.output.size_placeholder = size_placeholder
+
+  @classmethod
+  def _get_common_input_position_axes(cls, input_data, position_data, old_gather_axis):
+    """
+    Determine all common axes of input_data and position_data. All other axes will be specific axes then (except for
+    the gather axis of ``input_data``)
```
albertz, line 979:
```suggestion
    the gather axis of ``input_data``)

```
(Newline after description, before params.)

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    for position_axis, size in position_data.size_placeholder.items():
+      position_axis = position_data.get_batch_axis(position_axis)
+      output_axis = self.output.get_batch_axis_excluding_batch(self.translate_position_axis(position_axis, *axes))
+      size_placeholder.setdefault(output_axis, size)
+    self.output.size_placeholder = size_placeholder
+
+  @classmethod
+  def _get_common_input_position_axes(cls, input_data, position_data, old_gather_axis):
+    """
+    Determine all common axes of input_data and position_data. All other axes will be specific axes then (except for
```
albertz, line 978:
```suggestion
    Determine all common axes of input_data and position_data.
    All other axes will be specific axes then (except for
```
(Newline after sentence end.)

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    params = tf.transpose(
+      input_data.placeholder,
+      common_axes_input + input_axes[:gather_axis-batch_dims] + [old_gather_axis] + input_axes[gather_axis-batch_dims:])
+    # (BatchAxes.., PositionAxes..)
+    indices = tf.transpose(position_data.placeholder, common_axes_position + position_axes)
+    # (BatchAxes.., PositionAxes.., InputAxes..)
+    self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
+
+    # Take size_placeholder from input_data if they exist there, otherwise from position_data
+    size_placeholder = dict()
```
albertz, line 961:
```suggestion
    size_placeholder = {}
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    :param get_layer:
+    """
+    super(GatherLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
+
+    if isinstance(d["position"], int):
+      kwargs = dict(name="constant_position", network=network, sources=[], value=d["position"], dtype='int32')
+      d["position"] = ConstantLayer(output=ConstantLayer.get_out_data_from_opts(**kwargs), **kwargs)
+    else:
+      assert isinstance(d["position"], str), "other constant types not supported yet."
+      d["position"] = get_layer(d["position"])
```
albertz, line 1108:
```suggestion
    if not isinstance(d["position"], int):
      assert isinstance(d["position"], str), "invalid 'position' type %s" % type(d["position"])
      d["position"] = get_layer(d["position"])
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    return super(GatherLayer, self).get_dep_layers() + [self.position]
+
+  @classmethod
+  def get_out_data_from_opts(cls, name, sources, position, axis, **kwargs):
     """
     :param str name:
     :param list[LayerBase] sources:
-    :param str axis:
+    :param LayerBase position:
+    :param str|int axis:
```
albertz, line 1054:
```suggestion
    :param str axis:
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
   """
   layer_class = "gather"
 
-  def __init__(self, axis, position, **kwargs):
+  def __init__(self, position, axis, **kwargs):
     """
-    :param str axis:
-    :param int position:
+    :param LayerBase position: Layer containing the indices used to select the slices of the input from.
+    If another layer, must be of type ``int32`` or ``int64``. Can also specify a constant ``int``.
```
albertz, line 933:
```suggestion
      If another layer, must be of type ``int32`` or ``int64``.
      Can also specify a constant ``int``.
```

albertz:
The `GatherNdLayer` is exactly for the case of batched dims. I'm not sure if it is a good idea to duplicate that functionality in `GatherLayer`.

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      if input_data.get_dim_tag(input_axis).is_equal(position_data.get_dim_tag(position_axis), **is_equal_opts)
+    ]
+    common_axes_input, common_axes_position = zip(*common_axes_pairs) if common_axes_pairs else ([], [])
+    common_axes_input, common_axes_position = list(common_axes_input), list(common_axes_position)
+    specific_input_axes = [
+      axis for axis in range(input_data.batch_ndim) if axis not in common_axes_input and not axis == old_gather_axis]
+    specific_position_axes = [axis for axis in range(position_data.batch_ndim) if axis not in common_axes_position]
+    return common_axes_input, common_axes_position, specific_input_axes, specific_position_axes
 
+  # noinspection PyUnusedLocal
```
albertz, line 1001: Don't use `noinspection`. Why do you need that?

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,199 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      return None  # the gather axis will not be present in the output
+    else:
+      return len(common_axes_input) + len(position_axes) + input_axes.index(input_axis)
+
+  # noinspection PyUnusedLocal
+  @classmethod
+  def translate_position_axis(
+    cls, position_axis, common_axes_input, common_axes_position, input_axes, position_axes):
+    """
+    :param position_axis: batch axis of position_data
```
albertz, line 1028: Type?

Zettelkasten:
> The `GatherNdLayer` is exactly for the case of batched dims. I'm not sure if it is a good idea to duplicate that functionality in `GatherLayer`.

Oh true, I oversaw that - that is confusing. I thought `GatherNdLayer` expected a `position` tensor with a feature dim equal to the number of dims of the `values` (just like `tf.gather_nd` does).
If `gather_nd` does not do that, what is the difference to `GatherLayer`?
From my intuition, this is exactly what the behavior of `GatherLayer` should look like, while `GatherNd` takes `position` of a different format. I.e. I thought it would be exactly analogue to how it is in the Tensorflow API.

Also, this implementation supports multiple batch axes as well as as specific axes for both `position` and `input`.
In my use case, I need these features.
(Well okay, I can also flatten multiple batch dims into one and then revert that back, but that's pretty inconvenient).

albertz:
> > The `GatherNdLayer` is exactly for the case of batched dims. I'm not sure if it is a good idea to duplicate that functionality in `GatherLayer`.
> 
> Oh true, I oversaw that - that is confusing. I thought `GatherNdLayer` expected a `position` tensor with a feature dim equal to the number of dims of the `values` (just like `tf.gather_nd` does).

No. I think the name is badly chosen. `BatchedGatherLayer` or so might have been better.

Also, this was at a time where the original `tf.gather` did not support the argument `batch_dims `, so that is why I had to use `tf.gather_nd`. Thus also the original name, because the idea was also to maybe generalize it further. Although the current interface is really mostly intended for a batched-gather, and I'm not sure how to easily generalize (to also not make it too confusing). I guess a separate `RealGatherNdLayer` (ofc some better name...) would make more sense.

The current `GatherNdLayer` (batched-gather), or actually the function `batch_gather`, might be reimplemented using `tf.gather` now (for newer TF version) -- which might be cleaner, easier and faster, although it still should support the old code for older TF versions.

> If `gather_nd` does not do that, what is the difference to `GatherLayer`?

Your understanding of `gather_nd` is correct. It's just that this function was used internally in `GatherNdLayer`, to implement a batched gather.

`GatherLayer`, when I introduced it, only supported a single int. And I thought about extending it to any layer, but not using batching (because that would have overlap with the current `GatherNdLayer`).

> From my intuition, this is exactly what the behavior of `GatherLayer` should look like, while `GatherNd` takes `position` of a different format. Essentially I thought it would be exactly analogue to how it is in the Tensorflow API.

Yes, the names are bad, and yes, it would be much better if they are analogue to how it is in the TF API.
Unfortunately the names are like they are now. I think we should not change the names anymore.

So we could generalize `GatherLayer` like with your suggestion but then have the overlap with `GatherNdLayer`. Not sure. Or at least improve on the documentation to clarify this.

> Also, this implementation supports multiple batch axes as well as as specific axes for both `position` and `input`.
> In my use case, I need these features.

Yes, I planned to extend `GatherNdLayer` exactly by this functionality, but then I never needed it so far.

> (Well okay, I can also flatten multiple batch dims into one and then revert that back, but that's pretty inconvenient).
> Plus I think being able to specify the axis to gather on explicitly is nice.

Yes, this is also sth which I wanted to add for `GatherNdLayer`.

We could also merge in your changes to the current `GatherNdLayer` (you would need to be careful that it doesn't break old configs, I'm not exactly sure on the behavior in some edge cases, but there are enough test cases, this should be fine if all tests are passing).

I'm not sure what is better. What do you think?

@JackTemaki @patrick-wilken do you have an opinion on this?

Zettelkasten:
Ah okay!

> `GatherLayer`, when I introduced it, only supported a single int. And I thought about extending it to any layer, but not using batching (because that would have overlap with the current `GatherNdLayer`).

I think it does not make much sense to have two layers that both take inputs similar to what `tf.gather` takes: No matter if the `position` input is a constant int or another layer, has a batch axis or not, etc., the functionality is essentially the same and I wouldn't separate that into two different layers.
The same applies to what we right now call `GatherLayer` and `GatherNdLayer`: Both essentially do exactly the same, and I would want to keep it a single layer I think.

What about doing as follows:
- use the code from my PR and call it `GatherLayer`, and support both constant/non-constant as well as batched/non-batched inputs there (essentially how it is in this PR now), but also stay backward compatible to the API of the existing `GatherNdLayer` (should be already, but we have to test this more)
- deprecate the `GatherNdLayer`, and make it an effective alias for the `GatherLayer` until we maybe remove it entirely in the future
- update my proposed implementation for `GatherLayer` in such a way that it remains backwards compatible to older setups: In particular to TF versions where `tf.gather` does not support the `batch_dims` parameter

This would mean that we essentially do not have duplicate code, and updating configs is super simple: Just rename `gather_nd` to `gather`.
But on the other hand, the implementation in this PR will become more complicated if we ensure backwards compatibility - should be okay though I think.

If we in the future decide that we need a "real" `GatherNdLayer` that exactly accepts the different `indices` layout used in `tf.gather_nd`, then we can also think about making this a flag to `GatherLayer` or something. Or name this layer something entirely else then.

But of course I'm also open for other suggestions ^^

returnn/tf/layers/basic.py
Zettelkasten: is this how it should be? `__init__` itself can only handle `position` arguments of type `LayerBase`, not `int`. But yes, one can of course write it as a `int` within a config.

returnn/tf/layers/basic.py
Zettelkasten: This is for the common_axes_position argument, which I just added for completeness s.t. I can call the function with `translate_input_axis(ax, gather_axis, *axes)`.
I think Pycharm gave me a warning that the argument was unused even though I had the `del common_axes_position` statement there. I can try without :D

returnn/tf/layers/basic.py
albertz: `__init__` should be able to handle `int` directly.

albertz:
> I think it does not make much sense to have two layers that both take inputs similar to what `tf.gather` takes: No matter if the `position` input is a constant int or another layer, has a batch axis or not, etc., the functionality is essentially the same and I wouldn't separate that into two different layers.

My intention for `GatherLayer`, when this was to be extended for layer input, was to assert that the layer input does not have a batch axis. Exactly because of this reason, to separate it in functionality from `GatherNdLayer`.

But of course, we can deviate from this original intention.

> The same applies to what we right now call `GatherLayer` and `GatherNdLayer`: Both essentially do exactly the same, and I would want to keep it a single layer I think.

Actually not. Right now, they do not do the same. `GatherNdLayer` is the batched version, `GatherLayer` is the non-batched version. That's currently the main difference.

> What about doing as follows:
> 
> * use the code from my PR and call it `GatherLayer`, and support both constant/non-constant as well as batched/non-batched inputs there (essentially how it is in this PR now), but also stay backward compatible to the API of the existing `GatherNdLayer` (should be already, but we have to test this more)
> * deprecate the `GatherNdLayer`, and make it an effective alias for the `GatherLayer` until we maybe remove it entirely in the future
> * update my proposed implementation for `GatherLayer` in such a way that it remains backwards compatible to older setups: In particular to TF versions where `tf.gather` does not support the `batch_dims` parameter

Yes, sounds all good to me. (@JackTemaki @patrick-wilken what do you say?)

Backward compatibility is important for us, so we can deprecate things, and also remove it and make it an alias to `GatherLayer` if it is fully compatible (and we should be able to do so -- but this should be tested). But not completely remove it. Old configs should not break. That's important.

The update for older TF versions is less important now. We could already merge it before that. It only becomes important once we make `GatherNdLayer` an alias. But we can also do that later. (And then maybe at that later time, we can reconsider what TF compatibility we want anyway.)

> This would mean that we essentially do not have duplicate code, and updating configs is super simple: Just rename `gather_nd` to `gather`.
> But on the other hand, the implementation in this PR will become more complicated if we ensure backwards compatibility - should be okay though I think.

But I thought you said it already works?

But anyway, you don't have to make it an alias right away. A PR should be kept simple. It can be another PR. (Unless you think, maybe we would mess up now and then it becomes difficult or impossible later to get the right behavior, without further options. Maybe there are edge cases you did not consider here. Did you actually test it by just making it an alias?)

> If we in the future decide that we need a "real" `GatherNdLayer` that exactly accepts the different `indices` layout used in `tf.gather_nd`, then we can also think about making this a flag to `GatherLayer` or something. Or name this layer something entirely else then.

Yes. It depends how much it would complicate the logic in the layer. If this is simple (to add to the layer, and also to use for the user), then we can do that. Otherwise maybe it is better as a separate layer (with some new name). (I never actually needed this, so not sure how important that is.)

returnn/tf/layers/basic.py
albertz: I don't understand. Why would you call the function with random arguments you don't need?

Zettelkasten:
> > But on the other hand, the implementation in this PR will become more complicated if we ensure backwards compatibility - should be okay though I think.
> 
> But I thought you said it already works?

With this I mean being compatible to TF versions where `tf.gather` does not support batching.
If one uses such old TF version, the current `GatherNdLayer` implementation right now should still work, but the implementation in this PR will probably not as it relies on batching in `tf.gather`. But that's the only thing that is not compatible to the current `GatherNdLayer` implementation I think.

> But anyway, you don't have to make it an alias right away. A PR should be kept simple. It can be another PR. (Unless you think, maybe we would mess up now and then it becomes difficult or impossible later to get the right behavior, without further options. Maybe there are edge cases you did not consider here. Did you actually test it by just making it an alias?)

That sounds good! In this case, it does not matter at all that this implementation is not compatible with some old TF versions.

I think there only is a single test case using a `gather_nd` layer in this repo (namely `test_extra_scatter_nd_search_train`), and I verified for that one that the current `GatherNdLayer` implementation and the one in this PR compute the same results.
They are also compatible API-wise, with the difference that you have to specifiy a `axis`-parameter in this PR (if we decide to make this an alias in the future, we would just set that automatically).

I don't have any configs myself that used the current `GatherNdLayer`, so I can't really test this further.

> Yes. It depends how much it would complicate the logic in the layer. If this is simple (to add to the layer, and also to use for the user), then we can do that. Otherwise maybe it is better as a separate layer (with some new name). (I never actually needed this, so not sure how important that is.)

Yes, let's worry about that when there is actually a need for it :)

albertz:
> But on the other hand, the implementation in this PR will become more complicated if we ensure backwards compatibility - should be okay though I think.
> 
> With this I mean being compatible to TF versions where `tf.gather` does not support batching.
> If one uses such old TF version, the current `GatherNdLayer` implementation right now should still work, but the implementation in this PR will probably not as it relies on batching in `tf.gather`. But that's the only thing that is not compatible to the current `GatherNdLayer` implementation I think.

I see. For the `GatherNdLayer`, I implemented this `batch_gather` function, which can be useful for `GatherLayer` in the future to support older TF versions. But this can be done later.

> > But anyway, you don't have to make it an alias right away. A PR should be kept simple. It can be another PR. (Unless you think, maybe we would mess up now and then it becomes difficult or impossible later to get the right behavior, without further options. Maybe there are edge cases you did not consider here. Did you actually test it by just making it an alias?)
> 
> That sounds good! In this case, it does not matter at all that this implementation is not compatible with some old TF versions.

Yea. At least for now. We should clean it up at some point. Also, we should already extend the documentation right now in both layers to explain and clarify all of this. E.g. that `GatherNdLayer` is deprecated and `GatherLayer` should be used instead.
 
> I think there only is a single test case using a `gather_nd` layer in this repo (namely `test_extra_scatter_nd_search_train`), and I verified for that one that the current `GatherNdLayer` implementation and the one in this PR compute the same results.
> They are also compatible API-wise, with the difference that you have to specifiy a `axis`-parameter in this PR (if we decide to make this an alias in the future, we would just set that automatically).

The default value `axis="T"` might be reasonable. And then they are really the same.

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
-  This is basically ``x[position]`` in some axis.
-  See also :class:`SliceLayer`.
-  See also :class:`GatherNdLayer`, which works on the time-axis, but supports different positions per batch.
+  Gathers slices on a specified axis from the input layer using indices from a ``position`` layer.
+  If the input is a layer of the shape ``[B,D,F1]``, and indices of shape ``[B,F2]``, this will yield output of the
+  shape ``[B,F2,F1]`` where
 
-  (Currently `position` can just be an integer, so it's much less generic than ``tf.gather``.
-   However, we might extend this to lists, arrays, or even other layers.)
+  ``output[b,f2,f1] = input[b,indices[b,f2],f1]``
```
albertz, line 920: I would also explain the simple case, i.e. just give the example `input[position]`.

And extend it to explain the difference to `GatherNdLayer` (can be short, just mention this is a more generic version of `GatherNdLayer`).

And extend the `GatherNdLayer` documentation to say it is deprecated and `GatherLayer` should be used instead.

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
   layer_class = "gather"
 
-  def __init__(self, axis, position, **kwargs):
+  def __init__(self, position, axis, **kwargs):
     """
-    :param str axis:
-    :param int position:
+    :param LayerBase|int position: Layer containing the indices used to select the slices of the input from.
+    If another layer, must be of type ``int32`` or ``int64``.
+    Can also specify a constant ``int``.
```
albertz, line 934: Indent.
```suggestion
      If another layer, must be of type ``int32`` or ``int64``.
      Can also specify a constant ``int``.
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
-    axis = self.input_data.get_axis_from_description(axis, allow_int=False)
-    assert isinstance(position, int)  # nothing else implemented yet
-    slices = [slice(None, None)] * axis + [position]  # type: typing.List[typing.Union[slice, int]]
-    self.output.placeholder = self.input_data.placeholder[slices]
+    self.position = position
+
+    input_data = self.input_data
+    if isinstance(position, int):
+      constant_kwargs = dict(name="constant_position", network=kwargs["network"], value=position, dtype='int32')
+      position = ConstantLayer(output=ConstantLayer.get_out_data_from_opts(**constant_kwargs), **constant_kwargs)
```
albertz, line 943: Do not create a dummy layer. This is ugly. And no real need for it.

You could operate on `Data` directly. Use `Data.from_tensor`.

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
+
+    # determine all common axes of input_data and position_data
+    common_axes_input, common_axes_position, input_axes, position_axes = (
+      cls._get_common_input_position_axes(input_data, position_data, gather_axis))
+
+    # (BatchAxes.., PositionAxes.., InputAxes..)
+    shape = (
+      [input_data.batch_shape[ax] for ax in common_axes_input if not ax == input_data.batch_dim_axis] +
+      [position_data.batch_shape[ax] for ax in position_axes if not ax == position_data.batch_dim_axis] +
+      [input_data.batch_shape[ax] for ax in input_axes if not ax == input_data.batch_dim_axis])
```
albertz, line 1077:
```suggestion
      [input_data.batch_shape[ax] for ax in common_axes_input if ax != input_data.batch_dim_axis] +
      [position_data.batch_shape[ax] for ax in position_axes if ax != position_data.batch_dim_axis] +
      [input_data.batch_shape[ax] for ax in input_axes if ax != input_data.batch_dim_axis])
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    out_type = input_data.get_kwargs()
+    out_type["name"] = "%s_output" % name
+    out_type["shape"] = shape
+    out_type["beam"] = SearchBeam.get_combined_beam(input_data.beam, position_data.beam)
+
+    # Take axes from input_data if they exist there, otherwise from position_data
+    for axis_kind in ["batch_dim_axis", "time_dim_axis", "feature_dim_axis"]:
```
albertz, line 1084:
```suggestion
    for axis_kind in Data.SpecialAxesNames:
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
     """
-    out = get_concat_sources_data_template(sources, name="%s_output" % name)
-    axis = out.get_axis_from_description(axis, allow_int=False)
-    return out.copy_template_excluding_axis(axis)
+    input_data = get_concat_sources_data_template(sources)
+    gather_axis = input_data.get_axis_from_description(axis, allow_int=False)
+
+    if isinstance(position, int):
+      position = ConstantLayer.get_out_data_from_opts(
+        name="constant_position", network=kwargs["network"], value=position, dtype="int32")
```
albertz, line 1065: You don't need to do that so complicated.
```suggestion
      position = Data(name="constant_position", dtype="int32", shape=())
```

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
+
+    # (BatchAxes.., PositionAxes.., InputAxes..)
+    shape = (
+      [input_data.batch_shape[ax] for ax in common_axes_input if not ax == input_data.batch_dim_axis] +
+      [position_data.batch_shape[ax] for ax in position_axes if not ax == position_data.batch_dim_axis] +
+      [input_data.batch_shape[ax] for ax in input_axes if not ax == input_data.batch_dim_axis])
+    out_type = input_data.get_kwargs()
+    out_type["name"] = "%s_output" % name
+    out_type["shape"] = shape
+    out_type["beam"] = SearchBeam.get_combined_beam(input_data.beam, position_data.beam)
```
albertz, line 1081: You also should set `available_for_inference`.

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
+
+    # Take axes from input_data if they exist there, otherwise from position_data
+    for axis_kind in ["batch_dim_axis", "time_dim_axis", "feature_dim_axis"]:
+      input_axis, position_axis = getattr(input_data, axis_kind), getattr(position_data, axis_kind)
```
albertz, line 1085: Actually, there is some special logic for `feature_dim_axis`, which you should also follow.
In that case, you should check `input_data.feature_dim_axis_or_unspecified is NotSpecified`, and if so, just do not set it explicitly to an integer. 
(This is somewhat not nice here, as this needs a special case... But whether it is not specified, i.e. automatically derived or explicitly specified can change the behavior in some cases, e.g. when dropping some dimension, when reordering the axes, etc.)

returnn/tf/layers/basic.py
```diff
@@ -913,37 +913,204 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      if input_axis is not None and not input_axis == gather_axis:
```
albertz, line 1086:
```suggestion
      if input_axis is not None and input_axis != gather_axis:
```

albertz:
There is one actual difference to the old implementation of `GatherLayer`: In case that `position` is an `int`, it would not change the order of the axes at all. And this was an important property (for performance reasons, and also for some behavior expectations in other [external code](https://github.com/rwth-i6/pytorch-to-returnn)).

I think we should preserve this behavior (maybe in general, if `common_axes_input` is empty). If `common_axes_input` is empty, no extra `transpose` should be needed (and also not be done).

Zettelkasten:
> Yea. At least for now. We should clean it up at some point. Also, we should already extend the documentation right now in both layers to explain and clarify all of this. E.g. that `GatherNdLayer` is deprecated and `GatherLayer` should be used instead.

Great yes, I tried to clarify this now.

> The default value `axis="T"` might be reasonable. And then they are really the same.

Yes maybe. The current documentation of `GatherNdLayer` says: "indices into first axis (excluding batch) of the input" though, so I think it was position based instead. I didn't look at the code though.

> There is one actual difference to the old implementation of `GatherLayer`: In case that `position` is an `int`, it would not change the order of the axes at all. And this was an important property (for performance reasons, and also for some behavior expectations in other [external code](https://github.com/rwth-i6/pytorch-to-returnn)).
>
> I think we should preserve this behavior (maybe in general, if `common_axes_input` is empty). If `common_axes_input` is empty, no extra `transpose` should be needed (and also not be done).

Oh yes! That was actually an error in my implementation - now (as it should be) the output tensor always has the shape `[BatchAxes, InputAxesBeforeGatherAxis, PositionAxes, InputAxesAfterGatherAxis]`. So when `position` has dim 0, then this shape exactly matches the input shape minus the gather axis. Furthermore this case, all permutations passed to `tf.transpose` are identities. We don't really know if this has a performance impact then (as we already discussed in #406), but if it does, this sounds like something we should fix globally (e.g. by adding a global wrapper around this somewhere that only calls `tf.transpose` if the permutation is not the identity, or else returns a no-op) I think.

returnn/tf/layers/basic.py
Zettelkasten: @albertz Ah okay, thanks, I didn't know about this. Can you check if it's correct now? I'm still unsure.

returnn/tf/layers/basic.py
```diff
@@ -913,44 +913,227 @@ def transform_config_dict(cls, d, network, get_layer):
...
+        continue
+      output_axis = self.output.get_batch_axis_excluding_batch(
+        self._translate_input_axis(input_axis, old_gather_axis, common_axes_input, input_axes, position_axes))
+      size_placeholder[output_axis] = size
+    for position_axis, size in position_data.size_placeholder.items():
+      position_axis = position_data.get_batch_axis(position_axis)
+      output_axis = self.output.get_batch_axis_excluding_batch(self._translate_position_axis(
+        position_axis, old_gather_axis, common_axes_position, position_axes))
+      size_placeholder.setdefault(output_axis, size)
+    self.output.size_placeholder = size_placeholder
```
albertz, line 982: The `size_placeholder` should be set in `get_out_data_from_opts`, not here in `__init__`.
(Always for existing dyn sizes. The only exception is currently for new dyn sizes, and even that will change.)

tests/test_TFNetworkLayer.py
```diff
@@ -1805,6 +1805,63 @@ def test_SliceLayer_NCHW():
...
+    print('values [B, D, T, F1]:', repr(values_seqs))
+    print('position [B, T, F2]:', position_seqs)
+    print('produced output [B, T, F2, F1]:', out_seqs)
+
+    # test values
+    for b in range(batch_dim):
+      for t in range(seq_lens[b]):
+        for f2 in range(feature_dim2):
+          for f1 in range(feature_dim1):
+            np.testing.assert_almost_equal(out_seqs[b, t, f2, f1], values_seqs[b, position_seqs[b, t, f2], t, f1])
```
albertz, line 1862: Can you add another test, much simpler, where you use `position=5` or so (i.e. just a single integer)?

returnn/tf/layers/basic.py
```diff
@@ -913,44 +913,227 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      if input_axis is not None and input_axis != old_gather_axis:
+        out_type[axis_kind] = cls._translate_input_axis(
+          input_axis, old_gather_axis, common_axes_input, input_axes, position_axes)
+      elif position_axis is not None:
+        out_type[axis_kind] = cls._translate_position_axis(
+          position_axis, old_gather_axis, common_axes_position, position_axes)
+      else:
+        out_type[axis_kind] = None
+    if input_data.feature_dim_axis_or_unspecified is NotSpecified:
+      out_type["feature_dim_axis"] = NotSpecified
```
albertz, line 1109: Maybe the code should be restructured and simplified. Or this case should just be kept separate. Not sure how to simplify it in a clean way. But the logic should be like:
```suggestion
    if input_data.feature_dim_axis_or_unspecified is NotSpecified and position_data.feature_dim_axis_or_unspecified is NotSpecified:
      out_type["feature_dim_axis"] = NotSpecified
    elif input_data.feature_dim_axis_or_unspecified is NotSpecified:
      assert position_data.feature_dim_axis_or_unspecified is not NotSpecified
      if position_data.feature_dim_axis is not None:
        out_type["feature_dim_axis"] = cls._translate_position_axis(
          position_data.feature_dim_axis, old_gather_axis, common_axes_position, position_axes)
      else:
        out_type["feature_dim_axis"] = NotSpecified
    elif position_data.feature_dim_axis_or_unspecified is NotSpecified:
      assert input_data.feature_dim_axis_or_unspecified is not NotSpecified
      if input_data.feature_dim_axis is not None:
        out_type["feature_dim_axis"] = cls._translate_position_axis(
          input_data.feature_dim_axis, old_gather_axis, common_axes_position, input_axes)
      else:
        out_type["feature_dim_axis"] = NotSpecified
    else:
      assert input_data.feature_dim_axis_or_unspecified is not NotSpecified
      assert position_data.feature_dim_axis_or_unspecified is not NotSpecified
      pass  # keep the logic as before
```

returnn/tf/layers/basic.py
```diff
@@ -913,44 +913,227 @@ def transform_config_dict(cls, d, network, get_layer):
...
+
+    return Data(**out_type)
```
albertz, line 1111: The `size_placeholder` should be set here as well.

tests/test_TFNetworkLayer.py
Zettelkasten: Probably a good idea. I added one now that also verifies that the output axes are not shuffled around if `position` is a single `int`.

returnn/tf/layers/basic.py
Zettelkasten: What a beast! :D For now I just used your suggestion (there was a typo in line 1120, it should be `cls._translate_input_axis(input_data.feature_dim_axis, ...)`), we may change that later.

PR #410, opened 2020-12-11
Don't use depthwise conv if input data dim is 1

vieting:
Currently, when the input data dim (`n_in`) is 1 (e.g. raw waveform) and the default value of `groups` (also 1) is used, the depthwise convolution is used. This does not work, when `strides` is different from 1, because for the second dimension the stride is set to 1 and tensorflow does not yet support different strides for row and column dimensions:
> InvalidArgumentError: Current implementation only supports equal length strides in the row and column dimensions.
See also https://github.com/tensorflow/tensorflow/issues/33005

The simplest fix is not to use `tf.nn.depthwise_conv2d` when `n_in` is 1. Not sure about performance when `strides=(1,)`, in which case `tf.nn.depthwise_conv2d` works, but having that as another special case would add complexity which might not be needed.


vieting:
Oh, I just saw that @JackTemaki already did the same thing.

PR #412, opened 2020-12-16
Fix subset param import

curufinwe:
Fix bug where custom_parameter_importer = "subset" would not operate on parameters inside rec layers.

### DIFF
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -2042,8 +2042,10 @@ class TFNetwork(object):
       print('iter: ', l)
       yield l
       if isinstance(l, RecLayer) and isinstance(l.cell, _SubnetworkRecCell):
-        yield from l.cell.net.iter_layers_recursively()
-        yield from l.cell.input_layers_net.iter_layers_recursively()
+        for l in l.cell.net.iter_layers_recursively():  # TODO: replace with yield from once python2.7 support is dropped
+          yield l
+        for l in l.cell.input_layers_net.iter_layers_recursively():
+          yield l
 
 
 class TFNetworkParamsSerialized(object):
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -2036,6 +2036,15 @@ class TFNetwork(object):
       return root_net.register_search_choices_for_beam(beam, search_choices)
     self._map_search_beam_to_search_choices[beam] = search_choices
 
+  def iter_layers_recursively(self):
+    from .layers.rec import RecLayer, _SubnetworkRecCell
+    for l in self.layers.values():
+      print('iter: ', l)
+      yield l
+      if isinstance(l, RecLayer) and isinstance(l.cell, _SubnetworkRecCell):
+        yield from l.cell.net.iter_layers_recursively()
+        yield from l.cell.input_layers_net.iter_layers_recursively()
+
 
 class TFNetworkParamsSerialized(object):
   """
@@ -2691,7 +2700,7 @@ class CustomCheckpointLoader:
     self.obsolete_var_names = [v for v in sorted(self.var_ckpt_names) if v not in self.var_net_names]
     self.custom_param_importers = [
       self.CustomParamImporter(layer=layer, checkpoint_loader=self)
-      for layer in network.layers.values() if layer.custom_param_importer] if network else []
+      for layer in network.iter_layers_recursively() if layer.custom_param_importer] if network else []
 
   def __repr__(self):
     keys = ["filename", "params_prefix", "load_if_prefix", "ignore_missing", "network"]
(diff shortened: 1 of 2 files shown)


albertz:
See failing test, i.e. code style.

curufinwe:
There is one remaining warning that is not easily fixable: I import _SubnetworkRecCell to do an isinstance check. Can this warning be ignored or how should I fix that?

albertz:
Actually, we should not do it this way.
We should introduce more generally a function `iter_sub_layers` on `LayerBase`. And this should be overridden by `SubnetworkLayer` and `RecLayer`.
Also, `_SubnetworkRecCell ` actually internally has several subnetworks, and you probably should iterate through all of them. Specifically, `output_layers_net`, `input_layers_net` and `net`. See `get_layer_from_outside`. At least for this use case, you probably want all. Btw, I'm not exactly sure whether there can be duplicates, and whether you want to care about that (keep track in some `visited = set()` var).

Note that this is a bit inconsistent to the behavior of `get_sub_layer`, though, which requires that the layer (output) also can be used (which is not the case for layers within a loop). Not sure if we should maybe introduce two separate functions `iter_all_sub_layers` vs `iter_sub_layers` or so. Or just do not care about that. TF would anyway tell you when you use it wrongly. So I would vote to just implement `iter_sub_layers` and iterate all.

returnn/tf/network.py
```diff
@@ -2036,6 +2036,20 @@ def register_search_choices_for_beam(self, beam, search_choices):
...
 
+  def iter_layers_recursively(self):
+    """
+    Yield all sub-layers of this network (including layers inside rec-layers)
+    """
+    from .layers.rec import RecLayer, _SubnetworkRecCell
+    for layer in self.layers.values():
+      yield layer
+      if isinstance(layer, RecLayer) and isinstance(layer.cell, _SubnetworkRecCell):
+        # TODO: replace with yield from once python2.7 support is dropped
```
albertz, line 2047: Remove the TODO.

returnn/tf/network.py
```diff
@@ -2036,6 +2036,20 @@ def register_search_choices_for_beam(self, beam, search_choices):
...
+        for sublayer in layer.cell.net.iter_layers_recursively():
```
albertz, line 2048: See my comment. Don't do that. Introduce some function `LayerBase.iter_sub_layers`.

returnn/tf/network.py
```diff
@@ -2036,6 +2036,20 @@ def register_search_choices_for_beam(self, beam, search_choices):
       return root_net.register_search_choices_for_beam(beam, search_choices)
     self._map_search_beam_to_search_choices[beam] = search_choices
 
+  def iter_layers_recursively(self):
+    """
+    Yield all sub-layers of this network (including layers inside rec-layers)
+    """
+    from .layers.rec import RecLayer, _SubnetworkRecCell
+    for layer in self.layers.values():
```
albertz, line 2044: Do this in a deterministic way (deterministic also in Python <=3.6).
I.e. like:
```suggestion
    for _, layer in sorted(self.layers.items()):
```

returnn/tf/network.py
```diff
@@ -2036,6 +2036,20 @@ def register_search_choices_for_beam(self, beam, search_choices):
       return root_net.register_search_choices_for_beam(beam, search_choices)
     self._map_search_beam_to_search_choices[beam] = search_choices
 
+  def iter_layers_recursively(self):
+    """
+    Yield all sub-layers of this network (including layers inside rec-layers)
```
albertz, line 2041: Add the return type. I think it should be:
```suggestion
    :return: Yield all layers and sub-layers of this network
    :rtype: typing.Iterator[LayerBase]
```

returnn/tf/network.py
```diff
@@ -2036,6 +2036,20 @@ def register_search_choices_for_beam(self, beam, search_choices):
...
+    from .layers.rec import RecLayer, _SubnetworkRecCell
+    for layer in self.layers.values():
+      yield layer
+      if isinstance(layer, RecLayer) and isinstance(layer.cell, _SubnetworkRecCell):
+        # TODO: replace with yield from once python2.7 support is dropped
+        for sublayer in layer.cell.net.iter_layers_recursively():
+          yield sublayer
+        for sublayer in layer.cell.input_layers_net.iter_layers_recursively():
+          yield sublayer
+
```
michelwi, line 2052: You should also iterate over the output layers:
```suggestion
        for sublayer in layer.cell.output_layers_net.iter_layers_recursively():
          yield sublayer
          
```
or however you resolve albertz comment.

albertz:
What is the state here?

albertz:
Note: There is now the function `get_all_layers_deep`, which can be used instead of `iter_layers_recursively`.

albertz:
Ok, I removed `iter_layers_recursively` now, and just used `get_all_layers_deep`.

@curufinwe @michelwi can you check it's ok, so that we can merge?

PR #414, opened 2020-12-17
eval output_per_seq: fixed printing long values

patrick-wilken:


returnn/tf/engine.py
```diff
@@ -1802,6 +1802,8 @@ def extra_fetches_callback(seq_idx, seq_tags, **extra_fetches_out):
             res = ';'.join(value_list)
             f.write(res + '\n')
         elif output_per_seq_file_format == "py":
+          print_threshold = numpy.get_printoptions()["threshold"]
+          numpy.set_printoptions(threshold=sys.maxsize)  # make sure we don't print abbreviated values containing  '...'
```
albertz, line 1806: You should use a context manager for this, like:
`with numpy.printoptions(threshold=sys.maxsize):`

albertz: But also maybe more localized to exactly the place where this matters. Not sure exactly where?

albertz (approved):

PR #416, opened 2020-12-22
LinearLayer, allow passing custom weights/bias, not using param

vieting:
Add option to pass custom weights/bias to `LinearLayer`, similar to https://github.com/rwth-i6/returnn/commit/0645636034151288f5bdda72a155cde8051a7ece


albertz:
This should not be needed. We have `DotLayer` for that (for the `matmul`), and for bias, you can simply use `CombineLayer`.

vieting:
Oh I somehow oversaw that...

PR #422, opened 2021-01-20
SplitDimsLayer: Proper splits of axes with dynamic length

Zettelkasten:
If an axis has a dynamic length, and this is a non-trivial split (i.e. all other constant dims are not 1, where this is just expanding the dims), we pad it to the multiples of the product of all constant dims, via `pad_to_multiples=True`.

This commit also fixes that the axes for `size_placeholder` were incorrectly calculated when the split axis < batch dim axis, introduced in aa3b92a4873191460dde6547a7f38f8ee366c7d2 (I think.)


Zettelkasten:
Btw, does somebody know why the [`Util` test](https://github.com/rwth-i6/returnn/runs/1737082641?check_suite_focus=true) is failing here?

albertz:
> Btw, does somebody know why the [`Util` test](https://github.com/rwth-i6/returnn/runs/1737082641?check_suite_focus=true) is failing here?

Not really. I made a new issue about it, #424. We should fix that first, such that the tests run through again. Then better rebase, such that this PR passes all tests.

returnn/tf/layers/basic.py
```diff
@@ -2759,45 +2759,89 @@ class SplitDimsLayer(_ConcatInputLayer):
...
+  Also see :class:`MergeDimsLayer` which can undo this operation.
   """
   layer_class = "split_dims"
 
-  def __init__(self, axis, dims, **kwargs):
+  def __init__(self, axis, dims, pad_to_multiples=None, pad_value=0, **kwargs):
     """
     :param str axis: e.g. "F"
     :param tuple[int]|list[int] dims: what the axis should be split into. e.g. (window, -1)
+    :param Optional[bool] pad_to_multiples: If true, will pad along the axis to the next multiple of the remaining dims.
```
albertz, line 2781:
```suggestion
    :param bool|None pad_to_multiples: If true, will pad along the axis to the next multiple of the remaining dims.
```
Also, from this description, I don't quite understand what this means.

returnn/tf/layers/basic.py
```diff
@@ -2759,45 +2759,89 @@ class SplitDimsLayer(_ConcatInputLayer):
...
+      assert len([d for d in dims if d == -1]) == 1
+      old_size = old_shape[axis]
+      pad_size = (-old_size) % constant_size
+
+      paddings = [(0, 0)] * axis + [(0, pad_size)] + [(0, 0)] * (data.batch_ndim - axis - 1)
+      data.placeholder = tf.pad(data.placeholder, paddings=paddings, constant_values=pad_value)
+
+      axis_wo_batch = data.get_batch_axis_excluding_batch(axis + dims.index(-1))
+      dyn_size = (old_size + pad_size) // constant_size
+      self.output.size_placeholder[axis_wo_batch] = dyn_size
```
albertz, line 2815: This is wrong in general, right?
You should use the old `size_placerholder` instead of `old_size`, right?
Also, I thought this code is also for the case that it's not of dynamic length? (Or if not supported, just check and throw `NotImplementedError` otherwise.)

returnn/tf/layers/basic.py
```diff
@@ -2818,29 +2862,35 @@ def _map_old_axis_to_new_axis(cls, split_axis, dims, old_axis, require_exact=Tru
...
     return old_axis
 
   @classmethod
-  def get_out_data_from_opts(cls, name, axis, dims, sources=(), **kwargs):
+  def get_out_data_from_opts(cls, name, axis, dims, pad_to_multiples=None, sources=(), **kwargs):
     """
     :param str name:
     :param str|int axis:
     :param tuple[int] dims:
+    :param Optional[bool] pad_to_multiples:
```
albertz, line 2876:
```suggestion
    :param bool|None pad_to_multiples:
```

tests/test_TFNetworkLayer.py
```diff
@@ -1123,13 +1123,19 @@ def test_SplitDimsLayer_simple_time2():
...
-  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, -1, 5)), (2, 3, 5))
-  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, 3, -1)), (2, 3, 5))
-  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, 3, -1, 1)), (2, 3, 5, 1))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=3 * 5, new_dims=(3, -1), pad_to_multiples=False), (3, 5))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=3 * 5, new_dims=(3, 5), pad_to_multiples=False), (3, 5))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=3 * 5, new_dims=(5, -1), pad_to_multiples=False), (5, 3))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(-1, 3, 5), pad_to_multiples=False), (2, 3, 5))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, -1, 5), pad_to_multiples=False), (2, 3, 5))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, 3, -1), pad_to_multiples=False), (2, 3, 5))
+  assert_equal(SplitDimsLayer._resolve_dims(old_dim=2 * 3 * 5, new_dims=(2, 3, -1, 1), pad_to_multiples=False), (2, 3, 5, 1))
```
albertz, line 1132: Why did you change them? `pad_to_multiples=False` is anyway the default, right? Then better just leave it as-is.

returnn/tf/layers/basic.py
Zettelkasten: Ah yes, this wrong. I will have to fix that, and use `size_placeholder` instead (if it exists, else just the static dim)

tests/test_TFNetworkLayer.py
Zettelkasten: I set the default `pad_to_multiples=None` for `__init__` and `get_out_data_from_opts`, but not for `_resolve_dims` as that function does not know the actual input `Data` (thus we cannot infer it when the argument is `None`)
I can change it and set the default `pad_to_multiples=False` for `_resolve_dims`.

albertz: I see. Yes, maybe that's better (default pad_to_multiples=False for _resolve_dims).

albertz:
Can you rebase to master?

PR #423, opened 2021-01-20
EvalLayer: Pass eval_locals also to out_type

Zettelkasten:


returnn/tf/layers/basic.py
```diff
@@ -5776,7 +5777,15 @@ def get_out_data_from_opts(cls, n_out=NotSpecified, out_type=None, sources=(), *
...
       elif callable(out_type):
-        out_type_ = out_type  # just overwrite
+        def call_out_type_with_eval_locals(**out_type_kwargs):
+          """
+          :param out_type_kwargs:
+          :rtype: Data
+          """
+          out_type_kwargs = out_type_kwargs.copy()
+          out_type_kwargs.update(eval_locals or {})
+          return out_type(out_type_kwargs)
```
albertz, line 5787:
```suggestion
          return out_type(**out_type_kwargs)
```
?
(Did you test this?)

Zettelkasten: Sorry, my original code did not run in Python 2, and I did not test the change I pushed just now :sweat_smile:

albertz:
In principle this is fine.

Note however, `eval_locals` is somewhat not really needed. E.g. when you have `"eval_locals": {"a": my_magic_constant}, "eval": "source(0) * a"`, you could just write instead:

1. `"eval": "source(0) * %f" % my_magic_constant`
2. ```
   def my_func(source, **kwargs):
     return source(0) * my_magic_constant

   ...
   "eval": my_func ...
   ...
   ```
   (I think I prefer the last version. But that is maybe personal taste.)

Also, in general, once the usage of `EvalLayer` becomes too complex (almost always when you need to have a custom `out_type`), I wonder whether we could instead introduce some other simple generic layer that you don't need this custom eval. Such that this simplifies the config, and also makes it more robust for any errors.

Zettelkasten:
> Note however, `eval_locals` is somewhat not really needed. E.g. when you have `"eval_locals": {"a": my_magic_constant}, "eval": "source(0) * a"`, you could just write instead [...]

That is true. One use case for this could be that you want to use an `out_type`-function multiple times, but with different parameters in each use. (Then again, in this case it is probably also worth writing a sophisticated layer for this.)

> Also, in general, once the usage of `EvalLayer` becomes too complex (almost always when you need to have a custom `out_type`), I wonder whether we could instead introduce some other simple generic layer that you don't need this custom eval. Such that this simplifies the config, and also makes it more robust for any errors.

I agree, this is really not ideal. However I find it very useful to prototype with, and then later I can iterate and think of what new layers would be necessary for a clean solution. It is much more easier to "hack" something together in such an eval function (but okay, you could as well define your own layer in your config, that has almost the same effect with slightly more overhead though).

My current use case (maybe we should discuss this elsewhere) is inside a `RecLayer`, and I want to calculate different things (with different out types) depending on whether the loop is unrolled or not.
Some layers already do that (e.g. `DotLayer`, when using an axis `T?` somewhere, or `GatherLayer`), but most other's will not work that simply.

More specifically, I use this in a implementation for LSH self-attention.
I use a `RecHistoryLayer` (using my WIP version of #391) to accumulate all past attention keys, and then "chunk" this into a fixed window size using `SplitDimsLayer` along the rec-time-axis (using #422).
Then, I use my eval Layer to calculate the window slices that correspond to keys that also contain the query of the same time step, which I can then select using a `GatherLayer`.
But depending on whether the `RecLayer` is in a loop or unrolled, we calculate the attention for a single query or all queries at once (just like `:i` has a time-axis or not).
Thus in my case, my eval layers take as input `:i` and have `chunk_size` as parameter in `eval_locals`.
Then I want to calculate `tf.range(ceil(input_length / chunk_size))` if the input has a time axis with `input_length` many entries (when the loop is unrolled), or `x // chunk_size` if the input is a single position `x` (when the loop is recurrent).
And then of course I also need to adapt the output shape accordingly.

Maybe we can add a layer that takes two sub networks, and selects one of those depending on whether we are in a `RecLayer` or not.

albertz:
> Maybe we can add a layer that takes two sub networks, and selects one of those depending on whether the RecLayer is unrolled or not.

The original idea (hope) was that sth like that should never be needed.

> ... I want to calculate different things depending on whether the loop is unrolled or not.
> Some layers already do that (e.g. DotLayer or GatherLayer), but most other's will not work that simply.

Consider this a bug. Most (all) layers should in principle behave (logically) in the same way, whether they are unrolled (optimized out of loop) or not.

The idea is that whatever you calculate, you could do using (atomic) layers, and if RecLayer optimizes sth, it should result in the same results.

But yes, this is offtopic in this PR. Let's move the discussion elsewhere (other issue, or maybe [discussions](https://github.com/rwth-i6/returnn/discussions)).

Zettelkasten:
> But yes, this is offtopic in this PR. Let's move the discussion elsewhere (other issue, or maybe [discussions](https://github.com/rwth-i6/returnn/discussions)).

Okay yes, sounds good! I will think about this more myself first and try to narrow down what I actually need.
Then, if I have any further problem, I'm happy to start a thread in discussions or something alike.

PR #426, opened 2021-01-24
SwitchLayer: Constant `true_from`/`false_from` params

Zettelkasten:
This is especially useful to use `SwitchLayer` as a generic masking
layer. If you have some sort of `"mask"` for some `"input"` layer,
use `condition="mask", true_from="input", false_from=float("-inf")`
to e.g. mask out all unmasked values with -inf.

This commit also moves the logic from `ConstantLayer.get_out_data_from_opts` to `Data.template_from_constant` as it is very reusable. Sorry, that's a bit messy.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1649,6 +1649,8 @@ class SoftmaxOverSpatialLayer(_ConcatInputLayer):
     self.output_before_activation = OutputWithActivation(
       energy, act_func=tf.nn.log_softmax if log_space else tf.nn.softmax)  # (...,T)
     self.output.placeholder = self.output_before_activation.y
+    # Never allow inf in output, as softmax should remove all -inf values used for masking
+    self.allow_inf_in_output = False
 
   def get_dep_layers(self):
     """
@@ -1721,6 +1723,7 @@ class SeqLenMaskLayer(_ConcatInputLayer):
   """
   Masks some values away given the seq_len_source with mask_value.
   Also see :class:`SoftmaxOverSpatialLayer`.
+  Also see :class:`SwitchLayer`, which can be used to apply a generic mask.
   """
   layer_class = "seq_len_mask"
 
@@ -2134,23 +2137,7 @@ class ConstantLayer(LayerBase):
     :param bool with_batch_dim:
     :rtype: Data
     """
-    import numpy
-    if dtype is None:
-      if isinstance(value, int):
-        dtype = "int32"
-      elif isinstance(value, float):
-        dtype = "float32"
-      elif isinstance(value, bool):
-        dtype = "bool"
-      elif isinstance(value, numpy.ndarray):
-        dtype = str(value.dtype)
-      else:
-        raise TypeError("cannot handle value %r of type %r" % (value, type(value)))
-    shape = value.shape if isinstance(value, numpy.ndarray) else ()
-    return Data(
-      name="%s_const" % name,
-      shape=shape, batch_dim_axis=0 if with_batch_dim else None, time_dim_axis=None,
-      dtype=dtype)
+    return Data.template_from_constant(value, name="%s_const" % name, dtype=dtype, with_batch_dim=with_batch_dim)
 
 
 class GatingLayer(_ConcatInputLayer):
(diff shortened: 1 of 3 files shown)


returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
     self.false_from = false_from
     if isinstance(condition, bool):
       if condition:
-        self.output = true_from.output.copy("%s_output" % self.name)
+        true_data = true_from.output if isinstance(true_from, LayerBase) else Data.from_tensor(tf.identity(true_from))
```
albertz, line 6094: Why `tf.identity`? Should be `tf.constant`. Also you might want to set a name.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
...
+        self.output = true_data.copy("%s_output" % self.name)
+        if self._get_source_allow_inf_in_output(true_from):
+            self.allow_inf_in_output = True
       else:
-        self.output = false_from.output.copy("%s_output" % self.name)
+        false_data = false_from.output if isinstance(false_from, LayerBase) else Data.from_tensor(
+          tf.identity(false_from))
+        self.output = false_data.copy("%s_output" % self.name)
+        if self._get_source_allow_inf_in_output(false_from):
+            self.allow_inf_in_output = True
```
albertz, line 6103: You have a lot of unnecessary duplicate code here. Don't do that if you can avoid that.

You could do:
```
src = true_from if condition else false_from
```
And then the same code.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
...
     else:
       assert isinstance(condition, LayerBase)
       assert condition.output.dtype == "bool"
+      true_data = true_from.output if isinstance(true_from, LayerBase) else Data.from_tensor(tf.identity(true_from))
+      false_data = false_from.output if isinstance(false_from, LayerBase) else Data.from_tensor(tf.identity(false_from))
```
albertz, line 6108: No `tf.identity`.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
...
         condition=condition.output.copy_compatible_to(self.output, check_dtype=False, check_sparse=False).placeholder,
-        x=true_from.output.copy_compatible_to(self.output).placeholder,
-        y=false_from.output.copy_compatible_to(self.output).placeholder)
+        x=true_data.copy_compatible_to(self.output).placeholder,
+        y=false_data.copy_compatible_to(self.output).placeholder)
+      if self._get_source_allow_inf_in_output(true_from) or self._get_source_allow_inf_in_output(false_from):
+        self.allow_inf_in_output = True
+
+  @classmethod
+  def _transform_source_name(cls, source_name, get_layer):
```
albertz, line 6117: Make this function local to `transform_config_dict`. You are not using it anywhere else.
Also then you don't need the `get_layer` argument.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
...
+    :param ((str) -> LayerBase) get_layer: function to get or construct another layer
+    :return: how the input layer should appear in the transformed config dict
+    :rtype: LayerBase|float|int
+    """
+    if isinstance(source_name, str):
+      return get_layer(source_name)
+    return source_name
+
+  @classmethod
+  def _get_source_template(cls, source, name):
```
albertz, line 6129: Make the function local to `get_out_data_from_opts`, you are not using it anywhere else.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
     self.false_from = false_from
     if isinstance(condition, bool):
       if condition:
-        self.output = true_from.output.copy("%s_output" % self.name)
+        true_data = true_from.output if isinstance(true_from, LayerBase) else Data.from_tensor(tf.identity(true_from))
+        self.output = true_data.copy("%s_output" % self.name)
+        if self._get_source_allow_inf_in_output(true_from):
+            self.allow_inf_in_output = True
```
albertz, line 6097: Indent is wrong.

returnn/tf/layers/basic.py
albertz: Also, indent is wrong.

returnn/tf/layers/basic.py
```diff
@@ -6097,16 +6091,61 @@ def __init__(self, condition, true_from, false_from, **kwargs):
...
+    :param LayerBase|float|int source:
+    :param str name:
+    :rtype: Data
+    """
+    if isinstance(source, LayerBase):
+      return source.output.copy(name)
+    return Data.template_from_constant(source, name=name)
+
+  @classmethod
+  def _get_source_allow_inf_in_output(cls, source):
```
albertz, line 6140: Make the function local to `__init__`, you are not using it anywhere else.

albertz (changes requested):
See comments.

albertz (approved):

Issue #428, opened 2021-01-28
Data.copy_compatible_to does not add feature dim


Zettelkasten:
I have some tensor without a feature-dim, e.g. `[T]`, and want to copy it compatible to one with a feature dim, e.g. `[T,F]`, like in this test case here (`test_Data_copy_compatible_to_add_feature_dim`):

```python3
common_data = Data(name='energy', shape=(None, 3), batch_dim_axis=None)  # [T,F]
d1 = Data(name='mask', shape=(None,), batch_dim_axis=None, time_dim_axis=0, feature_dim_axis=None)  # [T]
d2 = d1.copy_compatible_to(common_data)
```

That however fails:
```
  File "/u/petrick/software/returnn/official/returnn/tf/util/data.py", line 1278, in copy_compatible_to
    line: assert data.feature_dim_axis is not None and v.feature_dim_axis is not None
    locals:
      data = <local> Data(name='energy', shape=(None, 3), batch_dim_axis=None, time_dim_axis=None, batch_shape_meta=[?,F|3])
      data.feature_dim_axis = <local> 1
      v = <local> Data(name='mask', shape=(1, None), batch_dim_axis=None, time_dim_axis=1, feature_dim_axis=None, batch_shape_meta=[1,T|?])
      v.feature_dim_axis = <local> None
```

I tried figuring out what went wrong in `Data.copy_compatible_to`, but I can't figure it out.
It would be great if somebody could take a look!

I pushed three test cases to 03416cdc3f0fe1b34bd20506ea0d88456b69528a. Note that `test_Data_copy_compatible_to_add_time_dim` already passes, this issue only occurs with the feature dim but not when adding a time dim.

albertz:
From briefly looking at it, I think `copy_compatible_to` gets confused about the spatial dim, which is unknown (without dim tag / seq len).

albertz:
We miss the logic (workaround / heuristic) which we have in `find_matching_dim_map`. I.e. to match an unknown time-dim if it's unique.

Zettelkasten:
I tried fiddling around with this for a while now, but ran into a few issues: I think the problem is that I don't quite know how `copy_compatible` exactly worked before, and how exactly it should behave in the future.

[...]

PR #429, opened 2021-01-28
Rewrite Data.copy_compatible_to, adds test cases & fixes for #428

Zettelkasten:
Three test cases for #428.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -114,8 +114,7 @@ class DimensionTag(object):
     return True
 
   def is_equal(self, other, ignore_feature_dim=False, allow_same_feature_dim=False, allow_same_spatial_dim=None,
-               treat_feature_as_spatial=False, broadcast_matches=False, unknown_spatial_matches=False,
-               different_static_matches=False):
+               treat_feature_as_spatial=False, broadcast_matches=False, unknown_spatial_matches=False):
     """
     Compares self to other for equality.
     Note that the default behavior is very restrictive.
@@ -129,7 +128,6 @@ class DimensionTag(object):
     :param bool treat_feature_as_spatial:
     :param bool broadcast_matches:
     :param bool unknown_spatial_matches:
-    :param bool different_static_matches:
     :rtype: bool
     """
     if allow_same_spatial_dim is None:
@@ -141,8 +139,6 @@ class DimensionTag(object):
     self_kind = self.kind
     other_kind = other.kind
     if self_kind == other_kind == self.Types.Feature and ignore_feature_dim:
-      return True
-    if different_static_matches and self.dimension is not None and other.dimension is not None:
       return True
     if treat_feature_as_spatial:
       if self_kind == self.Types.Feature:
@@ -1228,8 +1224,7 @@ class Data(object):
       raise ValueError("copy_compatible_to: self %r already has more dims than target data %r" % (self, data))
 
     is_equal_opts = dict(
-      allow_same_feature_dim=True, allow_same_spatial_dim=True, treat_feature_as_spatial=True, ignore_feature_dim=True,
-      different_static_matches=True)
+      allow_same_feature_dim=True, allow_same_spatial_dim=True, treat_feature_as_spatial=True, ignore_feature_dim=True)
     mapped_axes = data.find_matching_dim_map(v, list(range(v.batch_ndim)), is_equal_opts)  # maps v -> data
     assert len(mapped_axes) == v.batch_ndim
 
@@ -2531,8 +2526,6 @@ class Data(object):
         matching = [
           self_axis for self_axis in self.find_matching_dims(other_axis_dim_tag, is_equal_opts)
           if self_axis not in taken_self_axes]
-        assert len(matching) <= 1, 'cannot match the axes %s from %s to %s. Failing to match axis %s' % (
-          other_axes, other, self, other_axis)
       if not matching:
         # If still not, then also allow one single dyn_size to be unknown
         is_equal_opts["unknown_spatial_matches"] = True
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -114,7 +114,8 @@ class DimensionTag(object):
     return True
 
   def is_equal(self, other, ignore_feature_dim=False, allow_same_feature_dim=False, allow_same_spatial_dim=None,
-               treat_feature_as_spatial=False, broadcast_matches=False, unknown_spatial_matches=False):
+               treat_feature_as_spatial=False, broadcast_matches=False, unknown_spatial_matches=False,
+               different_static_matches=False):
     """
     Compares self to other for equality.
     Note that the default behavior is very restrictive.
@@ -128,6 +129,7 @@ class DimensionTag(object):
     :param bool treat_feature_as_spatial:
     :param bool broadcast_matches:
     :param bool unknown_spatial_matches:
+    :param bool different_static_matches:
     :rtype: bool
     """
     if allow_same_spatial_dim is None:
@@ -139,6 +141,8 @@ class DimensionTag(object):
     self_kind = self.kind
     other_kind = other.kind
     if self_kind == other_kind == self.Types.Feature and ignore_feature_dim:
+      return True
+    if different_static_matches and self.dimension is not None and other.dimension is not None:
       return True
     if treat_feature_as_spatial:
       if self_kind == self.Types.Feature:
@@ -167,7 +171,7 @@ class DimensionTag(object):
           return True
         if broadcast_matches and (self.dimension == 1 or other.dimension == 1):
           return True
-      if unknown_spatial_matches and ((self.dyn_size is None) != (other.dyn_size is None)):
+      if unknown_spatial_matches and ((self.dyn_size is None) or (other.dyn_size is None)):
         return True
     if self.description == other.description:
       return True
@@ -824,14 +828,16 @@ class Data(object):
     assert self.time_dim_axis is not None
     return self.copy_move_axis(self.time_dim_axis, time_dim_axis)
 
-  def copy_move_axis(self, old_axis, new_axis):
+  def copy_move_axis(self, old_axis, new_axis, keep_other_axes_intact=False):
     """
     :param int old_axis: counted with batch-dim
     :param int new_axis: counted with batch-dim
+    :param bool keep_other_axes_intact: if False, will shift all axes between old_axis and new_axis.
+      if True, all axes except old_axis and new_axis will remain unchanged
     :return: copy of myself with moved axis (see :func:`move_axis`)
     :rtype: Data
     """
-    from .basic import move_axis
+    from .basic import move_axis, swapaxes
     if old_axis < 0:
       old_axis += self.batch_ndim
       assert old_axis >= 0
@@ -851,6 +857,12 @@ class Data(object):
       """
       if axis is None:
         return None
+      if keep_other_axes_intact:
+        if axis == old_axis:
+          return new_axis
+        if axis == new_axis:
+          return old_axis
+        return axis
       if old_axis == new_axis:
         return axis
       if axis < min(old_axis, new_axis) or axis > max(old_axis, new_axis):
@@ -865,7 +877,10 @@ class Data(object):
 
     data = self.copy()
     if data.placeholder is not None:
-      data.placeholder = move_axis(data.placeholder, old_axis, new_axis)
+      if keep_other_axes_intact:
+        data.placeholder = swapaxes(data.placeholder, old_axis, new_axis)
+      else:
+        data.placeholder = move_axis(data.placeholder, old_axis, new_axis)
     data.batch_dim_axis = translate_axis(self.batch_dim_axis)
     new_feature_dim_axis = translate_axis(self.feature_dim_axis)
     if new_feature_dim_axis != data.feature_dim_axis:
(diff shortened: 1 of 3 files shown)


returnn/tf/util/data.py
```diff
@@ -824,14 +824,16 @@ def copy_with_time_dim_axis(self, time_dim_axis):
     assert self.time_dim_axis is not None
     return self.copy_move_axis(self.time_dim_axis, time_dim_axis)
 
-  def copy_move_axis(self, old_axis, new_axis):
+  def copy_move_axis(self, old_axis, new_axis, keep_other_axes_intact=False):
     """
     :param int old_axis: counted with batch-dim
     :param int new_axis: counted with batch-dim
+    :param bool keep_other_axes_intact: if False, will shift all axes between old_axis and new_axis.
+      if True, all axes except old_axis and new_axis will remain unchanged
```
albertz, line 832: Maybe it is more clear if we make a separate function `copy_swap_axes` instead?

returnn/tf/util/data.py
```diff
@@ -1088,25 +1099,42 @@ def get_default_new_axis_for_dim_tag(self, dim_tag):
...
       return self.batch_ndim  # add it at the end
 
-  def copy_add_dim_by_tag(self, dim_tag, unbroadcast=False, axis=None):
+  def copy_add_dim_by_tag(self, dim_tag, unbroadcast=False, axis=None, dyn_size=None):
     """
     :param DimensionTag dim_tag:
     :param bool unbroadcast:
     :param int|None axis:
+    :param tf.Tensor|int|None dyn_size:
+      For unbroadcast, if we do not want to rely on dim_tag.dimension.
```
albertz, line 1108: In what cases can that happen? If you know the `dyn_size`, maybe you would (should) just add it to the `dim_tag`?

returnn/tf/util/data.py
```diff
@@ -1088,25 +1099,42 @@ def get_default_new_axis_for_dim_tag(self, dim_tag):
...
     :rtype: Data
     """
     if axis is None:
       axis = self.get_default_new_axis_for_dim_tag(dim_tag=dim_tag)
+    if dyn_size is None:
+      dyn_size = dim_tag.dimension
+
+    def maybe_unbroadcast_placeholder(data):
```
albertz, line 1116: You don't need the `data` argument. You can directly access `res`.
Try to keep the code as simple as possible.

returnn/tf/util/data.py
albertz: max(dim_tag.dyn_size) == dyn_size here, or not?

Zettelkasten: That was what I was wondering about. If that is the case, we can do that.

albertz: I think that should always be the case.

Zettelkasten: Thanks! One issue maybe: What if it is the batch dim? Then, `dim_tag.dyn_size` and `dim_tag.dimension` are both unset.
Or maybe should we could use `tf.shape(dim_tag.src_data.placeholder)[dim_tag.src_axis]` if that exists?
But what if the `placeholder` does not exist?

I really don't know enough about the attributes of `DimensionTag` to tell there is any other way to consistently find the dim of the batch axis, or whether we actually need to add a parameter in this case.

albertz: If there is no good easy way to get the batch dim, we maybe should add such attribute.

In some cases, we use `LayerBase.get_recent_layer().get_batch_dim()`. But this is somewhat ugly. Although maybe a good simple workaround for now, and easy to replace later.

Zettelkasten: Okay, thanks, sure!

One question that remains is how we should handle the (now somewhat obsolete) `data_dyn_shape` parameter of `Data.copy_compatible_to`: Just as with `Data.copy_add_dim_by_tag`, this information can be inferred using the DimensionTags of the input.
All tests also seem to pass with just entirely ignoring the `data_dyn_shape` param.
Maybe we should just deprecate and ignore it then (but keep it for backwards compatibility to the API)?

The alternative would be a somewhat ugly hack to set the `dyn_size` (or `dimension`, if its static) on the dimension tags according to `data_dyn_shape` within `Data.copy_compatible_to`.

returnn/tf/util/data.py
```diff
@@ -824,16 +824,39 @@ def copy_with_time_dim_axis(self, time_dim_axis):
     assert self.time_dim_axis is not None
     return self.copy_move_axis(self.time_dim_axis, time_dim_axis)
 
-  def copy_move_axis(self, old_axis, new_axis, keep_other_axes_intact=False):
+  def _copy_template_translate_axes(self, translate_axis):
+    """
+    :param (int|None)->(int|None) translate_axis: function translating an axis to another, both counted with batch-dim.
```
albertz, line 829: I think just a dict or list would be easier, right?

returnn/tf/util/data.py
```diff
@@ -824,16 +824,39 @@ def copy_with_time_dim_axis(self, time_dim_axis):
     assert self.time_dim_axis is not None
     return self.copy_move_axis(self.time_dim_axis, time_dim_axis)
 
-  def copy_move_axis(self, old_axis, new_axis, keep_other_axes_intact=False):
+  def _copy_template_translate_axes(self, translate_axis):
```
albertz, line 827: This could be `copy_transpose`, and could be public, and also not just on template.

returnn/tf/util/data.py
```diff
@@ -871,27 +888,46 @@ def translate_axis(axis):
...
-        for (i, size) in data.size_placeholder.items()}
-      assert None not in data.size_placeholder
-    new_shape = [None] * data.ndim
-    for i, dim in enumerate(self.shape):
-      new_shape[data.get_batch_axis_excluding_batch(translate_axis(self.get_batch_axis(i)))] = dim
-    data.shape = tuple(new_shape)
+    data = self._copy_template_translate_axes(translate_axis)
+    if self.placeholder is not None:
+      from .basic import move_axis
+      data.placeholder = move_axis(self.placeholder, old_axis, new_axis)
```
albertz, line 894: I think `tf.transpose` might be easier. But actually, if you directly implement `copy_transpose`, you don't need any additional logic here.

returnn/tf/util/data.py
```diff
@@ -871,27 +888,46 @@ def translate_axis(axis):
...
+      if axis == axis1:
+        return axis2
+      if axis == axis2:
+        return axis1
+      return axis
+
+    data = self._copy_template_translate_axes(translate_axis)
+    if self.placeholder is not None:
+      from .basic import swapaxes
+      data.placeholder = swapaxes(self.placeholder, axis1, axis2)
```
albertz, line 930: Same here. I think `tf.transpose` might be easier. But actually, if you directly implement `copy_transpose`, you don't need any additional logic here.

returnn/tf/util/data.py
```diff
@@ -1201,13 +1246,15 @@ def copy_compatible_to(self, data, unbroadcast=False, except_feature=False,
...
-      For unbroadcast, if we do not want to rely on tf.shape(data.placeholder).
+    :param tf.Tensor|list[tf.Tensor|int]|tuple[tf.Tensor|int]|None data_dyn_shape: deprecated, will be ignored.
+      If unbroadcast=True, use the DimensionTags of `data` to specify the shape to unbroadcast to.
+      See :func:`copy_add_axis_by_dim_tag`.
     :param bool check_sparse:
     :param bool check_dtype:
     :returns: Data, might add broadcast dimensions
     :rtype: Data
     """
+    del data_dyn_shape  # ignored
```
albertz, line 1257: We better should better search for any code using this function, and clean that code up, instead of leaving this in.

returnn/tf/util/data.py
albertz: I assume it was not always the case that it could be inferred completely.

We should clean that up and remove the param. I don't think this API is really used externally, and internally we can simply clean that up.

tests/test_TFUtil.py
```diff
@@ -1357,6 +1431,7 @@ def test_scatter_nd():
   session.run(ref_grad)
 
 
+@unittest.skip('v.copy_compatible_to(x) in rel_embed tries to match a [B,2,F|5] tensor to [B,T,F|5]. We do not allow that!')
```
albertz, line 1434: I wonder whether we should instead just remove the test case. We anyway have it in history. I'm not sure it is really helpful to keep. The test file is already too big.
(@JackTemaki or @patrick-wilken , any opinions?)

Also, I still wonder where this comes from originally. I assume I expected that this should work at some point. But I guess it's really much better if we disallow that (and the user can use `reinterpret_data` layer or so to enforce T=2 or whatever).

returnn/tf/util/data.py
Zettelkasten: Yes, okay, I will remove it entirely (was not used much anyway). I will also remove the `out_shape` argument from `Data.get_common_shape`.

tests/test_TFUtil.py
Zettelkasten: I removed the test case for now, we can undo that commit again if @JackTemaki or @patrick-wilken think otherwise.

Zettelkasten:
Btw, the Pycharm inspections seem to fail, but I don't really understand why. The [line it is failing at](https://github.com/rwth-i6/returnn/pull/429/files#diff-da05c7386845d8a9d4509bc7304d5b0306b2c85c74b216e2120bcaf87dd893baR866) is just copied from the current code from master. It's complaining about ["continuation line under-indented for visual indent"](https://github.com/rwth-i6/returnn/pull/429/checks?check_run_id=1849374995). How should the indent be?

returnn/tf/util/data.py
```diff
@@ -824,48 +824,35 @@ def copy_with_time_dim_axis(self, time_dim_axis):
...
-      new_axis += self.batch_ndim
-      assert new_axis >= 0
-    assert 0 <= new_axis < self.batch_ndim
-    if old_axis == new_axis:
+    assert len(perm) == self.batch_ndim
+    assert set(perm) == set(range(self.batch_ndim))
+    if all(perm[axis] == axis for axis in range(self.batch_ndim)):
       return self.copy()
+    inv_perm = [[new_axis for new_axis in range(self.batch_ndim) if perm[new_axis] == old_axis][0]
+      for old_axis in range(self.batch_ndim)]
```
albertz, line 839:
```suggestion
    inv_perm = [
      [new_axis for new_axis in range(self.batch_ndim) if perm[new_axis] == old_axis][0]
      for old_axis in range(self.batch_ndim)]
```
(That might fix also the indent warning.)

However, I think this code is too difficult to read. I would use sth like:
```suggestion
    perm_ = dict(enumerate(perm))
    inv_perm_ = {j: i for (i, j) in perm_.items()}
    inv_perm = [j for (i, j) in sorted(inv_perm_.items())]
```

Zettelkasten: Thanks! `dict(enumerate(perm))` also gave me a PyCharm warning, [apparently a false positive](https://stackoverflow.com/questions/61792979/pycharm-warning-unexpected-type-for-filter-on-dictionary). I ended up using `enumerate(perm)` directly instead of `perm_.items()`

albertz: Yea I already reported that problem upstream. (You also should always report any PyCharm problems you encounter.) But it should be ignored in our CI test case.

albertz (approved):
Looks all good now to me.

albertz:
Maybe someone else should review as well. @JackTemaki or @patrick-wilken ?

patrick-wilken:
I will take a look, but will probably take some time.

tests/test_TFUtil.py
```diff
@@ -293,6 +293,65 @@ def test_Data_copy_compatible_to_time_major():
...
+  d2 = Data(name='p2_output', shape=(5, 1, 1), batch_dim_axis=None, time_dim_axis=None)  # [5,1,F|1]
+
+  # default should not match
+  is_equal_opts = dict(allow_same_feature_dim=True, allow_same_spatial_dim=True, treat_feature_as_spatial=True)
+  assert_equal(d1.find_matching_dims(d2.get_dim_tag(0), is_equal_opts=is_equal_opts), [0,1])
+  assert_equal(d1.find_matching_dims(d2.get_dim_tag(1), is_equal_opts=is_equal_opts), [])
+  assert_equal(d1.find_matching_dims(d2.get_dim_tag(2), is_equal_opts=is_equal_opts), [])
+
+  failed = False
+  try:
```
patrick-wilken: You could use `nose.tools.assert_raises` here.

patrick-wilken: Oh, nevermind, this is only an intermediate commit.

Zettelkasten: Thanks anyway, I didn't know that existed. Good to know definitely!

albertz:
@patrick-wilken did you complete the review? I don't see the approve.

returnn/tf/util/data.py
```diff
@@ -1190,123 +1243,57 @@ def copy_compatible_to(self, data, unbroadcast=False, except_feature=False,
...
-      break
-    if data.feature_dim_axis_or_unspecified is NotSpecified and v.feature_dim_axis_or_unspecified is not NotSpecified:
-      if v._default_feature_dim_axis() == v.feature_dim_axis:
-        v.feature_dim_axis = NotSpecified
+    assert all(mapped_axes[ax] == ax for ax in range(v.batch_ndim))
+
+    # Ensure time_dim_axis and feature_dim_axis is same as in data
+    assert v.batch_dim_axis == data.batch_dim_axis  # there is only at most one batch_dim_axis
+    v.time_dim_axis = data.time_dim_axis
+    if data.feature_dim_axis_or_unspecified is NotSpecified:
```
patrick-wilken, line 1282: Doesn't `feature_dim_axis_or_unspecified` handle both cases? So just `v.feature_dim_axis = data.feature_dim_axis_or_unspecified`?

patrick-wilken:
Looks excellent in terms of cleaning up and refactoring! I can now read `copy_compatible_to` 😄 And couldn't spot errors.

About the removed test case:
It was the only test of `nd_indices()`. It could be "saved" by using
```
if x.batch_dim_axis == 1:
  v = v.copy_move_axis(old_axis=1, new_axis=0)
```
instead of `v = v.copy_compatible_to(x)`. I couldn't figure out what's the intention of the test. What is rel_embed? Relative embedding? Whatever it is, it adds slices of a tensor into specific locations of another tensor. If we are not interested in this specific case we could instead test nd_indices more isolated.

albertz:
> About the removed test case:
> It was the only test of `nd_indices()`. It could be "saved" by using 
> ```
> if x.batch_dim_axis == 1:
>   v = v.copy_move_axis(old_axis=1, new_axis=0)
> ```
> instead of `v = v.copy_compatible_to(x)`.

Why not just `v.copy_as_time_major`?
(I think the intention of this code is that it not just tries to match the batch dim, but in general to make them compatible, and thus `copy_compatible_to`.)

> I couldn't figure out what's the intention of the test. What is rel_embed? Relative embedding? Whatever it is, it adds slices of a tensor into specific locations of another tensor. If we are not interested in this specific case we could instead test nd_indices more isolated.

This test (`test_nd_indices_scatter_nd_time_major`) was probably some use case I had for nd_indices and scatter_nd. I don't remember the details anymore.
But actually `copy_compatible_to` is really not the focus of this test. You are right. I overlooked this. We should just fix the test, as `copy_compatible_to` is totally unrelated here.
@Zettelkasten so can you put the test back in, and replace the `copy_compatible_to`?

returnn/tf/util/data.py
albertz: Yes right, that would be simpler. @Zettelkasten please also change that.

albertz:
Btw, overall, very nice cleanup! :) We need more of such clean rewrites. Some of these functions just have grown historically.

Zettelkasten:
Thanks a lot for your comments, as well as your effort in reviewing this!

> About the removed test case:
> It was the only test of `nd_indices()`. It could be "saved" by using
> 
> ```
> if x.batch_dim_axis == 1:
>   v = v.copy_move_axis(old_axis=1, new_axis=0)
> ```
> 
> instead of `v = v.copy_compatible_to(x)`.

I put the test back in and now use `v = v.copy_move_axis(v.batch_dim_axis, x.batch_dim_axis)`. After this, both the `batch_dim_axis` and the `feature_dim_axis` of `x` and `v` should match.

albertz:
Can you maybe cleanup the commits? Leave most of them as they are (it's good to have multiple commits here), but merge "remove test_nd_indices_scatter_nd_time_major" commit together with the "revert ..." commit, and rephrase that. And also, fixup the "PR review" commits into other corresponding commit (or maybe leave them, but rephrase them).

patrick-wilken (approved):

Zettelkasten:
> Can you maybe cleanup the commits?

Yes, of course! Does this look good now? I can also squash all `Data.copy_compatible_to` commits, but they also add `Data.copy_transpose` etc. - maybe it's nicer to keep it a little separated.

albertz:
Thanks, look good now. I will merge now.

PR #431, opened 2021-01-29
removed empty tokens when there are [XXX] in the sentence 

jiangj-apptek:


albertz:
Why did you open a new PR? Don't do that.

Also, do not push multiple commits. Make this a single commit.

jiangj-apptek:
yes, the same issue. what to do with #431 then?

albertz:
> yes, the same issue. what to do with #431 then?

I don't quite understand. We only need one PR, not two. (So now that you incorrectly created two, close one of them.)
In the future, do not create multiple PRs. Just update the PR instead.

jiangj-apptek:
Anything to do at this time?

albertz:
> Anything to do at this time?

Yes, as I explained:

* Close all but the original PR
* Update the original PR
* Cleanup the commits (make it a single commit)

PR #432, opened 2021-02-01
fix hdf_dump without target, MetaDataset workaround

JackTemaki:
I thought fixing `hdf_dump` to dump some of my datasets was simpler, but as there were now some additional problems to be solved a PR seems to be the safer option.

So `hdf_dump` did not allow for Datasets having no targets (but only data), and I did not work with MetaDatasets at all, as the child datasets did not get a `init_seq_order` call passed.

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -1364,12 +1364,15 @@ class HDFDatasetWriter:
     hdf_dataset.create_dataset(attr_seqLengths, shape=(num_seqs, 2), dtype="int32")
     for i, seq_len in enumerate(seq_lens):
       data_len = seq_len[default_data_input_key]
-      targets_len = seq_len[default_data_target_key]
-      for data_key in data_target_keys:
-        assert seq_len[data_key] == targets_len, "different lengths in multi-target not supported"
-      if targets_len is None:
-        targets_len = data_len
-      hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
+      if len(data_target_keys) > 0:
+        targets_len = seq_len[default_data_target_key]
+        for data_key in data_target_keys:
+          assert seq_len[data_key] == targets_len, "different lengths in multi-target not supported"
+        if targets_len is None:
+          targets_len = data_len
+        hdf_dataset[attr_seqLengths][i] = [data_len, targets_len]
+      else:
+        hdf_dataset[attr_seqLengths][i] = [data_len]
       if use_progress_bar:
         progress_bar_with_time(float(i) / num_seqs)
 
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -333,7 +333,7 @@ class MetaDataset(CachedDataset2):
     :param list[int]|None seq_order: List of corpus sequence indices, to set a predefined order.
     :rtype: bool
     """
-    need_reinit = self.epoch is None or self.epoch != epoch or seq_list
+    need_reinit = self.epoch is None or self.epoch != epoch or seq_list or self.expected_load_seq_start > 0
     super(MetaDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
 
     if not need_reinit:


returnn/datasets/hdf.py
```diff
@@ -1403,6 +1405,7 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
 
     # Again iterate through dataset, and set the data
     print("Write data...", file=log.v3)
+    dataset.epoch = None  # force re-init (inconsistency in MetaDataset)
```
patrick-wilken, line 1408: Can you elaborate on what this does?

albertz, line 1408: Why a workaround for another bug?
Why not fix that other bug ("inconsistency in MetaDataset")?

JackTemaki:
To answer you both, I will describe the problem. I did not see a feasible solution to fix this, so I just added this one line.

The problem is that the MetaDataset checks if it needs a re-init:

``` python
  def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
    """
    :param int|None epoch:
    :param list[str]|None seq_list: List of sequence tags, to set a predefined order.
    :param list[int]|None seq_order: List of corpus sequence indices, to set a predefined order.
    :rtype: bool
    """
    need_reinit = self.epoch is None or self.epoch != epoch or seq_list
    super(MetaDataset, self).init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)

    if not need_reinit:
      self._num_seqs = len(self.seq_list_ordered[self.default_dataset_key])
      return False

    seq_order_dataset = None
    if seq_order:
      seq_index = seq_order
    elif seq_list:
      seq_index = [self.tag_idx[tag] for tag in seq_list]
    elif self.seq_order_control_dataset:
      seq_order_dataset = self.datasets[self.seq_order_control_dataset]
      assert isinstance(seq_order_dataset, Dataset)
      seq_order_dataset.init_seq_order(epoch=epoch)
      seq_index = seq_order_dataset.get_current_seq_order()
```

As the dataset is already initialized here:
``` python
    print("Write data...", file=log.v3)
    dataset.epoch = None  # force re-init (inconsistency in MetaDataset)
    dataset.init_seq_order(epoch)
```
The `dataset.epoch = None` ensures that all childs of `MetaDataset` get a `init_seq_order` call.

If this does not happen, some datasets such as all that depend on `CachedDataset2` will not work:
``` python
  File "/u/rossenbach/src/returnn/returnn/datasets/hdf.py", line 1413, in dump_from_dataset
    line: dataset.load_seqs(seq_idx, seq_idx + 1)
    locals:
      dataset = <local> <MetaDataset 'train' epoch=1>
      dataset.load_seqs = <local> <bound method Dataset.load_seqs of <MetaDataset 'train' epoch=1>>
      seq_idx = <local> 0
  File "/u/rossenbach/src/returnn/returnn/datasets/basic.py", line 329, in load_seqs
    line: self._load_seqs(start, end)
    locals:
      self = <local> <MetaDataset 'train' epoch=1>
      self._load_seqs = <local> <bound method MetaDataset._load_seqs of <MetaDataset 'train' epoch=1>>
      start = <local> 0
      end = <local> 1
  File "/u/rossenbach/src/returnn/returnn/datasets/meta.py", line 409, in _load_seqs
    line: self.datasets[dataset_key].load_seqs(start_, end)
    locals:
      self = <local> <MetaDataset 'train' epoch=1>
      self.datasets = <local> {'audio': <OggZipDataset 'train_audio' epoch=1>}
      dataset_key = <local> 'audio'
      load_seqs = <not found>
      start_ = <local> 0
      end = <local> 1
  File "/u/rossenbach/src/returnn/returnn/datasets/basic.py", line 329, in load_seqs
    line: self._load_seqs(start, end)
    locals:
      self = <local> <OggZipDataset 'train_audio' epoch=1>
      self._load_seqs = <local> <bound method CachedDataset2._load_seqs of <OggZipDataset 'train_audio' epoch=1>>
      start = <local> 0
      end = <local> 1
  File "/u/rossenbach/src/returnn/returnn/datasets/cached2.py", line 112, in _load_seqs
    line: assert start >= self.expected_load_seq_start
    locals:
      start = <local> 0
      self = <local> <OggZipDataset 'train_audio' epoch=1>
      self.expected_load_seq_start = <local> 584
AssertionError
```
Because `self.expected_load_seq_start` is set to 0 only in the `init_seq_order` call to `CachedDataset2`, which is not passed from the `MetaDataset` to its child datasets.

So I thought of:
 - remove this `need_reinit` shortcut completely
 - add a flag to `init_seq_order` to not reinit (this needs to be changed in all datasets then)
 - try to change the behaviour of `CachedDataset2` (I did not follow this further)
 - set `epoch` to `None` to avoid the shortcut
 
So the last one seemed to be the safest...

patrick-wilken:
Ah, I see. One clean way would be to always call `init_seq_order` for the subdatasets in `MetaDataset`, even if `need_reinit==False`, assuming that all (sub-) dataset classes where sequence ordering is expensive themselves have this `need_reinit` logic implemented.
Or we have to separate initializing the sequence order, which is potentially costly, from "resetting counters". E.g. move what is now in `Dataset.init_seq_order` and `CachedDataset2.init_seq_order` to `init_epoch` methods and call this `init_epoch` within the `init_seq_order` functions. `MetaDataset` could then call the `init_epoch` methods from the subdatasets directly in case `init_seq_order` is not needed.

But by just setting `epoch=None` you make hdf_dump inefficient for all datasets, because sequence ordering is computed twice.

patrick-wilken:
Or maybe we could allow going through an epoch multiple times without reinitializing the sequence order. by setting `expected_load_seq_start = 0` when reaching the end in `is_less_than_num_seqs`. But this seems error-prone and inelegant if we only need it for hdf_dump.

albertz:
> I did not see a feasible solution to fix this, so I just added this one line.

Never do that. Never add a workaround because the workaround requires just one-line, while the real fix takes maybe more effort.

> The problem is that the MetaDataset checks if it needs a re-init ...

So it sounds like this check is buggy? Why not simply fix this check? This would solve the problem, or not?

JackTemaki:
> Or we have to separate initializing the sequence order, which is potentially costly, from "resetting counters"
 
 In the long run this would be the nicest solution, but I think changing this will affect nearly all code parts and might not be easy...

> One clean way would be to always call init_seq_order for the subdatasets in MetaDataset, even if need_reinit==False, assuming that all (sub-) dataset classes where sequence ordering is expensive themselves have this need_reinit logic implemented.

I guess this is the way to go then, if there are no other unwanted effects...

JackTemaki:
> > I did not see a feasible solution to fix this, so I just added this one line.
> 
> Never do that. Never add a workaround because the workaround requires just one-line, while the real fix takes maybe more effort.

I am fully aware of that, this is why we are discussing this now. I listed my thoughts of alternatives above, and this "one line" was at first glance the best solution with the least damage to other code/functionality
 
> > The problem is that the MetaDataset checks if it needs a re-init ...
> 
> So it sounds like this check is buggy? Why not simply fix this check? This would solve the problem, or not?

This would be more hacky, as we then would need to check which kind of datasets are childs of the MetaDataset, writing more code only for one specific edge-case. Patricks suggestion to just remove the check w.r.t. to calling init_seq_order on child datasets might be the best solution that does not require changing the way the dataset functions are working right now. But I guess this might slow down dataset initialization in some other scenarios (if they exist).

albertz:
> > > The problem is that the MetaDataset checks if it needs a re-init ...
> > 
> > So it sounds like this check is buggy? Why not simply fix this check? This would solve the problem, or not?
> 
> This would be more hacky, as we then would need to check which kind of datasets are childs of the MetaDataset,

Why? Of course not. Never do that.

> writing more code only for one specific edge-case. Patricks suggestion to just remove the check w.r.t. to calling init_seq_order on child datasets might be the best solution ...

I don't understand why this is needed. Why do you want to make it slower/suboptimal?

Why not just fix the check instead?

Clearly there is a bug in the check, unless I'm misunderstanding sth? The check checks whether init_seq_order is needed for the childs. You say it currently incorrectly does not call init_seq_order on the childs. So clearly the check is wrong? Just fix the check (totally independent of what the child datasets are).

patrick-wilken:
We could make it `need_reinit = self.epoch is None or self.epoch != epoch or self.expected_load_seq_start > 0`.
Which means that as soon as we read the meta-dataset we will always reinit, admitting that we don't have the logic to reset the sub-datasets without recomputing the sequence order yet.

JackTemaki:
> We could make it need_reinit = self.epoch is None or self.epoch != epoch or self.expected_load_seq_start > 0 in MetaDataset.
Which means that as soon as we read the meta-dataset we will always reinit, admitting that we don't have the logic to reset the sub-datasets without recomputing the sequence order yet.

Hmm yes, we could do this, I did not consider that the MetaDataset has the same counter. This is a cleaner workaround than my initial solution, but of course could result in some cases where the sequence ordering is unnecessarily recomputed. But as I am not sure these cases even exists (I did not find them in the TFEngine tests at least), I guess it is okay to fix it with this.

albertz (approved):

PR #437, opened 2021-02-10
MergeDimsLayer, unset time-dim if it got merged

albertz:
This was the expected behavior from most users,
after a discussion whether [B,T,F|5] merge "except_batch"
should result in [B,T|F] or [B,F] or [B,T].

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2625,10 +2625,10 @@ class MergeDimsLayer(_ConcatInputLayer):
     data.shape = tuple(new_shape)
     data.time_dim_axis = cls._old_axis_to_new_axis(
       input_data=input_data, merge_axes=axes, old_axis=input_data.time_dim_axis)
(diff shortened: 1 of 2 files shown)


JackTemaki (approved):
I am quite sure that this will not break any of my existing setups, so I would give this a go...

patrick-wilken:
Ok.
So there are no other cases where two of the batch/time/feature dim attributes can have the same value, right? Probably good that way.

patrick-wilken (approved):

albertz:
> So there are no other cases where two of the batch/time/feature dim attributes can have the same value, right? Probably good that way.

There could be (I don't know). These dims/axes are totally independent. Setting one of them usually does not affect the other. Also, you could have `{class: reinterpret_data, set_axes: {t: "enc-time"}}` or so, and that would not touch the feature-dim. I.e. if the input to that would be [B,F|"enc-time"], afterwards it would be [B,T|F|"enc-time"].

Anyway, I will merge this then.

PR #438, opened 2021-02-10
HDFDataset / HDFDatasetWriter: Support for only using targets

patrick-wilken:
Necessary changes to support HDF files without 'inputs'.

Also, there was a check that targets have to be of equal sequence length in the HDFDatasetWriter. I tested targets with different lengths and I didn't find a reason why this check was there.

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -1297,14 +1297,17 @@ class HDFDatasetWriter:
       data_keys.remove("raw")
     data_target_keys = [key for key in dataset.get_target_list() if key in data_keys]
     data_input_keys = [key for key in data_keys if key not in data_target_keys]
-    assert len(data_input_keys) > 0
-    if len(data_input_keys) > 1:
-      if "data" in data_input_keys:
-        default_data_input_key = "data"
+    default_data_input_key = None
+    progress_bar_data_key = None
+    if data_input_keys:
+      if len(data_input_keys) > 1:
+        if "data" in data_input_keys:
+          default_data_input_key = "data"
+        else:
+          raise Exception("not sure which input data key to use from %r" % (data_input_keys,))
       else:
-        raise Exception("not sure which input data key to use from %r" % (data_input_keys,))
-    else:
-      default_data_input_key = data_input_keys[0]
+        default_data_input_key = data_input_keys[0]
+      progress_bar_data_key = default_data_input_key
     print("Using input data key:", default_data_input_key)
     default_data_target_key = None
     if len(data_target_keys) > 1:
@@ -1315,9 +1318,14 @@ class HDFDatasetWriter:
           data_target_keys, attr_numLabels), file=log.v2)
     elif len(data_target_keys) == 1:
       default_data_target_key = data_target_keys[0]
+    progress_bar_data_key = progress_bar_data_key or default_data_target_key or data_target_keys[0]
     print("Using target data key:", default_data_target_key)
 
-    hdf_data_key_map = {key: key for key in data_keys if key != default_data_input_key}
+    # All but one of the inputs have to be treated as targets because our HDF format only supports one input.
+    data_target_keys += [key for key in data_input_keys if key != default_data_input_key]
+    data_input_key = default_data_input_key
+
+    hdf_data_key_map = {key: key for key in data_keys if key != data_input_key}
     if "data" in hdf_data_key_map:
       hdf_data_key_map["data"] = "classes"  # Replace "data" which is reserved for input key in HDFDataset.
       assert "classes" not in hdf_data_key_map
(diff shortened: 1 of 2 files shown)


returnn/datasets/hdf.py
```diff
@@ -270,14 +278,16 @@ def get_data(self, seq_idx, key):
     end_pos = self.file_seq_start[file_idx][real_file_seq_idx + 1]
 
     if key == "data":
+      assert 'inputs' in fin, "'data' key is reserved for 'inputs' in the HDF file, but 'inputs' not found in HDF file."
```
albertz, line 281: Maybe we might want to allow `"data"` as a possible entry in `targets/`?

returnn/datasets/hdf.py
```diff
@@ -1287,27 +1297,35 @@ def dump_from_dataset(self, dataset, epoch=1, start_seq=0, end_seq=float("inf"),
...
+        else:
+          raise Exception("not sure which input data key to use from %r" % (data_input_keys,))
       else:
-        raise Exception("not sure which input data key to use from %r" % (data_input_keys,))
-    else:
-      default_data_input_key = data_input_keys[0]
+        default_data_input_key = data_input_keys[0]
+      progress_bar_data_key = default_data_input_key
     print("Using input data key:", default_data_input_key)
+    default_data_target_key = None
```
albertz, line 1312: Do we actually still need `default_data_target_key`?

albertz:
Note that a HDF created with this new code might not be compatible with an old RETURNN (e.g. when "inputs" is missing). This is ok?

I think it's not necessary. You might also put just some dummy entry into "inputs", with 0 dim or so. Maybe the code is actually simpler then, as you don't need this more complex logic for `first_target_idx`, `num_input_keys`, etc.

JackTemaki (commented):
The HDF using no targets will not work with the MetaDataset right now, as this will do calls like `dataset.get_seq_length(idx)["data"]` on the HDFDataset, which would now return an error.

I am working on the PR for the MetaDataset right now, then this should work. I think we should add test-cases for this, many dataset-related things have no coverage at all right now.

EDIT: Of course this error only happens in the case that the HDFDataset is used as "default dataset" inside the MetaDataset. Still, this should be supported.

albertz:
@JackTemaki your comment is pretty independent from this PR here, right?

JackTemaki:
> @JackTemaki your comment is pretty independent from this PR here, right? Then better let's discuss this elsewhere, and not mix up things, to keep the discussion focused here about this specific PR.

This was somewhat related to the question if we should allow the "data" key in the targets, and this is also something Patrick needs to consider when testing his code. I did not want to start a discussion about this issue here.

albertz:
> > @JackTemaki your comment is pretty independent from this PR here, right?...
> 
> This was somewhat related to the question if we should allow the "data" key in the targets, ...

Why?

returnn/datasets/hdf.py
patrick-wilken: Only to create this `attr_numLabels = 'numLabels'` field in the HDF. I don't know if anyone uses it anymore though, it has been replaced by `'targets/size'`, see line 163.

returnn/datasets/hdf.py
patrick-wilken: Yes, would probably be good. So just
```
if key == "data" and 'targets/data/data' not in fin:
```
here?

To actually create an HDF with data as target we would need to change the HDFDatasetWriter.

returnn/datasets/hdf.py
albertz: Yea, since a long time. I think it would be safe to remove.

returnn/datasets/hdf.py
albertz: Yea. I'm not sure we need that logic now, but we should keep that in mind.

patrick-wilken:
>The HDF using no targets will not work with the MetaDataset right now.

True, I fixed the same thing for CombinedDataset by having `default_data_key` as a parameter of the dataset. Maybe we can think of something more general.

>Note that a HDF created with this new code might not be compatible with an old RETURNN (e.g. when "inputs" is missing). This is ok?

For me, yes 😄 

>I think it's not necessary. You might also put just some dummy entry into "inputs", with 0 dim or so. Maybe the code is actually simpler then, as you don't need this more complex logic for first_target_idx, num_input_keys, etc.

Hmm, you would also need the sequence lengths for the "data" key. So store an additional 0 for each sequence. I would rather keep the HDF file clean than avoiding this logic in the code.

returnn/datasets/hdf.py
patrick-wilken: What about `inputPattSize`? Shall I remove both in one go?

JackTemaki:
> > The HDF using no targets will not work with the MetaDataset right now.
> 
> True, I fixed the same thing for CombinedDataset by having `default_data_key` as a parameter of the dataset. Maybe we can think of something more general.

Okay, then this is identical to how I planned to solve this (giving this parameter to the "Dataset")

returnn/datasets/hdf.py
patrick-wilken: Ah no, for sparse "data" it seems to be used.

albertz (approved):

albertz:
What's the state here? @JackTemaki ? Ready to merge?

JackTemaki (approved):

JackTemaki:
> What's the state here? @JackTemaki ? Ready to merge?

I thought there would be still some open discussion, but for me it is fine.

PR #439, opened 2021-02-10
SplitLayer template construction fails if used after SplitDimsLayer

Zettelkasten:
See the test case I added, [which currently fails](https://github.com/rwth-i6/returnn/runs/1875491241?check_suite_focus=true).
I get the error:
```
  File "/u/petrick/software/returnn/official/returnn/tf/network.py", line 708, in _create_layer
    line: assert output_template_special_axes == output_special_axes, "%s %r: not equal: %r == %r, from data %r -> %r" % (
            layer_class.__name__, name,
            output_template_special_axes, output_special_axes,
            output_template, layer.output)
    locals:
      output_template_special_axes = <local> {'time_dim_axis': 1, 'feature_dim_axis': 2}
      output_special_axes = <local> {'time_dim_axis': 1, 'feature_dim_axis': 3}
      layer_class = <local> <class 'returnn.tf.layers.basic.SplitLayer'>
      layer_class.__name__ = <local> 'SplitLayer', len = 10
      name = <local> 'split_qkv', len = 9
      output_template = <local> Data(name='split_qkv_output', shape=(None, 2, None), batch_shape_meta=[B,T|?,F|2,?])
      layer = <local> <SplitLayer 'split_qkv' out_type=Data(shape=(None, 2, 20), batch_shape_meta=[B,T|'time:var:extern_data:data',2,F|20])>
      layer.output = <local> Data(name='split_heads_output', shape=(None, 2, 20), batch_shape_meta=[B,T|'time:var:extern_data:data',2,F|20])
AssertionError: SplitLayer 'split_qkv': not equal: {'time_dim_axis': 1, 'feature_dim_axis': 2} == {'time_dim_axis': 1, 'feature_dim_axis': 3}, from data Data(name='split_qkv_output', shape=(None, 2, None), batch_shape_meta=[B,T|?,F|2,?]) -> Data(name='split_heads_output', shape=(None, 2, 20), batch_shape_meta=[B,T|'time:var:extern_data:data',2,F|20])
```

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2707,6 +2707,14 @@ class SplitLayer(_ConcatInputLayer):
     return self._sub_layers.get(layer_name, None)
 
   @classmethod
+  def get_out_data_from_opts(cls, sources, **kwargs):
+    """
+    :param list[LayerBase] sources:
+    :rtype: Data
+    """
+    return get_concat_sources_data_template(sources)
+
+  @classmethod
   def get_sub_layer_out_data_from_opts(cls, layer_name, parent_layer_kwargs):
     """
     :param str layer_name: name of the sub_layer (right part of '/' separated path)
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -1107,6 +1107,22 @@ def test_SoftmaxOverSpatialLayer_window():
     numpy.testing.assert_allclose(out_np[~mask], 0, rtol=1e-5)  # check if masking worked
 
 
+def test_SplitLayer_after_SplitDimsLayer():
+  n_batch, n_time, n_in = 7, 3, 40
+  config = Config({
+    "extern_data": {"data": {"dim": n_in}},
+    "debug_print_layer_output_template": True,
+  })
+  with make_scope():
+    net = TFNetwork(config=config)
+    net.construct_from_dict({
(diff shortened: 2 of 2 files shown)


Zettelkasten:
I pushed a commit where I added `get_out_data_from_opts` to `SplitLayer`, and now it seems to work again.

Is this expected to be necessary? Or is this actually a bug in `_base_get_out_data_from_opts`, which I overrode?

albertz:
> I pushed a commit where I added `get_out_data_from_opts` to `SplitLayer`, and now it seems to work again.
> 
> Is this expected to be necessary? Or is this actually a bug in `_base_get_out_data_from_opts`, which I overrode?

This has grown historically. Originally, no layer had their own custom `get_out_data_from_opts` and all used this base function. This base function mostly has some heuristics to derive the right output format, where it uses the input format, and maybe replaces the dimension (if layers have `n_out`) or other things (if you provide some custom `out_type`). This was fine for basic layers like `LinearLayer`. But over time it became more and more generic, and the heuristics were not suitable in many cases. Actually I'm not sure if we really need them anymore, as most layers define their own.

But currently, the idea is, if your layer doesn't do special things on the format, but just keeps it as it is (maybe only replacing the dim), you don't need to override this. It's ok if you do (so this PR is fine), but there seems to be a bug in `_base_get_out_data_from_opts`. But no need to invest much time into that, unless the bug is obvious.

albertz:
I already merged anyway.

PR #440, opened 2021-02-11
MetaDataset rework

JackTemaki:
This should enable a free choice of dataset-key-mappings for the meta-dataset, and also enable future support for child-datasets that do not provide any "data" key, but instead define a "default_data_key" which is used in all cases where no explicit choices are made.

I also removed setting `num_inputs` in the MetaDataset. I am not sure if this has any additional consequences, so I put this into a separate commit for testing both changes separately (and keeping my previous edit of the line). If there is no problem, I will merge both commits.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -114,6 +114,7 @@ class Dataset(object):
     self.rnd_seq_drop = None  # type: typing.Optional[Random]
     self.num_inputs = 0  # usually not used, but num_outputs instead, which is more generic
     self.num_outputs = None  # type: typing.Optional[typing.Dict[str,typing.Tuple[int,int]]]  # tuple is num-classes, len(shape).  # nopep8
+    self.default_data_key = "data"  # type: str  # Allows Datasets to re-define their input/default data key
     self.window = window
     self.seq_ordering = seq_ordering  # "default", "sorted" or "random". See self.get_seq_order_for_epoch().
     if random_seed_offset is None:
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -180,6 +180,7 @@ class MetaDataset(CachedDataset2):
                seq_list_file=None,
                seq_order_control_dataset=None,
                seq_lens_file=None,
+               default_data_key="data",
                data_dims=None,
                data_dtypes=None,  # noqa  # not used
                window=1, **kwargs):
@@ -197,6 +198,8 @@ class MetaDataset(CachedDataset2):
     :param str|None seq_order_control_dataset: if set, this dataset will define the order for each epoch.
     :param str|None seq_lens_file: filename. json. dict[str,dict[str,int]], seq-tag -> data-key -> len.
       Use if getting sequence length from loading data is too costly.
+    :param str|None default_data_key: if seq_order_control_dataset is not set, this will determine which dataset
+      is the default dataset, and which data stream is used for the sequence length computation
     :param dict[str,(int,int)] data_dims: self-data-key -> data-dimension, len(shape) (1 ==> sparse repr).
        Deprecated/Only to double check. Read from data if not specified.
     :param dict[str,str] data_dtypes: self-data-key -> dtype. Read from data if not specified.
@@ -208,9 +211,11 @@ class MetaDataset(CachedDataset2):
     self.data_map = data_map
     self.dataset_keys = set([m[0] for m in self.data_map.values()])  # type: typing.Set[str]
     self.data_keys = set(self.data_map.keys())  # type: typing.Set[str]
-    assert "data" in self.data_keys
-    self.target_list = sorted(self.data_keys - {"data"})
-    self.default_dataset_key = seq_order_control_dataset or self.data_map["data"][0]
+    self.target_list = sorted(self.data_keys)
+    self.default_dataset_key = seq_order_control_dataset or self.data_map[default_data_key][0]
+    assert default_data_key in self.data_map.keys(), (
+      "default data key %s was not found in data map: %s" % (default_data_key, self.data_map))
+    self.default_data_key = default_data_key
     self.seq_order_control_dataset = seq_order_control_dataset
 
     # This will only initialize datasets needed for features occuring in data_map
@@ -237,7 +242,6 @@ class MetaDataset(CachedDataset2):
     if data_dims:
       data_dims = convert_data_dims(data_dims)
       self.data_dims = data_dims
-      assert "data" in data_dims
       for key in self.target_list:
         assert key in data_dims
     else:
@@ -251,7 +255,7 @@ class MetaDataset(CachedDataset2):
       if dataset_data_key in dataset.labels:
         self.labels[data_key] = dataset.labels[dataset_data_key]
 
-    self.num_inputs = self.data_dims["data"][0]
+    self.num_inputs = self.data_dims[default_data_key][0] if default_data_key in self.data_dims.keys() else 0
     self.num_outputs = self.data_dims
 
     self.orig_seq_order_is_initialized = False
@@ -317,14 +321,13 @@ class MetaDataset(CachedDataset2):
     return seq_list
 
   def _get_dataset_seq_length(self, seq_idx):
+    default_dataset = self.datasets[self.default_dataset_key]
     if not self.orig_seq_order_is_initialized:
       # To use get_seq_length() we first have to init the sequence order once in original order.
       # If sequence lengths are not needed by get_seq_order_for_epoch this is never executed.
-      self.datasets[self.default_dataset_key].init_seq_order(
-        epoch=self.epoch, seq_list=self.seq_list_original[self.default_dataset_key])
+      default_dataset.init_seq_order(epoch=self.epoch, seq_list=self.seq_list_original[self.default_dataset_key])
       self.orig_seq_order_is_initialized = True
-
-    return self.datasets[self.default_dataset_key].get_seq_length(seq_idx)["data"]
+    return default_dataset.get_seq_length(seq_idx)[default_dataset.default_data_key]
 
   def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
     """
@@ -357,7 +360,8 @@ class MetaDataset(CachedDataset2):
           :param int s:
           :rtype: int
           """
-          return self._seq_lens[self.seq_list_original[self.default_dataset_key][s]]["data"]
+          default_dataset_default_data_key = self.datasets[self.default_dataset_key].default_data_key
+          return self._seq_lens[self.seq_list_original[self.default_dataset_key][s]][default_dataset_default_data_key]
       elif self._seq_order_seq_lens_file:
         get_seq_len = self._get_seq_order_seq_lens_by_idx
       else:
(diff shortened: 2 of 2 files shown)


returnn/datasets/basic.py
```diff
@@ -114,6 +114,7 @@ def __init__(self, name=None,
     self.rnd_seq_drop = None  # type: typing.Optional[Random]
     self.num_inputs = 0  # usually not used, but num_outputs instead, which is more generic
     self.num_outputs = None  # type: typing.Optional[typing.Dict[str,typing.Tuple[int,int]]]  # tuple is num-classes, len(shape).  # nopep8
+    self.default_data_key = "data"  # type: str  # Allows Datasets to re-define their input/default data key
```
albertz, line 117: Why does the base `Dataset` class need this?
(And don't just explain here in the comment. This should be explained in the code. What is this used for?)

returnn/datasets/meta.py
```diff
@@ -180,6 +180,7 @@ def __init__(self,
                seq_list_file=None,
                seq_order_control_dataset=None,
                seq_lens_file=None,
+               default_data_key="data",
```
albertz, line 183: Why is the option here, but the attrib is in the base class? The option should be right where you have the attrib.

returnn/datasets/meta.py
```diff
@@ -208,9 +211,11 @@ def __init__(self,
     self.data_map = data_map
     self.dataset_keys = set([m[0] for m in self.data_map.values()])  # type: typing.Set[str]
     self.data_keys = set(self.data_map.keys())  # type: typing.Set[str]
-    assert "data" in self.data_keys
-    self.target_list = sorted(self.data_keys - {"data"})
-    self.default_dataset_key = seq_order_control_dataset or self.data_map["data"][0]
+    self.target_list = sorted(self.data_keys)
```
albertz, line 214: Now this includes any potential input (e.g. "data") as well. Which implies  `available_for_inference=False`. But this would be wrong.

Why do you change the behavior?

If you want to make this more generic, then make it more generic. But don't just change the one specific behavior to some other specific behavior.

returnn/datasets/meta.py
```diff
@@ -197,6 +198,8 @@ def __init__(self,
     :param str|None seq_order_control_dataset: if set, this dataset will define the order for each epoch.
     :param str|None seq_lens_file: filename. json. dict[str,dict[str,int]], seq-tag -> data-key -> len.
       Use if getting sequence length from loading data is too costly.
+    :param str|None default_data_key: if seq_order_control_dataset is not set, this will determine which dataset
```
albertz, line 201: Do you really allow `None`?

returnn/datasets/meta.py
```diff
@@ -208,9 +211,11 @@ def __init__(self,
     self.data_map = data_map
     self.dataset_keys = set([m[0] for m in self.data_map.values()])  # type: typing.Set[str]
     self.data_keys = set(self.data_map.keys())  # type: typing.Set[str]
-    assert "data" in self.data_keys
-    self.target_list = sorted(self.data_keys - {"data"})
-    self.default_dataset_key = seq_order_control_dataset or self.data_map["data"][0]
+    self.target_list = sorted(self.data_keys)
+    self.default_dataset_key = seq_order_control_dataset or self.data_map[default_data_key][0]
+    assert default_data_key in self.data_map.keys(), (
```
albertz, line 216: If `default_data_key is None`, this would be wrong.

returnn/datasets/meta.py
```diff
@@ -434,9 +437,8 @@ def _collect_single_seq(self, seq_idx):
     :rtype: DatasetSeq
     """
     seq_tag = self.seq_list_ordered[self.default_dataset_key][seq_idx]
-    features = self._get_data(seq_idx, "data")
-    targets = {target: self._get_data(seq_idx, target) for target in self.target_list}
-    return DatasetSeq(seq_idx=seq_idx, seq_tag=seq_tag, features=features, targets=targets)
+    features = {target: self._get_data(seq_idx, target) for target in self.target_list}
```
albertz, line 440: This is very confusing. Why `self.target_list`? This is not supposed just the targets, but all features (I assume). So then use `self.data_keys`.

returnn/datasets/basic.py
albertz: From later usage in the code, it seems like this is used as the default seq length data key, e.g. for sorting. Then better call it `seq_length_data_key` or so. Why anway have "default" in the name?

albertz: Also, why is this not configurable? (I.e. not a param in `__init__` here in the base class?)

returnn/datasets/meta.py
```diff
@@ -317,14 +320,13 @@ def _load_seq_list(self, seq_list_file=None):
...
     if not self.orig_seq_order_is_initialized:
       # To use get_seq_length() we first have to init the sequence order once in original order.
       # If sequence lengths are not needed by get_seq_order_for_epoch this is never executed.
-      self.datasets[self.default_dataset_key].init_seq_order(
-        epoch=self.epoch, seq_list=self.seq_list_original[self.default_dataset_key])
+      default_dataset.init_seq_order(epoch=self.epoch, seq_list=self.seq_list_original[self.default_dataset_key])
       self.orig_seq_order_is_initialized = True
-
-    return self.datasets[self.default_dataset_key].get_seq_length(seq_idx)["data"]
+    return default_dataset.get_seq_length(seq_idx)[default_dataset.default_data_key]
```
albertz, line 329: If you use that here, I think all the usages of `get_seq_order_for_epoch` should be adopted as well to use it, to keep this consistent.

returnn/datasets/basic.py
JackTemaki: I call it default, because the sequence sorting is not the only potential usage. I did not make it configurable, because then I would need to make sure that all datasets treat this correctly when changed by the user. As "data" as default data key is used in more than 100 locations across all datasets, this would be a lot of work. If you want, I can extend the docstring (although for me the explanation is quite obvious), but I would not make this configurable for all datasets, but only for datasets which actually treat this correctly (in this case, the MetaDataset).

albertz:
> I also removed setting `num_inputs` in the MetaDataset. I am not sure if this has any additional consequences, so I put this into a separate commit for testing both changes separately (and keeping my previous edit of the line). If there is no problem, I will merge both commits.

We don't have too much test cases for all the ways you can use a dataset (which is bad; we should add some more).
So that means, you should check the code. See all usages of `num_inputs`, to just know what consequences this has.
If `num_inputs` is not used anywhere, you should clean up other datasets as well.

returnn/datasets/meta.py
JackTemaki: No, this is a leftover from an earlier approach, this will be removed.

returnn/datasets/meta.py
JackTemaki: This is exactly the reason why I made "default_data_key" configurable for the MetaDataset only, the possible amount of usages (which will not only be for the sequence order) is huge.

If you say this should be changed consistently all at once, I can do that, but this may take several days.

JackTemaki:
> > I also removed setting `num_inputs` in the MetaDataset. I am not sure if this has any additional consequences, so I put this into a separate commit for testing both changes separately (and keeping my previous edit of the line). If there is no problem, I will merge both commits.
> 
> We don't have too much test cases for all the ways you can use a dataset (which is bad; we should add some more).
> So that means, you should check the code. See all usages of `num_inputs`, to just know what consequences this has.
> If `num_inputs` is not used anywhere, you should clean up other datasets as well.

I checked most (if not all) locations inside the Dataset class, and it seems that it is only used for windowing, which is disallowed for the MetaDataset, but not for other Datasets. This is why it is safe here, but not for other datasets.

returnn/datasets/basic.py
albertz: > I call it default, because the sequence sorting is not the only potential usage.

I don't understand. What are potential usages? We don't really need this I think. And we should be more explicit. If there ever comes up another usage, I think it would be more clear to just have a separate flag for it. But I think it would really be bad if there would be other use cases where one data stream is somewhat special.

> I did not make it configurable, because then I would need to make sure that all datasets treat this correctly when changed by the user.

I don't really understand. If this is not configurable, it cannot really be used at all. Now it looks like you can just configure it for the MetaDataset. But actually it would be relevant to configure for the childs which you use in MetaDataset, e.g. HDFDataset, ExternSprintDataset, whatever.

> As "data" as default data key is used in more than 100 locations across all datasets, this would be a lot of work.

I doubt that there are so much hardcoded usages of "data". And it shouldn't really be much work. Basically (almost) just a global replace `"data"` by `seq_length_sort_data_key`.

> If you want, I can extend the docstring (although for me the explanation is quite obvious),

For me it is not obvious at all. If you call it `seq_length_sort_data_key`, and then shortly explain it is used for sorting the sequences, then it would be clear.

> but I would not make this configurable for all datasets, but only for datasets which actually treat this correctly (in this case, the MetaDataset).

But this doesn't really make sense, as explained.

returnn/datasets/meta.py
albertz: `default_dataset` here is probably never a `MetaDataset`. I don't see how the whole approach is useful then.

returnn/datasets/meta.py
JackTemaki: Yes, that is true

albertz:
> > > I also removed setting `num_inputs` in the MetaDataset. ...
> > 
> > So that means, you should check the code. See all usages of `num_inputs`, to just know what consequences this has.
> > If `num_inputs` is not used anywhere, you should clean up other datasets as well.
> 
> I checked most (if not all) locations inside the Dataset class, and it seems that it is only used for windowing, which is disallowed for the MetaDataset, but not for other Datasets. This is why it is safe here, but not for other datasets.

Just replace the usage in windowing by `get_data_dim("data")` or sth like that (as long as "data" is hardcoded in windowing).

returnn/datasets/basic.py
JackTemaki: > But I think it would really be bad if there would be other use cases where one data stream is somewhat special.

Okay, but fixing this is definitely beyond the scope of this PR. 

> But actually it would be relevant to configure for the childs

This would require to make it configurable for all childs, which again is beyond the scope of this PR. I agree that I chose the "easy" way (for now) by setting this fixed in the base class. But to e.g. support `seq_length_sort_data_key` for the HDFDataset would require the changes currently on review in #438. I would have to check (and maybe change) all datasets.

JackTemaki:
I think it does not make sense to proceed here without test-cases. If fixing the MetaDataset requires to do changes that may break more that the MetaDataset itself, it is not a good idea to do it without (I already did not see that I changed the behaviour of`available_for_inference`, so there can be many more ways to break the code). So for now I am closing this PR, and reactivate this when I wrote some test-cases.

albertz:
This PR can also be in draft mode.

Also, I think adding some `seq_length_sort_data_key` attrib which is `"data"` by default in the base `Dataset` class, and then changing all hardcoded `"data"` to that attrib whenever it is used for seq sorting via `get_seq_order_for_epoch`, this is pretty simple and also pretty safe (I don't see how you could mess that up). But you might maybe want to do that as a separate PR, so that the review becomes easier, and we can focus on one change at a time. Also, that is what the reviews are for. Even without much test cases, I think this would be fine (most other changes in the dataset were done just like that).

But ok, if you feel safer, then first add some test cases. Maybe write some generic checking functions for certain expected behavior (like `_check_init_seq_order_behavior(dataset)` etc), and then some functions which generate some dummy instances of `HDFDataset`, `MetaDataset`, and some other important datasets you care about, and then pass them to these `_check_...` functions. This also should be another separate PR, so that review again is simpler and we don't need to look at multiple changes at once.

albertz:
Note that it will probably be really tricky to really check for all expected behavior. And this particular change in behavior you mentioned (now the input becomes a target, i.e. that implies `available_for_inference=False`), how would you have catched that? That behavior is totally fine and valid. No test case would fail. It's just different from the earlier behavior.

PR #443, opened 2021-02-17
Check size_placeholder in Data.sanity_check

Zettelkasten:
Adds some basic checks about `size_placeholder` in `Data.sanity_check`.

E.g. in `MergeDimsLayer`, this was set incorrectly in `get_out_data_from_opts`, but only in `__init__` (because it needs to create a new size tensor).
I now changed it so that at least the axes are correct.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2627,6 +2627,12 @@ class MergeDimsLayer(_ConcatInputLayer):
         # Time got merged with feature or batch.
         data.time_dim_axis = None
     data.feature_dim_axis = new_feature_dim_axis
+    # Set size_placeholder, here for now without merging any axes. Merged axes are then updated in __init__.
+    data.size_placeholder = {}
+    for old_axis, dyn_size in sorted(input_data.size_placeholder.items()):
+      new_axis = cls._old_axis_to_new_axis(input_data=input_data, merge_axes=axes, old_axis=old_axis)
+      if new_axis != data.batch_dim_axis:
+        data.size_placeholder[data.get_batch_axis_excluding_batch(new_axis)] = dyn_size
     return data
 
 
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -2627,6 +2627,12 @@ def get_out_data_from_opts(cls, name, axes, sources=(), n_out=NotSpecified, out_
         # Time got merged with feature or batch.
         data.time_dim_axis = None
     data.feature_dim_axis = new_feature_dim_axis
+    # Temporarily set size_placeholder, dyn_size values will be set correctly in __init__
+    data.size_placeholder = {}
+    for old_axis, dyn_size in input_data.size_placeholder.items():
+      new_axis = cls._old_axis_to_new_axis(input_data=input_data, merge_axes=axes, old_axis=old_axis)
+      if not new_axis == data.batch_dim_axis:
```
albertz, line 2634:
```suggestion
      if new_axis != data.batch_dim_axis:
```

returnn/tf/layers/basic.py
```diff
@@ -2627,6 +2627,12 @@ def get_out_data_from_opts(cls, name, axes, sources=(), n_out=NotSpecified, out_
         # Time got merged with feature or batch.
         data.time_dim_axis = None
     data.feature_dim_axis = new_feature_dim_axis
+    # Temporarily set size_placeholder, dyn_size values will be set correctly in __init__
```
albertz, line 2630: The comment is a bit confusing. This is often not temporarily.
You might say that it would be updated for merged axes in `__init__` or so.

returnn/tf/layers/basic.py
```diff
@@ -2627,6 +2627,12 @@ def get_out_data_from_opts(cls, name, axes, sources=(), n_out=NotSpecified, out_
...
+    data.size_placeholder = {}
+    for old_axis, dyn_size in input_data.size_placeholder.items():
```
albertz, line 2632: Maybe make this deterministic with `sorted()`. (Although not so relevant, as `__init__` should anyway then overwrite this, but just nicer...)

returnn/tf/layers/basic.py
Zettelkasten: Okay, jep sure!

PR #444, opened 2021-02-17
Data.copy_add_dim_tag: Fix unbroadcast batch dim

Zettelkasten:
`res.batch_shape[res.batch_dim_axis]` will always be `None`,
therefore the existing check did not work if we wanted to
unbroadcast an added batch dimension.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1179,9 +1179,7 @@ class GatherNdLayer(_ConcatInputLayer):
     if position.batch_dim_axis is not None:
       position = position.copy_as_batch_major()
     else:
-      position = position.copy_add_batch_dim(batch_dim_axis=0)
-      position.placeholder = tf.tile(
-        position.placeholder, [tf.shape(x.placeholder)[0]] + [1] * (position.batch_ndim - 1))
+      position = position.copy_add_batch_dim(batch_dim_axis=0, batch_dim=tf.shape(x.placeholder)[x.batch_dim_axis])
     self.output.placeholder = batch_gather(x.placeholder, position.placeholder)  # (B,...)
 
   def get_dep_layers(self):
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -295,6 +295,25 @@ class DimensionTag(object):
         res.append(tag)
     return res
 
+  def get_dim_value(self):
+    """
+    Infers the dim this axis should have if unbroadcasted.
+    If `self.src_data` has a placeholder, will use the shape from there.
+    Otherwise, uses `self.dimension` (if static) or `self.dyn_size` (if dynamic).
+    :rtype: int|tf.Tensor
+    """
+    if self.dimension is not None:
+      return self.dimension
+    if self.kind == self.Types.Batch:
+      from returnn.tf.layers.base import LayerBase
+      return LayerBase.get_recent_layer().get_batch_dim()
+    if self.src_data is not None and self.src_axis is not None and self.src_data.placeholder is not None:
+      from returnn.tf.util.basic import get_shape
+      return get_shape(self.src_data.placeholder)[self.src_axis]
+    if self.dyn_size is not None:
+      return tf.math.reduce_max(self.dyn_size)
+    raise Exception('%s: need placeholder, self.dimension or self.dyn_size for dim value' % self)
+
 
 class SearchBeam:
   """
@@ -1014,9 +1033,10 @@ class Data(object):
     assert self.feature_dim_axis is not None
     return self.copy_with_feature_dim_axis(-1)
 
-  def copy_add_batch_dim(self, batch_dim_axis):
+  def copy_add_batch_dim(self, batch_dim_axis, batch_dim):
     """
     :param int batch_dim_axis:
+    :param int|tf.Tensor batch_dim: the batch dim to unbroadcast to
     :return: copy of myself with added batch-dim
     :rtype: Data
     """
@@ -1028,6 +1048,10 @@ class Data(object):
     data = self.copy()
     if data.placeholder is not None:
       data.placeholder = tf.expand_dims(data.placeholder, batch_dim_axis, name="%s_add_batch_dim" % self.name)
+      if not isinstance(batch_dim, int) or batch_dim != 1:
+        assert batch_dim is not None
+        tiles = [1] * batch_dim_axis + [batch_dim] + [1] * (self.batch_ndim - batch_dim_axis)
+        data.placeholder = tf.tile(data.placeholder, tiles)
     data.batch_dim_axis = batch_dim_axis
     other_special_axes = self.get_special_axes_dict(counted_with_batch_dim=True, only_available=True)
     for k, a in other_special_axes.items():
@@ -1150,8 +1174,7 @@ class Data(object):
     """
     :param DimensionTag dim_tag:
     :param bool unbroadcast: If True unbroadcast the newly added axis.
-      If `dim_tag.src_data` has a placeholder, will use the shape from there.
-      Otherwise, uses `dim_tag.dimension` (if static) or `dim_tag.dyn_size` (if dynamic) to infer shape.
+      Will infer the unbroadcast shape via :func:`DimensionTag.get_dim_value`
     :param int|None axis:
     :rtype: Data
     """
@@ -1164,26 +1187,15 @@ class Data(object):
       """
       if res.placeholder is None:
         return
+      assert dim_tag.kind != DimensionTag.Types.Batch
       assert res.batch_shape[axis] == 1
       with tf.name_scope("copy_add_dim_by_tag_unbroadcast"):
-        dyn_size = None
-        if dim_tag.src_data is not None and dim_tag.src_axis is not None and dim_tag.src_data.placeholder is not None:
-          dyn_size = tf.shape(dim_tag.src_data.placeholder)[dim_tag.src_axis]
-        if dyn_size is None:
-          dyn_size = dim_tag.dimension
-        if dyn_size is None and dim_tag.dyn_size is not None:
-          dyn_size = tf.math.reduce_max(dim_tag.dyn_size)
-        if dyn_size is None and dim_tag.kind == DimensionTag.Types.Batch:
-          from returnn.tf.layers.base import LayerBase
-          dyn_size = LayerBase.get_recent_layer().get_batch_dim()
-        assert dyn_size is not None, 'need placeholder, dim_tag.dimension or dim_tag.dyn_size for unbroadcast'
-        tiles = [1] * axis + [dyn_size] + [1] * (res.batch_ndim - axis - 1)
+        tiles = [1] * axis + [dim_tag.get_dim_value()] + [1] * (res.batch_ndim - axis - 1)
         res.placeholder = tf.tile(res.placeholder, tiles)
 
     if dim_tag.kind == DimensionTag.Types.Batch:
-      res = self.copy_add_batch_dim(batch_dim_axis=axis)
-      if unbroadcast:
-        maybe_unbroadcast_placeholder()
+      batch_dim = dim_tag.get_dim_value() if unbroadcast else 1
+      res = self.copy_add_batch_dim(batch_dim_axis=axis, batch_dim=batch_dim)
       return res
     # Note: if dim_tag is feature, but we are sparse, we just treat is as spatial, handled below.
     if dim_tag.kind == DimensionTag.Types.Feature and not self.sparse:


albertz:
Where does this fail?
I think this is wrong. If the batch dim exists, we also would expect that it is not 1, i.e. already unbroadcasted, so it doesn't make sense to call this func then.

Zettelkasten:
> Where does this fail?
> I think this is wrong. If the batch dim exists, we also would expect that it is not 1, i.e. already unbroadcasted, so it doesn't make sense to call this func then.

The batch dim does not already exist. For example, consider that we call `data.copy_add_dim_by_tag(dim_tag, .., unbroadcast=True)` on some `data` of the form `[F|5]` (where `dim_tag` is of type Batch).
Then, `copy_add_dim_by_tag` first calls `copy_add_batch_dim`. After this, `res` is of the form `[B,F|5]`.
Then afterwards, we unbroadcast by calling `maybe_unbroadcast_placeholder`.
At this point, `res.batch_shape` is `(None,5)` (because `res.shape` is just `(5,)`, and `res.batch_shape` just adds `None` for the batch axis).

We could also fix this by changing the assert to:
`assert res.get_batch_shape(batch_dim=1)[axis] == 1`
Would that be easier to understand?

albertz:
But that looks inconsistent. In `maybe_unbroadcast_placeholder`, we should not just assume that the batch dim is 1. Why would we assume that? The batch dim is usually never 1. That would be wrong if it is 1.

So that sounds like `copy_add_dim_by_tag` does sth wrong about the batch dim. It should definitely not introduce a batch axis of dim 1.

Zettelkasten:
> So that sounds like `copy_add_dim_by_tag` does sth wrong about the batch dim. It should definitely not introduce a batch axis of dim 1.

Hm okay, I see. If that is wrong, than also the behavior of `copy_add_batch_dim` is wrong, [which also simply adds a batch axis of dim 1](https://github.com/rwth-i6/returnn/blob/fe4a0a4da832893b0b4f0bb6d233834ad3f0e9dd/returnn/tf/util/data.py#L1037):
```python
if data.placeholder is not None:
  data.placeholder = tf.expand_dims(data.placeholder, batch_dim_axis, name="%s_add_batch_dim" % self.name)
```

What else should it do? Always unbroadcast? If so, to which dim? Using `LayerBase.get_recent_layer().get_batch_dim()`?

albertz:
Yea, looks wrong as well.

I mean, it is up to us. We can either say:

* Batch-dim should always be predefined globally. I.e. by the mini-batch construction, and can have an optional factor for beam. Whenever you have a batch dim, it must be exactly this value. This is currently always the case. (Or maybe almost always. Maybe some rare uses of `copy_add_batch_dim` just didn't result in a problem because then it was broadcasted away.)
* Batch-dim is totally arbitrary. You cannot expect anything. I don't quite like this. But on the other side, `Data` is really generic, and doesn't really expect any global state.

I'm not sure. Maybe `copy_add_batch_dim` should at least have an explicit parameter for the batch dim. And `copy_add_dim_by_tag` should reuse the same batch dim when you pass a batch dim tag. The batch dim tag should somehow have a reference to the batch dim.

Currently the batch dim tag does not have a reference to the batch dim. That's another issue we are having here. In that case, we fall back to the somewhat ugly way to access the global state. But maybe that's okay for now. For the future, we probably should keep a reference to the batch dim in the batch dim tag.

returnn/tf/layers/basic.py
```diff
@@ -1179,9 +1179,7 @@ def __init__(self, position, **kwargs):
     if position.batch_dim_axis is not None:
       position = position.copy_as_batch_major()
     else:
-      position = position.copy_add_batch_dim(batch_dim_axis=0)
-      position.placeholder = tf.tile(
-        position.placeholder, [tf.shape(x.placeholder)[0]] + [1] * (position.batch_ndim - 1))
+      position = position.copy_add_batch_dim(batch_dim_axis=0, batch_dim=tf.shape(x.placeholder)[0])
```
albertz, line 1182:
```suggestion
      position = position.copy_add_batch_dim(batch_dim_axis=0, batch_dim=tf.shape(x.placeholder)[position.batch_dim_axis])
```
?

Zettelkasten: Hm yes that is nicer (but with `x.batch_dim_axis`). I just copied it from how it was before, and it shouldn't matter because `x` is copied as batch major anyway. But still, yup!

returnn/tf/util/data.py
```diff
@@ -295,6 +295,23 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
         res.append(tag)
     return res
 
+  def get_unbroadcast_dim(self):
```
albertz, line 298: (Meta: This is somewhat annoying to review on GitHub, because you don't see which class this belongs too... (Here `DimensionTag`.) I currently expand it until I see the class. Not sure what to do about it. Maybe tell GitHub as a feature request to show the class.)

The function name is not good. Why "unbroadcast"? You are simply returning the dimension here. Maybe `get_dimension`. Or maybe `get_dim_value`. Or so.

returnn/tf/util/data.py
```diff
@@ -295,6 +295,23 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
...
+    """
+    Infers the dim this axis should have if unbroadcasted.
+    If `self.src_data` has a placeholder, will use the shape from there.
+    Otherwise, uses `self.dimension` (if static) or `self.dyn_size` (if dynamic).
+    :rtype: int|Tensor
+    """
+    if self.src_data is not None and self.src_axis is not None and self.src_data.placeholder is not None:
+      return tf.shape(self.src_data.placeholder)[self.src_axis]
+    if self.dimension is not None:
+      return self.dimension
```
albertz, line 308: Move `self.dimension` to the top of this function. This should always be preferred if set.

returnn/tf/layers/basic.py
Zettelkasten: Hm yes that is nicer (but with `x.batch_dim_axis`). I just copied it from how it was before, and it shouldn't matter because `x` is copied as batch major anyway. But still, yup!

returnn/tf/util/data.py
```diff
@@ -295,6 +295,23 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
...
+    if self.dyn_size is not None:
+      return tf.math.reduce_max(self.dyn_size)
+    if self.kind == self.Types.Batch:
```
albertz, line 311: This should be the second check. `LayerBase.get_recent_layer().get_batch_dim()` should be preferred over `tf.shape(self.src_data.placeholder)[self.src_axis]`. `get_batch_dim` will just depend on the input data (extern data), nothing else, so is immediately available, and so TF doesn't need to wait until `self.src_data.placeholder` is available.

returnn/tf/util/data.py
```diff
@@ -295,6 +295,23 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
...
 
+  def get_unbroadcast_dim(self):
+    """
+    Infers the dim this axis should have if unbroadcasted.
+    If `self.src_data` has a placeholder, will use the shape from there.
+    Otherwise, uses `self.dimension` (if static) or `self.dyn_size` (if dynamic).
+    :rtype: int|Tensor
+    """
+    if self.src_data is not None and self.src_axis is not None and self.src_data.placeholder is not None:
+      return tf.shape(self.src_data.placeholder)[self.src_axis]
```
albertz, line 306: You should use `get_shape` instead.

returnn/tf/util/data.py
```diff
@@ -295,6 +295,23 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
...
+    if self.dimension is not None:
+      return self.dimension
+    if self.dyn_size is not None:
+      return tf.math.reduce_max(self.dyn_size)
+    if self.kind == self.Types.Batch:
+      from returnn.tf.layers.base import LayerBase
+      return LayerBase.get_recent_layer().get_batch_dim()
+    assert False, 'need placeholder, self.dimension or self.dyn_size for unbroadcast'
```
albertz, line 314:
```suggestion
    raise Exception('%s: need placeholder, self.dimension or self.dyn_size for dim value' % self)
```
Is more clean, and also adds information about `self`.

returnn/tf/util/data.py
```diff
@@ -1014,9 +1031,10 @@ def copy_with_feature_last(self):
     assert self.feature_dim_axis is not None
     return self.copy_with_feature_dim_axis(-1)
 
-  def copy_add_batch_dim(self, batch_dim_axis):
+  def copy_add_batch_dim(self, batch_dim_axis, batch_dim=1):
```
albertz, line 1034: Remove the default value. We really should enforce this.
(Yes, that means, maybe some other code needs to be fixed. But that is what we should do then.)

returnn/tf/util/data.py
```diff
@@ -1028,6 +1046,10 @@ def copy_add_batch_dim(self, batch_dim_axis):
     data = self.copy()
     if data.placeholder is not None:
       data.placeholder = tf.expand_dims(data.placeholder, batch_dim_axis, name="%s_add_batch_dim" % self.name)
+      if batch_dim != 1:
```
albertz, line 1049:
```suggestion
      if not isinstance(batch_dim, int) or batch_dim != 1:
```

returnn/tf/util/data.py
```diff
@@ -1150,8 +1172,7 @@ def copy_add_dim_by_tag(self, dim_tag, unbroadcast=False, axis=None):
     """
     :param DimensionTag dim_tag:
     :param bool unbroadcast: If True unbroadcast the newly added axis.
-      If `dim_tag.src_data` has a placeholder, will use the shape from there.
-      Otherwise, uses `dim_tag.dimension` (if static) or `dim_tag.dyn_size` (if dynamic) to infer shape.
+      Will infer the unbroadcast shape via :func:`DimensionTag.get_unbroadcast_dim`
```
albertz, line 1175: `get_dim_value` or whatever...

returnn/tf/util/data.py
```diff
@@ -1164,26 +1185,15 @@ def maybe_unbroadcast_placeholder():
...
-        tiles = [1] * axis + [dyn_size] + [1] * (res.batch_ndim - axis - 1)
+        tiles = [1] * axis + [dim_tag.get_unbroadcast_dim()] + [1] * (res.batch_ndim - axis - 1)
         res.placeholder = tf.tile(res.placeholder, tiles)
 
     if dim_tag.kind == DimensionTag.Types.Batch:
-      res = self.copy_add_batch_dim(batch_dim_axis=axis)
-      if unbroadcast:
-        maybe_unbroadcast_placeholder()
+      # Always unbroadcast batch dim axis
+      res = self.copy_add_batch_dim(batch_dim_axis=axis, batch_dim=dim_tag.get_unbroadcast_dim())
```
albertz, line 1196: Hm, now I see. I'm actually reconsidering whether we should do that here.
Maybe better:
```suggestion
      res = self.copy_add_batch_dim(batch_dim_axis=axis, batch_dim=dim_tag.get_dim_value() if unbroadcast else 1)
```
This is for efficiency. E.g. when you do `a *  b` (in terms of layers, e.g. via `CombineLayer`), and `b` doesn't have a batch-dim, and then this adds the dimensions (with `unbroadcast=False`), it would be way more efficient to not unbroadcast the batch-dim (both in terms of computation (unnecessary copying) and in terms of memory).

albertz: This is somewhat inconsistent to what I said before. But this should really only play a role for an intermediate `Data` instance. A layer output would never have that. So I don't think this is a problem here.

albertz: (This should also fix the failing test cases.)

Zettelkasten: Okay sure! Let's do that then.

returnn/tf/util/data.py
Zettelkasten: (About GitHub: We can [file a suggestion here](https://support.github.com/contact/feedback) if it is annoying. I personally have never reviewed anything yet though)

About the name: Yes. Let's go for `get_dim_value` then, I think `get_dimension` is a little bit confusing as there is also the attribute `dimension`.

returnn/tf/util/data.py
```diff
@@ -1014,9 +1033,10 @@ def copy_with_feature_last(self):
     assert self.feature_dim_axis is not None
     return self.copy_with_feature_dim_axis(-1)
 
-  def copy_add_batch_dim(self, batch_dim_axis):
+  def copy_add_batch_dim(self, batch_dim_axis, batch_dim):
     """
     :param int batch_dim_axis:
+    :param int|Tensor batch_dim: the batch dim to unbroadcast to
```
albertz, line 1039:
```suggestion
    :param int|tf.Tensor batch_dim: the batch dim to unbroadcast to
```

returnn/tf/util/data.py
```diff
@@ -295,6 +295,25 @@ def get_uniq_collection(cls, tags, is_equal_opts=None):
...
+    if self.dimension is not None:
+      return self.dimension
+    if self.kind == self.Types.Batch:
+      from returnn.tf.layers.base import LayerBase
+      return LayerBase.get_recent_layer().get_batch_dim()
+    if self.src_data is not None and self.src_axis is not None and self.src_data.placeholder is not None:
+      from returnn.tf.util.basic import get_shape
+      return get_shape(self.src_data.placeholder)[self.src_axis]
+    if self.dyn_size is not None:
+      return tf.math.reduce_max(self.dyn_size)
```
albertz, line 314: Btw, since when is it `tf.math.reduce_max`? Already since TF 1.8? (I'm also not sure what minimum TF version we want to support. If all tests are passing, and noone complains, we can just leave it.)

albertz (approved):
Looks good now (except my minor comments).

returnn/tf/util/data.py
Zettelkasten: I don't know. Tests seem to pass though :D

Zettelkasten:
Thanks a lot @albertz for the swift and detailed review and all your comments! It was very helpful.

Issue #447, opened 2021-02-19
Reusing parameter with different variable name not working


Zettelkasten:
As discussed in #446, `ReuseParams.get_absolute_name_scope_prefix` does not infer the name of a reused parameter correctly if a `custom_func` is used.

I have added a simple test case, where I share the `W` parameter of a `LinearLayer` with the `QKV` parameter of a `SelfAttentionLayer`:
```python
"self_att": {"class": "self_attention", "num_heads": n_heads, "total_key_dim": n_total, "n_out": n_total},
"linear": {"class": "linear", "n_out": n_total * 3, "activation": None, "with_bias": False,
  "reuse_params": {
    "auto_create_missing": False,  # should not matter as we do not have any bias
    "map": {"W": {"reuse_layer": "self_att", "custom": custom}}}}
```
with
```python
def custom(reuse_layer, *args, **kwargs):
  return reuse_layer.params['QKV']
```

This gives the error:
```
  File "/home/runner/work/returnn/returnn/returnn/tf/network.py", line 703, in _create_layer
    line: layer = layer_class(**layer_desc)
    locals:
      layer = <not found>
      layer_class = <local> <class 'returnn.tf.layers.basic.LinearLayer'>
      layer_desc = <local> {'n_out': 40, 'activation': None, 'with_bias': False, 'reuse_params': <ReuseParams reuse_layer None, map {'W': <ReuseParams reuse_layer <SelfAttentionLayer 'self_att' out_type=Data(shape=(None, 40), batch_shape_meta=[B,T|'time:var:extern_data:data',F|40])>, map None>}>, 'sources': [<SourceLayer '..., len = 8
  File "/home/runner/work/returnn/returnn/returnn/tf/layers/basic.py", line 1430, in __init__
    line: weights = self.add_param(tf_compat.v1.get_variable(
            name="W", shape=weights_shape, dtype=tf.float32, initializer=fwd_weights_initializer))
    locals:
      weights = <not found>
      self = <local> <LinearLayer 'linear' out_type=Data(shape=(None, 40), batch_shape_meta=[B,T|'time:var:extern_data:data',F|40])>
      self.add_param = <local> <bound method LayerBase.add_param of <LinearLayer 'linear' out_type=Data(shape=(None, 40), batch_shape_meta=[B,T|'time:var:extern_data:data',F|40])>>
      tf_compat = <global> <module 'returnn.tf.compat' from '/home/runner/work/returnn/returnn/returnn/tf/compat.py'>
[...]

PR #448, opened 2021-02-19
Test ReuseParams with different variable names

Zettelkasten:
The test case from #447, currently failing.


albertz:
So, we basically have outlined a possible solution for this in the discussion in #447.

Basically, what we want (`self` being the layer):

* Let's assume you have `param = self.add_param(tf.get_variable(name))` (or sth like that).
* `param` is what you would directly use in this code, for whatever purpose, e.g. matmul in `LinearLayer`.
* `self.params[name] is param` should be true after this call. This implies:
  *  `self.params` includes potential non-owned vars (shared/reuse) and/or potential transformations (transpose, weight noise, weight norm, etc) (i.e. not the original variable, but just some tensor).
* Weight noise, weight norm and other transformations, and also L2 losses etc, should only be applied if the tensor or variable is owned by the layer (determined by looking at the namespace).

So we need to change the logic of `add_param` to implement this logic for `self.params`. The difficulty is to infer the `name`. This should be doable by extending `var_creation_scope`:
* In case there is no reuse/sharing or transformation, no custom getter, let's not change anything. Then `add_param` would also keep the code for this simple case.
* With a custom getter, we can grab `name`, and store it somewhere.

We might need a separate way to get a list of variables of a layer. (Or do we really? What would be the use case?) Or we could also simply iterate through all existing variables (via the global collections) and check for the namespace if we need to filter.

Btw, you might ask, why do we need `self.add_param(tf.get_variable(name))`, and why not simply do sth like `self.create_param(name, ...)` instead, and not use `tf.get_variable`. This is because we want to be able to use other TF code, which makes use of `tf.get_variable`. E.g. when we use the original LSTM by TF, or other things. We want to support that. (Btw, if this was unclear, you might also extend the documentation in this PR for that.)

Actually, when this is external code, note that it could create multiple variables, i.e. multiple calls to `get_variable`. We should be careful to handle that correctly. Maybe the whole logic of adding it to `self.params` should be in `var_creation_scope`. Or we can keep it in `add_param` but call to that in `var_creation_scope`.

Are you going to try to implement this? (I'm somewhat short on time right now, not sure when I get to this.)

Zettelkasten:
> Are you going to try to implement this? (I'm somewhat short on time right now, not sure when I get to this.)

Thanks for your detailed notes! Yes, I can certainly give it a try implementing this. If I run into issues/open questions I of course will let you know. Thanks!

returnn/tf/layers/base.py
```diff
@@ -836,14 +836,18 @@ def add_param(self, param, custom_update=None, trainable=None, saveable=None, ax
       custom_update.set_on_var(param)
     if axes_split_info:
       tf_util.set_param_axes_split_info(param, axes_split_info)
-    if self.reuse_params:
-      name_scope_prefix = self.reuse_params.get_absolute_name_scope_prefix(base_layer=self, param=param)
+    if getattr(param, "RETURNN_layer_map_name", None) is not None:
```
albertz, line 839: It should be `_RETURNN_layer_map_name`.

albertz: But also, I think you don't need this at all, and this approach only leads to problems.

returnn/tf/layers/base.py
```diff
@@ -1755,8 +1759,12 @@ def custom_getter(getter, name, *args, **kwargs):
     assert name.startswith(abs_scope_prefix)
     param_name = name[len(abs_scope_prefix):]  # e.g. "W" (not "rec/W")
     if self.custom_func:
-      return self.custom_func(
+      variable = self.custom_func(
         base_layer=base_layer, reuse_layer=self.reuse_layer, name=param_name, getter=getter, full_name=name, **kwargs)
+      # The name of the variable created by custom_func might not match param_name.
+      # We store it here for LayerBase.add_param.
+      variable.RETURNN_layer_map_name = param_name
```
albertz, line 1766: I suggested `_RETURNN_layer_map_name` because this actually should be a map (dict). This cannot be the name directly, because this param might be used by multiple layers, under different names. So I thought about a mapping layer -> name. However, I'm now thinking, it might even be used by only one single layer under different names...

Zettelkasten: Hm I see, I kind of thought that there are no unrelated calls to `custom_getter` before the corresponding `add_param` is called. That's of course not so nice (and probably breaks if e.g. a `custom_func` calls `get_variable` multiple times or so).
Maybe we can make it a layer -> list[name] dict (and then always use the last entry in `add_param`?) I don't really know a good solution.

albertz: > and then always use the last entry ...

Such heuristics are never nice. They mostly work, but in some rare cases, or when we do sth more crazy, they suddenly fail. And then this is very annoying to debug. (Unfortunately, we have a couple of such heuristics, which have worked for the initial use case, but then later on failed, and caused quite some extra work. "Technical debt" is the keyword for this...)

But I think we don't need that here.

We can know when we are in the direct `get_variable` call from the layer. (Via `var_creation_scope`.)
And in there, we have the original name, and we can get the param, and we can directly do `self.params[name] = param`, and `add_param` doesn't need to do anything anymore.

And when `var_creation_scope` does not create a custom getter, this implies there will be no special handling. Then `add_param` can catch this.

PR #449, opened 2021-02-21
SplitLayer: Test beam search

Zettelkasten:
A test case for #450.

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -1123,6 +1123,31 @@ def test_SplitLayer_after_SplitDimsLayer():
     assert out_t.shape.as_list() == [None, None, 2, 5]
 
 
+def test_SplitLayer_search():
+  n_batch, n_time, n_in, n_out = 7, 3, 10, 10
+  beam_size = 4
+  config = Config({
+    "extern_data": {
+      "data": {"dim": n_in},
+      "classes": {"dim": n_out, "sparse": True, "available_for_inference": False}},
+    "debug_print_layer_output_template": True
+  })
+  with make_scope():
+    net = TFNetwork(config=config, search_flag=True, train_flag=False, eval_flag=True)
+    net.construct_from_dict({
+      "encoder_seq": {"class": "linear", "activation": "tanh", "n_out": 5},
+      "encoder": {"class": "reduce", "mode": "sum", "from": ["encoder_seq"], "axis": "T"},
+      "output": {"class": "rec", "from": [], "target": "classes", "max_seq_len": 20, "unit": {
+        "embed": {"class": "linear", "from": ["prev:output"], "activation": None, "n_out": 10},
+        "split": {"class": "split", "size_splits": (5, 5), "axis": "F", "from": ["embed"]},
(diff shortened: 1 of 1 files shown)


albertz:
This was not included in #453, right? Can you rebase this? So this test should pass after rebase, right? (It is passes, we can just merge the test in as well.)

albertz:
Btw, please remove the `base.py.save` file from the commits. Also, a single commit should be enough. And then you can mark the PR as ready-for-review (if it looks fine).

tests/test_TFNetworkLayer.py
```diff
@@ -1135,7 +1135,8 @@ def test_SplitLayer_search():
   with make_scope():
     net = TFNetwork(config=config, search_flag=True, train_flag=False, eval_flag=True)
     net.construct_from_dict({
-      "encoder": {"class": "linear", "activation": "tanh", "n_out": 5},
+      "encoder_seq": {"class": "linear", "activation": "tanh", "n_out": 5},
+      "encoder": {"class": "reduce", "mode": "sum", "from": ["encoder_seq"], "axis": "T"},
```
patrick-wilken, line 1139: Wait, does this work now? What you put into "from" is available in the rec unit as "data:source", but you don't use that. Maybe it works by chance because the encoder and rec layer time dims now match and therefore "base:encoder" is fed per timestep into the loop?

albertz: What do you mean? Here encoder has time completely removed (i.e. just [B,D]). And it is in the rec-layer just as-is.

(Btw, off topic here, but if you need to have more generic unroll logic in the RecLayer, I thought about implementing sth like an UnrollLayer. Which would check of course that the time axis match such that this is possible. And which would use tf.TensorArray internally.)

patrick-wilken: Oops, sorry, totally misread the code :D Yes, should be fine.

albertz: But even without the reduce, this should work. Then you would just have this extra time dim inside the rec layer. (It would not do automatic unrolling. I think this would be confusing. It should be explicit.)

Zettelkasten:
> Btw, please remove the `base.py.save` file from the commits.

Äh ups, sorry! This should be ready now.

Btw, what is more convenient for you if I add onto a PR based on feedback? Add new commits so you can see what exactly changed, or amending + force-pushing so that the commits are cleaned up like they would appear if merged?

albertz:
I prefer cleaning up (amending, squashing, whatever) + force-pushing, so that I can directly merge if it is fine.

Or it doesn't matter if the PR is so small that I anyway use the "squash and merge" function on GitHub. But once you think it make sense to keep it over multiple commits, then please make sure it is cleaned up.

I anyway want to see the whole change again every time I review, to get the whole picture. If there is no change in some file, I can click this "viewed file" button in GitHub, and it remembers that from my last review. In general, that's why I also prefer to break down a PR as much as possible, and focus only on one single change. We sometimes had huge PRs which changed multiple things together, and then also they had many problems, and this was really a pain to review, again and again. But so far, your PRs are really fine!

PR #452, opened 2021-02-22
HDFDataset: support having no target labels

patrick-wilken:
Needed when only having dense targets. In this case, although `self.labels` is unset we don't intend to use the deprecated(?) `fin["labels"]`.


albertz (approved):

patrick-wilken:
Didn't mean to make @JackTemaki a reviewer, feel free to merge.

albertz:
> Didn't mean to make @JackTemaki a reviewer, feel free to merge.

Why not? It doesn't always need to be me. And also, at least two reviewers per PR would be better.

But anyway, this is such a trivial change, so I'm just merging now.

patrick-wilken:
Ok, noted :)

Issue #450, opened 2021-02-21
SplitLayer: Beam search not working


Zettelkasten:
Inside a `RecLayer`, I am accessing the output `split/0` of a `SplitLayer`:
```python
"split": {"class": "split", "size_splits": (5, 5), "axis": "F", "from": ["embed"]},
"output_prob": {"class": "softmax", "from": ["split/0", "base:encoder"], "target": "classes", "loss": "ce"},
```

During search, the construction of `output_prob` fails:
```
  File "/home/runner/work/returnn/returnn/returnn/tf/network.py", line 657, in _create_layer_layer_desc
    line: layer_desc = SearchChoices.translate_to_common_search_beam(layer_desc)
    locals:
      layer_desc = <local> {'target': 'classes', 'sources': [<InternalLayer output/'split/0' out_type=Data(shape=(5,), time_dim_axis=None, beam=SearchBeam(name='output/prev:output', beam_size=4), batch_shape_meta=[B,F|5])>, <LinearLayer 'encoder' out_type=Data(shape=(None, 5), batch_shape_meta=[B,T|'time:var:extern_data:da...
      SearchChoices = <local> <class 'returnn.tf.layers.base.SearchChoices'>
      SearchChoices.translate_to_common_search_beam = <local> <bound method SearchChoices.translate_to_common_search_beam of <class 'returnn.tf.layers.base.SearchChoices'>>
  File "/home/runner/work/returnn/returnn/returnn/tf/layers/base.py", line 1988, in translate_to_common_search_beam
    line: assert search_choices, "layer %r has beam %r but no search choices; from layer desc\n%s" % (
            layer, layer.output.beam, pformat(layer_desc))
    locals:
      search_choices = <local> None
      layer = <local> <InternalLayer output/'split/0' out_type=Data(shape=(5,), time_dim_axis=None, beam=SearchBeam(name='output/prev:output', beam_size=4), batch_shape_meta=[B,F|5])>
      layer.output = <local> Data(name='split/0_output', shape=(5,), time_dim_axis=None, beam=SearchBeam(name='output/prev:output', beam_size=4), batch_shape_meta=[B,F|5])
      layer.output.beam = <local> SearchBeam(name='output/prev:output', beam_size=4)
      pformat = <local> <function pformat at 0x7f9baaeafb90>
[...]

PR #453, opened 2021-02-23
Fixes for #450

patrick-wilken:
The problem was that in template construction we re-initialize some "partially_finished" layers in the end and the sub-layers were not included here. Only after re-initialization the beam is set.

The test still fails because the `base:encoder` has a different time-dim which is kept in the loop, but this is a separate issue and can be fixed by using `"from": "encoder"` in the rec layer and then `"data:source"` in `"output_prob"`. (Or by using this "same dim tag" in extern_data I guess)

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -1156,6 +1156,12 @@ class _SubnetworkRecCell(object):
         if layer_ in ConstructCtx.partially_finished:
           if lself.got_uninitialized_deps_count == 0:  # in this case, we safely know that it is finished
             ConstructCtx.partially_finished.remove(layer_)
+            # Sub-layers were also added to 'partially_finished' to re-init them as well if necessary.
+            # But 'add_templated_layer' is not called for sub-layers, so clean them up here as well.
+            for sub_layer in layer_.sub_layers.values():
+              if sub_layer in ConstructCtx.partially_finished:
+                ConstructCtx.partially_finished.remove(sub_layer)
+
         return layer_
 
       # noinspection PyMethodParameters
@@ -1222,6 +1228,8 @@ class _SubnetworkRecCell(object):
                 # Note, we don't add dependencies to the sub-layer, instead, the dependency graph continues via the root
                 # layer (see above).
                 ConstructCtx.layers[-1].add_dependency(sub_layer, is_prev_time_frame=is_prev_time_frame)
+                if sub_layer not in ConstructCtx.partially_finished:
+                  ConstructCtx.partially_finished.append(sub_layer)
             return sub_layer
         # Need to create layer instance here now to not run into recursive loops.
         # We will extend it later in add_templated_layer().
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -3243,6 +3243,7 @@ class _TemplateLayer(LayerBase):
     self.construct_stack = construct_stack
     self._template_base = None  # type: typing.Optional[_TemplateLayer]
     self._cell = cell
+    self.sub_layers = {}  # type: typing.Dict[str,_TemplateLayer]  # layer name -> layer
 
   def __repr__(self):
     if self.is_initialized:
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/rec.py
```diff
@@ -1156,6 +1156,12 @@ def add_templated_layer(lself, name, layer_class, **layer_desc):
         if layer_ in ConstructCtx.partially_finished:
           if lself.got_uninitialized_deps_count == 0:  # in this case, we safely know that it is finished
             ConstructCtx.partially_finished.remove(layer_)
+            # Sub-layers were also added to 'partially_finished' to re-init them as well if necessary.
+            # But 'add_templated_layer' is not called for sub-layers, so clean them up here as well.
+            for sub_layer in layer_.sub_layers.values():
+              if sub_layer in ConstructCtx.partially_finished:
+                ConstructCtx.partially_finished.remove(sub_layer)
```
patrick-wilken, line 1163: @albertz This is not needed for the test to succeed. Is it needed at all? It felt like it. Also, it might be not the right place to remove the sub-layers. Maybe in `__call__()`? This template construction is still to complex for me 😆

returnn/tf/layers/rec.py
```diff
@@ -3294,10 +3303,13 @@ def get_sub_layer(self, layer_name):
     assert res, "Could not get out data for sub-layer template {}.".format(full_layer_name)
     output, network, sub_layer_class = res
 
-    sub_layer_template = _TemplateLayer(self.network, full_layer_name)
+    # During template construction templates might be re-initialized by calling init() several times. For sub-layers
+    # we have to do the same, so we have to store the sub-layers created so far.
```
albertz, line 3307: Why do the sub layer templates need to be stored?

returnn/tf/layers/rec.py
```diff
@@ -3243,6 +3251,7 @@ def __init__(self, network, name, construct_stack=None, cell=None):
     self.construct_stack = construct_stack
     self._template_base = None  # type: typing.Optional[_TemplateLayer]
     self._cell = cell
+    self.sub_layers = {}  # type: typing.Dict[str,LayerBase]  # layer name -> layer
```
albertz, line 3254: Not `_TemplateLayer`?

returnn/tf/layers/rec.py
albertz: I don't know... The whole logic about sublayers seems somewhat complex. But without looking at it more deeply, I don't really see a good solution now.

returnn/tf/layers/rec.py
patrick-wilken: Because of this trial and error approach of `_construct_template()`, `get_sub_layer()` might be called several times.
Then in `move_outside_loop` there is the check:
```
assert self.layer_data_templates[layer.name] is layer
```
where `layer` is taken from the layer `_TemplateLayer.dependencies` of another layer. If we create a new `_TemplateLayer` in each `get_sub_layer` call this check would fail, because we have created a new object. We have to update the existing object instead, also because it is referenced in other layers in `dependencies`.

albertz: I see, the `... is layer` check. Is this the only place where we expect the same instance?

Anyway, this is relevant for the comment. This is the crucial information you need to explain in the comment, why we cannot recreate the instances. (And just saying "because we have some `assert` for it elsewhere" does not explain anything. Then the question is, why do we have the `assert`...)

returnn/tf/layers/rec.py
patrick-wilken: Yes, sublayers are complicated to handle. But also because `_construct_template()` is already very complex...

returnn/tf/layers/rec.py
patrick-wilken: > Is this the only place where we expect the same instance?

It seems to be the only explicit check, yes. Moving out layers could be done by checking dependencies via the layer name only, so without the actual instance (although sometimes we check for the name, sometimes for the instance in the current code). But I think to get search choices we need the dependency instances, and probably if it's not the latest correct ones this will cause bugs. So I think the check is good.

But yes, let me update the comment.

albertz (approved):
Ok, from a glance, it looks ok. I guess it should be fine to merge.
(We anyway should probably clean this up at some point. Although I don't really know how... The template construction will always stay tricky.)

albertz:
@Zettelkasten or @JackTemaki are you also reviewing?

JackTemaki:
> @Zettelkasten or @JackTemaki are you also reviewing?

I am not familiar with the layer construction code, so I can provide no useful feedback here.

Zettelkasten (commented):
I don’t think I can add much here, I am not very familiar with this part of code.

I can confirm that the changes work for my (more complex) use case though. Thanks for the swift fix!

albertz:
> I am not familiar with the layer construction code, ...

I think it would be good if there are more people who get familiar with it, though. (And if the complaint is, it is too complicated or messy, that's even more a reason for it. And then we should improve it...)
(Anyway, this gets off-topic here.)

Ok, I will just merge then.

PR #457, opened 2021-03-02
SoftmaxOverSpatialLayer: support integers as window start

patrick-wilken:
I needed attention over encoder positions from 0 to `n(t)`, where `n` is determined by the network in each decoder step `t`.
Using `window_start` and `window_size` of `SoftmaxOverSpatialLayer` is the best way to do this, right? I have a layer calculating `n`, I just need to set `window_start` to 0 which is now possible (without creating a ConstantLayer or similar).

What is the difference between `start` and `window_start` by the way? They seem to do the same thing.


JackTemaki (approved):

albertz (approved):

albertz:
> What is the difference between `start` and `window_start` by the way? They seem to do the same thing.

`start` is not for a window, but until the end of the sequence.
Maybe a bit counter-intuitive. But then maybe we should remove `window_start` instead. I don't know...

PR #458, opened 2021-03-03
import_ add import path env var

JackTemaki:
Adds another env-var to be able to specify the _pkg_import_path from outside.


albertz (approved):

albertz:
I merged this just for completeness.

However, I don't really understand why you need that. And I think using that might even be a bad idea.

PR #459, opened 2021-03-05
Avoid zero division during peak normalization in the audio extraction pipeline

dthulke:


albertz:
I wonder, did you really had this case? An audio segment of complete silence (all samples exactly 0.0)?

dthulke:
Yes, I had this case. The corpus I used had annotations for segments at the end of the recording where the actual recording already had ended.

PR #460, opened 2021-03-05
Use output_before_softmax_flat in get_output_target_scores

dthulke:
The output may not have a time dimension and seq_lens and thus _flatten_or_merge fails if output_flat is None.

This is already handled in `init` and either output_flat or output_before_softmax_flat are always not None.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -7900,8 +7900,8 @@ class CrossEntropyLoss(Loss):
     """
     output_flat = self.output_flat
     if output_flat is None:
-      output_flat = self._flatten_or_merge(
-        self.output.placeholder, self.output_seq_lens, time_major=self.output.is_time_major)
+      output_flat = self.output_before_softmax_flat
+    assert output_flat is not None
     target_flat_exp = tf.stack(
       [tf.range(tf.shape(self.target_flat)[0], dtype=tf.int32),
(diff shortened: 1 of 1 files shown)


albertz:
Why did you close this? Instead of opening another PR, just update the original one when you need to update sth. (Also e.g. after review.)

dthulke:
I just saw that some test were failing and closed the PR till I had time to fix it. Didn't know till today that you can't reopen PRs after force pushing to the branch to update it.

albertz:
Ah, maybe I missunderstood. You plan to reopen this once you finished it? In such a case, you can put the PR into draft mode. Then it's clear that you are still working on it. Closing it is a signal that this is obsolete or can be ignored.

albertz:
But also, I think reopening does in fact not work for this specific PR in this specific case, because the PR belongs to a specific branch, and you are using the branch now for another PR (which you should never do -- a branch should belong to exactly one PR only).

dthulke:
In general I totally agree. I pushed it and noticed that some test are failing. Since I was in a meeting I didn't have time to directly look at it and closed it to indicate that the PR should be ignored for now.
A bit later I fixed the issue, amended the commit and force pushed the branch. The plan was then to reopen this PR, but then I noticed that GitHub does not allow to reopen PRs if you force pushed to the branch (you probably need to reopen the PR first and the force push to the branch). Thus I created then new PR.

albertz:
Use the draft mode for this purpose, but don't close it.

PR #462, opened 2021-03-05
Use output_before_softmax_flat in get_output_target_scores

dthulke:
Somehow forgot the softmax in #460 

The output may not have a time dimension and seq_lens and thus _flatten_or_merge fails if output_flat is None.
This is already handled in init so we don't have to do this case distinction here again.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -7900,8 +7900,8 @@ class CrossEntropyLoss(Loss):
     """
     output_flat = self.output_flat
     if output_flat is None:
-      output_flat = self._flatten_or_merge(
-        self.output.placeholder, self.output_seq_lens, time_major=self.output.is_time_major)
+      output_flat = tf.nn.softmax(self.output_before_softmax_flat)
+    assert output_flat is not None
     target_flat_exp = tf.stack(
       [tf.range(tf.shape(self.target_flat)[0], dtype=tf.int32),
        tf.cast(self.target_flat, tf.int32)], axis=1)  # (time,2)


returnn/tf/layers/basic.py
```diff
@@ -7900,8 +7900,8 @@ def get_output_target_scores(self):
     """
     output_flat = self.output_flat
     if output_flat is None:
-      output_flat = self._flatten_or_merge(
-        self.output.placeholder, self.output_seq_lens, time_major=self.output.is_time_major)
+      output_flat = tf.nn.softmax(self.output_before_softmax_flat)
+    assert output_flat is not None
```
albertz, line 7904: Can you comment on this? Why? Is this faster?

albertz: Ah, you wrote in your PR message:

> The output may not have a time dimension and seq_lens and thus _flatten_or_merge fails if output_flat is None.
> This is already handled in init so we don't have to do this case distinction here again.

This is not sth you should write in the PR message, but instead should be added as a comment directly to the code. No-one will ever re-read the PR message.

But anyway, efficiency is always the top priority for us. The code is simpler but potentially less efficient is not a good argument (e.g. to avoid the case distinction here again). Maybe we should just rewrite and cleanup the code then. Or it just has to be complicated when needed.

So, most importantly, we should check performance. I'm not really sure which variant is more efficient. Most importantly also not just for your case (without time dim) but also for the more standard case (with time dim). Maybe your new variant is actually more efficient, then it's all good. Also that should be noted in a comment in the code (when it is not obvious).

dthulke: Good point, my main motivation was to make the code simpler / cleaner and to make it consistent with how this is handled in the MSE loss (https://github.com/rwth-i6/returnn/blob/109895d0c3bcb4d2c66548fb12e66988223c587d/returnn/tf/layers/basic.py#L8692).
But if I don't miss anything this should actually be slightly (may be negligible) more efficient since the softmax is applied to the already flattened output. On the other hand, it could be slower if the output of the layer is used somewhere else and the softmax is computed twice (once here and once as the activation of the layer).

Do you have any tips / recommendations on actually profiling things like this?

albertz: > this should actually be slightly (may be negligible) more efficient since the softmax is applied to the already flattened output.

Yes, probably.

> it could be slower if the output of the layer is used somewhere else and the softmax is computed twice

Yea, I also thought about this, but I assume this is more rare. So we can maybe ignore this case for now. The common cases are most important.

> Do you have any tips / recommendations on actually profiling things like this?

You could run some very simple dummy setup. See e.g. `demo-tf-lstm-benchmark.py`.

But maybe you are right. I currently don't see how this could be slower for the common cases.

patrick-wilken:
Looks good. This is the natural fix and I would think it does not lead to issues regarding speed.

However, can you say how you even got into that case? Because you can get to that line either when using the focal loss or when this is false: `if self.use_fused and self.output_before_softmax_flat is not None:`. But as far as I see, `self.output_before_softmax_flat is None` implies `self.output_flat is not None`, so did you set `use_fused` to False?

patrick-wilken (approved):

albertz:
@dthulke So are you still working on the benchmark? Or do you consider it good for now? You did not answer my last comment.

returnn/tf/layers/basic.py
dthulke: I did run a small benchmark. By adasting the `demo-tf-lstm-benchmark.py` script (i.e. only using a single lstm layer reducing the hidden size, but increased the output dim to 200k to make the influence of the softmax significant and used the DummyDataset).

As pointed out by @patrick-wilken we can only get into this case if either use_fused is set to False or focal loss is used. Thus, I benchmarked both cases:

With `use_fused=False`:
```
Current   New
   10.9   09.8
```
With `focal_loss_factor=1.0`:
```
Current   New
   13.5   11.3
```

Surprisingly, the new variant is faster in both cases (maybe the additional _flatten_or_merge causes a significant slow down?).

dthulke:
> However, can you say how you even got into that case?

I was doing an experiment with focal loss when encountering the issue.

albertz:
Ok, I will merge then.

Issue #466, opened 2021-03-08
Direct support for flattened/packed arrays


albertz:
# What is that, and why would we want it

Flattened/packed/unpadded arrays/tensors are tensors consisting of multiple sequences of different length, but without padding. I.e. instead of a tensor [B,T,D1...Dn], you would have a tensor [B',D1...Dn] where B' = sum(seq_len). They represent the same data (you are supposed to ignore the padded values anyway). This can also be generalized for multiple dynamic axes [B,T1,T2,...,D1...Dn] and seq_lens1, seq_lens2, ..

We already make use of packed arrays in RETURNN in some places internally. See usages of `flatten_with_seq_len_mask`.

A packed array can **save memory**, and potentially **speed up** some computations. While other computations might be slower though. And in any case, you often need special implemented ops.

For purely feed-forward operations (basically all layers in RETURNN with `recurrent=False` flag, e.g. `LinearLayer`) it should not matter at all and just be faster. Recurrent operations, convolutional operations, softmax-over-spatial, and many other things would need special implementations.

(E.g. CuDNN LSTM early only supported to work on packed arrays. And also the sequences must be sorted by sequence length, longest first. But then they extended that to support also padded arrays, and now they even deprecated the packed format. Not sure why. Maybe it was just too unconvenient to use, because padded arrays are really the standard. The packed arrays cannot even be used anymore in TensorFlow. In PyTorch on the other side, packed arrays are still the standard for RNNs.)

I specifically want to use this for a more memory-friendly RNN-T loss implementation. I.e. there it would have two dynamic length axis.

# Proposal

RETURNN should more directly support packed arrays, i.e. via `Data`. Layers could potentially return such data, and get such data as input.

# Details (to be clarified)

## Implementation/API in `Data`

How exactly should we represent it in `Data`?

`placeholder` as `tf.Tensor` is clear, that will be the packed array.

`shape`, `batch_shape` etc directly describe `placeholder`, i.e. there is no time-axis anymore but the (batch) shape is [B',D1...Dn].

I think `size_placeholder` should be consistent to `shape`, so that would not contain the sequence length. We should store them separately somehow/somewhere.

We might want to add a new attribute `batch_dim`, which references a `DimensionTag`, and extend the `DimensionTag` for the batch dim. Or maybe use a separate or new object `BatchDim`. That could also store the beam search information, so that we can easily differentiate between different kinds of batch dims. That object can store the sequence lengths.

[...]

PR #467, opened 2021-03-08
Direct support for flattened/packed arrays

albertz:
This partly implements #466.


albertz:
These (and related) changes will probably get big. So maybe we should split this PR up?

So far this contains just the basic extensions for `Data`, as proposed. It's somewhat incomplete, but it's done in a way not to break existing code/configs. This is probably not so debatable anyway. So maybe we can merge this already, and then improve further with other separate PRs.

albertz:
(It's enough if at least one of you reviews. This will only be the base for future changes anyway.)

albertz:
As no-one has commented yet, and I already restructured some parts of `BatchInfo`, I directly pushed this to this PR.

Zettelkasten (commented):
This looks good I think! I did not spot anything that seems off.

PR #470, opened 2021-03-14
Add init_missing_vars

Zettelkasten:
If this is set to true, the variables not found in the current checkpoint will be ignored during loading and then (randomly) initialized by TF.

In my case I trained a model, and then wanted to do search with another network dict which included a random initialized, non-trainable variable.

### DIFF
--- a/docs/configuration_reference/model_loading.rst
+++ b/docs/configuration_reference/model_loading.rst
@@ -46,3 +46,8 @@ preload_from_files
           }
         }
 
+load_ignore_missing_vars
+    If enabled, it will ignore missing variables when loading a checkpoint.
+    Otherwise it will error on missing variables.
+    Non-loaded variables are using the standard variable initialization (e.g. random init).
+    By default, this is disabled.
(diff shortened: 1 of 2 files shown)


albertz:
I don't quite understand. This should already be the case, or not?

It's just that when loading a checkpoint, and when we expect all vars to exists, then it will throw an exception.

I would rename the option to also reflect what we have in the code, which makes more sense. And also prefix to reflect that this is about loading the variables. I.e. sth like `load_ignore_missing_vars` or so.

docs/configuration_reference/model_loading.rst
```diff
@@ -46,3 +46,6 @@ preload_from_files
           }
         }
 
+init_missing_vars
+    If set to ``True``, any variables not found in the checkpoint will be randomly initialized.
```
albertz, line 50: This is misleading. This is always the case.

It's just that we will throw an error when we do not found expected variables.

albertz (changes requested):
See comments. The option name should be changed, and the documentation as well.

albertz (approved):
I think I still would have formulated it a bit different in the doc, but ok anyway.

Zettelkasten:
> I think I still would have formulated it a bit different in the doc, but ok anyway.

I can of course still change it. How would you write it?

albertz:
Maybe sth like:

> If enabled, it will ignore missing variables when loading a checkpoint. Otherwise it will error on missing variables.

albertz (approved):

PR #476, opened 2021-03-20
RangeInAxisLayer: Small fixes

Zettelkasten:
Fixes output dim if no feature dim.
Fixes that placeholder shape is consistent with output shape if axis is static.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2021,7 +2021,8 @@ class RangeInAxisLayer(LayerBase):
     source = self.sources[0].output
     axis = source.get_axis_from_description(axis)
     axis_wo_b = source.get_batch_axis_excluding_batch(axis)
-    source_shape = tf.shape(source.placeholder)
+    from returnn.tf.util.basic import get_shape
+    source_shape = get_shape(source.placeholder)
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -2059,6 +2060,8 @@ def get_out_data_from_opts(cls, name, sources, axis, dtype="int32", keepdims=Tru
         out.feature_dim_axis = None
       out.size_placeholder = {0: size for (i, size) in source.size_placeholder.items() if i == axis_wo_b}
     out.sparse = sparse
+    if out.feature_dim_axis is None:
+      out.dim = None
```
albertz, line 2064: This is wrong. `dim` specifies the number of classes in case it is sparse (i.e. when it does not have a feature dim).
```suggestion
    if out.feature_dim_axis is None and not out.sparse:
      out.dim = None
```

albertz (changes requested):
See comment.

returnn/tf/layers/basic.py
Zettelkasten: Oops, thanks! I added your suggestion.

albertz (approved):

PR #479, opened 2021-03-22
Adds support for None in py-to-pickle.cpp

dthulke:
Currently, py-to-pickle crashes results in a parse error when the py object contains None.

I needed support for this since some of my oggzip files contained None s (speaker information).


albertz (approved):

albertz:
Nice, thanks.

PR #481, opened 2021-03-26
Error for --dataset eval & dev

jotix16:
`eval` and `dev` are set to `None` at the end of `init(config_str, config_dataset, verbosity)` so using `config:dev` or `config:eval` caused an error. 

Using `config.typed_value("dev",None)` instead, works fine.

### DIFF
--- a/tools/dump-dataset.py
+++ b/tools/dump-dataset.py
@@ -222,7 +222,7 @@ def init(config_str, config_dataset, verbosity):
     config.set("train", dataset_dict)
   elif config_dataset and config_dataset != "train":
     print("Use dataset %r from config." % config_dataset)
-    config.set("train", "config:%s" % config_dataset)
+    config.set("train", config.typed_value(config_dataset, None))
   else:
     print("Use train dataset from config.")
     assert config.value("train", None)
--- a/tools/dump-dataset.py
+++ b/tools/dump-dataset.py
@@ -222,7 +222,7 @@ def init(config_str, config_dataset, verbosity):
     config.set("train", dataset_dict)
   elif config_dataset and config_dataset != "train":
     print("Use dataset %r from config." % config_dataset)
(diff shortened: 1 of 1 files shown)


albertz:
I don't understand. When you don't define a `dev` set in your config, and then you call `dump-dataset config --dataset dev`, this is the obvious right behavior, that it should produce an error, or not?

tools/dump-dataset.py
```diff
@@ -222,7 +222,7 @@ def init(config_str, config_dataset, verbosity):
     config.set("train", dataset_dict)
   elif config_dataset and config_dataset != "train":
     print("Use dataset %r from config." % config_dataset)
-    config.set("train", "config:%s" % config_dataset)
+    config.set("train", config.typed_value(config_dataset, None))
```
albertz, line 225: This is not the same.

We also allow to pass e.g. `--dataset 'get_dataset("train")'`, which we commonly use for configs where the dataset is given by such a `get_dataset` function. The given string is then `eval`d in the context of the config.

With your change, this breaks.

jotix16: I see. Having defined `dev` as a dict doesn't work, though. Maybe we should make sure we don't set `dev` to `None` if we specify `--dataset dev`.

albertz: > I see. Having defined `dev` as a dict doesn't work, though. Maybe we should make sure we don't set `dev` to `None` if we specify `--dataset dev`.

I would avoid heuristics. There could be other ways how this fails then. But my other suggestion (replacing `rnn.init_data`) is anyway cleaner and simpler and should not have such problems. See `dump-forward-stats.py` as an example.

albertz:
Ok, I already merged this now. Feel free to improve further with my suggestion, or if not so important, leave it as it is.

PR #490, opened 2021-04-26
Data: add dim tag format stag-single:<idx>:<name>

vieting:
I added the format `stag-single:<idx>:<name>` for dim tags to be able to distinguish spatial axes in pytorch-to-returnn. `idx` is relative to the matching stags, i.e., it is the index among the list of spatial dims matching `name`.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -2574,9 +2574,8 @@ class Data(object):
       elif axes.startswith("stag-single:"):  # spatial tag which possibly matches multiple spatial axes
         # in this case, a name of form "stag-single:<idx>:<name> is expected.
         # idx is relative to the matching stags, i.e., it is the index among the list of spatial dims matching the name
-        name_split = axes.split(":")
-        idx = int(name_split[1])
-        name = ":".join(name_split[2:])
+        idx_s, name = axes.split(":", 1)
+        idx = int(idx_s)
         axes = self.get_axes_by_tag_name(name, spatial_only=True)[idx]
       else:
         raise Exception("invalid axis mode %r" % axes)
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -2989,10 +2989,9 @@ class Data(object):
       return "%s/%s" % (scope_name, self.name)
     return self.name
 
-  def get_dim_tag(self, axis, unique_spatial_dims=False):
+  def get_dim_tag(self, axis):
     """
     :param int axis: counted with batch-dim
-    :param bool unique_spatial_dims: return dim tags that are unique for spatial dims which would otherwise be identical
     :rtype: DimensionTag
     """
     name = self.get_full_name()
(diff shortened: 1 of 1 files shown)


vieting:
fixed issue with python 2 compatibility

vieting:
Find related PR in pytorch-to-return here https://github.com/rwth-i6/pytorch-to-returnn/pull/36

returnn/tf/util/data.py
```diff
@@ -2571,6 +2571,13 @@ def get_axes_from_description(self, axes, allow_int=True):
         return self.get_axes_from_description(list(axes))
       elif axes.startswith("stag:"):  # spatial tag
         axes = self.get_axis_by_tag_name(axes[len("stag:"):], spatial_only=True)
+      elif axes.startswith("stag-single:"):  # spatial tag which possibly matches multiple spatial axes
+        # in this case, a name of form "stag-single:<idx>:<name> is expected.
+        # idx is relative to the matching stags, i.e., it is the index among the list of spatial dims matching the name
+        name_split = axes.split(":")
+        idx = int(name_split[1])
+        name = ":".join(name_split[2:])
```
albertz, line 2579:
```suggestion
        idx_s, name = axes.split(":", 1)
        idx = int(idx_s)
```

returnn/tf/util/data.py
```diff
@@ -2973,9 +2990,10 @@ def get_full_name(self):
       return "%s/%s" % (scope_name, self.name)
     return self.name
 
-  def get_dim_tag(self, axis):
+  def get_dim_tag(self, axis, unique_spatial_dims=False):
```
albertz, line 2993: I don't really understand this change. What do you need this for? This behavior here in this function seems counter-intuitive to me. This function should just do what it says, i.e. return the dim tag for an axis.

vieting: In pytorch-to-returnn, I use it to obtain a dim tag where multiple spatial axes get dim tags with different names so that they can be kept apart. Would you prefer to keep this only on pytorch-to-returnn side?

albertz: Such logic should be implemented in pytorch-to-returnn but not here.

Also, it's actually wrong. It must be the same `DimensionTag` object. That's the idea of dimension tags. If the dimensions are the same, the dimension tags are the same.

The only way to different dimensions then is by `stag-single`.

albertz:
Btw, `stag-single` is the first prefix I came up with. But I'm not sure if this is really good. Do you maybe have better ideas? Or you think it's ok?

returnn/tf/util/data.py
vieting: I did this on pytorch-to-returnn side now and removed this part here.

vieting:
I think it's okay. Maybe something like `stag-unique` or `stag-index` otherwise? I added the comments to make clear what this is intended for.

returnn/tf/util/data.py
```diff
@@ -2571,6 +2571,12 @@ def get_axes_from_description(self, axes, allow_int=True):
         return self.get_axes_from_description(list(axes))
       elif axes.startswith("stag:"):  # spatial tag
         axes = self.get_axis_by_tag_name(axes[len("stag:"):], spatial_only=True)
+      elif axes.startswith("stag-single:"):  # spatial tag which possibly matches multiple spatial axes
+        # in this case, a name of form "stag-single:<idx>:<name> is expected.
+        # idx is relative to the matching stags, i.e., it is the index among the list of spatial dims matching the name
+        idx_s, name = axes.split(":", 1)
```
albertz, line 2577: I just notice, this is wrong.
```suggestion
        _, idx_s, name = axes.split(":", 2)
```

PR #492, opened 2021-04-27
extend DummyDatasetMultipleDataKeys

jotix16:
Added `get_all_tags`, `init_seq_order` and `get_current_seq_order` for **DummyDatasetMultipleSequenceLength**


returnn/datasets/generating.py
```diff
@@ -756,6 +757,34 @@ def generate_seq(self, seq_idx):
                            for i in range(i1, i2)])
     return DatasetSeq(seq_idx=seq_idx, features=features, targets=targets)
 
+  def get_all_tags(self):
+    """
+    :rtype: list[str]
+    """
+    return ["seq-%i" % i for i in range(self._num_seqs)]
```
albertz, line 764: `self._num_seqs` might be `inf`. It should handle that correctly (fall back to the default behavior).

returnn/datasets/generating.py
```diff
@@ -756,6 +757,34 @@ def generate_seq(self, seq_idx):
...
+    :param list[int]|None seq_order: List of corpus sequence indices, to set a predefined order.
+    :rtype: bool
+    :returns whether the order changed (True is always safe to return)
+    """
+    super(DummyDatasetMultipleSequenceLength, self).init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
+
+    def get_seq_len(i):
+      return self.seq_len['data']
+
+    self._seq_order = self.get_seq_order_for_epoch(epoch, self._num_seqs, get_seq_len=get_seq_len)
```
albertz, line 782: I'm not sure this makes sense to support for this dataset.
Also you are not using it, i.e. ignoring it (but maybe just because it's incomplete so far).

albertz:
Why is this specific for `DummyDatasetMultipleDataKeys`? Why not for the base class `DummyDataset`? Or even better its base class `GeneratingDataset`.

albertz:
Please also see failing tests.
```
  returnn/datasets/generating.py:779: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/datasets/generating.py:779: WEAK WARNING PyUnusedLocalInspection: Parameter 'i' value is not used
  returnn/datasets/generating.py:785: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
```

returnn/datasets/generating.py
jotix16: The  `self._seq_order` you mean?

returnn/datasets/generating.py
jotix16: What is the default behaviour? 

Does it even make sense for a dummy dataset to be initialized with `num_seqs=inf`?  
In `GeneratingDataset` it is per default `inf` but here it has to be set by the user.

jotix16:
> Why is this specific for `DummyDatasetMultipleDataKeys`? Why not for the base class `DummyDataset`? Or even better its base class `GeneratingDataset`.

I don't see any reason why not to use it for `GeneratingDataset` directly.

returnn/datasets/generating.py
albertz: > What is the default behaviour?

Whatever `super` of this function does. Just check the code.

> Does it even make sense for a dummy dataset to be initialized with `num_seqs=inf`?

Yes it does. Why not? It's usually only used for testing. In any case, this is explicitly supported.

PR #494, opened 2021-04-28
ChoiceLayer: added alpha for length normalization

patrick-wilken:
Normalize by length^alpha instead of length to be able to tune the length of the generated sequences. I thought this is quite common in practice (including alpha as the variable name), however I could only find a reference to a more complicated formula here: https://arxiv.org/pdf/1609.08144.pdf equation 14.
I needed this to tune the length of translations of partial sentences ("chunks") that we use for simultaneous MT.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -4363,6 +4363,7 @@ class ChoiceLayer(BaseChoiceLayer):
                input_type="prob",
                prob_scale=1.0, base_beam_score_scale=1.0, random_sample_scale=0.0,
                length_normalization=True,
+               length_normalization_exponent=1.0,
                custom_score_combine=None,
                source_beam_sizes=None, scheduled_sampling=False, cheating=False,
                explicit_search_sources=None,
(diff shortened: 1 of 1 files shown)


albertz:
So does this help?

Patrick Wilken ***@***.***> schrieb am Mi., 28. Apr. 2021,
13:59:

patrick-wilken:
Yes :D My model tends to overtranslate small chunks (speaking 1-3 words) and by using an alpha of e.g. 0.6 the BLEU score increased significantly.

patrick-wilken:
Ah, fairseq has the `--lenpen` option which does the same thing, found this in the code: `score /= (target_len) ** float(lenpen)`. OpenNMT has it too but uses the more complicated "Google Formula".

returnn/tf/layers/rec.py
```diff
@@ -4363,6 +4363,7 @@ def __init__(self, beam_size, keep_beams=False,
                input_type="prob",
                prob_scale=1.0, base_beam_score_scale=1.0, random_sample_scale=0.0,
                length_normalization=True,
+               length_normalization_alpha=1.0,
```
albertz, line 4366: Is this a good name? Maybe better `length_normalization_exp`, or `..._exponent`? I don't really like names like "alpha", "beta", "x", "y", "a", "b", which are all really arbitrary.

albertz (approved):

PR #495, opened 2021-04-28
reuse_name_scope_of_tensor: fix invalid name if no scope

patrick-wilken:
If `get_name_scope_of_tensor(x)` returns empty string, the resulting `name_scope` was invalid because of the leading `/`. This happened to me when using `compile_tf_graph.py`. I can check again for which tensor this happened if important.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -466,10 +466,16 @@ def reuse_name_scope_of_tensor(x, prefix="", postfix="", add_tensor_name=False):
   :param bool add_tensor_name:
   :return: reuse the name scope of x, e.g. "layer0/rec", yields scope
   """
+  name_scope = get_name_scope_of_tensor(x)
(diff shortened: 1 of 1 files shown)


albertz:
> If `get_name_scope_of_tensor(x)` returns empty string, the resulting `name_scope` was invalid because of the leading `/`. This happened to me when using `compile_tf_graph.py`. I can check again for which tensor this happened if important.

Instead of explaining this here in the PR, this should rather be explained in the code as a comment.

But I also wonder whether we can make the code more clear to reflect this case. E.g. some very explicit `if get_name_scope_of_tensor(x) == "": ...`.

patrick-wilken:
Like this? I think there was also the bug before that the postfix was added twice if add_tensor_name was set. But it should not be critical to change scope names, right?

albertz (approved):

PR #496, opened 2021-04-28
DataNotAvailableLayer fixes

patrick-wilken:
1. Fixed one place where trying to access the size placeholder fails in `get_out_data_from_opts()` when the input is a `DataNotAvailableLayer`. I suspect there might be more such cases in the code.

2. `DataNotAvailableLayer` has to do "register_as_extern_data" because the data template might be needed at inference time, e.g. as the target of a ChoiceLayer. Usually this happens in `LayerBase.post_init()`, however this is not called for `DataNotAvailableLayer`.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -191,20 +191,21 @@ class LayerBase(object):
     :param dict[str] layer_desc: kwargs as they are passed to self.__init__
     """
     self.kwargs = layer_desc
-    if self.use_batch_norm:
-      opts = {}
-      if isinstance(self.use_batch_norm, dict):
-        opts = self.use_batch_norm
-      self.output.placeholder = self.batch_norm(self.output, **opts)
-    if self.control_dependencies_on_output:
-      control_deps = self.control_dependencies_on_output(self)
-      if not isinstance(control_deps, (list, tuple)):
-        assert isinstance(control_deps, (tf.Operation, tf.Tensor))
-        control_deps = [control_deps]
-      assert all([isinstance(dep, (tf.Operation, tf.Tensor)) for dep in control_deps])
-      if control_deps:
-        with tf.control_dependencies(control_deps):
-          self.output.placeholder = tf.identity(self.output.placeholder)
+    if self.output.placeholder is not None:  # unset e.g. in DataNotAvailableLayer
+      if self.use_batch_norm:
+        opts = {}
+        if isinstance(self.use_batch_norm, dict):
+          opts = self.use_batch_norm
+        self.output.placeholder = self.batch_norm(self.output, **opts)
+      if self.control_dependencies_on_output:
+        control_deps = self.control_dependencies_on_output(self)
+        if not isinstance(control_deps, (list, tuple)):
+          assert isinstance(control_deps, (tf.Operation, tf.Tensor))
+          control_deps = [control_deps]
+        assert all([isinstance(dep, (tf.Operation, tf.Tensor)) for dep in control_deps])
+        if control_deps:
+          with tf.control_dependencies(control_deps):
+            self.output.placeholder = tf.identity(self.output.placeholder)
     if self.register_as_extern_data:
       self.network.extern_data.extra_added_keys.add(self.register_as_extern_data)
       self.network.extern_data.data[self.register_as_extern_data] = self.output
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3686,7 +3686,8 @@ class ReinterpretDataLayer(_ConcatInputLayer):
         setattr(out, s, i)
         if s == "feature_dim_axis":
           out.dim = out.batch_shape[out.feature_dim_axis]
-    if size_base:
+    if out.size_placeholder is not None and size_base:  # size_placeholder might be None, e.g. via DataNotAvailableLayer
+      assert size_base.output.size_placeholder is not None
       assert len(out.size_placeholder) == len(size_base.output.size_placeholder)
       # Keep same indices. Assumes same order of spatial dims.
       out.size_placeholder = {
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -951,10 +951,12 @@ class TFNetwork(object):
         output_template_special_axes = output_template.get_special_axes_dict()
         if not output_template.available_for_inference and not self.eval_flag:
           from returnn.tf.layers.base import DataNotAvailableLayer
-          return DataNotAvailableLayer(
+          layer = DataNotAvailableLayer(
             name=layer_desc['name'], network=layer_desc['network'], output=output_template,
-            layer_class=layer_class, layer_desc=layer_desc)
-        layer = layer_class(**layer_desc)
+            layer_class=layer_class, layer_desc=layer_desc,
+            register_as_extern_data=layer_desc.get('register_as_extern_data'))
+        else:
+          layer = layer_class(**layer_desc)
         layer.post_init(layer_desc)
         layer.output.sanity_check()
         # The axes should not have moved now.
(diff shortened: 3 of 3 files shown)


returnn/tf/layers/base.py
```diff
@@ -1478,6 +1478,11 @@ def __init__(self, layer_class, layer_desc, **kwargs):
     self.layer_class_ = layer_class
     self.layer_desc = layer_desc
 
+    register_as_extern_data = layer_desc.get("register_as_extern_data")
```
JackTemaki, line 1481: maybe add a comment here then (or in the class docstring) that it needs to manage the `register_as_extern_data`

JackTemaki (approved):
This actually solves a problem in one of the setups I avoided by explicitly giving `n_out` for output layers relying on non existent data, so thanks for finding that.

returnn/tf/layers/basic.py
```diff
@@ -3686,7 +3686,8 @@ def map_axis_name(s):
         setattr(out, s, i)
         if s == "feature_dim_axis":
           out.dim = out.batch_shape[out.feature_dim_axis]
-    if size_base:
+    if out.size_placeholder is not None and size_base:  # size_placeholder is not set if source is DataNotAvailableLayer
```
albertz, line 3689: I don't quite understand this comment. Why is this only for DataNotAvailableLayer, and maybe not in other cases? And why is that way at all? Shouldn't we maybe fix this in DataNotAvailableLayer? This sounds like you do some workaround here.

albertz:
Do not explain things here in a comment in the PR, but directly in the code.

returnn/tf/layers/base.py
```diff
@@ -1478,6 +1478,11 @@ def __init__(self, layer_class, layer_desc, **kwargs):
...
+    if register_as_extern_data:
+      self.network.extern_data.extra_added_keys.add(register_as_extern_data)
+      self.network.extern_data.data[register_as_extern_data] = self.output
```
albertz, line 1484: I don't understand. Why is this needed here?

albertz: > Usually this happens in `LayerBase.post_init()`, however this is not called for DataNotAvailableLayer.

Why? And shouldn't we fix that? This again sounds like you do a workaround here.

returnn/tf/layers/basic.py
JackTemaki: I think the `size_placeholder` being `None` is a valid case. This can not be fixed in the DataNotAvailableLayer, because the (correct) content for the size_placeholder does not exist, and the code would fail in any way. Introducing an arbitrary filled size_placeholder does not sound like a valid solution.

albertz: But then this comment is wrong (misleading), as this is not about DataNotAvailableLayer.

albertz: Also, as I always say, we should never just accept it as it is. We should always ask: Does this makes sense? Should it really be this way? Does it make sense that `size_placeholder` can be `None`?

returnn/tf/layers/base.py
JackTemaki: @patrick-wilken Is there any unwanted side effect in calling `post_init()` on the `DataNotAvailableLayer`? I am not sure what happens if we call `self.batch_norm` on an output which has no defined placeholder. if there are no side effects, we could call `post_init` before returning the `DataNotAvailableLayer`, if there are, this seems to be the correct solution.

returnn/tf/layers/basic.py
JackTemaki: But it does make sense. What should the `size_placeholder` be if the data does not exist?  I think `None` is valid here. Of course you could add a `size_placeholder` with fake entries, but I do not think this is a good idea. Any variable should have a reasonable content, or not?

The comment is also correct, because as long as we are not using `DataNotAvailableLayer`, the size_placeholder should be correctly defined.

returnn/tf/layers/base.py
albertz: If `LayerBase.post_init()` does not work on `DataNotAvailableLayer`, then why do another workaround at a different place? Why not just define `DataNotAvailableLayer.post_init()`? Or fix `LayerBase.post_init()`?

returnn/tf/layers/basic.py
albertz: But there might be other layers as well. It would be wrong to say this is only for one specific layer. The `LayerBase` class is the base and should never have specific code for specific sub instances.

Either we say `size_placeholder` can be `None` in general, or not. But saying it can only be for one specific layer is very hacky and will surely lead to other problems.

What you could do (much better):
```suggestion
    if out.size_placeholder is not None and size_base:  # size_placeholder could be unset eg. with DataNotAvailableLayer
```

patrick-wilken:
Ok, thanks, this should be cleaner now.

returnn/tf/layers/base.py
patrick-wilken: `register_as_extern_data` is already mentioned in the doc string of `DataNotAvailableLayer`, check if a comment is still necessary in the current code.

returnn/tf/network.py
```diff
@@ -949,14 +949,18 @@ def _create_layer(self, name, layer_class, **layer_desc):
...
             name=layer_desc['name'], network=layer_desc['network'], output=output_template,
-            layer_class=layer_class, layer_desc=layer_desc)
-        layer = layer_class(**layer_desc)
+            layer_class=layer_class, layer_desc=layer_desc,
+            register_as_extern_data=layer_desc.get('register_as_extern_data'))
+        else:
+          layer = layer_class(**layer_desc)
         layer.post_init(layer_desc)
-        layer.output.sanity_check()
+        layer.output.sanity_check(ignore_placeholder=not placeholder_is_available)
```
albertz, line 963: Why is this needed? If it is not set, `sanity_check` will also work.

returnn/tf/network.py
```diff
@@ -949,14 +949,18 @@ def _create_layer(self, name, layer_class, **layer_desc):
           print("layer %s/%r output: %r" % (self.name, name, output_template))
         output_template.sanity_check(ignore_placeholder=True)  # placeholder might be overwritten later
         output_template_special_axes = output_template.get_special_axes_dict()
+        placeholder_is_available = True
```
albertz, line 952: Why do we need this variable? Can't we just check for `layer.output.placeholder is not None` below?

JackTemaki:
Isn't this now way more complicated than before? The change in `post_init` is good, but I would still return the `DataNotAvailableLayer` directly where it is defined, just applying post_init at that point.

returnn/tf/network.py
patrick-wilken: Oh, didn't notice.

returnn/tf/network.py
patrick-wilken: Fixed now.

returnn/tf/network.py
```diff
@@ -972,22 +974,23 @@ def _create_layer(self, name, layer_class, **layer_desc):
...
         from returnn.tf.util.basic import identity_with_check_numerics
         layer.output.placeholder = identity_with_check_numerics(
           layer.output.placeholder,
           name="%s_identity_with_check_numerics_output" % layer_class.cls_get_tf_scope_name(name))
     assert layer.output
-    assert layer.output.placeholder is not None
-    layer.output.placeholder.set_shape(layer.output.batch_shape)
-    assert layer.output.size_placeholder is not None
+    if layer.output.placeholder is not None:
+      assert layer.output.placeholder is not None
```
patrick-wilken, line 991: Oops, let me remove this line :D

albertz (approved):
Ok looks fine now. Let's wait for the tests.

PR #498, opened 2021-04-29
HDFDataset profiling

patrick-wilken:
I finally followed up on #434 and did some profiling of HDFDataset. I had observed that when training small networks it can be the bottleneck.

I wrote this code to go through a dummy dataset with 200000 sequences, all one data stream with content `[0]`:

```
    hdf_dataset = HDFDataset(files=[file_name], seq_ordering="default", cache_whole_file=False)
    hdf_dataset.init_seq_order(epoch=1)

    start_time = time.time()
    seq_idx = 0
    while hdf_dataset.is_less_than_num_seqs(seq_idx):
        hdf_dataset.load_seqs(seq_idx, seq_idx + 1)
        data = hdf_dataset.get_data(seq_idx, data_key)
        assert data is not None
        seq_idx += 1

    time_needed = time.time() - start_time
    print(f"Needed {time_needed:.1f} seconds to go through the RETURNN HDFDataset once.")
```

Runs in 63.5s. The file was on a local ssd, so file access should have been fast.

When I set `cache_whole_file=True` (see commit), it runs in 1.9s (~most of which is initialisation, I guess~).

For comparison, when I run
```
fin = h5py.File(file_name, "r")

start_time = time.time()
sequence_lengths = list(fin["seqLengths"][...])
sequence_tags = list(fin["seqTags"][...])
data = list(fin[f"/targets/data/{data_key}"][...])
size = list(fin["/targets/size"].attrs[data_key])

time_needed = time.time() - start_time
print(f"Needed {time_needed:.1f} seconds to load the file in one go.")
```
it takes 0.1s.

In the first case this is the beginning of the output of cProfile:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   200001   15.695    0.000   41.155    0.000 dataset.py:476(__getitem__)
   200011    8.447    0.000   16.636    0.000 group.py:255(__getitem__)
   800004    6.496    0.000    6.637    0.000 dataset.py:282(shape)
   200002    5.857    0.000    6.987    0.000 dataset.py:395(__init__)
        1    4.046    4.046   67.433   67.433 test_reading_speed.py:9(test_returnn_reading_speed)
   200002    3.302    0.000    3.302    0.000 {method 'reduce' of 'numpy.ufunc' objects}
   200000    3.171    0.000   61.370    0.000 hdf.py:262(get_data)
```
 So almost all time (61s) is spent within `HDFDataset.get_data()` where the hdf file access happens.
 
 I also tested `cache_whole_file` in a real-life experiment: sequence labelling with a 3-layer LSTM, 512 units on 150k sequences per epoch (I reduced the epoch size for profiling). Here, enabling `cache_whole_file` improved the run time of one epoch from 5.5 minutes to 2.5 minutes.
 
Obviously caching the whole file is only possible if it isn't too big. So as touched upon in #434, I wonder whether implementing an improved version of the caching already implemented in HDFDataset (which also caches targets...) would be worth it...

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -28,7 +28,7 @@ class HDFDataset(CachedDataset):
   This was the main original dataset format of RETURNN.
   """
 
-  def __init__(self, files=None, use_cache_manager=False, cache_whole_file=False, **kwargs):
+  def __init__(self, files=None, use_cache_manager=False, cache_whole_file=False, cache_hdf_file_handles=False, **kwargs):
     """
     :param None|list[str] files:
     :param bool use_cache_manager: uses :func:`Util.cf` for files
@@ -44,7 +44,9 @@ class HDFDataset(CachedDataset):
     self.data_dtype = {}  # type: typing.Dict[str,str]
     self.data_sparse = {}  # type: typing.Dict[str,bool]
     self._cache_whole_file = cache_whole_file
+    self._cache_hdf_file_handles = cache_hdf_file_handles
     self._cached_files = []
+    self._cached_file_handles = []
     if files:
       for fn in files:
         self.add_file(fn)
@@ -98,6 +100,7 @@ class HDFDataset(CachedDataset):
     self.files.append(filename)
     self.h5_files.append(fin)
     self._cached_files.append({})
+    self._cached_file_handles.append({})
     print("parsing file", filename, file=log.v5)
     if 'times' in fin:
       if self.timestamps is None:
@@ -278,11 +281,14 @@ class HDFDataset(CachedDataset):
     end_pos = self.file_seq_start[file_idx][real_file_seq_idx + 1]
 
     if key == "data":
-      assert 'inputs' in fin, "'data' key is reserved for 'inputs' in the HDF file, but 'inputs' not found in HDF file."
       if self._cache_whole_file:
         if "inputs" not in self._cached_files[file_idx]:
           self._cached_files[file_idx]["inputs"] = fin['inputs'][:]
         inputs = self._cached_files[file_idx]["inputs"]
+      elif self._cache_hdf_file_handles:
+        if "inputs" not in self._cached_file_handles[file_idx]:
+          self._cached_file_handles[file_idx]["inputs"] = fin['inputs']
+        inputs = self._cached_file_handles[file_idx]["inputs"]
       else:
         inputs = fin['inputs']
       data = inputs[start_pos[0]:end_pos[0]]
@@ -293,6 +299,10 @@ class HDFDataset(CachedDataset):
         if key not in self._cached_files[file_idx]:
           self._cached_files[file_idx][key] = fin['targets/data/' + key][:]
         targets = self._cached_files[file_idx][key]
+      elif self._cache_hdf_file_handles:
+        if key not in self._cached_file_handles[file_idx]:
+          self._cached_file_handles[file_idx][key] = fin['targets/data/' + key]
+        targets = self._cached_file_handles[file_idx][key]
       else:
         targets = fin['targets/data/' + key]
       first_target_idx = 1 if self.num_inputs > 0 else 0  # self.num_inputs == 0 if no 'inputs' in HDF file
@@ -350,6 +360,10 @@ class HDFDataset(CachedDataset):
       if "#tags" not in self._cached_files[file_idx]:
         self._cached_files[file_idx]["#tags"] = list(self.h5_files[file_idx]["seqTags"][:])
       s = self._cached_files[file_idx]["#tags"][real_file_seq_idx]
+    elif self._cache_hdf_file_handles:
+      if "#tags" not in self._cached_file_handles[file_idx]:
+        self._cached_file_handles[file_idx]["#tags"] = self.h5_files[file_idx]["seqTags"]
+      s = self._cached_file_handles[file_idx]["#tags"][real_file_seq_idx]
     else:
       s = self.h5_files[file_idx]["seqTags"][real_file_seq_idx]
     s = self._decode(s)
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -28,7 +28,7 @@ class HDFDataset(CachedDataset):
   This was the main original dataset format of RETURNN.
   """
 
-  def __init__(self, files=None, use_cache_manager=False, **kwargs):
+  def __init__(self, files=None, use_cache_manager=False, cache_whole_file=False, **kwargs):
     """
     :param None|list[str] files:
     :param bool use_cache_manager: uses :func:`Util.cf` for files
@@ -43,6 +43,8 @@ class HDFDataset(CachedDataset):
     self.file_seq_start = []  # type: typing.List[numpy.ndarray]
     self.data_dtype = {}  # type: typing.Dict[str,str]
     self.data_sparse = {}  # type: typing.Dict[str,bool]
+    self._cache_whole_file = cache_whole_file
+    self._cached_files = []
     if files:
       for fn in files:
         self.add_file(fn)
@@ -95,6 +97,7 @@ class HDFDataset(CachedDataset):
         "expected " + str(len(self.labels['classes'])) + " got " + str(len(labels)))
     self.files.append(filename)
     self.h5_files.append(fin)
+    self._cached_files.append({})
     print("parsing file", filename, file=log.v5)
     if 'times' in fin:
       if self.timestamps is None:
@@ -276,13 +279,22 @@ class HDFDataset(CachedDataset):
 
     if key == "data":
       assert 'inputs' in fin, "'data' key is reserved for 'inputs' in the HDF file, but 'inputs' not found in HDF file."
-      inputs = fin['inputs']
+      if self._cache_whole_file:
+        if "inputs" not in self._cached_files[file_idx]:
+          self._cached_files[file_idx]["inputs"] = fin['inputs'][:]
+        inputs = self._cached_files[file_idx]["inputs"]
+      else:
+        inputs = fin['inputs']
       data = inputs[start_pos[0]:end_pos[0]]
       if self.window > 1:
         data = self._sliding_window(data)
     else:
-      assert 'targets' in fin
-      targets = fin['targets/data/' + key]
+      if self._cache_whole_file:
+        if key not in self._cached_files[file_idx]:
+          self._cached_files[file_idx][key] = fin['targets/data/' + key][:]
+        targets = self._cached_files[file_idx][key]
+      else:
+        targets = fin['targets/data/' + key]
       first_target_idx = 1 if self.num_inputs > 0 else 0  # self.num_inputs == 0 if no 'inputs' in HDF file
       ldx = first_target_idx + self.target_keys.index(key)
       data = targets[start_pos[ldx]:end_pos[ldx]]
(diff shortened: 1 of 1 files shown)


albertz:
I don't really understand your first comparison. In one case, you create 200k numpy arrays. In the other case, you do not. This doesn't make sense.

Also, this is somewhat artificial. More realistic is to have seqs of len 100 or so. For such a benchmark, create 2k seqs each of len 100, not 200k seqs each of len 1.

Also, sure, when you can load the whole dataset into memory, this will be faster. But this is not possible in the usual case.

Also, when your network is really really tiny, then yes, the dataset can always become the bottleneck. But also this is again not the usual case. And for HDFDataset, I expect this to be really extreme to get into effect (at least for realistic cases).

Note, when we use the new `tf.data` pipeline, we could also write a C++ op which directly reads from a HDF. (Should be very straight forward.) Then all this Python bottleneck and numpy array creation etc goes away.

About your new code: This should be the same as what we already have with the old caching logic (when the cache is configured large enough). I don't think it makes sense to have two separate caching logics implemented which do the same thing (just their implementation differs).

Also, HDF (H5Py) also has caching logic itself. Have you tried and compared to that? [Some doc](https://docs.h5py.org/en/stable/high/file.html#chunk-cache). E.g. use `rdcc_nbytes=1024**3` or so.

patrick-wilken:
> 2k seqs each of len 100

2K sequences is unrealistic for a training dataset of text. But I ran the first code snippet on the train and dev file of my real experiment. The dev  set has 1.4K sequences, the training set 1.2M sequences. Average sequence length is around 20.

```
       cache_whole_file   running time (s)
dev    False                0.44
dev    True                 0.012
train  False              324.8
train  True                 9.5
```

So again, in both cases the difference is around factor 30.

> Also, when your network is really really tiny, then yes, the dataset can always become the bottleneck.

As I said, 3 layer LSTM, 512 units. It may be small for todays standards, but we have several applications where we use networks of this size. Caching the whole file here makes a difference of more than factor 2. So it's not factor 30, which it apparently would be in the limit of making the network smaller and smaller, but nevertheless noticeable.

> In one case, you create 200k numpy arrays. In the other case, you do not.

In both cases I call `get_data()` for all sequences. So the data is split. That when accessing the file once per sequence more numpy arrays are created is part of the problem I want to point out.

> This should be the same as what we already have with the old caching logic.

Yes, I'm aware of this, that's why this is a draft 😉. But I can't use that caching at the moment because only inputs are cached. So my question is whether you would support rewriting/extending it.

I tried
```
fin = h5py.File(filename, "r", rdcc_nbytes=1024**3)
```
in HDFDataset, but it didn't improve the loading speed.

albertz:
> 2K sequences is unrealistic for a training dataset of text

But this is not relevant for such a benchmark, right?

> > Also, when your network is really really tiny, then yes, the dataset can always become the bottleneck.
>
> As I said, 3 layer LSTM, 512 units.

But I mean for a realistic dataset benchmark (not sequences of length 1). I don't expect that you see this effect then.

> > In one case, you create 200k numpy arrays. In the other case, you do not.

> In both cases I call get_data() for all sequences. So the data is split. That when accessing the file once per sequence more numpy arrays are created is part of the problem I want to point out.

`get_data` will always create Numpy arrays per sequence, no matter if you do caching or not. I just wanted to point out that your second code snippet is not comparable because it does not do this ("Needed ... seconds to load the file in one go.").

> I tried `rdcc_nbytes=1024**3` but it didn't improve the loading speed.

But then sth is wrong? This should do the same what you are doing. It should cache the whole thing in memory.

patrick-wilken:
No, sorry, everything I said about the LSTM was done with data of average sequence length 20.

albertz:
I still think we should understand this better. What exactly happens in `get_data` (or in h5py) that this is so slow? I would assume on the OS level, if the file is small, it should be in the OS filesystem cache in memory. On the h5py level with an option like `rdcc_nbytes=1024**3`, it should definitely also be in memory. So why should one variant be slower than the others?

Or maybe it is only slow the first time you read, and then not anymore (in further epochs) because it stays in the cache? Did you test this? Maybe there is also some way in h5py to preload the file into its cache. (Or you can just do manually by calling `list(fin[f"/targets/data/{data_key}"][...])` or so. (Of course, for the real code, needs to be done in a better way, this is only for testing.)

patrick-wilken:
I'm somewhat confident now this is not about OS caching at all. It simply seems that a statement like
```
fin['inputs'][start_pos[0]:end_pos[0]]
```
has a certain overhead in the h5py library which gets significant when you do it thousands of times.

First clue is the cProfile output:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  1295996   86.063    0.000  220.134    0.000 dataset.py:476(__getitem__)
  1296013   45.315    0.000   89.906    0.000 group.py:255(__getitem__)
  5183984   33.781    0.000   34.553    0.000 dataset.py:282(shape)
  1295998   31.461    0.000   37.964    0.000 dataset.py:395(__init__)
        1   21.519   21.519  361.310  361.310 test_reading_speed.py:9(test_returnn_reading_speed)
  1295998   18.528    0.000   18.528    0.000 {method 'reduce' of 'numpy.ufunc' objects}
  1295993   16.459    0.000  328.449    0.000 hdf.py:262(get_data)
  1295996    8.406    0.000   24.154    0.000 selections.py:250(__getitem__)
  1295996    7.580    0.000    7.734    0.000 selections.py:147(__init__)
  1296002    6.601    0.000    6.601    0.000 base.py:85(is_empty_dataspace)
  1295996    6.373    0.000   44.365    0.000 selections.py:27(select)
  1295996    5.624    0.000   14.468    0.000 selections.py:444(_handle_simple)
  1295998    5.497    0.000   26.931    0.000 fromnumeric.py:70(_wrapreduction)
  1295998    5.301    0.000    5.301    0.000 filters.py:246(get_filters)
  1296002    4.718    0.000    4.718    0.000 {built-in method h5py.h5t.py_create}
  1295996    4.039    0.000   12.492    0.000 selections.py:244(__init__)
 18189630    3.885    0.000    3.885    0.000 {built-in method builtins.isinstance}
  1295993    3.576    0.000    8.532    0.000 cached.py:173(load_seqs)
  1295996    3.151    0.000    5.315    0.000 selections.py:421(_expand_ellipsis)
  ```
  
 I don't want to spend much time getting into the h5py code. But I think all `__init__` functions do not access the file but "just" create internal data structures. `group.py:255(__getitem__)` also should not read data, or at least not much, it just accesses a group in the file. E.g. `base.py:85(is_empty_dataspace)` and `filters.py:246(get_filters)` also sound like pure overhead.
(code is here: https://github.com/h5py/h5py/tree/master/h5py/_hl)

Second clue: When I alter my "cache_whole_file" code such that instead of really reading all the data I just cache the h5py dataset, i.e. caching `fin['targets/data/' + key]` instead of `fin['targets/data/' + key][:]` I already get a speed-up of 50%! This means half of the speed difference is due to initializing the `fin['targets/data/' + key]` object, whatever happens there.
 
The overhead is also mentioned in this answer: https://github.com/h5py/h5py/issues/994#issuecomment-364767614
 
Also, by chance I noticed that that h5py version 2.10 is 25% faster than 3.10 for my test case...
 
I experimented with different "chunking" of the HDF file, no major differences, as maybe expected because the data is 1-dimensional (except for sequence lengths).

Second time loading the HDF file seems not to be faster than the first time.

patrick-wilken:
I implemented a version of HDFDataset that accesses the HDF file less often by getting data for multiple sequences at once using fancy indexing. Turns out I can't get it to run more efficient than the baseline in general. Because fancy indexing with h5py seems to be not that efficient in general - after all it's random access - and also the necessary logic of calculating indices etc. (h5py requires indices to be sorted...) has some overhead. So I'll be closing this soon...

Nevertheless, I found the speed-ups described above. I ran my real word task again (3-Layer LSTM, sparse in- and outputs of length around 20, hidden sizes 512, 690K sequences per epoch) and get these training times:

Baseline (current master):
`train epoch 1, finished after 15409 steps, 0:16:39 elapsed (28.5% computing time)`
Using `cache_whole_file`:
`train epoch 1, finished after 15409 steps, 0:07:06 elapsed (74.1% computing time)`
Using `cache_hdf_file_handles`:
`train epoch 1, finished after 15409 steps, 0:09:39 elapsed (47.9% computing time)`

At least some version of `cache_hdf_file_handles` should find its way into the master because it is a speed-up for free. `cache_whole_file` was mainly for debugging but it is actually useful sometimes...

@albertz, you did not want to have another caching logic on top of what we have. Makes sense. But CachedDataset only caches inputs and after looking at the code I decided against extending it to targets to implement this loading of multiple sequences. Instead I based `HDFDataset` on `CachedDataset2` and implemented a different `load_seqs()`. Do you see any chance we would want to do that? Would be hard to make everything backwards compatible but in terms of code quality would be a big improvement in my opinion. Caching up to a certain memory limit could then be done simply by calling `load_seqs` with a wider range. The data is not as tightly packed as in these alloc_intervals in `CachedDataset` now, but I'm not sure whether in practice this makes any difference.

Anyway, should I make a pull request with a clean version of this `cache_hdf_file_handles` option?

albertz:
First of all, maybe someone should overtake the ownership of this class (`HDFDataset`). I'm only rarely using it and don't feel too responsible for it. Although I care about code quality and maintainability in general. @patrick-wilken you are maybe one candidate, but there should be also someone from ASR, to cover this case (dense input data) as well. Maybe @curufinwe or @JackTemaki.

From the user perspective, I think it should not be confusing. Why do we need an option for `cache_hdf_file_handles`? Why can't it be always enabled? Or what are the downsides? So if there are no downsides, I would recommend to just always use this code.

W.r.t. `cache_whole_file`, yes, I don't like that we would have this logic duplicated (both via the "old" `CachedDataset` caching logic, via the `cache` option, and now your new option) (because code quality/maintainability) and this also might cause confusion to the user. I understand that the implementation in `CachedDataset` only caches the input data. So we should either extend that, or remove the whole implementation and replace it by a new one. (You mention `CachedDataset2` but I think that one also does not support caching the whole file well, right?) When this is going to be reimplemented, I think we again should do some benchmarks. But maybe anyway make this separate from `cache_hdf_file_handles`.

patrick-wilken:
> maybe someone should overtake the ownership of this class

Yes, maybe me, and ideally someone who uses the current caching implementation regularly (if someone does 😄). Because I don't, and we would need feedback if want to change it.

> You mention CachedDataset2 but I think that one also does not support caching the whole file well, right?
One could call `load_seqs()` with `end=num_seqs`. This would store sequences as a list of `DatasetSeq` in memory instead of all packed in a single numpy array, so less efficient, but simpler and I assume good enough.

albertz:
> > maybe someone should overtake the ownership of this class
> 
> Yes, maybe me, and ideally someone who uses the current caching implementation regularly (if someone does 😄). Because I don't, and we would need feedback if want to change it.

You should check maybe in the RETURNN Slack Channel, who is using it. I always was recommending to not to use it, because the way it was implemented was also slow (it loaded everything into cache size, then went through it until cache was empty, and then filled up cache again, etc) and also obsolete because the RETURNN TF FeedDictDataProvider cache also solved the same thing (it loaded it async for any dataset). It was also buggy and I didn't fully trust the code.

> > You mention CachedDataset2 but I think that one also does not support caching the whole file well, right?
> 
> One could call `load_seqs()` with `end=num_seqs`. This would store sequences as a list of `DatasetSeq` in memory instead of all packed in a single numpy array, so less efficient, but simpler and I assume good enough.

This loads the seqs for the current epoch (in the current order defined by/for the epoch). I think at the end of the epoch, when loading the next epoch, it would discard the cache. It also cannot really reuse it in general.

I think the cache in `CachedDataset` also works across epochs. But as said, it is maybe buggy (there are two or three indirections for the seq indices to work around this I think...).

JackTemaki:
> Maybe @curufinwe or @JackTemaki.

I would be interested in this, as I am using HDFDataset (mostly in "meta" combination with OggZipDataset) in nearly every setup I have.

patrick-wilken:
Closing this for now, as #560 improved things already and further optimization of file access seems hard to do.

We can discuss refactoring the code and extending the caching separately.

Issue #500, opened 2021-04-30
Problems when declaring rec-time independent layers within rec-layer


Zettelkasten:
I am using a basic cross-attention setup inside a rec-layer, essentially as in e.g. [`test_reclayer_optimize_out_dot` here](https://github.com/rwth-i6/returnn/blob/522ce85122103fc8a21df5798e9596476250a96d/tests/test_TFNetworkRecLayer.py#L3435).
In that test case, the layers for calculating keys and value from the encoder states are declared outside the rec layer, as these are completely independent from the rec-layer time.
However, in my case I would like to declare them *within* the rec layer, right next to the rest of the attention calculation.
(That would simplify my code as everything related to the attention would be in one place, and I don't think there really is a reason this should not be supported. I use `optimize_move_layers_out=True` always anyway.)

I added a test case in PR #501:
```python3
  check_reclayer_optimize_out(
    {"class": "linear", "activation": None, "from": ["att"]},
    other_subnet_layers={
      "s": {"class": "linear", "activation": None, "with_bias": False, "from": ["data:source"],
            "n_out": EncKeyTotalDim},  # (B, D)  -- Q (query). D should be same as enc_ctx
      "att_query": {"class": "split_dims", "axis": "F", "dims": (AttNumHeads, EncKeyPerHeadDim),
                    "from": ["s"]},  # (B, H, D/H)
      # this does not depend on the rec-time, but you should still be able to define it here.
      "enc_ctx0": {"class": "linear", "activation": None, "with_bias": False, "from": ["base:encoder"],
                   "n_out": EncKeyTotalDim},  # (B, enc-T, D)
      "enc_ctx": {"class": "split_dims", "axis": "F", "dims": (AttNumHeads, EncKeyPerHeadDim),
                  "from": ["enc_ctx0"], "is_output_layer": True},  # (B, enc-T, H, D/H)
      "enc_value0": {"class": "linear", "activation": None, "with_bias": False, "from": ["base:encoder"],
                     "n_out": EncValueTotalDim},  # (B, enc-T, D)
      "enc_value": {"class": "split_dims", "axis": "F", "dims": (AttNumHeads, EncValuePerHeadDim),
                    "from": ["enc_value0"], "is_output_layer": True},  # (B, enc-T, H, D/H)
      "energy": {"class": "dot", "red1": -1, "red2": -1, "var1": "T", "var2": "T?",  # Note the "T?".
                 "from": ["enc_ctx", "att_query"]},
[...]

PR #501, opened 2021-04-30
Test case for #500

Zettelkasten:
See #500.

### DIFF
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -3499,7 +3499,8 @@ def test_reclayer_optimize_out_dot_kv_in_rec():
       "att": {"class": "merge_dims", "axes": "static", "from": ["att0"]},  # (B, H*V); Use "static" here.
       },
     shared_base_net={
-      "encoder": {"class": "copy", "from": ["data"]}
+      # need to mark 'encoder' as output layer, otherwise it will not be constructed.
+      "encoder": {"class": "copy", "from": ["data"], "is_output_layer": True}
     },
     rtol=1e-3)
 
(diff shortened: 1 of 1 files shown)


tests/test_TFNetworkRecLayer.py
```diff
@@ -3469,6 +3469,41 @@ def test_reclayer_optimize_out_dot():
...
+      "enc_value": {"class": "split_dims", "axis": "F", "dims": (AttNumHeads, EncValuePerHeadDim),
+                    "from": ["enc_value0"], "is_output_layer": True},  # (B, enc-T, H, D/H)
+      "energy": {"class": "dot", "red1": -1, "red2": -1, "var1": "T", "var2": "T?",  # Note the "T?".
+                 "from": ["enc_ctx", "att_query"]},
+      "att_weights": {"class": "softmax_over_spatial", "from": ["energy"]},  # (B, enc-T, H, 1)
+      "att0": {"class": "generic_attention", "weights": "att_weights", "base": "enc_value"},  # (B, H, V)
+      "att": {"class": "merge_dims", "axes": "static", "from": ["att0"]},  # (B, H*V); Use "static" here.
+      },
+    shared_base_net={
+      "encoder": {"class": "copy", "from": ["data"]}
```
albertz, line 3502:
```suggestion
      "encoder": {"class": "copy", "from": "data", "is_output_layer": True}
```

returnn/tf/layers/rec.py
```diff
@@ -295,6 +295,7 @@ def transform_config_dict(cls, d, network, get_layer):
     :param ((str) -> LayerBase) get_layer: function to get or construct another layer
     """
     if isinstance(d.get("unit"), dict):
+      d.setdefault("from", [])
```
albertz, line 298: Let this be handled eventually by #519 but not here. This is inconsistent. Also a RecLayer with subnet can have an input (it would unpack it inside the subnet). This is a very valid case which we use often.

albertz: So instead, just set explicitly for now in the test and in your config.

albertz (approved):

Issue #376, opened 2020-10-21
Decouple `Dataset` and chunking logic and batching logic


albertz:
Currently it is mixed, via `Dataset.iterate_seqs` (chunking logic) and `Dataset._generate_batches` (batching logic), although the logic is already mostly separate, and in principle, it should be separate. (Historically, it might have been mixed, because the pipeline is different depending on train vs dev, just as the dataset itself.)

When this gets decoupled, it still needs to support separate pipelines depending if train or dev.

I think in PyTorch, `DataLoader` is similar.

In `tf.data` design, this would be just further steps in the pipeline (even the batching).

We could do it in a similar pipeline way, i.e. have a `ChunkingDataset`, which gets one dataset in and which does the chunking, and RETURNN can optionally plug this dataset when needed.

We could also directly implement this as part of the new `tf.data` pipeline, but then the question is, how to keep compatible with older setups.

In #375, there is the request for a `train_custom_iterate_seqs` config option, which can also be part of this.

JackTemaki:
For a start I would just move the functions away from the `Dataset`, the dataset dependent parameters `chunk_size` and `chunk_step` and `weights` can stills stay where they are,  and are just passed in `generate_batches`. If this is the case, we still have the different pipelines for train vs dev according to the dataset parameters, but at least the code is independent and can be overwritten with custom code. Just as suggested in #375.

albertz:
Hm, not sure if this will make it simpler, or cleaner, if you keep chunk attribs in the `Dataset`. This sounds even uncleaner then, because now part of chunking / batching is separate, while the attribs are still in `Dataset`.
Also I don't quite understand why this would make the overwrite simpler. In my original proposal, I said to check for `Dataset.custom_iterate_seqs` inside `Dataset.iterate_seqs`. This would be even uglier if `iterate_seqs` is separate but `custom_iterate_seqs` is an attrib of `Dataset`.

In general, if you have a separate function (`iterate_seqs` separate), which is good, but what you want then is that the external dependencies/expectations (e.g. like attribs in `Dataset`) as minimal as possible. But it would be heavily depend on all sorts of stuff from `Dataset`, like all these attribs, which are not really used by `Dataset`, but only by `iterate_seqs`.

albertz:
Related is also #292.

albertz:
Note that while we agreed in principle that more decoupling would be nice, there are still a number of open questions, or just details which need to be clarified further.

[...]

PR #502, opened 2021-05-02
add custom chunking logic

jotix16:
This commit tries to add custom chunking logic without breaking anything and solve #375. It is related to #376, but doesn't offer a solution for that, no reorganization, it just makes it more complicated.

The `iterate_seqs` gets now an extra input `chunking_logic` that provides with the extra information required. 

#### Inputs 
- sequence(`numpy.array`)
- chunks_step(`int`)
- chunk_size(`int`)
#### Outputs
- chunks_length(`int`): number of chunks that can be generated from given sequence, chunk_size and chunk_step
- custom_chunk_generators(`dict[str, ()->(int, int)]`): dict with generators for every data_key that has a custom chunking logic

The names are not final and can be changed. This commit only presents the logic. Will add Andre's RNNT chunking and some tests afterwards. 

@albert is the idea ok in general?

**Edit**: Is there any script to run the tests locally? Couldn't find any information in the wiki or contribution. 
**Edit**: I am not very acquainted with the tests, but looks like chunking is being tested as part of different tests. Should I create a new file for testing chunking?

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -95,7 +95,7 @@ class Dataset(object):
     :param int window: features will be of dimension window * feature_dim, as we add a context-window around.
       not all datasets support this option.
     :param None|int|dict|NumbersDict|(dict,dict) context_window: will add this context for each chunk
-    :param None|str|int|(int,int)|dict|(dict,dict) chunking: "chunk_size:chunk_step"
+    :param None|str|int|(int,int)|dict|(dict,dict,dict) chunking: "chunk_size:chunk_step"
     :param str seq_ordering: "batching"-option in config. e.g. "default", "sorted" or "random".
       See self.get_seq_order_for_epoch() for more details.
     :param int|None random_seed_offset:
@@ -139,7 +139,7 @@ class Dataset(object):
     self._estimated_num_seqs = estimated_num_seqs
     self.min_chunk_size = min_chunk_size
     self.chunking_variance = chunking_variance
-    self.chunk_size, self.chunk_step = self._parse_chunking(chunking)
+    self.chunk_size, self.chunk_step, self.dynamic_chunk_gen = self._parse_chunking(chunking)
     if isinstance(context_window, (tuple, list)):
       assert len(context_window) == 2
       for elem in context_window:
@@ -193,11 +193,12 @@ class Dataset(object):
   @staticmethod
   def _parse_chunking(chunking):
     """
-    :param None|str|int|(int,int)|dict|(dict,dict)|(NumbersDict,NumbersDict) chunking:
+    :param None|str|int|(int,int)|dict|(dict,dict,dict)|(NumbersDict,NumbersDict) chunking:
       as it comes from the config / from the user
     :return: chunk_size, chunk_step
-    :rtype: (NumbersDict,NumbersDict)
+    :rtype: (NumbersDict,NumbersDict,dict)
     """
+    dynamic_chunk_gen = {}
     if isinstance(chunking, str):
       if ":" in chunking:
         chunking = tuple(map(int, chunking.split(":")))
@@ -205,7 +206,9 @@ class Dataset(object):
         chunking = int(chunking)
     if not isinstance(chunking, (tuple, list)):
       chunking = (chunking, None)
-    chunk_size, chunk_step = chunking
+    chunk_size, chunk_step = chunking[:2]
+    if len(chunking) == 3:
+      dynamic_chunk_gen = chunking[2]
     if chunk_size is None:
       chunk_size = 0
     assert isinstance(chunk_size, (int, dict, NumbersDict))
@@ -218,7 +221,7 @@ class Dataset(object):
     if chunk_size != 0:
       assert sorted(chunk_step.keys()) == sorted(chunk_size.keys())
       assert chunk_step.max_value() > 0, "chunking step must be positive (for some key)"
-    return chunk_size, chunk_step
+    return chunk_size, chunk_step, dynamic_chunk_gen
 
   @staticmethod
   def _load_seq_list_file(filename, use_cache_manager=False, expect_list=True):
@@ -941,12 +944,13 @@ class Dataset(object):
       i += 1
     return numpy.array(priori / self.get_num_timesteps(), dtype=numpy.float32)
 
-  def iterate_seqs(self, chunk_size=None, chunk_step=None, used_data_keys=None):
+  def iterate_seqs(self, chunk_size=None, chunk_step=None, used_data_keys=None, chunking_logic=None):
     """
     Takes chunking into consideration.
     :param int|NumbersDict chunk_size:
     :param int|NumbersDict chunk_step:
     :param set(str)|None used_data_keys:
+    :param dict[str, (numpy.ndarray, int, int)->(int, dict[str,()->(int, int))])]|None chunking_logic:
     :return: generator which yields tuples (seq index, seq start, seq end)
     :rtype: list[(int,NumbersDict,NumbersDict)]
     """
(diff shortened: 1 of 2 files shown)


albertz:
This is supposed to be an alternative for #375?

This is still too specific. And also too complicated. And also not documented. Please provide documentation first, before you even implement it. It took a while to understand what you do. Before you even implement sth, please propose the idea first (in pseudo code if you think that is more helpful).

I don't think `chunking_logic` should be a dict. I think it is more generic and simpler to just allow the user to define an own custom `iterate_seqs` function somehow.

jotix16:
>This is supposed to be an alternative for #375? And an implementation for #376?

It is only an alternative for #375. I said it is only related to #376(but doesn't solve anything)


>It took a while to understand what you do. Before you even implement sth, please propose the idea first

The main comment holds the logic.
The chunking logic is an extension to the tuple (chunk_size, chunk_step). Now we have (chunk_size, chunk_step, chunking_logic).

It is a dict because you have to specify the data_key for which the chunking is implemented for.

Why do you think it is too specific?
You can just implement the logic however you wish. I dont think one requires more information than than one sequence and chunk_size, chunk_step to properly specify a chunking logic (or I am not thinking broader than transducer based models)

albertz:
> The main comment holds the logic.

Ah, I did not really read that. You should not explain such things in the comment. This should be documented in the code, docstrings, documentation. No-one will later check the comments in a PR to understand why sth was done this way.

> Why do you think it is too specific?

E.g. your assumptions:

- it is separate per data key (not all approaches might fit into this)
- it gets chunk size / chunk step (that might not really be relevant for more custom stuff)
- it returns chunk_step and chunk_size (not all approaches might fit into this)

This all is very close to the existing approach. What if you (or the user) would want to implement some completely different approach? The your generic function is not generic enough and we again need to have a second more generic function. Which is sth we should avoid.

There should not be any such assumptions. When we already introduce a custom function for the user config, it should be generic. E.g. the user could define a generic `iterate_seqs` function (separate for training and testing).

We might simplify it a bit and still keep it generic. E.g. I think this part of the loop can always be done by us:
```
def seq_idx_generator(self):
    seq_idx = 0
    while self.is_less_than_num_seqs(seq_idx):
       yield seq_idx
       seq_idx += 1
```

And the user defines a function `(Dataset, Iterable[int]) -> Iterable[Tuple[int, NumbersDict, NumbersDict]]` which gets the dataset + `seq_idx_generator()` and returns the same as `iterate_seqs`.

It might also be even more generic if the user directly defines the whole batching logic. E.g. currently this is still not generic enough such that the user could e.g. define/implement the bucketing logic as it is common in other frameworks.

Actually all of that is already possible with our alternative `tf.data` pipeline. So we maybe should also not over-engineer this now, as this is anyway all kind of deprecated.

But all of this discussion is exactly #376.

---

> I am not very acquainted with the tests, but looks like chunking is being tested as part of different tests. Should I create a new file for testing chunking?

There are tests for the dataset, and this is part of the dataset. No new file is needed.

---

Off topic here (use discussion forum for it, or open separate issue for missing documentation or so):

> Is there any script to run the tests locally? Couldn't find any information in the wiki or contribution.

You just run them directly. Or via `nosetests`. (You can also simply look at how GitHub CI runs them.)

jotix16:
``` python
def my_chunking(data, chunk_size, chunk_step):
  blank_indices = np.where(data == blank_index)[0]  # [T]
  blank_length = len(blank_indices)
  len = (blank_length - chunk_size) / chunk_step

  def chunk_generator():
  # ...
    yield chunk_step, chunk_size  

  return len, chunk_generator



chunking = (
   {"data": 12 * 2, "alignment": 22},  # chunk_size
   {"data": 12 * 2 // 2, "alignment": 22 // 2},  # chunk_step
   {"alignment": my_chunking}   # data has the default chunking logic, alignment has a custom one.
  )
```

> This all is very close to the existing approach.

Yes, but it doesn't have the problem that you can't change it anymore. The one you mentioned in #375 : "*But even that is hacky and ugly, and will break in some cases, e.g. when you define custom_iterate_seqs in some epochs, and later not anymore. Then it would not correctly reset this.*"

> We might simplify it a bit and still keep it generic. E.g. I think this part of the loop can always be done by us:

You are right, this looks much more flexible. But I don't really like the idea that it should be dataset dependent. Chunking is a simple thing, it shouldn't require the whole dataset instance to be done. It gets more difficult for the user to implement a custom_iterator, i.e. is not straightforward without a previous example.

Could you mention one such scenario when such a complicated chunking is required? Maybe then I can better understand what I am missing.

albertz:
> This is a dummy usage in case I didn't make it clear: ...

Note: Before you implement sth, this should exactly be the proposal, and then to be discussed.

When you implement sth and make a PR, this should not be a comment, but part of the code / docstring / documentation.

> > This all is very close to the existing approach.
> 
> Yes, but it doesn't have the problem that you can't change it anymore. The one you mentioned in #375 : "_But even that is hacky and ugly, and will break in some cases, e.g. when you define custom_iterate_seqs in some epochs, and later not anymore. Then it would not correctly reset this._"

This comment was purely about the specific implementation in the #375 PR, not about the generic approach to provide a custom `iterate_seqs` function.

> > We might simplify it a bit and still keep it generic. E.g. I think this part of the loop can always be done by us:
> 
> You are right, this looks much more flexible. But I don't really like the idea that it should be dataset dependent. Chunking is a simple thing, it shouldn't require the whole dataset instance to be done. It gets more difficult for the user to implement a custom_iterator,

I don't exactly understand. This is not dataset dependent?

> i.e. is not straightforward without a previous example.

I see this argument. But this is a problem with almost all solutions. Also for your case, a user would not really be able to do this without having seen such an example before.

So we just give this example. The example for such a custom iterator is just as simple/short, or maybe even shorter as your example.

> Could you mention one such scenario when such a complicated chunking is required? Maybe then I can better understand what I am missing.

E.g. to implement the standard batching behavior as it present in most other frameworks (bucketing).

Or to have some logic where you select from `classes` to contain some specific class indices, or be more uniform, and then the `data` accordingly to match the `classes` selection.

Or whatever. Maybe even you could do sort of data augmentation / shuffling or so at this point. Or image translation, where you select sub parts of source and target sentence according to some logic.

Maybe returning `Tuple[int, NumbersDict, NumbersDict]` is also already too specific, and you would just return directly the data (dict data key -> numpy array).

jotix16:
> Note: Before you implement sth, this should exactly be the proposal, and then to be discussed.

Now I know :)

> It might also be even more generic if the user directly defines the whole batching logic. E.g. currently this is still not generic enough such that the user could e.g. define/implement the bucketing logic as it is common in other frameworks.

> E.g. to implement the standard batching behavior as it present in most other frameworks (bucketing).

I am a little confused now. Isn't batching and chunking independent? Batching comes after chunking in the pipeline.
This flexibility you are saying involves only the batching logic in '_generate_batch`. 

**batching** is in sequence level
**chunking** is in frame level

```
      seqs                   seqs'                     batches
     ----->     chunking     ----->     batching      --------->
```  

is this right?

albertz:
> I am a little confused now. Isn't batching and chunking independent? Batching comes after chunking in the pipeline.
> This flexibility you are saying involves only the batching logic in '_generate_batch`.

They are in our pipeline. But chunking is just one custom post processing thing you can do (actually not so common, I'm not sure you find any other framework having this). There are many other potential post processing steps you could do. The general interface to allow any kind of post processing would be some function seqs -> seqs (output: yielding dict with data key -> numpy array).

Sure you can always keep batching as a separate step. Sorry, my suggestion was probably confusing. I thought you could also generalize this part just in the same way. And there could be some logic where you want to combine the pipeline. E.g. prepare chunks exactly such that it is optimal for batching or so.

> **batching** is in sequence level
> **chunking** is in frame level
> 
> ```
>       seqs                   seqs'                     batches
>      ----->     chunking     ----->     batching      --------->
> ```
> 
> is this right?

Yes, this is our custom pipeline (when we do not use `tf.data`).

Note that `tf.data` and our support in RETURNN is very much this generic pipeline already. The user just provides the final batched `tf.data.Dataset`.

I think it is also similar in PyTorch.

albertz:
Obsolete via #1075.

PR #506, opened 2021-05-18
Subnetwork: add construction warning

JackTemaki:
This is only a simple change, but I also wanted to ask with this about future logging messages. The problem is, that the errors/exceptions during construction are often not related to the actual error at all.
Sometimes it takes hours to work through the different Exceptions, especially if you have a new config and you are not sure about what exactly is correct.

In the case here a simple missing `concat_sources=False` in `SubnetworkLayer` caused 5 new completely unrelated Exceptions, because the construction continued after throwing a KeyError. The problem is, that the new Exceptions seemed to be much more relevant, as they were related to the (config) code that was changed.

So in the future I would like to add warnings like this one to make the construction log more verbose, and to give reasonable hints that may save many hours of useless debugging.

In the location of the code I changed the KeyError can only occur in 2 cases: You have an actual typo in the key, or your forgot to add `concat_sources=False` and RETURNN searches for the key in a wrong location, so for me it would be reasonable to give the user information about this second case (the first one is clear from the KeyError itself).

Another question: Are there actually cases where having a KeyError is okay, and the construction will continue to a valid solution? If not, we might want to directly fail the construction here.

### DIFF
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -2536,7 +2536,11 @@ class Subnetwork:
         return get_layer(layer_name)
 
       else:  # name != "data"
-        return base_get_layer("data:%s" % name)
+        try:
+          return base_get_layer("data:%s" % name)
+        except KeyError as e:
+          print("warning: possibly missing concat_sources=False in %s" % self, file=log.v3)
+          raise e
 
     # Not concat_sources
     for i, arg in enumerate(self._from_arg):


JackTemaki:
And the PyCharm test case is failing for an unreasonable error? This is strange...

patrick-wilken:
You mean this error is suppressed completely, so you add a warning? Or is it about seeing the warning at a better/earlier time?

In any case, I also had troubles with suppressed error messages. The layers in the recurrent net are constructed by trial and error, suppressing all exceptions because many are expected, for example if an input for a given layer is not yet constructed.

My intuition is that there should be a more deterministic way to construct the layers, but I haven't dug into this.
But at least we could raise a custom Exception for the cases where errors occur because of expected missing dependencies and only catch this Exception type during the construction loop. Then all other exceptions would be visible without problems. But this itself is a major effort...

albertz:
Instead of catching the very generic `KeyError`, why not throw a more specific exception at the place where this could occur? I assume this is in `ExternData.get_data`?

`concat_sources` is about `SubnetworkLayer`, right? `SubnetworkLayer` can also have an additional check, and then throw a more specific error/exception.

Or, you say you thought that the other later exceptions are more relevant. But this is in general wrong. It is usually always the first exception (not just in this case, very generic, even outside of RETURNN). I'm not sure if we should do sth about this. Maybe highlight the first exception somewhat more?

JackTemaki:
> why not throw a more specific exception at the place where this could occur

But this is the exact location where this occurs! This is why I am catching the KeyError there. Only at this location it is relevant, but no where else. The problem is that we have the overloaded "data" syntax, which could mean using ExternData or Subnetwork inputs (we can discuss if this is a good idea, I do not really like it because it causes a lot of confusion), so here is the earliest stage in the execution where I know that a KeyError might be thrown not because of a typo or not existing layer, but because the user forgot to add "concat_sources=False".

albertz:
> > why not throw a more specific exception at the place where this could occur
> 
> But this is the exact location where this occurs!

This cannot be. What is `base_get_layer`? I'm sure it is somewhere inside this, and not directly there. I assume it is thrown in `ExternData.get_data` or so.

> ... because of a typo

A typo will in most cases not lead to a `KeyError`. It depends where exactly this is thrown. This is what I asked before.

> or not existing layer

This also cannot be. This would lead to a `LayerNotFound` exception

> but because the user forgot to add "concat_sources=False".

You keep mentioning `concat_sources`. This is an option of `SubnetworkLayer`. This is what I mentioned/asked before, so this is actually about `SubnetworkLayer` here? And maybe specifically also `ExternData.get_data`?

JackTemaki:
> This cannot be. What is base_get_layer? I'm sure it is somewhere inside this, and not directly there. I assume it is thrown in ExternData.get_data or so.

You are mssunderstanding me

- yes, the KeyError is of course thrown in `in ExternData.get_data`, but this is absolutely the correct behavior
- so of course a typo in accessing a dataset also results in a KeyError
- so does this happen when a user did not specify the dataset he wants (sorry that I wrote Layer before which was wrong, I meant a data related layer input)
- and this happens when the user intended to access the "from" layer inputs, but either forgot to add the layers, or did not specify `concat_sources=False`

But in the ExternData.get_data we have no idea where we are coming from, so changing the Exception itself makes no sense. So I wanted just wanted to print a warning in the Subnetwork (= related to the Subnetwork Layer), so that the user sees what a possible reason for the KeyError might be, besides the obvious ones.

I do not think it is a good idea to check for correctness in the SubNetworkLayer itself, as this will require checking all parameters of all sublayers, which makes no sense when you can do it in a simple way here (maybe, I did not look at this in detail, as I did not want to spend hours on adding a simple warning).

albertz:
> yes, the KeyError is of course thrown in `ExternData.get_data`

That is what I asked in my first comment. So yes. This was not clear to me. Can you maybe show the full stack trace?

> so of course a typo in accessing a dataset also results in a KeyError

Now you speak about a dataset. Before you spoke about a layer. A typo in a layer name should give you a `LayerNotFound`.

And a invalid dataset name does not necessarily lead to a `KeyError`. I would even argue that `KeyError` should never be thrown for this. So then this is about `ExternData.get_data`, as I argued before.

> `concat_sources=False`

So again, you speak about `SubnetworkLayer`/`Subnetwork` here? This is still not clear to me. This is what I asked a couple of times now. But I assume so now. I will edit the PR description accordingly. Because this is a quite crucial information which is not mentioned at all. You should say this in the very beginning.

But I still don't understand, why not catching the problem where it occurs (`ExternData.get_data`) and directly throw a meaningful exception (not `KeyError`)? Why does this not solve your concerns?

albertz:
Another thing not exactly clear: When you say the net construction continues, this sounds like you speak about construction in `RecLayer`? This would be also relevant to mention. I assume so now. But please make these things very clear in the beginning. And further, this is specifically about the template layer construction? I also assume this now.

The `RecLayer` template construction logic continues on exception, as you write. But we could maybe directly stop on some specific exceptions, like `LayerNotFound`, and maybe also a new `DataNotFound` or so (but definitely not `KeyError`).

JackTemaki:
> Why does this not solve your concerns?

Because this was only about the mismatch between the Exception that is thrown and the error the user actually did (it does not matter if it is KeyError or a DataNotFound error, the information content of exception stays the same, and does not represent the "error" the user actually did).

As this discussion is not at all about what I intended it to be about I try again in a very general way:

I formulated two questions in my initial text:
 - How should I add additional information that is helpful to the user, but can not be part of the original Exception itself?
 -> my solution was to capture an Exception in a location where it may have a different meaning for the user, and add an additional warning. In the case I had, also DataNotFound would be not related at all to what I did wrong. Because I wanted to access a `Layer`, but because I did not have the correct parameter in the Subnetwork (`concat_sources=False`)  RETURNN tried to search in ExternData instead. As I said before, this is a problem of syntax overloading.
 
 - Should we stop template construction for the recurrent layer directly on some Exceptions.
 -> In this case we can talk about changing the KeyError to DataNotFound error, and add this to a "critical exception" set in the construction context. But I formulated this just as an additional side question, I now see that this was not a good idea to do in the same PR.

albertz:
Ah sorry, I understand now. So this is really specific about `SubnetworkLayer`, and that `data:` can change depending on its settings (which I don't like, but that is how it is; this is why I generally would also avoid to use `data:` in the `SubnetworkLayer` at all).

And you like to add more meaningful information to potential exceptions on wrong usages of `data:` (caused maybe indirectly by other wrong settings in `SubnetworkLayer`).

In general, we sometimes have such cases that we catch exceptions and extend them by further information. E.g. in the template construction, we catch all exceptions, and then maybe rethrow some combined new exception.

This is actually not so uncommon in Python. Some pattern like:

```
class FooException(Exception): pass

def foo():
  ...
  raise FooException()

class BarException(Exception): pass

def bar():
  ...
  try:
    foo()
  except FooException as foo_exc:
    throw BarException()
    # or since Python 3:
    throw BarException() from foo_exc
```

Note that it is important here that you use a specific (somewhat potentially expected) `FooException`, and not generic exceptions like `KeyError`.

Transferred to your case, it really means we need to introduce some `DataNotFound` exception. You definitely should not catch all `KeyError` exceptions.

JackTemaki:
As we want to discourage people from using the `data:` syntax to access subnetwork inputs there is no need to add this warning. I will close this pull request, and open a new one that deals with the Exception handling during the construction only (and of course adding the `DataNotFound` Exception).

PR #507, opened 2021-05-19
Fix hierarchical base access for subnetwork

JackTemaki:
When using subnetworks inside a recurrent net, from which then layers try to access layers outside the recurrent net, the dependencies will not be resolved correctly. Unfortunately, in some cases the target layer in the root dict will already exist, so this bug did not always appear and was difficult to track.

I added two test cases, one with using `base:base:layer` in the sub-layer and one with using the subnetwork `from: ["base:layer"]` and then using `data:[n]` in the sub-layer, as those two cases have to be handled differently.

The solution is possibly not optimal (basically just copied from the RecLayer), but after days of searching for this bug I was still not able to fully understand the construction code, and what is supposed to happen when and where.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -6600,6 +6600,42 @@ class SubnetworkLayer(LayerBase):
     d["_output"] = subnet.construct_layer("output", parent_get_layer=get_layer)
     d["_from"] = d.get("from", "data")  # cache this
     d["from"] = []  # disable now. we should get them in the template construction when needed
+
+    def sub_get_layer(name):
+      """
+      :param str name:
+      :rtype: LayerBase
+      """
+      # Only used to resolve deps to base network.
+      if name.startswith("data:"):
+        # check for "data:<n>" notation
+        try:
+          index = int(name.split(":")[-1])
+          return get_layer(d["_from"][index])
+        except ValueError as e:
+          # only resolve dependency if we access data defined by the subnetwork
+          if name[len("data:")] in d["_from"]:
+            return get_layer("base:" + name[len("data:"):])
+      elif name.startswith("base:"):
+        return get_layer(name[len("base:"):])  # calls get_layer of parent network
+      return InternalLayer(
+        name=name, network=subnet,
+        output=Data(name="dummy:SubnetworkLayer.transform_config_dict(%s)" % name, dim=1))
+
+    from returnn.tf.network import TFNetwork
+    subnet = TFNetwork(parent_net=network, extern_data=network.extern_data, name="%s(tmp)" % d["_name"])
+    for key, sub in d["subnetwork"].items():  # iterate over the layers of the subnet
+      assert isinstance(sub, dict)
+      if "class" in sub:
+        sub = sub.copy()
+        class_name = sub.pop("class")
+        cl = get_layer_class(class_name)
+        sub["_network"] = subnet
+        sub["_name"] = key
+        # Operate on a copy because we will transform the dict later.
+        # We only need this to resolve any other layer dependencies in the main network.
+        cl.transform_config_dict(sub, network=subnet, get_layer=sub_get_layer)
+
     super(SubnetworkLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
 
   @classmethod
--- a/tests/test_TFEngine.py
+++ b/tests/test_TFEngine.py
@@ -966,30 +966,7 @@ def test_engine_search_attention():
   check_engine_search_attention()
 
 
-def check_engine_train_simple_attention(lstm_unit):
-  net_dict = {
-    "lstm0_fw": {"class": "rec", "unit": lstm_unit, "n_out": 20, "dropout": 0.0, "L2": 0.01, "direction": 1},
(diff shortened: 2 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -6598,8 +6598,34 @@ def transform_config_dict(cls, d, network, get_layer):
     d["_subnet"] = subnet
     # In case of non-template construction, this will trigger the non-template construction of our "output" sublayer.
     d["_output"] = subnet.construct_layer("output", parent_get_layer=get_layer)
-    d["_from"] = d.get("from", "data")  # cache this
-    d["from"] = []  # disable now. we should get them in the template construction when needed
+    d["from"] = d.get("from", "data")
```
albertz, line 6601: But this is exactly what should **not** be done here.
The sources **must** only be accessed dynamically during the template construction of the subnet to correctly resolve dependencies only on those sub layers which really need a particular source.

albertz: I doubt that there is a problem on this part in the code. I think `transform_config_dict` is fine and should be left like it is.

tests/test_TFEngine.py
```diff
@@ -1096,6 +1099,80 @@ def test_attention_train_then_search():
   engine.finalize()
 
 
+def test_attention_subnetwork_base_dependency():
+    net_dict = {
```
albertz, line 1103: Indent is off.

tests/test_TFEngine.py
```diff
@@ -1096,6 +1099,80 @@ def test_attention_train_then_search():
...
+        'output': {'class': 'choice', 'target': 'classes', 'beam_size': 4, 'from': ["output_prob"]},
+        "end": {"class": "compare", "from": ["output"], "value": 0},
+        'orth_embed': {'class': 'linear', 'activation': None, 'from': ['output'], "n_out": 7},
+        "s": {"class": "rnn_cell", "unit": "LSTMBlock", "from": ["prev:att", "prev:orth_embed"], "n_out": 7},
+        "att": {
+          "class": "subnetwork",
+          "from": ["s", "prev:orth_embed", "base:encoder"],
+          "concat_sources": False,
+          "subnetwork": {
+            "c_in": {"class": "linear", "activation": "tanh", "from": ["data:0", "data:1"], "n_out": 5},
```
albertz, line 1163: Btw, (not related to this PR, just generic comment), instead of `data:0` here, you could also simply directly use `base:s`. This is what I usually do. Isn't this cleaner?

albertz:
Despite the wrong solution (just remove that for now, it should be done differently), can you also try to add a simplified test case? Simplified = as minimal as possible.

tests/test_TFEngine.py
JackTemaki: The best way is actually to define them in the "from" and then access it via the same name, so writing e.g. `data:s`. Because then you see exactly which external inputs the subnetwork has, but you also do not have to search for the index numbers.

JackTemaki:
> Despite the wrong solution (just remove that for now, it must be done differently), can you also try to add a simplified test case? Simplified = as minimal as possible.

The test case is already  (nearly) as minimal as possible. I can check again if the encoder really does need to be a subnetwork, but sometimes there were some changes in the construction order so that the bug does not appear.

tests/test_TFEngine.py
albertz: > access it via the same name, so writing e.g. `data:s`

This is actually the solution I dislike the most. This mixes up the data names and layer names. This could lead to strange cases when you have same data key names as layer names (unusual maybe, but so are all potential bugs in the beginning, until you get hit by them). And why is this better than just `base:s`? The `base:` prefix is clear what it does. What the `data:` prefix does is not so clear (you were searching long time for some bugs because of this, which would not have happened with `base:`).

returnn/tf/layers/basic.py
JackTemaki: Yes, editing the from was not correct, and I was aware of that. I added this to show what would fix one of the test cases. But why is the other change wrong? The `transform_config_dict`of the Recurrent layer has exactly the same code to resolve the dependecies, so why should it be correct for a recurrent subnetwork layer but be incorrect for a normal subnetwork layer?

FYI:
the "from" change was used to make `test_attention_subnetwork_from_dependency` work, and
copying the code from the recurrent layer that iterates the sublayers was used to make `test_attention_subnetwork_base_dependency` work.

albertz: Because that was the whole point of all the recent `SubnetworkLayer` changes, that the whole `SubnetworkLayer` itself does not directly have the dependencies, but just its sub layers have. This is important to be able to split it up e.g. for optimization (e.g. some sub layers are inside the loop, some are outside).

JackTemaki: But this is only related to the "from" change (which I understand why this was wrong).

The other code should have no effect on this, otherwise this would not work for the recurrent layer.

I will update the PR accordingly and then we can have a look again.

JackTemaki: Okay, then I am definitely not able to solve this

albertz:
New PR in #511. Closing this one.

PR #509, opened 2021-05-19
RepeatLayer: Support input without batch dim, constant int repetitions

Zettelkasten:
This PR extends `RepeatLayer` s.t. it also works if the input values / repetition count tensor do not have a batch dim.
Also, if the repetition count tensor does not have the axis we repeat along, it is simply broadcasted along the entire axis.
This allows you to now just give a single int as `repetitions` input.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3246,40 +3246,71 @@ class RepeatLayer(_ConcatInputLayer):
   def __init__(self, repetitions, axis="T", **kwargs):
     """
     :param LayerBase|int repetitions:
-      number of repetitions for each sequence and position in target axis, [B,T] or [T,B]
+      number of repetitions for each sequence and position in target axis.
+      Can be [B,T] or [T,B] or some subset of that shape
     :param str axis: (dynamic) axis for repetition (currently only time axis is supported)
     """
     from returnn.tf.util.data import DimensionTag
     super(RepeatLayer, self).__init__(**kwargs)
     self.repetitions = repetitions
     if isinstance(self.repetitions, int):
-      repetitions_data = Data.from_tensor(tf.constant(self.repetitions)).copy_add_batch_dim(0)
+      repetitions_data = Data.from_tensor(tf.constant(self.repetitions))
     else:
       assert isinstance(self.repetitions, LayerBase)
       repetitions_data = self.repetitions.output
-    input_data = self.input_data
-    assert self.input_data and self.input_data.have_batch_axis()
-    input_data = input_data.copy_as_batch_major()
-    original_axis = input_data.get_axis_from_description(axis, allow_int=False)
-    input_data = input_data.copy_move_axis(original_axis, 1)
-    repetitions_data = repetitions_data.copy_as_batch_major()
+    input_axis = self.input_data.get_axis_from_description(axis, allow_int=False)
     assert repetitions_data.dtype in ["int32", "int64"]
-    assert repetitions_data.have_batch_axis() and repetitions_data.batch_ndim == 2, (
-      "RepeatLayer only accepts repetitions of the shape (B, None|S) or (None|S, B)")
-    assert input_data.shape[0] == repetitions_data.shape[0], (
-      "Axis mismatch between input (%i) and repetitions (%i)" %
-      (input_data.shape[0], repetitions_data.shape[0]))
+    if repetitions_data.ndim == 0:  # no T dim: add a (un)broadcasted one.
+      axis_dim_tag = self.input_data.get_dim_tag(input_axis)
+      repetitions_data = repetitions_data.copy_add_dim_by_tag(axis_dim_tag, unbroadcast=True)
+    repetitions_axis = repetitions_data.get_axis_from_description(axis, allow_int=False)
+    assert repetitions_data.ndim == 1, "Repetitions %r must only have at most one non-batch axis" % repetitions
+    assert repetitions_data.batch_shape[repetitions_axis] == self.input_data.batch_shape[input_axis], (
+      "Axis mismatch between input (%i) and repetitions (%i)" % (
+        self.input_data.batch_shape[input_axis], repetitions_data.batch_shape[repetitions_axis]))
+
+    assert self.output.have_batch_axis() == (self.input_data.have_batch_axis() or repetitions_data.have_batch_axis())
+
+    def copy_placeholder_with_batch_axis(data, other_batch):
+      """
+      Adds a batch dim by unbroadcasting if our output also has a batch dim,
+      otherwise just expands the dims.
+
+      :param Data data:
+      :param BatchInfo|None other_batch: use this batch info if we should add some
+      :rtype: tf.Tensor
+      :return: the placeholder, with shape [B, T, ...], maybe with added batch dim.
+      """
+      if self.output.have_batch_axis():
+        if data.have_batch_axis():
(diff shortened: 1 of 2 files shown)


Zettelkasten:
I also tried to add a test case [here](https://github.com/rwth-i6/returnn/pull/509/commits/3e95de1253a1f3c9133f128608bf7948766e4584) where the input has no batch dim, but the repetitions tensor does. But that one currently fails: When I do `data.copy_add_batch_dim(batch_dim_axis=0)` on the input data, RETURNN infer the `batch`:
```
  File "/u/petrick/software/returnn/official/tests/test_TFNetworkLayer.py", line 1868, in test_RepeatLayerTF
    line: _run_repeat_layer(session, net, input_data_layer)
    locals:
      _run_repeat_layer = <global> <function _run_repeat_layer at 0x7f6126beca60>
      session = <local> <tensorflow.python.client.session.Session object at 0x7f6126c223d0>
      net = <local> <TFNetwork '<network via test_RepeatLayerTF>' train>
      input_data_layer = <local> <InternalLayer 'src' out_type=Data(batch_shape_meta=["spatial0:var:'Const:0':random_uniform/src_output",F|5])>
  File "/u/petrick/software/returnn/official/tests/test_TFNetworkLayer.py", line 1809, in _run_repeat_layer
    line: out_data = RepeatLayer.get_out_data_from_opts(**opts)
    locals:
      out_data = <not found>
      RepeatLayer = <global> <class 'returnn.tf.layers.basic.RepeatLayer'>
      RepeatLayer.get_out_data_from_opts = <global> <bound method RepeatLayer.get_out_data_from_opts of <class 'returnn.tf.layers.basic.RepeatLayer'>>
      opts = <local> {'network': <TFNetwork '<network via test_RepeatLayerTF>' train>, 'name': 'repeat_layer_test', 'sources': [<InternalLayer 'src' out_type=Data(batch_shape_meta=["spatial0:var:'Const:0':random_uniform/src_output",F|5])>], 'repetitions': <InternalLayer 'repetitions' out_type=Data(dtype='int32', batc...
  File "/u/petrick/software/returnn/official/returnn/tf/layers/basic.py", line 3346, in RepeatLayer.get_out_data_from_opts
    line: data = data.copy_add_batch_dim(batch_dim_axis=0)
    locals:
      data = <local> Data(name='repeat_layer_test_output', batch_shape_meta=["spatial0:var:'Const:0':random_uniform/src_output",F|5])
      data.copy_add_batch_dim = <local> <bound method Data.copy_add_batch_dim of Data(name='repeat_layer_test_output', batch_shape_meta=["spatial0:var:'Const:0':random_uniform/src_output",F|5])>
      batch_dim_axis = <not found>
  File "/u/petrick/software/returnn/official/returnn/tf/util/data.py", line 1571, in Data.copy_add_batch_dim
    line: batch = LayerBase.get_recent_layer().get_batch_info()
    locals:
      batch = <local> None
      LayerBase = <local> <class 'returnn.tf.layers.base.LayerBase'>
      LayerBase.get_recent_layer = <local> <bound method LayerBase.get_recent_layer of <class 'returnn.tf.layers.base.LayerBase'>>
      get_batch_info = <not found>
  File "/u/petrick/software/returnn/official/returnn/tf/layers/base.py", line 768, in LayerBase.get_batch_info
    line: batch = self.network.get_global_batch_info()
    locals:
      batch = <not found>
      self = <local> <InternalLayer 'repetitions' out_type=Data(dtype='int32', batch_shape_meta=[B?,T|"time:var:'Const_2:0':repetitions_output"])>
      self.network = <local> <TFNetwork '<network via test_RepeatLayerTF>' train>
      self.network.get_global_batch_info = <local> <bound method TFNetwork.get_global_batch_info of <TFNetwork '<network via test_RepeatLayerTF>' train>>
  File "/u/petrick/software/returnn/official/returnn/tf/network.py", line 2122, in TFNetwork.get_global_batch_info
    line: raise Exception("%s: Cannot get global batch info" % root)
    locals:
      Exception = <builtin> <class 'Exception'>
      root = <local> <TFNetwork '<network via test_RepeatLayerTF>' train>
Exception: <TFNetwork '<network via test_RepeatLayerTF>' train>: Cannot get global batch info
```
This is because there is no `extern_data` in the test case, but also for some reason the `repetitions` layer does not have any `batch` even though it has a `batch_dim_axis`. Is that expected? How can I add this info?

Zettelkasten:
I rewrote parts of this because I overlooked the case that both `input_data` and `repetitions_data` do not have a batch dim. That should be fixed now.

albertz:
There are tests failing?

albertz:
> RETURNN cannot infer the batch

It usually infers it from extern data. So this fails if there is no extern data. Instead of working around that in the test by this hack to explicitly set it (which has other potential problems which might break at some future point), I would just add some extern data.

returnn/tf/layers/basic.py
```diff
@@ -3246,40 +3246,71 @@ class RepeatLayer(_ConcatInputLayer):
...
+      """
+      Adds a batch dim by unbroadcasting if our output also has a batch dim,
+      otherwise just expands the dims.
+
+      :param Data data:
+      :param BatchInfo|None other_batch: use this batch info if we should add some
+      :rtype: tf.Tensor
+      :return: the placeholder, with shape [B, T, ...], maybe with added batch dim.
+      """
+      if self.output.have_batch_axis():
```
patrick-wilken, line 3284: Isn't this always true? `self.output` should come from `self.get_out_data_from_opts()` and there it seems you always add a batch axis.

returnn/tf/layers/basic.py
```diff
@@ -3310,21 +3346,29 @@ def transform_config_dict(cls, d, network, get_layer):
...
     :rtype: Data
     """
     data = get_concat_sources_data_template(sources, name="%s_output" % name)
-    data = data.copy_as_batch_major()
-    axis = data.get_axis_from_description(axis, allow_int=False)
-    data = data.copy_move_axis(axis, data.get_batch_axis(0))
+    if data.have_batch_axis():
+      data = data.copy_as_batch_major()
+    elif isinstance(repetitions, LayerBase):
+      assert repetitions.output.have_batch_axis()
```
albertz, line 3365: Why this assert? Just not implemented otherwise? Because in general, should also work without, right?

returnn/tf/layers/basic.py
albertz: In case `repetitions` is an `int`, there might be no batch.

returnn/tf/layers/basic.py
albertz: If this is just not implemented, either add such a comment:
```suggestion
      assert repetitions.output.have_batch_axis()  # not implemented
```
Or raise `NotImplementedError`.

returnn/tf/layers/basic.py
patrick-wilken: Ah, true.

returnn/tf/layers/basic.py
Zettelkasten: Ups, that should not have been an assert but part of the if-condition.

returnn/tf/layers/basic.py
Zettelkasten: My code in `self.get_out_data_from_opts` was also a bit wrong before (see @albertz's comment). In general, `self.output` has a batch dim exactly if the input or the repetitions have one.

albertz (approved):

PR #510, opened 2021-05-20
Modified Exception handling in template construction

JackTemaki:
As discussed in #506, we want to throw e.g. are more explicit Exception `DataNotFound` when there is an error in accessing the `ExternData.data` dict via `get_data()`.

I propose to catch errors like this directly, instead of continuing the construction which will defenitely fail. In that case, RETURNN will still display all the debug information from the template construction, but the Exception that is definitely indicating an error is displayed independently, directly visible to the user.

This is important because in many cases the "important" Exception will neither be the last nor the first one in the collection of Exceptions that is created during template construction.


returnn/tf/layers/rec.py
```diff
@@ -1043,11 +1043,14 @@ def _construct_template(self):
...
     from collections import OrderedDict
     from returnn.util.basic import StringIO
-    from returnn.tf.network import NetworkConstructionDependencyLoopException
+    from returnn.tf.network import NetworkConstructionDependencyLoopException, DataNotFound
     # The stack trace is not so interesting for these exceptions.
     skip_stack_trace_exception_types = (
       NetworkConstructionDependencyLoopException, LayerNotFound)
 
+    # These Exceptions always indicate incorrect construction, so fail directly instead of collecting them
+    fail_directly_exception_types = (DataNotFound, )
```
albertz, line 1052: Why the extra space after `,`?

Why is `LayerNotFound` not in here?

returnn/tf/layers/rec.py
```diff
@@ -1059,14 +1062,15 @@ class ConstructCtx:
...
-      def collect_exception(cls, layer_name):
+      def collect_exception(cls, exception, layer_name):
         """
         Collect most recent exception.
         Pretty generic exception handling but anything could happen.
         We don't do any output by default, as this could be very spammy,
         but we collect the traceback, in case we get some other error later.
         Then go on with the next get_layer.
 
+        :param Exception exception:
```
albertz, line 1073: Why is this needed? `value` below should be exactly the same.

returnn/tf/layers/rec.py
```diff
@@ -1152,8 +1161,8 @@ def construct(self, layer_name_):
         # noinspection PyBroadException
         try:
           self.__call__(layer_name_)
-        except Exception:
-          ConstructCtx.collect_exception(layer_name=layer_name_)
+        except Exception as e:
```
albertz, line 1164: Why?

returnn/tf/network.py
```diff
@@ -18,6 +18,12 @@
 from returnn.tf.util.basic import Data, DimensionTag, reuse_name_scope, VariableAssigner
 
 
+class DataNotFound(Exception):
+  """
+  Used for KeyErrors when accessing ExternData
```
albertz, line 23:
```suggestion
  When accessing non-existing ExternData data key
```
(Nothing about `KeyError` here. `KeyError` is totally irrelevant.)

returnn/tf/layers/rec.py
```diff
@@ -1082,6 +1086,11 @@ def collect_exception(cls, layer_name):
           else:
             out = StringIO()
             better_exchook.better_exchook(exc_type, value, tb, file=out)
+            if isinstance(value, fail_directly_exception_types):
+              assert False, (
```
albertz, line 1090: Do not use `assert False` but `throw Exception(...)` instead.

returnn/tf/layers/rec.py
```diff
@@ -1082,6 +1086,11 @@ def collect_exception(cls, layer_name):
...
+                  "Template construction failed with Exception:\n: %s\n"
+                  "Template construction failed with non-allowed Exception" % out.getvalue()
+              )
```
albertz, line 1093: The closing bracket should not be on an own line.

returnn/tf/layers/rec.py
JackTemaki: PyCharm gives a warning if there is no whitespace.

Yes, LayerNotFound probably can be moved.

returnn/tf/layers/rec.py
```diff
@@ -1043,10 +1043,13 @@ def _construct_template(self):
     from pprint import pformat
     from collections import OrderedDict
     from returnn.util.basic import StringIO
-    from returnn.tf.network import NetworkConstructionDependencyLoopException
+    from returnn.tf.network import NetworkConstructionDependencyLoopException, DataNotFound
     # The stack trace is not so interesting for these exceptions.
     skip_stack_trace_exception_types = (
-      NetworkConstructionDependencyLoopException, LayerNotFound)
+      NetworkConstructionDependencyLoopException)
```
albertz, line 1049:
```suggestion
      NetworkConstructionDependencyLoopException,)
```

returnn/tf/layers/rec.py
```diff
@@ -1082,6 +1085,11 @@ def collect_exception(cls, layer_name):
           else:
             out = StringIO()
             better_exchook.better_exchook(exc_type, value, tb, file=out)
+            if isinstance(value, fail_directly_exception_types):
+              raise Exception(
+                  "Template construction failed with Exception:\n: %s\n"
+                  "Template construction failed with non-allowed Exception "
+                  "while constructing layer: %s" % (out.getvalue(), layer_name))
```
albertz, line 1092: Indentation is off.
```suggestion
                "Template construction failed with Exception:\n: %s\n"
                "Template construction failed with non-allowed Exception "
                "while constructing layer: %s" % (out.getvalue(), layer_name))
```

returnn/tf/layers/rec.py
```diff
@@ -1082,6 +1085,11 @@ def collect_exception(cls, layer_name):
           else:
             out = StringIO()
             better_exchook.better_exchook(exc_type, value, tb, file=out)
+            if isinstance(value, fail_directly_exception_types):
+              raise Exception(
```
albertz, line 1089: Why actually do you throw a new exception?
Why not simply reraise the original exception, i.e. just `raise` or `raise value`?

JackTemaki: Is it possible to add text after the Exception message then?

albertz: Why do you want to edit it? It already has all the information. And when you pass it through, it will end up to the user as the final failing exception.

JackTemaki: The way it is now it will end up as the final exception, it is really just about the sentence...

```
[...]
    line: raise DataNotFound(
            "ExternData: unknown key %r. available keys: %s. config: %s" % (
              name, list(self.data.keys()), config_extern_data))
    locals:
      DataNotFound = <global> <class 'returnn.tf.network.DataNotFound'>
      name = <local> 'decoder_cells', len = 13
      list = <builtin> <class 'list'>
      self = <local> <ExternData data={'classes': Data(name='classes', dtype='int32', sparse=True, dim=54, available_for_inference=False, batch_shape_meta=[B]), 'data': Data(name='data', batch_shape_meta=[B,F|80]), 'speaker_label': Data(name='speaker_label', dtype='int32', sparse=True, dim=251, available_for_inferenc...
      self.data = <local> {'classes': Data(name='classes', dtype='int32', sparse=True, dim=54, available_for_inference=False, batch_shape_meta=[B]), 'data': Data(name='data', batch_shape_meta=[B,F|80]), 'speaker_label': Data(name='speaker_label', dtype='int32', sparse=True, dim=251, available_for_inference=False, batch_sh..., len = 7
      self.data.keys = <local> <built-in method keys of dict object at 0x7f082e1105c0>
      config_extern_data = <local> '<unknown>', len = 9
DataNotFound: ExternData: unknown key 'decoder_cells'. available keys: ['classes', 'data', 'speaker_label', 'ctx', 'padded_data_target', 'windowed_data_target', 'stop_token_target']. config: <unknown>

Template construction failed with non-allowed Exception while constructing layer: data:decoder_cells

Comment (albertz) [reply returnn/tf/layers/rec.py:1089]: All unhandled exceptions are always "non-allowed".
I still don't see how just passing through the exception (i.e. not handling it) is not just the same (but more clean).

Comment (JackTemaki) [reply returnn/tf/layers/rec.py:1089]: Okay, this is not super important for me so I can remove that.

Still, I need to figure out how to capture Exceptions and make them more verbose without raising a different exception / using `from` (because depending on the stack this makes it not nice to read). But we can discuss this offline.

File: returnn/tf/layers/rec.py [hunk 10]
```diff
@@ -1043,10 +1043,13 @@ def _construct_template(self):
...
     from returnn.util.basic import StringIO
-    from returnn.tf.network import NetworkConstructionDependencyLoopException
+    from returnn.tf.network import NetworkConstructionDependencyLoopException, DataNotFound
     # The stack trace is not so interesting for these exceptions.
     skip_stack_trace_exception_types = (
-      NetworkConstructionDependencyLoopException, LayerNotFound)
+      NetworkConstructionDependencyLoopException)
+
+    # These Exceptions always indicate incorrect construction, so fail directly instead of collecting them
+    fail_directly_exception_types = (DataNotFound, LayerNotFound)
```
patrick-wilken, line 1052: Are we certain that LayerNotFound is really never expected? Because e.g. in `Network.get_layer()` it is raised when a Layer does exist in the network dict, merely is not constructed yet. I think here doing template construction we should be fine, because the layers in the subnet are not expected to be constructed yet and any layer accessed outside should already be constructed. Just wanted to point out that it is not as simple as it sounds.

JackTemaki: This is why I did not move it, but at least all checks are passing...

returnn/tf/layers/rec.py
patrick-wilken: Oh, didn't see this.

returnn/tf/layers/rec.py
albertz: The construction code should always go through the dynamic `get_layer` (misleading name, `construct_layer` might have been better), so this should not really happen during template construction.

It might have happened at some point in the past. I don't exactly remember anymore. I actually wondered myself why I had it in `skip_stack_trace_exception_types`.

But when all checks are passing now (a lot of things got cleaned up here in the recent past), I think we should really just remove it.

returnn/tf/layers/rec.py
albertz: Not resolved yet?

returnn/tf/layers/rec.py
```diff
@@ -1082,6 +1085,8 @@ def collect_exception(cls, layer_name):
           else:
             out = StringIO()
             better_exchook.better_exchook(exc_type, value, tb, file=out)
+            if isinstance(value, fail_directly_exception_types):
+              raise
```
albertz, line 1089: You should move this two lines above. The `out` and `better_exchook` is not needed.

returnn/tf/layers/rec.py
JackTemaki: So it is okay like it is?

albertz: From my side, this seems good. Not sure if @patrick-wilken has further comments here.

patrick-wilken: Yes, should be fine.

albertz (approved):

PR #526, opened 2021-05-21
Fix optimizer selection code

JackTemaki:
Some bugfixing and extending features in the `updater.py` code:
 - Keras optimizers are now longer stored as wrapper function reference (caused problems due to inconsistencies of `type` vs `callable`)
 - Optimizer types can directly be used as "class" parameter -> allows to use external optimizers such as those from `tensorflow_addons` 
 - The documentation correctly displays also the Optimizers available via Keras (e.g. Nadam and Adamax)

### DIFF
--- a/returnn/tf/updater.py
+++ b/returnn/tf/updater.py
@@ -53,6 +53,8 @@ def _init_optimizer_classes_dict():
       continue
     if not issubclass(v, allowed_types):
       continue
+    if v is KerasOptimizerWrapper:
+      continue
     register_optimizer_class(v, name=name)
 
 
--- a/returnn/tf/updater.py
+++ b/returnn/tf/updater.py
@@ -17,11 +17,12 @@ import returnn.tf.util.basic as tf_util
 from returnn.tf.util.basic import tf_version_tuple, assert_min_tf_version, CustomUpdate, add_check_numerics_ops
 
 Optimizer = tf_compat.v1.train.Optimizer
+KerasOptimizer = None
 if tf_compat.v2:
   KerasOptimizer = tf_compat.v2.optimizers.Optimizer
 
 _OptimizerClassesDictInitialized = False
-_OptimizerClassesDict = {}  # type: typing.Dict[str, typing.Union[Optimizer, KerasOptimizer]]
+_OptimizerClassesDict = {}  # type: typing.Dict[str, typing.Type[typing.Union[Optimizer, KerasOptimizer]]]
 
 
 def _init_optimizer_classes_dict():
@@ -37,9 +38,9 @@ def _init_optimizer_classes_dict():
     except ImportError:  # TF 2
       pass
   allowed_types = (Optimizer,)
-  if tf_compat.v2:
+  if KerasOptimizer:
     potential_list += list(vars(tf_compat.v2.keras.optimizers).items())
-    allowed_types += (tf_compat.v2.keras.optimizers.Optimizer,)
+    allowed_types += (KerasOptimizer,)
   potential_list += list(globals().items())
   for name, v in potential_list:
     assert isinstance(name, str)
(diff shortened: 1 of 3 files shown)


JackTemaki:
@albertz was there any problem here about the PR?

returnn/tf/updater.py
```diff
@@ -17,9 +17,11 @@
 from returnn.tf.util.basic import tf_version_tuple, assert_min_tf_version, CustomUpdate, add_check_numerics_ops
 
 Optimizer = tf_compat.v1.train.Optimizer
+if tf_compat.v2:
+  KerasOptimizer = tf_compat.v2.optimizers.Optimizer
```
albertz, line 21:
```suggestion
KerasOptimizer = tf_compat.v2.optimizers.Optimizer if if tf_compat.v2 else None
```

returnn/tf/updater.py
```diff
@@ -47,25 +49,37 @@ def _init_optimizer_classes_dict():
...
     register_optimizer_class(v, name=name)
 
 
+def _check_valid_optimizer(optimizer_class):
+  """
+  :param type optimizer_class:
+  """
+  if tf_compat.v2:
+    assert (issubclass(optimizer_class, tf_compat.v2.keras.optimizers.Optimizer) or
+            issubclass(optimizer_class, Optimizer))
```
albertz, line 69: Do not use hanging indents.
Also, use `KerasOptimizer` directly in the `issubclass` check.
Also, is it not possible to directly specify the tuple in `issubclass`, like you would do for `isinstance`?
Also, check for `KerasOptimizer` instead of `tf_compat.v2`.
So:
```suggestion
  if KerasOptimizer:
    assert issubclass(optimizer_class, (KerasOptimizer, Optimizer))
```

returnn/tf/updater.py
```diff
@@ -517,6 +531,8 @@ def _create_optimizer(self, optimizer_opts):
       lr = lr * optimizer_opts.pop("learning_rate_multiplier")
     optimizer_opts["learning_rate"] = lr
     print("Create optimizer %s with options %r." % (optim_class, optimizer_opts), file=log.v2)
+    if tf_compat.v2 and issubclass(optim_class, tf_compat.v2.keras.optimizers.Optimizer):
```
albertz, line 534:
```suggestion
    if KerasOptimizer and issubclass(optim_class, KerasOptimizer):
```

returnn/tf/updater.py
```diff
@@ -926,9 +942,11 @@ def get_apply_grads_op(self, loss, var_list):
     return tf.group(*all_apply_grads)
 
 
-class KerasOptimizer(Optimizer):
+class KerasOptimizerWrapper(Optimizer):
```
albertz, line 945: Maybe the class name should also reflect that this is for internal use only, by putting an underscore.
```suggestion
class _KerasOptimizerWrapper(Optimizer):
```

returnn/tf/updater.py
```diff
@@ -926,9 +942,11 @@ def get_apply_grads_op(self, loss, var_list):
...
   """
   Wraps a TF optimizer into a standard TF optimizer.
+
+  For internal use only.
```
albertz, line 949: This comment is kind of obsolete when you directly use the class name `_KerasOptimizerWrapper` (where the `_` prefix exactly implies this).

returnn/tf/updater.py
albertz: Btw, with this, also later the code `allowed_types += (tf_compat.v2.keras.optimizers.Optimizer,)` should be changed to `allowed_types += (KerasOptimizer,)

albertz: Actually, is `tf_compat.v2.keras.optimizers.Optimizer` the same as `tf_compat.v2.optimizers.Optimizer`? If not, I think we should reflect this more clearly in the name.

albertz: Basically all further use of `tf_compat.v2.keras.optimizers.Optimizer` (or `tf_compat.v2.optimizers.Optimizer`) should be replaced by `KerasOptimizer`, and `if tf_compat.v2` should be replaced by `if KerasOptimizer`.

JackTemaki:
I think I got everything now...

albertz (approved):

PR #528, opened 2021-05-24
Add SearchSortedLayer

Zettelkasten:
Basically wraps around [tf.searchsorted](https://www.tensorflow.org/api_docs/python/tf/searchsorted).

The functionality should be done, and it's working in my own config as far as I can tell.
I still want to add some unit test, then I'll remove the draft status.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -6485,6 +6485,102 @@ class CondLayer(LayerBase):
     return layers
 
 
+class SearchSortedLayer(LayerBase):
+  """
+  Basically wraps :func:`tf.searchsorted`.
+
+  Takes a tensor `sorted_sequence` that is sorted along one axis, and a tensor `values`.
+  Will compute an output tensor with the same axes as `values`,
+  where each entry is the index of the value within the sorted sequence.
+  All (batch) axes of `sorted_sequence` except for the axis it is sorted along must be present in `values`.
+  """
+  layer_class = "search_sorted"
+  recurrent = True  # do not shuffle the sorted_sequence
+
+  def __init__(self, sorted_sequence, values, axis="T", side="left", **kwargs):
+    """
+    :param LayerBase sorted_sequence:
+    :param LayerBase values: search values
+    :param str axis: the axis along which `sorted_sequence` is sorted
+    :param str side: "left" or "right".
+      When one of the `values` exactly matches an element of the `sorted_sequence`,
+      whether to choose the lower or higher index.
+    """
+    super(SearchSortedLayer, self).__init__(**kwargs)
+    self.sorted_sequence = sorted_sequence  # e.g. [B,T]
+    self.values = values  # e.g. [B,F]
+    sorted_axis = sorted_sequence.output.get_axis_from_description(axis)  # = T
+    side = side.lower()
+    assert side in {"left", "right"}
+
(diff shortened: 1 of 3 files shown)


JackTemaki:
Out of curiousity, what are you doing with this layer exactly?

returnn/tf/layers/basic.py
```diff
@@ -6485,6 +6485,98 @@ def get_sub_layers(self):
...
+    :param LayerBase sorted_sequence:
+    :param LayerBase values: search values
+    :param str axis: the axis along which `sorted_sequence` is sorted
+    :param str side: "left" or "right".
+      When one of the `values` exactly matches an element of the `sorted_sequence`,
+      whether to choose the lower or higher index.
+    """
+    super(SearchSortedLayer, self).__init__(**kwargs)
+    self.sorted_sequence = sorted_sequence  # e.g. [B,T]
+    self.values = values  # e.g. [B',F']
```
JackTemaki, line 6511: I think `B'` is misleading in the comments, because the batch axis has to be exactly the same for both inputs.

In my understanding I would say just `B` for both to say it has to be the same axis.

Zettelkasten:
> Out of curiousity, what are you doing with this layer exactly?

I use it for an extension of [locality-sensitive hashed attention](https://arxiv.org/pdf/2001.04451.pdf): Basically you "embed" the keys/queries into a one-dimensional int each (a "hash"), with the hope that keys/queries close to each other get the same hash. Then you sort the key/query sequences according to the those hashes, and align the sorted queries to the sorted keys somehow (this is where I use this layer). This way, by only computing the attention for the keys with the same hash as the query, you avoid the quadratic memory usage.

returnn/tf/layers/basic.py
Zettelkasten: Hm yes, I see. `B` here is just an example, it could be an entire set of common axes (maybe permuted differently, and at different axes).
Is it clearer like I have it now?

JackTemaki:
> > Out of curiousity, what are you doing with this layer exactly?
> 
> I use it for an extension of [locality-sensitive hashed attention](https://arxiv.org/pdf/2001.04451.pdf): Basically you "embed" the keys/queries into a one-dimensional int each (a "hash"), with the hope that keys/queries close to each other get the same hash. Then you sort the key/query sequences according to the those hashes, and align the sorted queries to the sorted keys somehow (this is where I use this layer). This way, by only computing the attention for the keys with the same hash as the query, you avoid the quadratic memory usage.

Interesting... but this means you do hard choices, how does the gradient work in that case? Or do you use it only at inference time?

Zettelkasten:
> Interesting... but this means you do hard choices, how does the gradient work in that case? Or do you use it only at inference time?
> EDIT: The paper says it is for training, I guess I am not understanding something correctly then.

It is a bit weird indeed! We have it in such a way that a query can only attend to keys of the same hash class. This alignment that we compute on the sorted sequences is rather an memory efficient way to then implement this, but not so much a hard decision to which keys a query can attend. The only hard decision comes from introducing the hash classes.
But there you are completely right: There are no gradients for that decision.

Zettelkasten:
This PR should be ready now. I added a test, and also fixed a bug in `Data.get_sequence_mask_broadcast` (which caused my test to fail at first)

albertz (approved):

PR #533, opened 2021-05-28
tf.Engine.search: fixed output file write not considering use of seq_list_filter_file

npwynands:
I've encountered a bug when filtering out sequences from a dataset by their tags using the `seq_list_filter_file` feature of the [`HDFDataset`](https://returnn.readthedocs.io/en/latest/api/datasets.hdf.html#returnn.datasets.hdf.HDFDataset):
RETURNN's [`tf.Engine.search`](https://returnn.readthedocs.io/en/latest/api/tf.engine.html#returnn.tf.engine.Engine.search) method still assumed that the `out_cache` would contain an output for all the lines of the dataset. That caused `AssertionError`s and further `KeyError`s on the iteration over the `out_cache` entries by using `range(len(out_cache))`. The iteration fix should be safe. However, I'm not sure how important the assertions I deleted were.

### DIFF
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -2391,15 +2391,13 @@ class Engine(EngineBase):
       runner.num_steps, self.format_score(runner.score), self.format_score(runner.error)), file=log.v1)
     if output_file:
       assert out_cache
-      assert 0 in out_cache
-      assert len(out_cache) - 1 in out_cache
       if output_file_format == "txt":
-        for i in range(len(out_cache)):
+        for i in sorted(out_cache.keys()):
           output_file.write("%s\n" % out_cache[i])
       elif output_file_format == "py":
         from returnn.util.basic import better_repr
         output_file.write("{\n")
-        for i in range(len(out_cache)):
+        for i in sorted(out_cache.keys()):
           output_file.write("%r: %s,\n" % (seq_idx_to_tag[i], better_repr(out_cache[i])))
         output_file.write("}\n")
       else:


albertz:
But this check is there for a purpose to ensure nothing is missing. You should not just remove the check, but fix it. Or maybe only skip the check if `dataset.seq_tags_filter` is set.

patrick-wilken:
In #66 we asserted that filtering by length is disabled in search as we argued that when evaluating a test set you always want to get results for all sequences.
That's certainly true for academic test sets, but I can see that using a `seq_list_filter_file` you effectively want to define your own test set without changing the corpus files. @npwynands, is this vaguely what you are trying to do?
But then again from the output file, at least for "txt" output format, you don't know to which sequences the lines correspond if you filter sequences, right?
The code changes are not wrong of course, but I'm not sure whether it wouldn't be better to enforce creating a fixed, unfiltered test set. Because it can happen easily that to use filtering settings meant for training in search and then wonder why lines are missing in the output file.

albertz:
Normally we expect that there are no holes in the seq idxs, i.e. you get exactly all seq idxs from 0 to num seqs - 1. This is just a sanity check. And it's good to have it. We also e.g. check that the seq idxs are unique.

There are cases where this might not be fulfilled. Like `seq_list_filter_file`. Or also `epoch_split`. Or `max_seq_length`. Maybe other cases. Often this is actually not wanted by the user. So it would be bad to remove the check.

I think it doesn't make much sense for the "txt" output format anyway. It might make sense for the "py" output format to only search on some subset. For the "py" format, if the user explicitly sets `seq_list_filter_file`, I think we can assume that this is probably wanted. And due to the "py" format, it will be easy to see in the further pipeline when sth is missing.

npwynands:
No, I don't use the `seq_list_filter_file` feature to define myself a subset of my test set without changing the corpus file.
I use this feature to fetch single lines from my dev set and perform an analysis of a custom search algorithm I created.
That analysis cannot be easily done on the whole corpus at once, that's why the selection of single lines.
Disregarding this use case, I've found it useful to have the `seq_list_filter_file` feature for debug purposes...

albertz:
For such testing, debugging, analysis, I would probably recommend to write your own small script (probably you have that already) and then call the search more explicitly. E.g. there is `search_single` for simple testing. But otherwise, you could simply replicate what `search` is doing. There is not much magic to it.

However, if you want we can still go forward with this PR. As I suggested, we could disable these checks if `seq_list_filter_file` is set (but not in general as your change does it). Also, I would only allow that for the "py" output format.

albertz:
@npwynands ping? Do you plan to go forward with the PR as I suggested? Or close it?

npwynands:
No, currently I do not have the time to work on this, please excuse me.

Issue #519, opened 2021-05-21
Remove default "from":"data" layer argument, require it explicitly (or default `None` for layers where it make sense)


albertz:
#508

albertz:
This here is a case where we do not necessarily need a new `behavior_version` and instead could just print a deprecation warning.

JackTemaki:
close with #534

PR #534, opened 2021-05-31
Introduce behavior_version parameter

JackTemaki:
Also adds the first behavior_change by introducing: #519.

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -15,21 +15,19 @@ Each version change will be explicitly document here, and the documentation is i
 recommended usage with the latest ``behavior_version``, and not listing legacy/deprecated parameters.
 
 
-Current Version
+Version History
 ---------------
 
 Behavior Version 1:
-28.05.2021 - 0eb02dec9727f287b1305377c8e793353b0f0845
+28.05.2021
 
 Disallow not specifying "from" in layer definition dictionaries,
 thus making use of the hidden default "data" as layer input.
 
 "from" needs to be set explicitely now.
 
-See Issue `#519 <https://github.com/rwth-i6/returnn/issues/519`_
+See issue `#519 <https://github.com/rwth-i6/returnn/issues/519`_.
 
 
-History
--------
 
 
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -1,0 +1,35 @@
+.. _behavior_version:
+
+====================
+``behavior_version``
+====================
+
+Instead of maintaining different versions explictely, RETURNN has a ``behavior_version`` parameter to specify
+the default behavior of the configuration, as well as restricting the usage of legacy parameters/options.
+
+This "version" number is not related anyhow to new features or bugfixes,
+which are always added for all behavior versions.
+Still, it may change the default parameters of certain features such as of available layers.
+
+Each version change will be explicitly document here, and the documentation is indented to always reflect the
+recommended usage with the latest ``behavior_version``, and not listing legacy/deprecated parameters.
+
+
+Current Version
+---------------
+
+Behavior Version 1:
+28.05.2021 - 0eb02dec9727f287b1305377c8e793353b0f0845
+
+Disallow not specifying "from" in layer definition dictionaries,
+thus making use of the hidden default "data" as layer input.
+
+"from" needs to be set explicitely now.
+
+See Issue `#519 <https://github.com/rwth-i6/returnn/issues/519`_
+
+
+History
+-------
+
+
--- a/docs/configuration_reference/index.rst
+++ b/docs/configuration_reference/index.rst
@@ -12,6 +12,7 @@ Configuration Parameters
     :maxdepth: 1
     :titlesonly:
 
+    behavior_version.rst
     general_settings.rst
     training.rst
     optimizer_settings.rst
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -452,11 +452,10 @@ class LayerBase(object):
     which should be resolved.
     """
     from .basic import get_loss_class
-    if "from" not in d:
-      if BehaviorVersion.get() >= 1:
-        raise Exception('Missing "from" in layer definition: %s/%s' % (network.name, d.get("_name", "<UNKNOWN>")))
-      else:
-        log.print_deprecation_warning('Not specifying "from" in layer definitions is deprecated.', behavior_version=1)
+    BehaviorVersion.require(
+      condition="from" in d,
+      message='Missing "from" in layer definition: %s/%s' % (network.name, d.get("_name", "<UNKNOWN>")),
+      version=1)
     src_names = d.pop("from", ["data"])
     if not isinstance(src_names, (list, tuple)):
       src_names = [src_names]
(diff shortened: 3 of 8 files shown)


docs/configuration_reference/behavior_version.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Each version change will be explicitly document here, and the documentation is indented to always reflect the
+recommended usage with the latest ``behavior_version``, and not listing legacy/deprecated parameters.
+
+
+Current Version
+---------------
+
+Behavior Version 1:
+28.05.2021 - 0eb02dec9727f287b1305377c8e793353b0f0845
```
albertz, line 22: The hash (and date) will be wrong when this gets merged.
Not sure how to handle this.
Maybe just the approx date like in the changelog.

docs/configuration_reference/behavior_version.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Disallow not specifying "from" in layer definition dictionaries,
+thus making use of the hidden default "data" as layer input.
+
+"from" needs to be set explicitely now.
+
+See Issue `#519 <https://github.com/rwth-i6/returnn/issues/519`_
```
albertz, line 29:
```suggestion
See issue `#519 <https://github.com/rwth-i6/returnn/issues/519>`__.
```

docs/configuration_reference/behavior_version.rst
```diff
@@ -0,0 +1,35 @@
...
+
+
+History
```
albertz, line 32: What is this about?

docs/configuration_reference/index.rst
```diff
@@ -12,6 +12,7 @@ Configuration Parameters
     :maxdepth: 1
     :titlesonly:
 
+    behavior_version.rst
```
albertz, line 15: So this should be above all others?

returnn/tf/layers/base.py
```diff
@@ -452,6 +452,11 @@ def transform_config_dict(cls, d, network, get_layer):
     which should be resolved.
     """
     from .basic import get_loss_class
+    if "from" not in d:
+      if BehaviorVersion.get() >= 1:
+        assert False, "Missing \"from\" in layer definition: %s/%s" % (network.name, d.get("_name", "<UNKNOWN>"))
```
albertz, line 457:
```suggestion
        raise Exception('Missing "from" in layer definition: %s/%s' % (network.name, d.get("_name", "<UNKNOWN>"))
```

returnn/tf/layers/base.py
```diff
@@ -452,6 +452,11 @@ def transform_config_dict(cls, d, network, get_layer):
...
+      else:
+        log.print_deprecation_warning("Not specifying \"from\" in layer definitions is deprecated.", behavior_version=1)
```
albertz, line 459:
```suggestion
        log.print_deprecation_warning('Not specifying "from" in layer definitions is deprecated.', behavior_version=1)
```

albertz, line 459: The code looks a bit redundant.
I wonder whether we should introduce some generic helper, because code snippets like this will become common for other usages of `BehaviorVersion`.
Sth like this:

```suggestion
    BehaviorVersion.require(
      check="from" in d,
      message='%s/%s: missing "from"' % (network.name, d.get("_name", "<UNKNOWN>")),
      version=1)
```

Or so.

That `BehaviorVersion.require` function would basically do the logic you had before. Like:
```
@classmethod
def require(cls, check, message, version):
  if cls.get() >= version:
    raise Exception(...)
  else:
    log.print_exception_warning(...)
```

returnn/util/basic.py
```diff
@@ -180,6 +180,41 @@ def is_tensorflow_selected(cls):
...
 
+class BehaviorVersion:
+  """
+  Stores the global behavior_version
+
+  The version will be set after the config is defined at __main__.init_config() or Engine.__init__()
+  """
+
+  _behavior_version = 0  # type: typing.Optional[int]  # the selected global behavior_version, default 0 for tests
+  _is_set = False  # type:bool  # store if the behavior version was already set (e.g. re-check in engine init)
```
albertz, line 191:
```suggestion
  _is_set = False  # type: bool
```
(I don't like to have spammy documentation which is obvious.)

returnn/util/basic.py
```diff
@@ -180,6 +180,41 @@ def is_tensorflow_selected(cls):
...
 
 
+class BehaviorVersion:
+  """
+  Stores the global behavior_version
+
+  The version will be set after the config is defined at __main__.init_config() or Engine.__init__()
+  """
+
+  _behavior_version = 0  # type: typing.Optional[int]  # the selected global behavior_version, default 0 for tests
```
albertz, line 190:
```suggestion
  _behavior_version = 0  # type: typing.Optional[int]
```

docs/configuration_reference/behavior_version.rst
JackTemaki: I thought to only display the latest one above and put the rest under history

docs/configuration_reference/index.rst
JackTemaki: Yes

docs/configuration_reference/behavior_version.rst
albertz: I don't quite understand. Every new `behavior_version` will change one thing. Most of them will be independent from each other. I thought you will just describe the change above.

docs/configuration_reference/index.rst
albertz: Hm I do not agree. This is definitely not the most important thing a newcomer would be interested in when learning about RETURNN (which is the purpose of the documentation).

docs/configuration_reference/behavior_version.rst
JackTemaki: Ok I will just call it History and put everything in order below it.

docs/configuration_reference/behavior_version.rst
JackTemaki: I will add it with the date, and we can optionally add the hash later when it is part of the master branch.

docs/configuration_reference/index.rst
JackTemaki: The configuration reference is not about learning RETURNN, but for looking up configuration parameters. And this will be one of the most important parameters, so I think it should be first. (e.g. when you write a new config, you need to think about setting this first before you do the rest, because other parameters might depend on this one).

albertz: But I would rather say it's one of the least important parameters. In the remaining documentation, I would document the most recent behavior version always first, and if at all mention earlier behaviors when needed.

But also, I think this will be some feature which we would/should only rarely use. I expect most people will copy some example config or from someone else which just has this set to some recent version and otherwise will not care about it.

Also, I don't think people will need to look this up so often. Or if they need to, then this would be bad, because it should be mostly straight-forward. If it is not straight-forward, we did sth wrong.

But I don't really have a strong opinion on this. If you really think this should be first, then do it. But maybe someone else should also comment his/her opinion. @patrick-wilken @christophmluscher @Zettelkasten @jotix16 @Atticus1806 maybe? In the end, this is not really so important.

Atticus1806: While I don‘t use the section often, I would personally put it somewhere in the upper parts of the list. After version changes it might be searched for quite a bit.

albertz (approved):

PR #535, opened 2021-06-02
small fix in TransposedConvLayer

vieting:
I assume it was intended this way as the default value for `remove_padding` is 0 and later `len(remove_padding)` is checked, which crashes with the default value right now.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4252,7 +4252,7 @@ class TransposedConvLayer(_ConcatInputLayer):
     padding = padding.lower()
     if strides is None:
       strides = filter_size
-    if not isinstance(remove_padding, (tuple, int)):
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -4252,7 +4252,7 @@ def __init__(self, filter_size, activation, strides=None,
     padding = padding.lower()
     if strides is None:
       strides = filter_size
-    if not isinstance(remove_padding, (tuple, int)):
+    if not isinstance(remove_padding, (tuple, tuple)):
```
albertz, line 4255: I assume this was supposed to mean:
```suggestion
    if not isinstance(remove_padding, (list, tuple)):
```

vieting: Oh yes of course, sorry...

albertz (approved):

PR #538, opened 2021-06-08
Explicit definition of the batch norm parameters in the BatchNorm Layer

Atticus1806:
This PR copies the parameters used for `batch_norm()` in the `LayerBase` directly into the `BatchNorm` layer, instead of them being extracted from the `kwargs` in the `BatchNorm` layer. This change should not make functional differences and is in order to make the generated `BatchNorm` layer in the `returnn_common` interfaces easier to use. This change enables directly defining the parameters of the layer in the new interface instead of handing over a dict of strings.

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -4604,6 +4604,14 @@ def test_automatic_seq_lengths2():
     assert out_lens is in_data.size_placeholder[0]
 
 
+def test_batch_norm_args():
+  from returnn.tf.layers.basic import BatchNormLayer
+  from returnn.util.basic import getargspec
+  batch_norm_args = getargspec(BatchNormLayer.batch_norm).args[2:] #drop self and data
+  layer_args = getargspec(BatchNormLayer.__init__).args[1:] #drop self
+  assert batch_norm_args == layer_args #different arguments in BatchNormLayer and LayerBase.batch_norm()
+
+
 if __name__ == "__main__":
   try:
     better_exchook.install()
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -4607,9 +4607,9 @@ def test_automatic_seq_lengths2():
 def test_batch_norm_args():
   from returnn.tf.layers.basic import BatchNormLayer
   from returnn.util.basic import getargspec
-  batch_norm_args = getargspec(BatchNormLayer.batch_norm).args[2:] #drop self and data
-  layer_args = getargspec(BatchNormLayer.__init__).args[1:] #drop self
-  assert batch_norm_args == layer_args #different arguments in BatchNormLayer and LayerBase.batch_norm()
+  batch_norm_args = getargspec(BatchNormLayer.batch_norm).args[2:]  # drop self and data
+  layer_args = getargspec(BatchNormLayer.__init__).args[1:]  # drop self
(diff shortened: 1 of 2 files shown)


albertz:
I wonder, don't you get warnings about unused local vars? Or is it clever enough to see and understand the use of `locals()`?

Also, I don't like that we need to copy all the default values (which are bad in certain ways). We might later change the default values and then this could become inconsistent, which is bad. One possible solution is to use `NotSpecified` as default here and check for that.

Atticus1806:
So for me Pycharm does not mark them as unused. I assume, since locals() is called, it marks all local variables as used. 

I see your reasoning with the default values and agree. Also I dont see why this would have impact on the goal of this PR so I will change that. In the long run, this might be a bit confusing, so maybe we should put a reference to the function where the default parameters are set into the documentation, so someone who sees the layer for the first time knows where to look. I will write a small addition to the comments, feel free to extend or change it.

As an addition: Do you think it would make sense to write a test checking wether the input to the `BatchNormLayer` `__init__` equals the input to the `batch_norm` function? Otherwise we maybe should extend the documentation also in the `batch_norm` function so changes made to variables in the function are also made in the `BatchNormLayer`.

albertz (approved):

Atticus1806:
In my opinion it makes sense to have the parameters explicitly defined in the `BatchNormLayer` because of readability and documentation. I think otherwise its tedious and confusion when wanting to use BatchNorm as this layer to search for the possible parameters. So even though this PR was written for the wrapper generator, thinking about it I would say it also makes sense for usability reasons.

albertz:
Ok.

Should we merge now, or you want to add the test first?

Atticus1806:
I would try writing a test today. Thus I would suggest waiting until then. If not problems occure I will append the test to the PR.

Atticus1806:
I added the test for the args. So now everything should be complete.

tests/test_TFNetworkLayer.py
```diff
@@ -4604,6 +4604,14 @@ def test_automatic_seq_lengths2():
     assert out_lens is in_data.size_placeholder[0]
 
 
+def test_batch_norm_args():
+  from returnn.tf.layers.basic import BatchNormLayer
+  from returnn.util.basic import getargspec
+  batch_norm_args = getargspec(BatchNormLayer.batch_norm).args[2:] #drop self and data
```
albertz, line 4610:
```suggestion
  batch_norm_args = getargspec(BatchNormLayer.batch_norm).args[2:]  # drop self and data
```

tests/test_TFNetworkLayer.py
```diff
@@ -4604,6 +4604,14 @@ def test_automatic_seq_lengths2():
...
+  layer_args = getargspec(BatchNormLayer.__init__).args[1:] #drop self
```
albertz, line 4611:
```suggestion
  layer_args = getargspec(BatchNormLayer.__init__).args[1:]  # drop self
```

tests/test_TFNetworkLayer.py
```diff
@@ -4604,6 +4604,14 @@ def test_automatic_seq_lengths2():
...
+  assert batch_norm_args == layer_args #different arguments in BatchNormLayer and LayerBase.batch_norm()
```
albertz, line 4612:
```suggestion
  assert batch_norm_args == layer_args  # different arguments in BatchNormLayer and LayerBase.batch_norm()
```

albertz (approved):

Issue #522, opened 2021-05-21
Fix batch norm default parameters


albertz:
#508

New potential defaults:

* `momentum=0.1`
* `update_sample_only_in_training=True`
* `delay_sample_update=True`
* `param_version=2` (with #898)
* `masked_time` mandatory

See also other frameworks for defaults. Some references:

* [PyTorch-to-RETURNN comment](https://github.com/rwth-i6/pytorch-to-returnn/blob/0d55be95da12fbe3beaba8205a760aa5f18de2db/pytorch_to_returnn/torch/nn/modules/batchnorm.py#L90)
* [PyTorch BatchNorm1d](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html)

albertz:
One open question here is on our `masked_time` option, which is currently `True` by default. Which is the correct thing to do. This flattens the tensor, i.e. removes any padded frames, and calculates all the statistics only on the valid frames.

However, that option makes it slower and requires more memory. (I don't really have numbers on this. Some benchmarks would be nice...)

Also, we currently do not use the fused TF op for batch norm, which would make it faster. Although we could also use it on the flattened tensor. I don't know if there is any interplay between the two things.

Also note: All other frameworks I know do not use this padding, and always use a simple fused op. E.g. PyTorch (https://github.com/pytorch/pytorch/issues/70529) and Keras (https://github.com/tensorflow/tensorflow/issues/30161).

albertz:
One note on the confusion of `momentum`: This probably comes from the different meaning of momentum.

RETURNN batch norm:
```
mean_new = (1 - momentum) * mean_old + momentum * mean_cur_batch
```

[PyTorch batch norm](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html):
```
mean_new = (1 - momentum) * mean_old + momentum * mean_cur_batch
```

[Keras batch norm](https://keras.io/api/layers/normalization_layers/batch_normalization/):
```
mean_new = momentum * mean_old + (1 - momentum) * mean_cur_batch
```

[PyTorch optimizer](https://pytorch.org/docs/stable/generated/torch.optim.SGD.html):
```
v_new = momentum * v_old + g
```

[Keras optimizer](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/SGD):
```
v_new = momentum * v_old - learning_rate * g
```

albertz:
[...]

PR #539, opened 2021-06-10
batch norm, cleanup unused args

albertz:
At least I assume they are unused.
Likely a user would not provide a tf.Tensor in a config.
Or the specified type was wrong
and this was supposed to be a float.
Anyway, I still don't see
how this would have been used potentially.
Maybe this was a relict from the Theano code conversion.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -1190,11 +1190,9 @@ class LayerBase(object):
   def batch_norm(self, data,
                  use_shift=True, use_std=True, use_sample=0.0, force_sample=False,
                  momentum=0.99, epsilon=1e-3,
-                 sample_mean=None, sample_variance=None,
                  update_sample_only_in_training=False,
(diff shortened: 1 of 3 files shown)


albertz:
Please check if this was used in any existing config, or whether this cleanup makes sense.

@Atticus1806 can you also check this?

albertz:
Ah, I see, this was used at least in the test. The test should be rewritten anyway...

JackTemaki:
As far as I see it, there is no obvious application for these parameters... so I guess they can be removed.

Atticus1806:
I agree with @JackTemaki , in the config I am porting to the `returnn_common` style they are not used. Now that we talk about it: If I remember correctly there were certain default values like momentum different in the `Returnn` default values than in the original paper. Maybe if we are modifying the args now we could also change these values?

albertz:
> [changing] default params

This is #522.

Removing unused options should not make any difference to anyone.

PR #540, opened 2021-06-10
cleanup/remove OfficialResNetLayer

albertz:
This should never have been directly in this repo.
Such a layer can be added by a user
via a user config.


JackTemaki:
It could also just be moved to `extra.py` layers or so, but it is definitely wrong in `basic.py`. But if you delete it here, shouldn't be the other code in the `extra` folder deleted as well?

albertz:
> It could also just be moved to `extra.py` layers or so, but it is definitely wrong in `basic.py`. But if you delete it here, shouldn't be the other code in the `extra` folder deleted as well?

Let's discuss these things individually. Each custom stuff might have some different history and reason (or not) to be there.

In this case, the `OfficialResNetLayer` was just used by one of my Hiwi for a custom test case. Only once. I wonder why we allowed to accept this layer already back then. I think we were less strict. And I doubt that anyone is really using this. It's also very much not how you would and should implement it.

JackTemaki:
> Let's discuss these things individually.

I thought this is directly related to each other (`Model` is only used here in this Layer).  But yes, I agree that the Layer can be removed.

JackTemaki (approved):

PR #543, opened 2021-06-13
Add test for optimize_out_slice_nd

jotix16:
This is an example case which shows that `slice_nd` layer doesn't get properly optimized out of the loop.

This pull request is meant to fix that.


tests/test_TFNetworkRecLayer.py
```diff
@@ -3598,6 +3598,21 @@ def test_reclayer_optimize_out_access_split():
...
+    from_="position",
+    other_subnet_layers={
+      "window": {"class": "slice_nd", "from": "base:encoder", "start": "data:source", "size": None, "min_size": 1, "is_output_layer": True},
+      "encoder_reduced": {"class": "reduce", "mode": "sum", "axis": "T", "from": ["base:encoder"], "is_output_layer": True}},
+    shared_base_net={
+      "encoder": {"class": "copy", "from": ["data"], "is_output_layer": True},
+      "position": {"class": "eval", "from": ["encoder"], "is_output_layer": True,
+                "eval": "tf.zeros(tf.shape(source(0, enforce_batch_major=True, auto_convert=False))[:-1], dtype=tf.dtypes.int32)",
+                "out_type": {"batch_dim_axis": 0, "time_dim_axis": 1, "shape": (None,),
+                             "sparse": True, "dtype": "int32", "dim": None}}})
```
albertz, line 3613: Do not indent it like that. Always use a new line + 2 space indent instead of these hanging indents.

tests/test_TFNetworkRecLayer.py
```diff
@@ -3598,6 +3598,21 @@ def test_reclayer_optimize_out_access_split():
...
 
+def test_reclayer_optimize_out_slice_nd():
+  check_reclayer_optimize_out(
+    {"class": "linear", "activation": None, "from": ["encoder_reduced"]},
+    from_="position",
+    other_subnet_layers={
+      "window": {"class": "slice_nd", "from": "base:encoder", "start": "data:source", "size": None, "min_size": 1, "is_output_layer": True},
+      "encoder_reduced": {"class": "reduce", "mode": "sum", "axis": "T", "from": ["base:encoder"], "is_output_layer": True}},
+    shared_base_net={
+      "encoder": {"class": "copy", "from": ["data"], "is_output_layer": True},
```
albertz, line 3609: Do not use `["data"]` if this is just a single source. Just use `"data"`.

tests/test_TFNetworkRecLayer.py
```diff
@@ -3598,6 +3598,21 @@ def test_reclayer_optimize_out_access_split():
...
+      "position": {"class": "eval", "from": ["encoder"], "is_output_layer": True,
+                "eval": "tf.zeros(tf.shape(source(0, enforce_batch_major=True, auto_convert=False))[:-1], dtype=tf.dtypes.int32)",
```
albertz, line 3611: I don't think just zeros is a good test case. Use e.g. `tf.random.uniform`.

jotix16:
https://github.com/rwth-i6/returnn/blob/6410df5eeb26cd9d77635a6e2af5a8330083198b/returnn/tf/layers/basic.py#L940 

already offers the logic of `slice_nd` but is a little more general.

I plan to use `GatherLayer` as base class and add the required constraints/checks for `slice_nd` in its `__init__()`.

albertz:
I don't think GatherLayer is a good base.

jotix16:
I've added the tf implementation of the new `slice_nd` in `util/basic` and some tests. I haven't yet cleaned the old ones out.


Before implementing the RETURNN layer, I wanted to clarify if we want to allow a `slice_axis` as input param to the function, similar to the `GatherLayer`.  If we don't, the user has to make sure that the axis of `input` and `start` have the same order. 

I.e.

``` python
"""
  :param tf.Tensor input: shape (B, T1, ..., Tn, D)
  :param tf.Tensor start: shape (B,T1 .., Tn-1), int32 which automatically indicates n as the slice-axis
"""
```

returnn/tf/layers/basic.py
```diff
@@ -1176,6 +1176,114 @@ def transform_config_dict(cls, d, network, get_layer):
       d["position"] = get_layer(d["position"])
 
 
+class SliceNdLayer2(_ConcatInputLayer):
```
albertz, line 1179: Why `2`? Never do that. Extend/generalize the existing `SliceNdLayer`.

jotix16: Will change the name and clean the rest once it is working. Let them both for comparison reasons.

tests/test_TFNetworkRecLayer.py
```diff
@@ -3598,6 +3598,31 @@ def test_reclayer_optimize_out_access_split():
...
+    enc_time_dim = enc_shape[enc.time_dim_axis]
+    return tf.random.uniform(enc_shape[:-1], 0, enc_time_dim-2, dtype=tf.dtypes.int32)  #[B,T]
+
+  check_reclayer_optimize_out(
+    {"class": "linear", "activation": None, "from": "my_layer"},
+    from_="position",
+    other_subnet_layers={
+      "my_layer": {"class": "gather_nd", "from": "base:data", "position": ":i"},
+      "window": {"class": "slice_nd2",    # no_opt: [B,4,D], opt: [B,T,4,D]
+                 "from": "base:data", "start": "data:source", "size": 4, "is_output_layer": True},
```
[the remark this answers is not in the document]
albertz: > I am not sure what **input to expect** when the `slice_nd2` layer gets **optimized-out**.

I don't really understand how this can not be clear? This should be totally obvious/straightforward or not?

`data:source` is the rec layer input. Inside the loop, it is [B]. Outside the loop, it is [T,B]. (It is `position` here.)

`base:data` is independent from the loop / rec layer, so obviously nothing changes.

> It is related to #550.

I don't really understand. #550 is not even a bug but was just a question.

> I would suggest adding an option to control the behaviour of the layers (e.g. the ones comming from base network) when being pulled of `rec`.

I don't understand. What behavior are you speaking about? There is only one straightforward behavior.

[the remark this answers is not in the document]
jotix16: > `data:source` is the rec layer input. Inside the loop, it is [B]. Outside the loop, it is [T,B]. (It is position **here.)**

This is clear.

> `base:data` is independent from the loop / rec layer, so obviously nothing changes.

Exactly. But, instead of `base:data`,  the input could be loop-dependent and we would get an extra time axis when pulled out. (first case above)
So, I am saying that depending on whether the input comes from inside the loop or not we may get 2 different inputs and `SliceNdLayer` has to handle both properly. 

#550 was indeed a question. 
My point is very simple, sorry that I am not being clear. **If you pull out a layer from the loop, you would expect its inputs to be accumulated too(extra time dim), no matter if they are loop-dependent or not.** Or at least have a parameter to force it.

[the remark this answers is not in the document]
albertz: > > `base:data` is independent from the loop / rec layer, so obviously nothing changes.
> 
> Exactly. But, instead of `base:data`, the input could be loop-dependent and we would get an extra time axis when pulled out. (first case above)

No. The `base:` prefix means it comes from the base, so it is always independently. This is independent how it is optimized (in loop or without).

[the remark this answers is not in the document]
jotix16: Is there any way how to make it dependent? For example copy it into an intermediate layer or so?

Otherwise, I don't know how making it work for all cases is possible. Maybe making sure that the `input` has one more time axis than `start` or `tf.tile` it otherwise?

albertz:
I don't really understand why you would want that? I would assume this
layer (like GatherNdLayer) only makes sense like this here. Do not care
about other cases, esp not if they don't really make sense.


Mikel Zhobro ***@***.***> schrieb am Fr., 25. Juni 2021,
15:17:

returnn/tf/layers/basic.py
```diff
@@ -1176,6 +1176,143 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    input_data = self.input_data.copy_as_batch_major()
+    start = start.output.copy_as_batch_major()
+
+    # make sure axis of start are in input
+    is_equal_opts = dict(ignore_feature_dim=True, allow_same_spatial_dim=True, broadcast_matches=True)
+    for start_axis in range(start.batch_ndim):
+      assert input_data.get_dim_tag(start_axis).is_equal(start.get_dim_tag(start_axis), **is_equal_opts)
+
+    # Handle the case when layer is pulled out of rec loop but the input hasn't change
+    if self.optimized_out_of_loop_and_unchanged_input(input_data, start):
```
albertz, line 1233: There should be nothing about "optimized-out-of-loop" in this layer.
It should not matter at all whether this is in a rec layer.
The layer just gets some input and should do the correct thing.

tests/test_TFNetworkRecLayer.py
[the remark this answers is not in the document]
albertz: Why would you want that? It doesn't really make sense. Also you do not have this case. Why do you want to cover cases which you do not need? Keep it as simple as possible.

returnn/tf/layers/basic.py
albertz: But you should not reimplement the whole function/layer but just extend it. I would assume this to be only a small or even almost minimal change to make it work.

albertz:
Can you post here what the actual problem/error is when you use these test cases with the original `slice_nd`? It is not really clear to me why there is actually a problem. (Just post the error with stack trace.)

jotix16:
The [Error strace](https://github.com/rwth-i6/returnn/runs/2930341551) you asked. The thing is that the `slice_nd` we had, expects `start` to have only one batch axis `[B]`, but if `slice_nd` gets pulled out, so does `start` and instead of shape `[B]`, start has shape `[B,T]`. So, it requires a loop over T.  That's what I added with this commit.

returnn/tf/layers/basic.py
jotix16: Well, I explained above the discrepancy. 

There exists both the possibility that the input layer to `slice_nd` gets pulled out or that it comes from `base:network` and stays the same, when `slice_nd` gets pulled out.

I can rename the function to something like: `input_needs_tiling`

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
-    assert x.time_dim_axis == 1, "currently only time-axis==1 supported"
+    start = start.output.copy_as_batch_major()
+
+    # make sure axis of start are in input
+    is_equal_opts = dict(ignore_feature_dim=True, allow_same_spatial_dim=True, broadcast_matches=True)
+    for start_axis in range(start.batch_ndim):
+      assert x.get_dim_tag(start_axis).is_equal(start.get_dim_tag(start_axis), **is_equal_opts)
+
+    # Handle the case when layer is pulled out of rec loop but the input hasn't change
+    if self.optimized_out_of_loop_and_unchanged_input(x, start):
```
albertz, line 891: Change the name of the function and the comment. This layer is not related to the rec loop (`RecLayer`) in any way. No comment should mention anything about rec loop here.

I don't really understand what you are actually testing here.

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
+      # add an axis after the last start axis and tile the input Tn-1 times: [B,T0,..,Tn-1,D] -> [B,T0,..,Tn-1,Tn-1,D]
+      tag = start.get_dim_tag(-1)
+      x = x.copy_add_dim_by_tag(tag, True, start.batch_ndim)  # tiles the input
```
albertz, line 894: I don't know what the arguments do here. Please use args by keywords (for all except the `tag` maybe).

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
+
+    start = start.get_placeholder_as_batch_major()
```
albertz, line 896: Do not change the type of `start` (before `Data`, now `tf.Tensor`). Call it `start_tensor` or so.

But anyway, why do you need it? I don't think you really need it. Just access `start.placeholder` directly.

Also, do not use `get_placeholder_as_batch_major`. Rather, use `copy_as_batch_major` or so. But you actually already have that, so it is already batch-major. So no need to do that again.

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
     seq_lens = x.get_sequence_lengths() if x.is_time_axis_dynamic() else None
-    self.start = start
-    assert start.output.have_batch_axis() and start.output.batch_shape == (None,)
-    start = start.output.get_placeholder_as_batch_major()
+    slice_axis = len(list(start.shape)) - 1  # slice_axis w/o batch
```
albertz, line 898: Do not access the `shape` of the tensor but rather from the `Data` instance. Also, use `batch_ndim` or so instead.

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
-      mask = expand_multiple_dims(mask, list(range(2, x.batch_ndim)))
+      mask = tf.greater_equal(
+        tf.range(size)[None, :] + tf.expand_dims(start, axis=-1), seq_lens[:, None])  # (B,T1,..,Tn)
+      mask = expand_multiple_dims(mask, list(range(slice_axis + 2, x.batch_ndim)))  # (B,T1,..,Tn,1,..)
       slices = where_bc(mask, tf.zeros_like(slices), slices)
+
     self.output.size_placeholder = x.size_placeholder.copy()
     if isinstance(size, tf.Tensor):
-      self.output.size_placeholder[0] = tf.maximum(seq_lens - tf.reshape(start, tf.shape(seq_lens)), 0)
+      self.output.size_placeholder[slice_axis] = size
```
albertz, line 917: This is wrong. `size` is just a scalar. But you need a vector of shape [B] here. What's wrong with the old code?

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
+      tag.set_tag_on_size_tensor(self.output.size_placeholder[slice_axis])
     else:
       assert isinstance(size, int)
-      self.output.size_placeholder.pop(0, None)  # static time axis
+      self.output.size_placeholder.pop(slice_axis, None)  # static time axis
+
     self.output.placeholder = slices
 
+  @classmethod
+  def optimized_out_of_loop_and_unchanged_input(cls, input_data, start):
```
albertz, line 929: As explained above (and before), this name must be changed to what this function actually does/checks.

returnn/tf/layers/basic.py
```diff
@@ -854,52 +854,85 @@ class SliceNdLayer(_ConcatInputLayer):
...
+    """
+    :rtype: bool
+    The idea is to check that the axis after the last common axis is a feature axis instead of spatial.
```
albertz, line 932: I don't really understand this. Why do you check this? Why is this relevant?
Also, what does this mean, "the idea"? Is this what this function returns, or how does this idea relate to what this function returns? If this is what the function returns, it should be like `:returns: True iff axis after last ... is ... instead of ...` or so.

returnn/tf/util/basic.py
```diff
@@ -3633,44 +3633,29 @@ def windowed_nd(source, window_size, window_left=None, window_right=None,
...
+    shape = tf.shape(x)
+    len_common_dims = len(start.shape)  # nr of common dims
+    slice_dim = shape[len_common_dims]  # dim of axis to be sliced
+    # assert size < slice_dim, "Slice size cannot be bigger than the dimension to be sliced."
+    # Create indexes for the slices where slice_idx[B,T1 .., Tn-1] = start[B,T1 .., Tn-1] + range(size)
+    # (B,T1 .., Tn-1, size)
+    slice_idx = tf.tile(tf.expand_dims(start, -1), [1] * len_common_dims + [size]) + tf.range(size)
+    mask = tf.logical_or(tf.greater(slice_idx, slice_dim - 1), tf.less(slice_idx, 0))  # (B,T1 .., Tn-1, size)
+    slice_idx = tf.clip_by_value(slice_idx, 0, slice_dim - 1)  # cliped slice idx
+    res = tf.gather(x, slice_idx, axis=len_common_dims, batch_dims=len_common_dims)
```
albertz, line 3656: I think `tf.gather` with `axis` and `batch_dims` is only supported in later TF versions (I don't remember since what version, can you/someone check?).

I'm not sure anymore which is the min TF version we want to support (I think we documented this somewhere; can someone check? @patrick-wilken ? @JackTemaki ?).

Maybe we need our own wrapper for `gather`. Or you use the more generic `gather_nd` (which might be slower for this case though).

returnn/tf/util/basic.py
```diff
@@ -3633,44 +3633,29 @@ def windowed_nd(source, window_size, window_left=None, window_right=None,
...
-    # (B, size, ...), we assume time-axis is/was 1
-    new_shape = [shape[0], size] + shape[2:]
-
-    # zero-pad
-    mask_bc = expand_multiple_dims(mask, [-1] * (slices.get_shape().ndims - 1))
-    slices = where_bc(mask_bc, tf.zeros_like(slices), slices)
-
-    slices = tf.reshape(slices, new_shape)  # (B, size, ...)
-    return slices
+    shape = tf.shape(x)
```
albertz, line 3647: Use `get_shape` instead (as before).

returnn/tf/layers/basic.py
jotix16: If we only expect `input_data` comes from `base_network`, no check is required. We can always assume that the `input_data` stays the same both when being pulled out of the loop or not.

Should I follow this logic?

returnn/tf/layers/basic.py
jotix16: >> Call it start_tensor or so.
>> But anyway, why do you need it? I don't think you really need it. Just access start.placeholder directly.

Right. Will do just that.

returnn/tf/layers/basic.py
jotix16: > But you need a vector of shape [B] here. What's wrong with the old code?

We cannot use `start` directly to calculate the size as it has shape `[B,T0,..]` instead of `[B,]`.
So I am not sure how to set `self.output.size_placeholder`.

returnn/tf/layers/basic.py
jotix16: It is now called `input_comes_from_base_network`.

returnn/tf/util/basic.py
jotix16: The arguments are present in both [tf_2.3.0](https://www.tensorflow.org/versions/r2.3/api_docs/python/tf/gather) and in [tf_2.4.0](https://www.tensorflow.org/versions/r2.4/api_docs/python/tf/gather). So it should be fine?

returnn/tf/layers/basic.py
jotix16: i reformulated it. Should be more clear now.

``` python
    """
    The idea is to check if the axis after the last common axis is the feature axis instead of another spatial axis.
    Because the input_data should normally have one extra spatial axis compared to start.
    """
```

returnn/tf/layers/basic.py
jotix16: What is the meaning of `self.output.size_placeholder` if we have more than 1 spatial axis? Should any element of `self.output.size_placeholder` have size `(B,)`?

Zettelkasten: `size_placeholder` is a dict, mapping each spatial axis (counted without batch dim) to such `[B]` tensor.
So if you have a tensor `[B,T1,T2,F]`, then for each batch entry `b` you find it's length of axis `T1` in `size_placeholder[0][b]`, and the one for `T2` in `size_placeholder[1][b]`.
However with this it's not possible that lengths depend on anything else then the batch entry, i.e. you cannot model that the some batch entry should have a different length in axis `T2` for different time steps `t1`.

jotix16: > you cannot model that the same batch entry should have a different length in axis T2 for different time steps t1

This is what we have here, though.


What is `self.output.size_placeholder` used for anyways? Does it cause any problems if set wrong?

Zettelkasten: The `size_placeholder` is mostly used for masking, e.g which entries to ignore when calculating the loss, or reducing a sequence (e.g. take the average/min/max over the sequence), or also for layers which concat in time and such.
Many layers consider this, if this is set wrong then weird things can happen and your calculations will be wrong and depend on batching.
So yes, I'd say setting this is kind of important.

I don't know a way to set it how you want it. Maybe @albertz knows more?

PR #545, opened 2021-06-15
Expanded SelfAttentionLayer to be more generic

mattiadg:
added attention_mask parameter, which is a superset of attention_left_only. It can set the existing behavior and additionally receive a function which maps the energy tensor to a mask.
This way, it's possible to pass arbitrary masks to the self-attention operation.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -5819,7 +5819,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
   def __init__(self, num_heads, total_key_dim,
                key_shift=None,
                forward_weights_init="glorot_uniform", attention_dropout=0.0,
-               attention_left_only=False, attention_mask="length", initial_state=None,
+               attention_left_only=False, attention_mask=None, initial_state=None,
                restrict_state_to_last_seq=False, state_var_lengths=None, **kwargs):
     """
     :param int num_heads:
@@ -5830,9 +5830,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
     :param str forward_weights_init: see :func:`TFUtil.get_initializer`
     :param float attention_dropout:
     :param bool attention_left_only: will mask out the future. see Attention is all you need.
-    :param str|Callable attention_mask: mask to apply to input sequence. Allowed strings:
-      "length" masks the positions outside the original length, "left_only" masks the future.
-      If a callable, it must be a function that receives only the input sequence and returns the mask.
+    :param Callable attention_mask: a function that receives the input sequence in input and returns an attention mask.
     :param str|float|int|None initial_state: see RnnCellLayer.get_rec_initial_state_inner().
     :param bool restrict_state_to_last_seq: see code comment below
     :param None|tf.Tensor|()->tf.Tensor state_var_lengths:
@@ -5843,8 +5841,6 @@ class SelfAttentionLayer(_ConcatInputLayer):
     self._restrict_state_to_last_seq = restrict_state_to_last_seq
     assert self._rec_previous_layer or self.input_data.time_dim_axis is not None, (
       "%s: This layer is expected to be used inside a RecLayer, or to have input with time." % self)
-    if attention_mask is None:
-      attention_mask = "left_only" if attention_left_only else "length"
     total_value_dim = self.output.dim
     assert total_key_dim % num_heads == 0, "must be divisible"
     assert total_value_dim % num_heads == 0, "must be divisible. total_value_dim = n_out"
@@ -5861,7 +5857,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
       prev_mask = None
       if self._rec_previous_layer:
         assert self.input_data.time_dim_axis is None
-        assert attention_mask == "left_only"
+        assert attention_left_only
         # (batch,heads,time,k-dim//heads)
         prev_k_left = self._rec_previous_layer.rec_vars_outputs["k_left"]
         # (batch,heads,time,v-dim//heads)
@@ -5956,10 +5952,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
       energy_ = tf.transpose(energy_, [1, 2, 0, 3])  # [batch,heads,num_queries|1,num_keys]
       energy += energy_
     if self.input_data.time_dim_axis is not None:
-      assert isinstance(attention_mask, str) or callable(attention_mask)
-      if isinstance(attention_mask, str):
-        assert attention_mask in ["left_only", "length"]
-      if attention_mask == "left_only":
+      if attention_left_only:
         # We also ignore the input data sequence length, because we expect that frames outside the seq length
         # are anyway ignored.
         from returnn.tf.util.basic import matrix_triangular
@@ -5970,12 +5963,13 @@ class SelfAttentionLayer(_ConcatInputLayer):
         if have_prev_kv_left:
           energy_mask_left = tf.ones((1, 1, num_queries, tf.shape(prev_k_left)[2]), dtype=tf.bool)
           energy_mask = tf.concat([energy_mask_left, energy_mask], axis=-1)
-      elif attention_mask == "length":
+      else:
         energy_mask = tf.sequence_mask(
           self.input_data.get_sequence_lengths(), maxlen=tf.shape(energy)[-1])  # (batch,time)
         energy_mask = tf.reshape(energy_mask, [tf.shape(energy)[0], 1, 1, tf.shape(energy)[-1]])  # (batch,1,1,time)
-      else:
-        energy_mask = attention_mask(energy)
+      if attention_mask:
+        new_mask = attention_mask(energy)
+        energy_mask = tf.math.logical_and(energy_mask, new_mask)
       if state_var_lengths is not None and have_prev_kv_left:
         if callable(state_var_lengths):
           state_var_lengths = state_var_lengths()
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -5819,7 +5819,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
   def __init__(self, num_heads, total_key_dim,
                key_shift=None,
                forward_weights_init="glorot_uniform", attention_dropout=0.0,
-               attention_left_only=True, attention_mask=None, initial_state=None,
+               attention_left_only=False, attention_mask="length", initial_state=None,
                restrict_state_to_last_seq=False, state_var_lengths=None, **kwargs):
     """
     :param int num_heads:
@@ -5831,7 +5831,7 @@ class SelfAttentionLayer(_ConcatInputLayer):
     :param float attention_dropout:
     :param bool attention_left_only: will mask out the future. see Attention is all you need.
     :param str|Callable attention_mask: mask to apply to input sequence. Allowed strings:
-      "length" masks the positions outside the original length, "only_left" masks the future.
+      "length" masks the positions outside the original length, "left_only" masks the future.
       If a callable, it must be a function that receives only the input sequence and returns the mask.
     :param str|float|int|None initial_state: see RnnCellLayer.get_rec_initial_state_inner().
     :param bool restrict_state_to_last_seq: see code comment below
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/rec.py
```diff
@@ -5819,8 +5819,8 @@ class SelfAttentionLayer(_ConcatInputLayer):
   def __init__(self, num_heads, total_key_dim,
                key_shift=None,
                forward_weights_init="glorot_uniform", attention_dropout=0.0,
-               attention_left_only=False, initial_state=None, restrict_state_to_last_seq=False,
-               state_var_lengths=None, **kwargs):
+               attention_left_only=True, attention_mask=None, initial_state=None,
```
albertz, line 5822: You should not change the default of `attention_left_only`.

returnn/tf/layers/rec.py
```diff
@@ -5856,7 +5861,7 @@ def __init__(self, num_heads, total_key_dim,
       prev_mask = None
       if self._rec_previous_layer:
         assert self.input_data.time_dim_axis is None
-        assert attention_left_only
+        assert attention_mask == "only_left"
```
albertz, line 5864:
```suggestion
        assert attention_mask == "left_only"
```
?

returnn/tf/layers/rec.py
```diff
@@ -5830,6 +5830,9 @@ def __init__(self, num_heads, total_key_dim,
     :param str forward_weights_init: see :func:`TFUtil.get_initializer`
     :param float attention_dropout:
     :param bool attention_left_only: will mask out the future. see Attention is all you need.
+    :param str|Callable attention_mask: mask to apply to input sequence. Allowed strings:
+      "length" masks the positions outside the original length, "only_left" masks the future.
```
albertz, line 5834: "left_only" missing here?

albertz:
I'm not sure that we want to make `SelfAttentionLayer` even more complex and complicated.
The idea (since a long time, many times mentioned in our videos/tutorials/lectures) is to deprecate it and break it down into more atomic building blocks such that no modification is ever needed on RETURNN side to implement whatever is wanted.

Although in this case, the change is still quite short, so maybe ok.

@curufinwe @JackTemaki @patrick-wilken @Zettelkasten what do you think?
Related is also #391.

On the change itself:

While a callable is generic, this is quite non-RETURNN-ic. It would be even more generic or better if this is a layer instead of a callable. You could just pass a custom `attention_mask: LayerBase`.

Also, test cases are failing. Please always check that (mark it as draft when the PR is not ready).

returnn/tf/layers/rec.py
Zettelkasten: Here, `attention_mask == "left_only" or callable(attention_mask)` would need to be supported too if we want to allow setting a callable `attention_mask` inside a rec layer.
Would the provided custom `attention_mask` then need to handle masking away the future, or should `SelfAttentionLayer` always apply such masking on top of the provided mask?
I think I'd slightly prefer the latter.

Zettelkasten (commented):
I have been using a WIP implementation of self attention with multiple layers instead of this `SelfAttentionLayer` for quite some time now, and I don't think I'd ever go back to using `SelfAttentionLayer` after we merge those changes into the master (we still need some changes to how new dim tags are created for that, as discussed in #391).

So if we continue with the plan to deprecate that layer altogether, I do not really have hard feelings about how we change it now. This change looks fine to me, it's small and rather generic.

> While a callable is generic, this is quite non-RETURNN-ic. It would be even more generic or better if this is a layer instead of a callable. You could just pass a custom `attention_mask: LayerBase`.

I agree here: I'd also make it another layer. Especially when used inside a rec layer, the `attention_mask` would need to have or not have a time dim depending on whether the loop is optimized or not, I think that's much easier to establish right now using other layers than with plain TF ops.

returnn/tf/layers/rec.py
mattiadg: Maybe `attention_mask` can be made only of type callable (or layer) and we leave the previous behavior in place?
This way the mask can be anything and be simple, then the length or past-only restrictions are enforced as before.
What do you think?

Zettelkasten: Yes, I think that's a good idea :+1:

albertz (changes requested):
Still the change is missing to make `attention_mask` a layer.

mattiadg:
> Still the change is missing to make `attention_mask` a layer.

Are there layers in returnn to create masks? If not, why would this solution be better?
The called layer would need to know exactly how the shape of the input tensor to the self-attention layer is transformed to the energy tensor shape, so maybe I'm wrong but it looks more complicated to me.
Also, at the moment I can't think of anything better than an EvalLayer encapsulating the callable that I would pass with the current implementation.

albertz:
> > Still the change is missing to make `attention_mask` a layer.
> 
> Are there layers in returnn to create masks? If not, why would this solution be better?

Please see our videos and slides and documentation about RETURNN on this.
If you miss any layer for any type of computation (creating masks, or anything really), please create an issue about this, and let's implement this. This should never be the case that there is anything which can not be computed via layers.

It does not make much sense to partly implement stuff in pure TF code (and then use `EvalLayer` or here this `callable`) and other parts via RETURNN layers. It should be consistent. 

> The called layer would need to know exactly how the shape of the input tensor to the self-attention layer is transformed to the energy tensor shape, so maybe I'm wrong but it looks more complicated to me.

But you know that? Why is it more complicated? It is just the same as the `callable` but directly via RETURNN layers.

> Also, at the moment I can't think of anything better than an EvalLayer encapsulating the callable that I would pass with the current implementation.

You should avoid `EvalLayer` if you can. If you can not, again, consider this a bug, and implement what is missing. You should have reported that. There are only rare cases where `EvalLayer` makes sense.

JackTemaki:
> If you miss any layer for any type of computation (creating masks, or anything really), please create an issue about this, and let's implement this.

This functionality is indeed missing, but this is because we just do not have all tf ops available directly in Returnn.

So the steps needed here would be:
 - take a tensor of size [B, T, F] with size placeholder S for axis T
 - get the actual size of B and T (tf.shape op)
 - create a tensor of ones like [B, T] (tf.ones) and keep the size placeholder
 - apply the diagonal masking op of TF (this could be an eval layer, as this is an elementary operation that does not change the tensor)
 
 So step 2 and 3 are not available natively.
 
 The problem here is the shape, so if the SelfAttentionLayer would be written in low-level-layers, you could directly apply step 3 and 4 on the energies (which are [B, T]) with an eval layer. But right now they are hidden inside the SelfAttentionLayer.
 
In general it is a problem that we do not have a RETURNN equivalent to all basic TF functions... so no matter the approach (introducing new layers, writing a quite complex eval layer), solving this issue here other than just passing a callable will be quite difficult.

albertz:
> So the steps needed here would be:
> 
> * take a tensor of size [B, T, F] with size placeholder S for axis T
> * get the actual size of B and T (tf.shape op)

We have `LengthLayer`.

We do not have some layer like `ShapeLayer` or so but if really needed we can (and should) just add it.

> * create a tensor of ones like [B, T] (tf.ones) and keep the size placeholder

We have `ConstantLayer`.

> * apply the diagonal masking op of TF (this could be an eval layer, as this is an elementary operation that does not change the tensor)

Which op do you mean? Applying the mask just corresponds to `SwitchLayer` (`tf.where` internally).

Note that we also have `SeqLenMaskLayer`.

> So step 2 and 3 are not available natively.

I'm not sure this is true.

> In general it is a problem that we do not have a RETURNN equivalent to all basic TF functions...

If anything is missing, we should just add such layer.

> so no matter the approach (introducing new layers, writing a quite complex eval layer), solving this issue here other than just passing a callable will be quite difficult.

We should not introduce workarounds because some other things are missing (or some other things do not work yet, or whatever). We should rather add whatever is missing (or fix whatever is needed).

Zettelkasten:
I think one main problem is that it is not clear what the time dim tags of the Layer you provide as mask should actually be.
In #391 we solve this issue by introducing a 'rec-history' dim tag for the key sequence (and just use the input time axis for the queries).
But here, this is really unclear to me because right now what the key dim tag should be, you cannot add any new tags yourself.

> This functionality is indeed missing, but this is because we just do not have all tf ops available directly in Returnn.

If you had access to a dim tag for the query and for the key sequence, you can simply use RangeInAxisLayer to create a [query-T] and a [key-T] tensor with the time indices, and then obtain the mask with a '<='-CompareLayer.
Or maybe add additional masking with LengthLayer, depending on what you really want to do.

albertz:
> I think one main problem is that it is not clear what the time dim tags of the Layer you provide as mask should actually be.
> In #391 we solve this issue by introducing a 'rec-history' dim tag for the key sequence (and just use the input time axis for the queries).

I'm not sure this is really a problem here. I assume `attention_left_only` is never used here but just a custom mask, and this layer is not used within a rec layer. So this is much simpler here.

> you cannot add any new tags yourself.

You can, by just putting `new_dim_tag = DimensionTag(...)` and then setting it on a custom `size_placeholder` via `EvalLayer`. Although this is surely quite non-intuitive. But anyway, I think this is not needed here.

albertz:
Note that #391 was fixed now (via `CumConcatLayer` and many other changes).
So we should adopt a similar logic for this use case here. I.e. not extend `SelfAttentionLayer`, but think about what is missing for basic layers to implement this:

Maybe `dyn_mask_ext` (earlier called `seq_mask_ext`) as discussed in #391, but not sure. I think this is actually not really needed here.

The masking could also be more explicit by the user. This can already be done via `SwitchLayer`. Sth like this:
```
"values": {...},
"energy": {...},
"mask": {...},  # whatever you want
"energy_masked": {"class": "switch", "condition": "mask", "true_from": "energy", "false_from": float("-inf")},
"att_weights": {"class": "softmax_over_spatial", "from": "energy_masked"},
"att": {"class": "dot", "from": ["values", "att_weights"], ...},
```
You get the idea.

Note that `SoftmaxOverSpatialLayer` also already has the options `window_start` and `window_size`. If your mask is supposed to be a window, this could also be used.

Is this already what you want? Then we can close this PR.

albertz:
For reference, an example for generalized self attention (non recursive) (via #656):
```
new_dim = DimensionTag(kind=DimensionTag.Types.Spatial, description="new_self_att_dim")
net_dict_new = {
  "qkv": {
    "class": "linear", "from": "data", "with_bias": False,
    "n_out": n_key_dim_total * 2 + n_value_dim_total},  # [B,T,2*K'+V']
  "qkv_": {
    "class": "split_dims", "from": "qkv",
    "axis": "F", "dims": (n_heads, n_key_dim_per_head * 2 + n_value_dim_per_head)},
  "qkv_split": {
    "class": "split", "from": "qkv_",
    "size_splits": [n_key_dim_per_head, n_key_dim_per_head, n_value_dim_per_head]},
  "q": {"class": "copy", "from": "qkv_split/0"},  # [B,T,H,K]
  "k": {"class": "copy", "from": "qkv_split/1"},  # [B,T,H,K]
  "v": {"class": "copy", "from": "qkv_split/2"},  # [B,T,H,V]
  "q_": {"class": "eval", "from": "q", "eval": "source(0) * %f" % ((n_key_dim_total // n_heads) ** -0.5)},
  "k_": {"class": "reinterpret_data", "from": "k", "set_dim_tags": {"T": new_dim}},  # [B,T_new,H,K]
  "v_": {"class": "reinterpret_data", "from": "v", "set_dim_tags": {"T": new_dim}},  # [B,T_new,H,V]
  "energy": {
    "class": "dot", "from": ["q_", "k_"],
    "red1": "static:-1", "red2": "static:-1",
    "var1": time_dim, "var2": new_dim},  # [B,H,T_new,T]
  "att_weights": {
    "class": "softmax_over_spatial", "from": "energy", "axis": new_dim},  # [B,H,T,T_new]
  "att": {
    "class": "dot", "from": ["att_weights", "v_"],
    "red1": new_dim, "red2": new_dim,
    "var1": time_dim, "var2": "static:-1"},  # [B,H,T,V]
  "output": {
    "class": "merge_dims", "from": "att", "axes": "static"},  # [B,T,V']
}
```

albertz:
I assume this can be closed now, given that this should be possible already, as described.

PR #546, opened 2021-06-16
compile_tf_graph: use absolute layer names

vieting:
In many of my setups, I use `compile_tf_graph.py` to compile the graph based on my RETURNN config and then use this graph in RASR for recognition. In RASR, I always used `output/output_batch_major` as tensor name to obtain the posteriors that I needed.

This does not work as expected anymore, since `network.layers` used to include only `SubnetworkLayers` but now also includes all layers inside the subnetwork. Therefore, if I have a subnetwork with a layer named "output" inside, the name of the tensor I actually want changes to `output/output_batch_major_1`.

We can avoid this by using absolute layer names in the script. Not sure if this breaks some setups people recently built, but obviously not doing this also breaks setups as in my case.

### DIFF
--- a/tools/compile_tf_graph.py
+++ b/tools/compile_tf_graph.py
@@ -701,7 +701,7 @@ def main(argv):
         continue
       if layer.output.batch_dim_axis is None:
         continue
-      with layer.cls_layer_scope(layer.name):
+      with layer.cls_layer_scope(layer.get_absolute_name()):
         tf.identity(layer.output.get_placeholder_as_batch_major(), name="output_batch_major")
 
     tf.group(*network.get_post_control_dependencies(), name="post_control_dependencies")


albertz:
> ... Therefore, if I have a subnetwork with a layer named "output" inside, the name of the tensor I actually want changes to `output/output_batch_major_1`.

I don't exactly understand this reasoning. Why? What is `output/output_batch_major` now?

> We can avoid this by using absolute layer names in the script. Not sure if this breaks some setups people recently built, but obviously not doing this also breaks setups as in my case.

I also don't quite understand why this fixes the problem... (But haven't looked deeply into it yet.)

patrick-wilken:
If I understand correctly, before, if you had the layers `my_layer` and `my_subnetwork/my_layer`, only the output of `my_layer` was given the name `my_layer/output_batch_major`. With recent code we also try to call the output of `my_subnetwork/my_layer` `my_layer/output_batch_major` which causes the name conflict.
If this is indeed what happens this change makes sense to me. I would also choose to break old setups here in favour of giving a name to all layers' outputs, not just the ones in the root network. (When I use graph exportation I tend to lookup the tensor names in tensorboard anyway 😆  Maybe for RASR there is a more automated approach though 😆 )

albertz:
Ah yes that makes sense.


Patrick Wilken ***@***.***> schrieb am Mi., 16. Juni 2021,
17:53:

albertz (approved):

PR #547, opened 2021-06-22
allow layers directly for target option

albertz:
Related:
https://github.com/rwth-i6/returnn_common/pull/8

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -452,7 +452,7 @@ class LayerBase(object):
     which should be resolved.
     """
     from .basic import get_loss_class
-    from ..network import DataNotFound
+    from ..network import LayerNotFound
     BehaviorVersion.require(
       condition="from" in d,
       message='Missing "from" in layer definition: %s/%s' % (network.name, d.get("_name", "<UNKNOWN>")),
@@ -493,6 +493,9 @@ class LayerBase(object):
             target_layers[target] = get_layer(target[len("layer:"):])
           else:
             try:
+              # Check whether the target itself is a layer
+              target_layers[target] = get_layer(target)
+            except LayerNotFound:
               # Note: This is a workaround for cases where we need to know about used data keys before the layer
               # itself is constructed (e.g. in _SubnetworkRecCell.get_output).
               # A nicer solution would be to not modify this here,
@@ -501,13 +504,6 @@ class LayerBase(object):
               # Also, if we are inside a rec layer, and doing search, we also cannot do that.
               if network.is_inside_rec_layer() and not network.search_flag:
                 network.get_extern_data(target, mark_data_key_as_used=True)
-              else:
-                # Anyway, to check for the name.
-                network.get_extern_data(target, mark_data_key_as_used=False)
-            except DataNotFound:
-              # Expect that this is a layer.
-              target_layers[target] = get_layer(target)
-            else:
               if not network.search_flag:
                 # Also, there are cases when we want to have the target as an explicit layer dep,
                 # e.g. when the target has a beam, to derive the search choices.
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -452,6 +452,7 @@ class LayerBase(object):
     which should be resolved.
     """
     from .basic import get_loss_class
+    from ..network import DataNotFound
     BehaviorVersion.require(
       condition="from" in d,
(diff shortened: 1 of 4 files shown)


albertz:
Not sure if this is ready yet. (I cannot make it a draft again? I don't find the button...)
I just wanted to run the checks for now.

Anyway, regarding the change: I'm not sure if this is the best approach. There are some nasty edge cases.

albertz:
@Atticus1806 This, or something like this, would be needed.

returnn/tf/layers/base.py
```diff
@@ -491,18 +492,26 @@ def transform_config_dict(cls, d, network, get_layer):
...
+              # Also, if we are inside a rec layer, and doing search, we also cannot do that.
+              if network.is_inside_rec_layer() and not network.search_flag:
+                network.get_extern_data(target, mark_data_key_as_used=True)
+              else:
+                # Anyway, to check for the name.
+                network.get_extern_data(target, mark_data_key_as_used=False)
+            except DataNotFound:
+              # Expect that this is a layer.
+              target_layers[target] = get_layer(target)
+            else:
```
JackTemaki, line 510: where does this `else` belong to?

JackTemaki: Ah nevermind, I did not know that there is try-except-else and not only try-except-finally

patrick-wilken:
I was kind of out of the loop regarding anything in return_common. But as I understand giving a plain layer name without a `layer:` prefix would simplify automatic creation of layers, right?

> There are some nasty edge cases.

Which ones do you think of? `network.get_extern_data()` with `mark_data_key_as_used=False` should have no side effects, right? But `get_layer(target)` could be unwanted if the target is actually never used in the layer. (But this is already the case with the current code when using the `layer:` prefix.)

returnn/tf/layers/base.py
```diff
@@ -491,18 +492,26 @@ def transform_config_dict(cls, d, network, get_layer):
...
+              if not network.search_flag:
```
patrick-wilken, line 511: Actually I don't understand the existing code here. Is it `not network.search_flag` because in the search case `ChoiceLayer` would add the dependency? If so, this is already a pretty bad. 😅

albertz:
> I was kind of out of the loop regarding anything in return_common. But as I understand giving a plain layer name without a `layer:` prefix would simplify automatic creation of layers, right?

returnn_common and the wrappers there was one motivation. But I also see this independent, as one attempt to make RETURNN more consistent and unified.

> > There are some nasty edge cases.
> 
> Which ones do you think of?

I'm not even sure I can think of all possible cases. The handling of extern data, used targets, etc was always somewhat special with many special cases.

One obvious problem here, which should only rarely occur: If you have e.g. a layer "classes" and the data key "classes", you have a conflict now. This current code would choose the extern data over the layer. But this problem might be rare. Maybe we could also do it the other way around and always prefer the layer if possible.

Getting the layer and then checking for `LayerNotFound` exception would be the other way around. However, I'm not 100% sure this works always correctly with the template construction (although I also don't see why it would not).

> `network.get_extern_data()` with `mark_data_key_as_used=False` should have no side effects, right? But `get_layer(target)` could be unwanted if the target is actually never used in the layer. (But this is already the case with the current code when using the `layer:` prefix.)

Yes, this is also one of the problems. It would just copy the template with `mark_data_key_as_used=False` and no use the data itself. We don't really have sth equivalent for layers.

But maybe this is still not really a problem. At inference time, it would still get the targets (as a layer) but it has `available_for_inference=False` but that should not really be a problem when it is not used.

returnn/tf/layers/base.py
albertz: I don't exactly remember anymore. I think it was just buggy, in a way that this would trigger `get_extern_data(target, mark_data_key_as_used=True)`, and then the rec layer would try to unroll the targets in the loop, which would not work because the loop can potentially run different number of steps than the target (in search mode).

Or maybe sth similar/related. You are right in principle. Or actually, it's not really well defined what the targets should be when search mode is enabled. It cannot unroll it. It can still provide the template (with `dim` and maybe `vocab`, but ignoring time-dim). So maybe that should be it, but then as template only (maybe using `DataNotAvailableLayer`).

I think when I wrote this code, I was just too lazy to really think about how it should be for search mode, and just implemented this case.

But anyway, I think solving this would be better for a separate PR.

PR #549, opened 2021-06-23
Layer trainable inherits from parent by default

albertz:


michelwi (approved):

albertz:
> This might break old configs if someone accidentally set it to be this way, but this is a feature we should have. (But I guess those configs are then wrong somehow anyway)

Yep. The question is, if there really exists such a case. And if we should increase `behavior_version`. Or just apply this PR as-is.

patrick-wilken (approved):

albertz:
Ok, as you all approved, I just merge now.

PR #551, opened 2021-06-25
get_initializer: Do not override seed set by user

Zettelkasten:
When adding a param, most layers pass sth like `self.network.random.randint(2 ** 31)` as seed to `get_initializer`: e.g.

https://github.com/rwth-i6/returnn/blob/master/returnn/tf/layers/basic.py#L6911

However, when the provided initializer already contains a seed, this is just overridden: E.g. 
```
{'class': 'variable', 'add_batch_axis': True, 'shape': (8, 1, 64, 2), 'trainable': False,
 'init': "variance_scaling_initializer(mode='fan_in', distribution='uniform', 'scale=1.0, seed=100)'}
```
will be initialized differently for different global `random_seed`s.

That is definitely not what I expect: If I explicitly provide a seed to the initializer, I expect it to always initialize my tensor to the same value - independent of the network seed.

In this PR, I change `get_initializer` to only override the `seed` if the initializer had none set before.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -1431,7 +1431,7 @@ def get_initializer(s, seed=None, eval_local_ns=None, dtype=tf.float32):
   :param str|dict[str]|float|numpy.ndarray s: e.g. "glorot_uniform" or "truncated_normal" or "orthogonal",
     or config dict with "class",
     or string to be `eval`ed if it contains "(". constant if a float is given.
-  :param int|tf.Tensor seed:
+  :param int|tf.Tensor seed: used in case the initializer has no explicit seed specified.
   :param dict[str]|None eval_local_ns:
   :param tf.DType|str dtype:
   :return: (function (shape) -> tf.Tensor) | tf.Initializer
@@ -1505,7 +1505,10 @@ def get_initializer(s, seed=None, eval_local_ns=None, dtype=tf.float32):
       raise Exception("invalid initializer: %r" % s)
     if seed is not None:
       assert isinstance(f, init_ops.Initializer)
-      if hasattr(f, "seed"):
+      if hasattr(f, "seed") and f.seed is None:
+        # If f.seed is not None, do not override the seed.
+        # This e.g. happens when the user explicitly specifies a seed in the initializer constructor:
+        # Use the user's seed instead.
         f.seed = seed
   except Exception:
     error()


albertz (approved):

albertz:
Ok, merged.

But I wonder, why do you ever want to do that? You can already control the global seed (option `random_seed` in config). Setting the seed per parameter runs into the potential problem that you then have all the same parameters everywhere.

Zettelkasten:
> But I wonder, why do you ever want to do that? You can already control the global seed (option `random_seed` in config). Setting the seed per parameter runs into the potential problem that you then have all the same parameters everywhere.

I used this for debugging: Because my network behaved very differently depending on the global `random_seed`, I wanted to narrow down if a specific layer was the problem. So I wanted to fix its initialization and vary the params of all other layers by using the global seed.

Zettelkasten:
For anyone interested: This does not work like I thought (making this PR pretty much useless for me at least.)

Apparently, for each variable, both its operation level and graph level seeds are considered. If you change either of those, the outcome will be different.
I thought that specifying the op level seed would cause it to ignore the network seed for that op and only use my specified seed. But that's not the case.

Instead, I think I'll have to specify my initializer matrix explicitly. Or load it from a file or so.

albertz:
I don't quite understand.

All the initializers use the basic random ops functions. E.g. `VarianceScaling` uses `random_ops.random_normal(shape, 0.0, stddev, dtype, seed=self.seed)`.

All those random ops functions have such a seed logic:
```
    seed1, seed2 = random_seed.get_seed(seed)
    rnd = gen_random_ops.random_standard_normal(
        shape_tensor, dtype, seed=seed1, seed2=seed2)
```
And `get_seed` will use this logic:
```
    global_seed = ops.get_default_graph().seed
    seeds = _truncate_seed(global_seed), _truncate_seed(op_seed)
```
or
```
      seeds = DEFAULT_GRAPH_SEED, _truncate_seed(op_seed)
```
The global graph seed would be set via `set_random_seed`.
We have this logic in RETURNN (in `Engine._init_network`):
```
    tf_random_seed = 42
    net_random_seed = epoch
    if self.config.opt_typed_value("random_seed", None):
      seed = self.config.int("random_seed", None)
      net_random_seed = (epoch * 3 + seed * 5 + 7) % (2 ** 31)
      tf_random_seed = (net_random_seed * 2 + 3) % (2 ** 31)
    tf_compat.v1.set_random_seed(tf_random_seed)
```
I assume you do not set `random_seed` explicitly. So then the global graph seed would always be 42.

So then together with your fixed op seed, this should indeed always give you exactly the same.

Zettelkasten:
Sorry, I should have said this more clearly: There is not really an issue with RETURNN, and I still believe this PR makes sense.
It is just that my understanding of how initializers in TF worked was wrong (now thats independent from RETURNN). I didn't know there were two seeds per each op.

What I want to achieve is the following:
- train three models with different parameter initializations, except that I want the init of one specific layer to be fixed.
- do the reverse: train three models with the same parameter inits for all layers except for one, and vary the init of the one specific layer.

This way I want to figure out whether a large variance that we saw can be explained by the init of this one specific layer that I suspect.

So in my misunderstanding, I thought that I could fix the init of my specific layer using the `seed` param I pass its initializer, and vary the seeds of the other layers using the global seed using the RETURNN global `random_seed`. But as I learned, that does not work.

Instead I ended up writing an external script which just calls the init for my specific layer with `global TF seed = 0` and a given specific op seed, which I write to a numpy file. Then in RETURNN I just use `numpy.load("init_hashXYZ.npy")` as initializer, which will guarantee the same init even for different global seeds.
For the reverse, I can fix the global seed and vary which numpy array I load. (Assuming that the construction order is the same, then all inits should be exactly the same for the same global seed then, right?)

Thanks for your detailed info on how the seeds work in RETURNN. It makes sense, I was just confused by the TF behavior.

albertz:
But still, we can make your initial idea work. There is no real reason why it should not work like that. So you want to set the `random_seed` (to get a different RETURNN net random seed) but you want to keep the TF random seed fixed. We could simply introduce another config option like `tf_random_seed` which you could set to some fixed value (e.g. `tf_random_seed = 42`), and from the code above (`Engine._init_network`), you would use sth like:
```
tf_random_seed = (self.config.int("tf_random_seed", None) or (net_random_seed * 2 + 3)) % (2 ** 31)
```

Although your other approach you use now sounds also fine.

Zettelkasten:
> [..] We could simply introduce another config option like `tf_random_seed` which you could set to some fixed value (e.g. `tf_random_seed = 42`) [..]

Ah cool, that'd be nice for the future. I won't use it right now as my new solution is working already, but I'll add such option if I stumble into this problem again.

PR #552, opened 2021-06-27
RecLayer multiple inputs via explicit unstacking

albertz:
Instead of using `from: x` in the RecLayer, which gets unstacked (via TensorArray) such that inside the rec loop you get `x[i]`, here is some approach to make this unstacking explicit. This would then also support multiple inputs to the RecLayer.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -4296,6 +4296,27 @@ class GetRecAccumulatedOutputLayer(LayerBase):
     return subnet.layers[sub_layer].output
 
 
+class IterLayer(LayerBase):
+  """
+  This is supposed to be used inside a :class:`RecLayer`.
+  The input is supposed to be outside the rec layer (i.e. via ``base:``).
+  Uses tf.TensorArray and then unstack on the inputs to make it available per-frame.
+  This is an alternative to making some input to the rec layer,
+  such that the rec layer can have multiple inputs (as long as they have the same time dim).
+  """
+  # TODO:
+  #  I'm not really sure whether this layer is the best approach.
+  #  We could also introduce a new syntax, like ``base-iter:`` instead of ``base:`` to do that,
+  #    and then hide the logic inside RecLayer.
+  #  Another problem: As there is no recurrent dependency on this layer (its only input is outside),
+  #    it will be optimized out.
+  #    In the optimized-out case, it would not use tf.TensorArray but simply copy the input.
+  #    So then this is just equivalent to CopyLayer.
+  #    TODO Why does CopyLayer actually not work already?
+  #      See also the logic in get_input_moved_out inside _construct.
+  # TODO ...
+
+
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/rec.py
```diff
@@ -4296,6 +4296,27 @@ def get_out_data_from_opts(cls, name, sources, sub_layer, **kwargs):
...
+  """
+  # TODO:
+  #  I'm not really sure whether this layer is the best approach.
+  #  We could also introduce a new syntax, like ``base-iter:`` instead of ``base:`` to do that,
+  #    and then hide the logic inside RecLayer.
+  #  Another problem: As there is no recurrent dependency on this layer (its only input is outside),
+  #    it will be optimized out.
+  #    In the optimized-out case, it would not use tf.TensorArray but simply copy the input.
+  #    So then this is just equivalent to CopyLayer.
+  #    TODO Why does CopyLayer actually not work already?
```
albertz, line 4315: This is the relevant question... Why does CopyLayer actually not work already?

patrick-wilken:
Related to #649 the question is, whether the additional inputs should be considered for the sequence length (dynamic vs. fixed) in `_SubnetworkCell.get_output()`. So if we copy a base-layer into a dynamic loop, does it inherit the fixed length? Can we allow several base-layers with different length? Etc.

albertz:
> Related to #649 the question is, whether the additional inputs should be considered for the sequence length (dynamic vs. fixed) in `_SubnetworkCell.get_output()`. So if we copy a base-layer into a dynamic loop, does it inherit the fixed length? Can we allow several base-layers with different length? Etc.

I'm not sure I understand. You mean whether `i` is an explicit argument to the layer or implicit via the rec layer (so basically hardcoded to `:i`)?

But even when it is hardcoded to `:i`, this can still support any input which is the same length or shorter, so it doesn't need to be the same.

Or what do you mean by "fixed length"? The length can still be dynamic in any case (of shape [B]).

The number of iterations (`max(i) + 1`) the `tf.while_loop` does is of course a scalar.

I think I would prefer to not have too much magic here. This layer (`IterLayer`) or such a layer should really just do `x[i]` (via `TensorArray`), nothing else. Whatever else you do should be up to you via other layers.

albertz:
Note on the layer name: I considered several options:

* `IterLayer`: because we iterate over some tensor, so you could think of `for x in xs: ...`.
* `UnstackLayer` because you unstack the tensor, and it basically wraps `tf.TensorArray.unstack`

I did not want to have `TensorArray` in the name as I felt this being too much implementation specific. The name should reflect the logical operation and not implementation details.

In the end, I choose `RecUnstackLayer`, to reflect that this is specific for a `RecLayer`. There is also `StackLayer`, and people might mistake this as the inverse of `StackLayer`, which would be misleading. There also might be some `RecStackLayer` later, although this is anyway what you get when you access some `RecLayer` sub layer from outside (when it was marked as output layer).

PR #553, opened 2021-06-30
Handle nan and inf in better_repr

dthulke:
To create search output files in the py format we currently kind of rely on `better_repr` to generate valid python code.

But this is not the case if the output contains infs or nans (e.g. beam scores (can happen in my case since I am doing a kind of constrained search where I mask invalid hypotheses)).
To fix this, this changes the output of `better_repr` from e.g. `-inf` to `float('-inf')`.

### DIFF
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -16,6 +16,7 @@ import inspect
 import os
 import sys
 import shlex
+import math
 import numpy as np
 import re
 import time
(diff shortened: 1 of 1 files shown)


patrick-wilken (approved):

albertz (approved):
I'm not 100% sure this will work in all future Python versions (i.e. the `repr` is well defined and will never change) but I guess this is fine for now.

PR #554, opened 2021-06-30
ChoiceLayer, enable dict in explicit_search_sources

jvhoffbauer:
Enable usage of dicts in the `explicit_search_sources` of the choice layer. This is useful when writing custom logic in `custom_score_combine`, that behaves differently, depending on the inputs.

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -5001,13 +5001,20 @@ class ChoiceLayer(BaseChoiceLayer):
       # Note that we don't look at d.get("search") here, because in case of search,
       # if there are other choice layers, we still need to add the scores to the beam.
       d["from"] = []
+
+    # Convert explicit_search_sources to layers
     if d.get("explicit_search_source"):
       assert "explicit_search_sources" not in d
       d["explicit_search_sources"] = [get_layer(d.pop("explicit_search_source"))] if network.search_flag else []
     elif d.get("explicit_search_sources"):
-      assert isinstance(d["explicit_search_sources"], (list, tuple))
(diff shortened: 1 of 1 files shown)


jvhoffbauer:
Would it also be fine to accept None values in `explicit_search_sources`?

returnn/tf/layers/rec.py
```diff
@@ -5010,7 +5017,11 @@ def get_dep_layers(self):
     # See also self.transform_config_dict where we might strip away the sources.
     ls = super(ChoiceLayer, self).get_dep_layers()
     if self.explicit_search_sources:
-      ls.extend(self.explicit_search_sources)
+      if isinstance(self.explicit_search_sources, dict):
+        additional_sources = self.explicit_search_sources.values()
```
albertz, line 5021: This is non-deterministic in earlier Python versions.
Better:
```suggestion
        additional_sources = [src for _, src in sorted(self.explicit_search_sources.items())]
```

albertz:
> Would it also be fine to accept None values in `explicit_search_sources`?

Yea, you can add that if you want.
Then you also need to extend `get_dep_layers` for that case.
```
additional_sources = [src for _, src in sorted(self.explicit_search_sources.items()) if src]
```

patrick-wilken:
This is just to access the layers by key instead of index in `custom_score_combine`?

There is `assert isinstance(d["explicit_search_sources"], (list, tuple))` in `compile_tf_graph.py`, maybe we want to keep it up to date.

jvhoffbauer:
Can we merge?

albertz (approved):
Looks good. Let's wait for the tests.

Issue #558, opened 2021-07-07
SplitLayer is never optimized as output layer


Zettelkasten:
Consider some net dict of a rec unit like this (see #559 for a full test case):
```
  input: {...., from: "prev:input"},  # cannot be optimized out
  split: {class: split, from: input, size_splits: [3, 7]},
  output: {class: linear, from: ["split/0", "split/1"]}  # can be optimized out
```
Then, `split` should also be optimized out as an output layer.
This is not the case however, because:
`split/0` and `split/1` have `collocate_with=["split"]` set by `_TemplateLayer.get_sub_layer`, but also an dependency on their parent layer `split`.

E.g. `split/0` is not optimized out because the logic of `output_can_move_out(split/0)` does the following:
1. it checks whether the collocated parent `split` can be moved out with `_collocated={split/0}`
2. however, `split` cannot be moved out because `split/1` depends on it. And `split/1` has not been moved out yet. [edited.]

Symmetrically, `split/1` is also not optimized out.

--------------------

(Actually, in my use case, I don't even have a layer `input` that recurrently depends on itself. But instead, I use a `:i` layer, which currently never is optimized out as an input layer:
```python3
      if layer.name in [":i", "end"]:  # currently not fully implemented
        return False
```
Somewhat related - what is missing of the implementation there?)

albertz:
> Then, `split` should also be optimized out as an output layer.

Yes.

> This is not the case however, because:
> `split/0` and `split/1` have `collocate_with=["split"]` ...

Hm, I think our whole logic with sub layers is a bit buggy, or not even well defined.

For the case of subnetwork sub layers, they are all completely independent (and the main subnet layer is just the same as the `output` sub layer of it).

Otherwise (basically all other layers which have sub layers), they all belong together. Like in this case for `SplitLayer`.

Probably this second case is buggy.

I remember that @patrick-wilken did some changes/fixes recently on this?

> (Actually, in my use case, I don't even have a layer `input` that recurrently depends on itself. But instead, I use a `:i` layer, which currently never is optimized out as an input layer:
> 
> ```python
>       if layer.name in [":i", "end"]:  # currently not fully implemented
>         return False
> ```
> 
> Somewhat related - what is missing of the implementation there?)

Let's keep this as a separate issue, and not mix them up here. This is really unrelated / orthogonal.

The layer responsible for `:i` (`RecStepInfoLayer` or sth like that) is just not implemented (as far as I remember).

[...]

PR #559, opened 2021-07-07
Test case / fix for #558

Zettelkasten:


Zettelkasten:
I'm not 100% sure if this catches everything. Especially not if there are transitive `collocate_with`s. But at least my new test passes, and the old test `test_SplitLayer_search` in which split layer should not be optimized out also passes.

returnn/tf/layers/rec.py
```diff
@@ -2904,6 +2904,12 @@ def output_can_move_out(layer, _collocated=None):
         if layer in other_layer.dependencies:
           return False
         if other_layer.name in layer.collocate_with:
+          # layer is collocated with other_layer.
+          # we can move out layer if other_layer has no dependencies on any layer in loop but on the sibling layers
+          # that themselves are collocated with other_layer
+          other_layer_collocated_with = [
+            sibling for sibling in layers_in_loop if other_layer.name in sibling.collocate_with]
+          _collocated.update(other_layer_collocated_with)
```
albertz, line 2912: I think the logic direction is wrong. `x in y.collocate_with` says that `y` should be collocated with `x`. So if `x` must be inside, `y` must be as well. But it has no implications on `x`. `x` could be inside or outside.

Zettelkasten: The variable name is misleading. I wanted this set to be the set of layers that should be collocated with `other_layer` (with the notion that they are "siblings" of `layer`).
`layers_collocated_with_other_layer` or so would be better, I'll change it.

albertz: > I think the logic direction is wrong. `x in y.collocate_with` says that `y` should be collocated with `x`. So if `x` must be inside, `y` must be as well. But it has no implications on `x`. `x` could be inside or outside.

Sorry, I was partly wrong here. It has implications on `x` as well. It goes both ways. It means they belong together. And if `collocate_with` is incorrectly set, i.e. not possible to have both collocated together, this should be an error.

So I think the logic of your code is partly correct. However, it should also check `can_move_out` for all the other collocated layers. So it is wrong to just directly call `_collocated.update`.

returnn/tf/layers/rec.py
```diff
@@ -2904,6 +2904,13 @@ def output_can_move_out(layer, _collocated=None):
         if layer in other_layer.dependencies:
           return False
         if other_layer.name in layer.collocate_with:
+          # layer is collocated with other_layer.
+          # we can move out layer if other_layer has no dependencies on any layer in loop but on the sibling layers
+          # that themselves are collocated with other_layer
+          layers_collocated_with_other_layer = [
+            sibling for sibling in layers_in_loop if other_layer.name in sibling.collocate_with]
+          assert layer in layers_collocated_with_other_layer
+          _collocated.update(layers_collocated_with_other_layer)
```
albertz, line 2913: See my added comment on the earlier code. I think calling `_collocated.update` directly without checking `output_can_move_out` is wrong.

We should have some similar code like `if layer.name in other_layer.collocate_with:` (the other way around). I think this would cover what you have here.

Also, I think we might/should move that `if layer in other_layer.dependencies:` below. And the `if other_layer in _collocated:` also after the `collocate_with` logic.

So I think this order:
```
if other_layer.name in layer.collocate_with: ...
if layer.name in other_layer.collocate_with: ...
if other_layer in _collocated: ...
if layer in other_layer.dependencies: ...
```

albertz:
Also adding Patrick as a reviewer here as he also worked on this code before. (Although I think it still needs some update, as I commented.)

patrick-wilken:
Before ea35a10ef `_construct_template.GetLayer.__call__()` contained this:

```
        if '/' in name:
          # this is probably a path to a sub-layer
          root_name = name.split('/')[0]
          # Get the root layer. Note, this will also add the root layer (and following layers) as a dependency
          # to ConstructCtx.layers[-1] which is what we want (see below).
          root_layer = lself.__call__(
            ("prev:%s" % root_name) if is_prev_time_frame else root_name)
          sub_layer = root_layer.get_sub_layer('/'.join(name.split('/')[1:]))  # get the sub-layer from the root-layer
          if sub_layer:  # get_sub_layer returns None by default (if sub-layer not found)
            # add to templates so we will collect output in self.get_output if this is an output layer
            if isinstance(sub_layer, _TemplateLayer):
              self.layer_data_templates[name] = sub_layer
              if ConstructCtx.layers:
                # Add the sub-layer to the dependencies just so we will visit it in self._move_outside_loop().
                # Note, we don't add dependencies to the sub-layer, instead, the dependency graph continues via the root
                # layer (see above).
                ConstructCtx.layers[-1].add_dependency(sub_layer, is_prev_time_frame=is_prev_time_frame)
                if sub_layer not in ConstructCtx.partially_finished:
                  ConstructCtx.partially_finished.append(sub_layer)
            return sub_layer
```

So the dependency graph skipped the sub-layers and directly added the root layer as a dependency. And then later added the sub-layers as a dependency of the root layer (wrong direction (?)) just so they are visited at all in `self._move_outside_loop()`. I did this with sub-layers like those of `ChoiceLayer` and `SplitLayer` in mind that are just an additional output of the root. And I think optimization of these layers worked that way. But not for optimization of individual layers in subnetworks of course, so this was changed...

I think @albertz is right, the other collocated layers have to be checked for dependencies.
Logically, what needs to happen is that all layers connected via `collocated_with` need to be treated as one node in the graph, depending on the union of the individual dependencies. But it's probably tricky to implement as `collocated_with` is only given in one direction. I will think about this... :smile

albertz:
> And then later added the sub-layers as a dependency of the root layer (wrong direction (?))

Yea, the dependency concept of layers + their sub layers doesn't really make sense. Dependency can only go one direction. I think the collocation concept (undirected) is the right approach here.

> Logically, what needs to happen is that all layers connected via `collocated_with` are treated as one node in the graph, depending on the union of the individual dependencies. But it's probably tricky to implement as `collocated_with` is only given in one direction. I will think about this... 😄

My suggested code above would already handle that, or not?

patrick-wilken:
Like this?:

```
      for other_layer in layers_in_loop:
        if other_layer in _collocated:
          continue
        if other_layer.name in layer.collocate_with or layer.name in other_layer.collocate_with:
          if output_can_move_out(other_layer, _collocated=_collocated):
            _collocated.add(other_layer)
            continue  # this case is ok, maybe we can move all collocated layers out
          else:
            return False
        if layer in other_layer.dependencies:
          return False
      return True
```

Works for the test case, but maybe I missed something. `if other_layer in _collocated:` has to come first in this version to avoid infinite recursion.

albertz:
> Like this?:
> 
> ```
>       for other_layer in layers_in_loop:
>         if other_layer in _collocated:
>           continue
>         if other_layer.name in layer.collocate_with or layer.name in other_layer.collocate_with:
>           if output_can_move_out(other_layer, _collocated=_collocated):
>             _collocated.add(other_layer)
>             continue  # this case is ok, maybe we can move all collocated layers out
>           else:
>             return False
>         if layer in other_layer.dependencies:
>           return False
>       return True
> ```

Ah yes, you can even combine both cases.
But your code is already too complicated. You can almost leave the existing code. You don't need the `_collocated.add` (it is covered by the recursive `output_can_move_out` call).
So:
```
if other_layer in _collocated:
  continue
if other_layer.name in layer.collocate_with or layer.name in other_layer.collocate_with:
  if output_can_move_out(other_layer, _collocated=_collocated):
    continue  # this is ok, we can move all out
  else:
    return False
if layer in other_layer.dependencies:
  return False
```

Zettelkasten:
Thanks you both for your input.
Wow, the solution you came up with looks great! Super clean, thanks a lot for working on this!

albertz (approved):

patrick-wilken:
> You don't need the _collocated.add

True, but it saves some recursive calls, right? Anyway, this is not critical.

albertz:
The recursive calls are needed in any case for the can-move-out logic.


Patrick Wilken ***@***.***> schrieb am So., 11. Juli 2021,
15:07:

PR #561, opened 2021-07-09
Fix unused TF operations in Batchnorm

JackTemaki:
A student worker at AppTek found that using batch norm during search has a lot of computational overhead (up tp 30%), and the graph includes nodes that should not be needed during search, as e.g. re-estimating the variance and mean.

This PR fixes this issue, reducing the computational share of the BN from 30% to 13% in this specific case. The problem is that TF does not detect multiplications with 0 automatically, and will still run the full graph, thus, zero cases need to be handled explicitely. 

The adding the epsilon also had a share of 2% computation, so I also disabled the addition when the epsilon is zero.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -1271,17 +1271,22 @@ class LayerBase(object):
       else:
         sample_variance = updated_sample_variance
       # If train or if force_sample, use default use_sample=0.0, otherwise use_sample=1.0.
-      if self.network.train_flag is not False or force_sample:
-        if force_sample:
-          pass  # leave use_sample as-is
-        else:
-          use_sample = tf.where(self.network.train_flag, use_sample, 1.0)
+      skip_updates = self.network.train_flag is False and update_sample_only_in_training and not force_sample
+      if skip_updates:
+        mean = sample_mean
+        variance = sample_variance
       else:
-        use_sample = 1.0
-      mean = (1. - use_sample) * mean + use_sample * sample_mean
-      variance = (1. - use_sample) * variance + use_sample * sample_variance
-      bn = (data.placeholder - mean) * tf_compat.v1.rsqrt(variance + epsilon)
-      if delayed_ops:
+        if self.network.train_flag is not False or force_sample:
+          if force_sample:
+            pass  # leave use_sample as-is
+          else:
+            use_sample = tf.where(self.network.train_flag, use_sample, 1.0)
+        else:
+          use_sample = 1.0
+        mean = (1. - use_sample) * mean + use_sample * sample_mean
+        variance = (1. - use_sample) * variance + use_sample * sample_variance
+      bn = (data.placeholder - mean) * tf_compat.v1.rsqrt(tf_util.optional_add(variance, epsilon))
+      if not skip_updates and delayed_ops and (isinstance(use_sample, tf.Tensor) or use_sample != 0.0):
         for op in delayed_ops:
           # Make sure we update after we calculated the batch norm.
           tf_util.add_control_input(op, control_input=bn.op)


returnn/tf/layers/base.py
```diff
@@ -1271,17 +1271,23 @@ def batch_norm(self, data,
...
+        if self.network.train_flag is not False or force_sample:
+          if force_sample:
+            pass  # leave use_sample as-is
+          else:
+            use_sample = tf.where(self.network.train_flag, use_sample, 1.0)
+        else:
+          use_sample = 1.0
+        mean = (1. - use_sample) * mean + use_sample * sample_mean
+        variance = (1. - use_sample) * variance + use_sample * sample_variance
+      variance_with_epsilon = variance if epsilon == 0.0 else variance + epsilon
```
albertz, line 1288: We have `opt_add` for that. And I personally would prefer to keep the code shorter and do it directly in the line `bn = ...`.

returnn/tf/layers/base.py
```diff
@@ -1271,17 +1271,23 @@ def batch_norm(self, data,
...
+      bn = (data.placeholder - mean) * tf_compat.v1.rsqrt(variance_with_epsilon)
+      if not skip_updates and delayed_ops and use_sample != 0.0:
```
albertz, line 1290: `use_sample != 0.0` is incorrect, as it might be a TF tensor here.

albertz:
> The adding the epsilon also had a share of 2% computation, so I also disabled the addition when the epsilon is zero.

Would that ever be the case? I'm not sure if it makes sense to have eps=0. And I'm not sure we should optimize for cases which should/would never be used.

curufinwe:
> > The adding the epsilon also had a share of 2% computation, so I also disabled the addition when the epsilon is zero.
> 
> Would that ever be the case? I'm not sure if it makes sense to have eps=0. And I'm not sure we should optimize for cases which should/would never be used.

eps is only needed during training to prevent diverging models from dividing by zero here. if the model truly has something close enough to zero in the pre-computed variance we could update the stored variance and not perform this addition of a fixed term.

vieting (approved):

returnn/tf/layers/base.py
JackTemaki: I think this is correct. When use_sample is a tensor, we can not optimize this, as it can have any value. So only when we "know" that use_sample is actually fixed 0.0 this step can be skipped. But for the originally intended case this is not needed, so if you want I can remove it.

albertz: You cannot do `if tf_tensor:` when `tf_tensor` is a `tf.Tensor` because `if` needs a `bool` (or sth which can be casted to `bool`). This is not eager-mode, so it would need a `session.run`.

JackTemaki: But why would I want to do that in the first place? I do not understand your comment.

albertz: I did not write this code. I don't want to do that. But the code does this.

`if ... and use_sample != 0.0:` is exactly the case I described.

JackTemaki: Nowhere in the code is `if tf_tensor`. What I can have is `tf_tensor != 0.0` which is True, and thus it behaves like I want it to behave.

albertz: > Nowhere in the code is `if tf_tensor`. What I can have is `tf_tensor != 0.0` which is True, and thus it behaves like I want it to behave.

You have this code here:
`if ... and use_sample != 0.0:`

Which is exactly this case. `use_sample != 0.0` is a `tf.Tensor`.

Actually it depends on the TF version, and might raise an exception if `use_sample` is a `tf.Tensor`.

Why are we arguing about this? Just do not do an `if`-check on a `tf.Tensor`.

Write the code in a way that does not do this. E.g. add some `not isinstance(use_sample, tf.Tensor)`. Then it would be fine. (But not sure if this is what you really want here.)

JackTemaki: Why should I write it more complicated than needed?

TF1:
```
In [10]: import tensorflow as tf

In [11]: tf.compat.v1.disable_eager_execution()

In [12]: a = tf.compat.v1.zeros(())

In [13]: a
Out[13]: <tf.Tensor 'zeros_1:0' shape=() dtype=float32>

In [14]: a != 0.0
Out[14]: True
```

TF2:
```
In [2]: import tensorflow as tf

In [3]: a = tf.zeros(())

In [4]: a
Out[4]: <tf.Tensor: shape=(), dtype=float32, numpy=0.0>

In [5]: a != 0.0
Out[5]: <tf.Tensor: shape=(), dtype=bool, numpy=False>

In [6]: bool (a != 0.0)
Out[6]: False
```

So the check is "can the variable be other than zero". Which is always True for graph mode, because I can not look into the Tensor. And might be "True" in eager mode when I can look inside. So no matter the TF version and mode this will evaluate as I want it to be.

But I do not care about this case that is covered by this (which is `force_sample=True` and `use_sample=0.0`), so I will just delete the check.

albertz: I think there were other TF versions where this resulted in exceptions, or unexpected behavior, or also a tensor in graph mode (I definitely had this case).

Also, `a != 0.0` reads as if you are checking if it is not equal to 0.0 which is not true if this is a tensor.

If you want to check `not isinstance(use_sample, tf.Tensor) and use_sample != 0.0`, then write that (more explicitly).

JackTemaki: > I think there were other TF versions where this resulted in exceptions, or unexpected behavior, or also a tensor in graph mode

At least since 1.13 this works, but I understand the concerns about readability. I was just very confused because you mentioned I would do an `if` check directly on a tensor and I did not understand what you were getting at.

returnn/tf/layers/base.py
```diff
@@ -12,7 +12,7 @@
 import returnn.tf.compat as tf_compat
 import returnn.tf.util.basic as tf_util
 from returnn.tf.util.data import Data, SearchBeam
-from returnn.tf.util.basic import OutputWithActivation, CustomUpdate, reuse_name_scope
+from returnn.tf.util.basic import OutputWithActivation, CustomUpdate, reuse_name_scope, optional_add
```
albertz, line 15: I would prefer to not import too much here, and rather use `tf_util.optional_add` below.

JackTemaki:
I squashed the fix commits, I think everything is okay now.

albertz (approved):

Issue #331, opened 2020-08-11
Arbitrary Data Format for HDFDataset


JackTemaki:
It seems to me that the `HDFDataset` does only support data streams with a single time-axis, and not data with no or multiple time axes. I know that I can still add data without time axis, by just using a sequence length of one, and then remove the time axis in a layer in the network.

Are there any possibilities right now to load data with 2 time axes (e.g. Images) or also 1 time axis and 1 fixed spatial dimension into RETURNN? And if not, are there any plans yet to change that?

albertz:
Yes, I do that a couple of times, and there are also some demos showing how to do that. You can simply flatten multiple dynamic axes into a single axis. When you dump to HDF (e.g. via `HDFDumpLayer`), this will be done automatically, and it will also save the information needed to unflatten it later. Then there is the `UnflattenNdLayer`, which unflattens the input. Search the tests for `unflatten_nd` to see some demos.

For fixed spatial dimensions, we do that very often, in a similar way. In that case, we merge it with the feature dim. E.g. when Sprint adds a window. Then you can simply use `SplitDimsLayer`.

JackTemaki:
Okay. I would still leave this Issue open, as it would be nice to support more data-formats in the datasets itself in the future, without having to rely on extra layers.

albertz:
I agree, we should simplify it for the user as far as possible. `HDFDumpLayer` and `HDFDataset` should complement each other in this case. (We should be a bit careful to not break old setups, though, when we change the behavior now.)

Another use case which came up a couple of times is to store a search beam in the HDF, e.g. N target hypotheses for one sequence. E.g. [this StackOverflow question](https://stackoverflow.com/questions/63411653/loading-ground-truth-and-hypotheses-as-training-target) (@npwynands @michelwi).

albertz:
For reference, [here](https://github.com/rwth-i6/returnn/discussions/434) is some discussion relevant for this.

albertz:
A note: I think our `Dataset` API itself is not really flexible enough.

The `Dataset.get_data_shape` function currently implies that there is always exactly one dynamic time axis in the beginning. So this is not only about `HDFDataset` but also about `Dataset` itself.

[...]

PR #562, opened 2021-07-13
HDFDumpLayer: Add batch dim to "extra" inputs

Zettelkasten:
I had `extra` inputs of a HDF dump layer which did not have any batch dim.
Here I change it so that all `extra` inputs without batch dim get an unbroadcasted batch dim of the main input.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -7677,7 +7677,11 @@ class HDFDumpLayer(LayerBase):
     self.filename = filename
     if extra is None:
       extra = {}
-    extra = {key: layer.output.copy_as_batch_spatial_major() for (key, layer) in extra.items()}
+    extra = {key: layer.output for (key, layer) in extra.items()}
+    extra = {
+      key: output.copy_add_batch_dim(0, batch=data.batch)
+      if not output.have_batch_axis() else output for (key, output) in extra.items()}
+    extra = {key: output.copy_as_batch_spatial_major() for (key, output) in extra.items()}
     self.extra = extra  # type: typing.Dict[str,Data]
     self.dump_whole_batches = dump_whole_batches
     self.num_seqs_written = 0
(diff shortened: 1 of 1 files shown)


patrick-wilken:
Yep, this is necessary.  Related to #331. Also here, we are limited by this `num_outputs = [dim, num_dims]` shape specification used by `Dataset` and also within the HDF file (`targets/size`). If we could specify `"batch_dim": None` etc. in the HDF file, we could implement it in a way that does not require unbroadcasting and therefore wasting space.

returnn/tf/layers/basic.py
```diff
@@ -7677,7 +7677,11 @@ def __init__(self, filename, extra=None, dump_whole_batches=False, labels=None,
     self.filename = filename
     if extra is None:
       extra = {}
-    extra = {key: layer.output.copy_as_batch_spatial_major() for (key, layer) in extra.items()}
+    extra = {key: layer.output for (key, layer) in extra.items()}
+    extra = {
+      key: output.copy_add_batch_dim(0, batch=data.batch)
+      if not output.have_batch_axis() else output for (key, output) in extra.items()}
```
albertz, line 7683:
```suggestion
      if not output.have_batch_axis() else output
      for (key, output) in extra.items()}
```

albertz:
> If we could specify `"batch_dim": None` etc. in the HDF file, we could implement it in a way that does not require unbroadcasting and therefore saves space.

No. Our dataset concept has one entry per sequence. If you have some layer without batch dim, it means the same value/tensor will be used (broadcasted) for all sequences in the batch. What do you think the dataset (no matter if HDF or whatever else) should save per sequence (seq idx)?

> Related to #331. Also here, we are limited by this `num_outputs = [dim, num_dims]` shape specification used by `Dataset` and also within the HDF file (`targets/size`).

No, this is not really related (because see above). And also, I think I already mentioned, if `num_outputs` of the `Dataset` is really a problem, this is kind of trivial to fix / change / extend. We should not keep repeating that it is a problem if it is so. It probably would have been simpler to already fix this. But anyway, this is unrelated here.

albertz (approved):

patrick-wilken:
A `Dataset` returns data per sequence and cannot provide a constant, I know. But I was talking about the HDF file structure. Per data key we store one big 1-dim array of values. What it represents is inferred from "metadata" (sequence lengths and the `size` field). I'm just saying it would be conceivable to add the information about a missing batch dim in the file and then interpret the whole data array for a given key as a constant sequence, and thus return it for each sequence index.

Issue #514, opened 2021-05-21
Disallow optimizer as boolean parameters


JackTemaki:
Discussion related to #508 

Right now, some of the optimizers can be enabled by using their name as boolean paramater. This is inconsistent as not all available optimizers are accessible this way. The correct and consistent way should be to allow only the use of the `optimizer` parameter as future behaviour.

PR #563, opened 2021-07-13
behavior version 2: disallowing boolean optimizer

JackTemaki:
related to #514


JackTemaki:
I just saw that we can not do this yet, as then "nadam" can not be used anymore.

The changes from #526 would be needed before.

albertz:
(Btw, when you fix some GitHub issue, then in the commit message, in the details section, you should add sth like "fix #514". Once this gets merged, it would automatically close the corresponding issue.)

returnn/tf/updater.py
```diff
@@ -534,6 +535,8 @@ def _create_default_optimizer(self):
       assert isinstance(optim_config, (dict, str))
       assert "class" in optim_config
       optimizer = self._create_optimizer(optim_config)
+    elif BehaviorVersion.get() >= 2:
+      raise Exception("Please definine an optimizer specifically via the 'optimizer=...' parameter")
```
mmz33, line 539: typo: `define` instead of `definine`

mmz33 (changes requested):

PR #565, opened 2021-07-21
Dataset attributes cleanup

patrick-wilken:
Some cleanups for the Dataset base class.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -132,7 +132,6 @@ class Dataset(object):
     self.labels = {}  # type: typing.Dict[str,typing.List[str]]
     self.weights = {}
     self.nbytes = 0
-    self.num_running_chars = 0  # CTC running chars.
     self._num_timesteps = 0
     self._num_codesteps = None  # type: typing.Optional[int]  # Num output frames, could be different from input, seq2seq, ctc.  # nopep8
     self._num_seqs = 0
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -131,7 +131,6 @@ class Dataset(object):
     self.timestamps = None
     self.labels = {}  # type: typing.Dict[str,typing.List[str]]
     self.weights = {}
-    self.nbytes = 0
     self._num_timesteps = 0
     self._num_codesteps = None  # type: typing.Optional[int]  # Num output frames, could be different from input, seq2seq, ctc.  # nopep8
     self._num_seqs = 0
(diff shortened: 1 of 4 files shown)


patrick-wilken:
Another cleanup candidate would be `Dataset.get_ctc_targets()` (and anything related to CTC in generic code actually, e.g. `DatasetSeq.ctc_targets`). It is only used by the Theano backend and only provided by `HDFDataset`. Is anyone still using it? If yes, could at least be disallowed via a `behavior_version`.

albertz:
Sorry for the late response.

I wonder, you move stuff from `Dataset` to `CachedDataset` or `HDFDataset`.
I assume it is never used anywhere for the generic `Dataset`. (Right? I have not checked this. I assume you have.)
But then, is it used at all? The `CachedDataset` or `HDFDataset` actually use this?

> Another cleanup candidate would be `Dataset.get_ctc_targets()` (and anything related to CTC in generic code actually, e.g. `DatasetSeq.ctc_targets`). It is only used by the Theano backend and only provided by `HDFDataset`. Is anyone still using it? If yes, could at least be disallowed via a `behavior_version`.

`behavior_version` is for the user config, not for internal RETURNN code.

Maybe the code can be moved to be purely on the Theano side where it is used, such that it is not in the generic dataset code anymore.

patrick-wilken:
`CachedDataset.timestamps` is only used in `theano/engine_task.py` via `get_times()`, yes. But it is read from a special `times` field in the HDF file, so this should probably stay part of `HDFDataset` - if we still need it.
`nbytes` is used by the caching mechanism of `CachedDataset`/`HDFDataset`.

returnn/datasets/basic.py
```diff
@@ -736,13 +721,6 @@ def has_ctc_targets(self):
     """
     return False
 
-  # noinspection PyMethodMayBeStatic
-  def get_max_ctc_length(self):
-    """
-    :rtype: int
-    """
-    return 0
-
```
albertz, line 745: In the comments, you suggest to also clean up CTC related stuff, but it sounds as if this is not done yet. However, here you partially remove some of it (and likely already break old Theano CTC setups by that). Is this on purpose? I don't quite understand why to remove only some part of it and not all (e.g. why not `has_ctc_targets` above?), or otherwise keep it as it is?

patrick-wilken: `get_max_ctc_length()` was only used here:
```
device.alloc_data(shapes=shapes, max_ctc_length=dataset.get_max_ctc_length())
```
(in `theano/engine_util.py`)
But it always returned 0, which is also the default value for the `max_ctc_length` parameter. So this should not break anything, except maybe for custom datasets which actually properly implement this function. I doubt that anyone uses one, but I can't be sure.

albertz (approved):

PR #567, opened 2021-08-04
Test cases and example fixes for tf_compile_graph with recurrent step

JackTemaki:
Tries to fix issues when compiling with `--rec_step_by_step` by doing the following changes:

 - Adapted the code to the new internal structure of the rec layer
 - Resolved the dependency issue on the target length for the `cond` state var
 - Added two test cases with a simple search decoder, one for default compilation, one for compiling with rec-step-by-step

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -994,8 +994,6 @@ class _SubnetworkRecCell(object):
     self.parent_rec_layer = None  # type: typing.Optional[RecLayer]
     self.parent_net = parent_net
     self.net_dict = safe_deep_copy(net_dict)
-    self.rec_layer_name = rec_layer_name
-    self.parent_get_layer = parent_get_layer
     from returnn.tf.network import TFNetwork, ExternData, LossHolder
     from returnn.tf.util.data import ControlFlowContext
     control_flow_ctx = ControlFlowContext(
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -994,6 +994,8 @@ class _SubnetworkRecCell(object):
     self.parent_rec_layer = None  # type: typing.Optional[RecLayer]
     self.parent_net = parent_net
     self.net_dict = safe_deep_copy(net_dict)
+    self.rec_layer_name = rec_layer_name
+    self.parent_get_layer = parent_get_layer
     from returnn.tf.network import TFNetwork, ExternData, LossHolder
     from returnn.tf.util.data import ControlFlowContext
     control_flow_ctx = ControlFlowContext(
@@ -1053,6 +1055,8 @@ class _SubnetworkRecCell(object):
 
   def set_parent_layer(self, parent_rec_layer):
     """
+    Only called from RecLayer._get_cell to define itself as parent
+
     :param RecLayer parent_rec_layer:
     """
     assert parent_rec_layer.network is self.parent_net
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1,0 +1,82 @@
+"""
+Test cases for different tools
+"""
+
+from __future__ import print_function
+
+import tempfile
+import _setup_test_env  # noqa
+from subprocess import Popen, PIPE, STDOUT, CalledProcessError
+import os
+import sys
+
+
+py = sys.executable
+print("Python:", py)
+
+
+def run(*args):
+  args = list(args)
+  print("run:", args)
+  # RETURNN by default outputs on stderr, so just merge both together
+  p = Popen(args, stdout=PIPE, stderr=STDOUT)
+  out, _ = p.communicate()
+  if p.returncode != 0:
+    print("Return code is %i" % p.returncode)
+    print("std out/err:\n---\n%s\n---\n" % out.decode("utf8"))
+    raise CalledProcessError(cmd=args, returncode=p.returncode, output=out)
+  return out.decode("utf8")
+
+
+###############################
+# Tests for compile_tf_graph.py
+###############################
+
+rec_decoder_config = """
+#!rnn.py
+network = {
+    "enc0": {"class": "linear", "activation": "sigmoid", "n_out": 3},
+    "enc1": {"class": "reduce", "mode": "max", "axis": "t", "from": "enc0"},
+    "output": {
+      "class": "rec", "from": [], "max_seq_len": 10, "target": "classes",
+      "unit": {
+        "embed": {"class": "linear", "from": "prev:output", "activation": "sigmoid", "n_out": 3},
+        "prob": {"class": "softmax", "from": ["embed", "base:enc1"], "loss": "ce", "target": "classes"},
+        "output": {"class": "choice", "beam_size": 4, "from": "prob", "target": "classes", "initial_output": 0},
+        "end": {"class": "compare", "from": "output", "value": 0}
+      }
+    },
+    "decision": {"class": "decide", "from": "output", "loss": "edit_distance"}
+}
+num_inputs = 5
+num_outputs = 3
+use_tensorflow = True
+"""
+
+
+def test_compile_tf_graph_basic():
+  tmp_dir = tempfile.mkdtemp()
+  with open(os.path.join(tmp_dir, "returnn.config"), "wt") as config:
+    config.write(rec_decoder_config)
+  args = [
+    "tools/compile_tf_graph.py",
+    "--output_file",
+    os.path.join(tmp_dir, "graph.metatxt"),
+    os.path.join(tmp_dir, "returnn.config")
+  ]
+  run(*args)
+
+
+def test_compile_tf_graph_recurrent_step():
+  tmp_dir = tempfile.mkdtemp()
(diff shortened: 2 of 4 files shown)


tools/compile_tf_graph.py
```diff
@@ -461,13 +460,25 @@ def add_stochastic_var(self, name):
 
   def _get_cell(self, unit, unit_opts=None):
     """
-    :param str|dict[str] unit:
+    Replace the unit by an identical SubnetworkRecCellSingleStep instance
+    Only allows for SubnetworkRecCell, not strings as the super method does
```
albertz, line 464: We should not make a new instance/copy.
I'm pretty sure there is several code which expects that it stays the same instance.
I'm also pretty sure such copy could potentially introduce some very nasty edge cases and buggy behavior.

Rather, this `RecStepByStepLayer` should directly create a `SubnetworkRecCellSingleStep` instance in the first place, instead of a `_SubnetworkRecCell`.
This creation happens in `RecLayer.transform_config_dict` via `subnet = _SubnetworkRecCell(...)`.
I just pushed 416209a2d8929667b7e8011d0413d0706d33ae0a which allows you to easily make use of `SubnetworkRecCellSingleStep` in the same way.

albertz: I think this whole function `_get_cell` can be removed here then.

tools/compile_tf_graph.py
```diff
@@ -633,11 +644,6 @@ def cond(i, net_vars, acc_tas, seq_len_info=None)
     # Note that this would be wrong in beam search.
     self.net.get_global_batch_info().dim = rec_layer.get_batch_dim_from_state()
 
-    with tf.name_scope("cond"):
-      rec_layer.create_state_var("cond", tf.constant(True))
-      res = cond(i, net_vars, acc_tas, seq_len_info)
-      rec_layer.set_state_var_final_value("cond", res)
-
```
albertz, line 640: You talked about this. But what exactly was the problem? Can you post some details about the error you get because of this?

Also, just removing this is not really an option anyway. We could maybe make it optional (some option like `add_cond` or so). But before we do that, we should understand what the real problem is. Because I would not expect that this should cause any problem. It should not be a problem to just have this always in the graph, such that it can be used when needed/wanted.

returnn/tf/layers/rec.py
```diff
@@ -985,6 +985,8 @@ def __init__(self, net_dict, source_data, time_dim_tag, rec_layer_name, parent_n
     self.parent_rec_layer = None  # type: typing.Optional[RecLayer]
     self.parent_net = parent_net
     self.net_dict = safe_deep_copy(net_dict)
+    self.rec_layer_name = rec_layer_name
+    self.parent_get_layer = parent_get_layer
```
albertz, line 989: I think copying `parent_get_layer` is a bad idea (can potentially lead to very nasty problems). You only need that for your copy of `_SubnetworkRecCell`? Then this can be removed after the other cleanup I suggested before.

tests/test_tools.py
```diff
@@ -0,0 +1,82 @@
...
+    os.path.join(tmp_dir, "graph.metatxt"),
+    "--rec_step_by_step",
+    "output",
+    os.path.join(tmp_dir, "returnn.config")
+  ]
+  result = run(*args)
+
+
+
+
```
albertz, line 82: I think there are too many empty lines at the end of the file.

tools/compile_tf_graph.py
albertz: (So you need to rebase the branch / PR on current master.)

albertz:
The tools-test file should also be added to the GitHub actions.

What is the state about the tests? They already run through?

JackTemaki:
> The tools-test file should also be added to the GitHub actions.
> 
> What is the state about the tests? They already run through?

The tests did run through with the current code, after rebasing and addressing your comments I will add them to the GitHub actions.

tools/compile_tf_graph.py
JackTemaki: I did not want to mess with this without your comments on that, so I will check now how that works with your change.

JackTemaki:
@albertz FYI:

The `cond` code causes the following error with the test case, without it works without crash now. I did not look into this yet.
```python
EXCEPTION
  File "/home/nick/github/returnn/tools/compile_tf_graph.py", line 750, in <module>
    line: main(sys.argv)
    locals:
      main = <local> <function main at 0x7f417d1b1310>
      sys = <local> <module 'sys' (built-in)>
      sys.argv = <local> ['tools/compile_tf_graph.py', '--output_file', '/tmp/tmpf6kfnzbg/graph.metatxt', '--rec_step_by_step', 'output', '/tmp/tmpf6kfnzbg/returnn.config'], len = 6, _[0]: {len = 25}
  File "/home/nick/github/returnn/tools/compile_tf_graph.py", line 679, in main
    line: RecStepByStepLayer.post_compile(
            rec_layer_name=args.rec_step_by_step, network=network, output_file_name=args.rec_step_by_step_output_file)
    locals:
      RecStepByStepLayer = <global> <class '__main__.RecStepByStepLayer'>
      RecStepByStepLayer.post_compile = <global> <bound method RecStepByStepLayer.post_compile of <class '__main__.RecStepByStepLayer'>>
      rec_layer_name = <not found>
      args = <local> Namespace(config='/tmp/tmpf6kfnzbg/returnn.config', train=0, eval=0, search=0, verbosity=4, summaries_tensor_name=None, rec_step_by_step='output', rec_step_by_step_output_file=None, output_file='/tmp/tmpf6kfnzbg/graph.metatxt', output_file_model_params_list=None, output_file_state_vars_list=None)
      args.rec_step_by_step = <local> 'output', len = 6
      network = <local> <TFNetwork 'root' train=False>
      output_file_name = <not found>
      args.rec_step_by_step_output_file = <local> None
  File "/home/nick/github/returnn/tools/compile_tf_graph.py", line 277, in RecStepByStepLayer.post_compile
    line: next_step_ops.append(var.final_op())
    locals:
      next_step_ops = <local> []
      next_step_ops.append = <local> <built-in method append of list object at 0x7f4266f7bcc0>
      var = <local> <StateVar 'cond', shape Data(name='cond', dtype='bool', batch_shape_meta=[]), initial <tf.Tensor 'output/rec/cond/Const:0' shape=() dtype=bool>>
      var.final_op = <local> <bound method RecStepByStepLayer.StateVar.final_op of <StateVar 'cond', shape Data(name='cond', dtype='bool', batch_shape_meta=[]), initial <tf.Tensor 'output/rec/cond/Const:0' shape=() dtype=bool>>>
  File "/home/nick/github/returnn/tools/compile_tf_graph.py", line 396, in RecStepByStepLayer.StateVar.final_op
    line: assert not path, "There should be no path from extern data to this final op value, but there is: %r" % (path,)
    locals:
      path = <local> [<tf.Operation 'output/rec/subnet_base/classes_transpose' type=Transpose>, <tf.Operation 'output/rec/subnet_base/check_data_len/Shape' type=Shape>, <tf.Operation 'output/rec/subnet_base/check_data_len/strided_slice' type=StridedSlice>, <tf.Operation 'output/rec/cond/loop_cond/i_less_max_seq_len' ...
AssertionError: There should be no path from extern data to this final op value, but there is: [<tf.Operation 'output/rec/subnet_base/classes_transpose' type=Transpose>, <tf.Operation 'output/rec/subnet_base/check_data_len/Shape' type=Shape>, <tf.Operation 'output/rec/subnet_base/check_data_len/strided_slice' type=StridedSlice>, <tf.Operation 'output/rec/cond/loop_cond/i_less_max_seq_len' type=Less>, <tf.Operation 'output/rec/cond/loop_cond/LogicalAnd' type=LogicalAnd>]
```

albertz:
You write this in the description of the PR:

> Tries to fix issues

So is this still incomplete? Or is the description wrong/outdated? Please describe in the description what the PR is supposed to do (when it is complete).

In comments, you can then write that it is still incomplete, or what the current problems are.

You can later also update the description of the PR, or mark comments as resolved or outdated. That makes it much easier to understand the current state.

> The `cond` code causes the following error ...

This is resolved now? Or still a problem? If it is still a problem, why is the PR marked as ready?

> tools-test file should also be added to the GitHub actions

This is not yet done? So it is not ready yet?

JackTemaki:
> You write this in the description of the PR:
> 
> > Tries to fix issues
> 
> So is this still incomplete? Or is the description wrong/outdated? Please describe in the description what the PR is supposed to do (when it is complete).

I would still call it "try", because this only covers one specific test cases. So maybe it works now for everyone, maybe not.

I will update the description to summarize what I did.

> 
> In comments, you can then write that it is still incomplete, or what the current problems are.
> 
> You can later also update the description of the PR, or mark comments as resolved or outdated. That makes it much easier to understand the current state.
> 
> > The `cond` code causes the following error ...
> 
> This is resolved now? Or still a problem? If it is still a problem, why is the PR marked as ready?

Yes, this is resolved.

> 
> > tools-test file should also be added to the GitHub actions
> 
> This is not yet done? So it is not ready yet?

You are right, I forgot to add the github action.

albertz (approved):

PR #568, opened 2021-08-09
More efficient get_seq_order_for_epoch()

patrick-wilken:
If training on huge datasets (e.g. 100M sequences), storing a full list of indices for the sequence order uses a significant amount of memory, especially when using a plain Python list and not a numpy array. With `default` sequence ordering this can be avoided by using a `range` instead. The full list is now only created after applying partition epoch, which usually means orders of magnitude less sequences.
I also made creating the sequence orderings a bit faster by using numpy. Again, with huge datasets that can make a difference. The random seed will be different, so to say, before and after this commit, not sure whether this is a problem.
Unfortunately, the code now sometimes goes back and forth between lists and arrays, but emulating `list.sort(key=...)` with numpy also wouldn't be too nice.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -402,6 +402,13 @@ class Dataset(object):
       reverse = -1 if self.seq_ordering == 'sorted_reverse' else 1
       seq_lens = [reverse * get_seq_len(i) for i in range(num_seqs)]
       seq_index = numpy.argsort(seq_lens, kind="stable")
+    elif self.seq_ordering.startswith('random'):
+      tmp = self.seq_ordering.split(':')
+      nth = int(tmp[1]) if len(tmp) > 1 else 1
+      # Keep this deterministic! Use fixed seed.
+      rnd_seed = (full_epoch - 1) // nth + 1
+      numpy.random.seed(rnd_seed)
+      seq_index = numpy.random.permutation(num_seqs)
     elif self.seq_ordering.startswith('sort_bin_shuffle'):
       # Shuffle seqs, sort by length, and shuffle bins (then shuffle seqs within each bin if sort_bin_shuffle_x2).
       assert get_seq_len
@@ -459,13 +466,6 @@ class Dataset(object):
         part.sort(key=get_seq_len, reverse=(i % 2 == 1))
         out_index += part
       seq_index = out_index
-    elif self.seq_ordering.startswith('random'):
-      tmp = self.seq_ordering.split(':')
-      nth = int(tmp[1]) if len(tmp) > 1 else 1
-      # Keep this deterministic! Use fixed seed.
-      rnd_seed = (full_epoch - 1) // nth + 1
-      numpy.random.seed(rnd_seed)
-      seq_index = numpy.random.permutation(num_seqs)
     else:
       assert False, "invalid batching specified: " + self.seq_ordering
     if self.unique_seq_tags:
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -397,14 +397,11 @@ class Dataset(object):
         seq_index[i::num] += i * (num_seqs // num)
     elif self.seq_ordering == 'reverse':
       seq_index = range(num_seqs - 1, -1, -1)
-    elif self.seq_ordering == 'sorted':
+    elif self.seq_ordering in ['sorted', 'sorted_reverse']:
       assert get_seq_len
-      seq_index = list(range(num_seqs))
-      seq_index.sort(key=get_seq_len)  # sort by length, starting with shortest
-    elif self.seq_ordering == "sorted_reverse":
-      assert get_seq_len
-      seq_index = list(range(num_seqs))
-      seq_index.sort(key=get_seq_len, reverse=True)  # sort by length, in reverse, starting with longest
+      reverse = -1 if self.seq_ordering == 'sorted_reverse' else 1
+      seq_lens = [reverse * get_seq_len(i) for i in range(num_seqs)]
+      seq_index = numpy.argsort(seq_lens, kind="stable")
     elif self.seq_ordering.startswith('sort_bin_shuffle'):
       # Shuffle seqs, sort by length, and shuffle bins (then shuffle seqs within each bin if sort_bin_shuffle_x2).
       assert get_seq_len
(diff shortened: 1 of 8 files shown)


patrick-wilken:
The reason why I'm working on this is that I started using multi-gpu training. In our current implementation the dataset is loaded once per GPU which makes reducing the memory footprint even more important.

patrick-wilken:
Converted to draft because I overlooked that what I actually needed was that `get_seq_order_for_epoch()` does not create the whole sequence list at all in case a range object would do. It works for me when I remove the `seq_index = list(seq_index)` line, but this will not work in general.

My use case is: I have several huge HDF files, I create one HDFDataset per file and then combine via `CombinedDataset` with sampling. I cannot afford to store sequence orderings for all full datasets as the total number of sequences is so high that it consumes too much memory. So I want to use "default" sequence ordering for the HDFDatasets, which can be represented by a python range, sample a constant amount of sequences from each of the HDFDatasets per epoch and only on CombinedDataset level create a sequence ordering list with e.g. 1M sequences.

I have thought about writing a `SequenceOrdering` class which would have the interface of a list, but internally would be more memory efficient if possible, mainly by using `range` for "default" and "reverse" sequence orderings. But somehow this seems to be too much code overhead for such a simple thing. Maybe allowing `get_seq_order_for_epoch()` return a `range` is easier after all...

albertz:
I haven't really looked into this too much yet. I just wanted to comment two ways how you could do shuffling on huge data:

- The way it is usually done in TensorFlow, on-the-fly. You have one (infinite) stream of data, and then a buffer, and keep shuffling the buffer. See our discussion on the TF dataset for this. It is an approximation but it works on infinitely large datasets.
- You could just do random sampling on the fly. So whenever some new seq is requested, you randomly generate one index. So one epoch will not guarantee that you visit all seqs. But after iterating a couple of epochs, this approximation should also be fine.

patrick-wilken:
Yes, what I'm doing is basically the first way. Except that HDFDataset is not iterator-like, however it has no overhead in representing an "almost infinite" map-like dataset because nothing is loaded on initialization.
With a few exceptions... 😄: the sequence ordering and a list of sequence start indices (which I was also able to get rid off by storing them inside the file instead of the sequence lengths, topic for a different PR...). So I now have a HDFDataset implementation where the memory consumption is really in no way dependent on the total amount of sequences in the file, so no limit to the dataset size. 😎

patrick-wilken:
Actually, I think, at no point in the code we use anything else than `len()` and `_getitem__` on the return value of `get_seq_order_for_epoch()`, so returning a range should already work. However we would have to change type docs to `collections.abc.Sequence` in many places...

patrick-wilken:
This should be acceptable now. The output of `get_seq_order_for_epoch()` is now marked as `typing.Sequence[int]`, it currently can be a list, a numpy array or a range, depending on what is the most efficient representation. In all our code we only use the `typing.Sequence` interface (`__getitem__`, `__len__` and iteration) for the return value, except for that one place in `CachedDataset` where the equality operator is used, which did not work for numpy arrays.

patrick-wilken:
I kind of forgot about this PR. @JackTemaki I did the requested changes, so please update the status.

returnn/datasets/basic.py
```diff
@@ -414,9 +414,9 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
         nth = 1
       else:
         nth = int(tmp[1])
-      rnd_seed = ((full_epoch - 1) // nth + 1) if full_epoch else 1
-      rnd = Random(rnd_seed + self.random_seed_offset)
-      rnd.shuffle(seq_index)  # Shuffle sequences.
+      rnd_seed = self._get_random_seed_for_epoch(epoch=epoch, num_epochs_fixed=nth)
+      numpy.random.seed(rnd_seed)
```
albertz, line 418: Do not set the global Numpy random seed. Rather, create an instance of `numpy.RandomState` or `random.Random`.

returnn/datasets/basic.py
```diff
@@ -414,9 +414,9 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
...
+      seq_index = numpy.random.permutation(num_seqs).tolist()  # type: typing.List[int]
```
albertz, line 419: Using `rnd.shuffle` is more efficient.

albertz: Ah sorry, this is using `num_seqs` as argument. No, I think then `rnd.permutation` is ok. But not the global Numpy rnd.

returnn/datasets/basic.py
```diff
@@ -425,18 +425,17 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
...
       for i in bin_ids:
         if i == bins - 1:
           part = seq_index[i * len(seq_index) // bins:][:]
         else:
           part = seq_index[i * len(seq_index) // bins:(i + 1) * len(seq_index) // bins][:]
         if self.seq_ordering.startswith('sort_bin_shuffle_x2'):
-          rnd.shuffle(part)  # Shuffle within the bin.
-        out_index += part
-      seq_index = out_index
+          numpy.random.shuffle(part)  # Shuffle within the bin.
```
albertz, line 436: Do not use the global Numpy rnd generator.

returnn/datasets/basic.py
```diff
@@ -425,18 +425,17 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
           bins = max(num_seqs // int(tmp[0][1:]), 2)
         else:  # the number of bins
           bins = int(tmp[0])
-      bin_ids = list(range(bins))
-      rnd.shuffle(bin_ids)  # Shuffle bins.
+      bin_ids = numpy.random.permutation(bins)  # Shuffle bins.
```
albertz, line 428: Do not use the Numpy global rnd generator.

returnn/datasets/basic.py
```diff
@@ -403,8 +403,8 @@ def get_seq_order_for_epoch(self, epoch, num_seqs, get_seq_len=None):
       nth = int(tmp[1]) if len(tmp) > 1 else 1
       # Keep this deterministic! Use fixed seed.
       rnd_seed = self._get_random_seed_for_epoch(epoch=epoch, num_epochs_fixed=nth)
-      numpy.random.seed(rnd_seed)
-      seq_index = numpy.random.permutation(num_seqs)
+      random_generator = numpy.random.default_rng(rnd_seed)
```
patrick-wilken, line 406: This is now the recommended way and works with numpy >= 1.17. Ok or rather use `numpy.random.RandomState`?

patrick-wilken: Ok, done.

albertz (approved):
Sorry for the delay. I think it's fine now.

albertz:
Merging now, as I think @JackTemaki concerns were also addressed (and he is on vacation currently).

PR #577, opened 2021-08-12
explicit dimension tags for Data

albertz:


albertz:
Note: This is just an initial preparation for the feature.

What we want in the end is that we can add some dim tag (without size placeholder) to `Data` and this will correctly be carried over in all layers and other operations. This is not the case. Even just `Data.copy` will forget about it.

There is currently a lot of code which assigns `Data.shape` directly. This makes the dim tags inconsistent.

We probably want a more generic `Data.shape` (or better the batched variant `Data.batch_shape`) but with `DimensionTag` instead of `int|None`.

Also some layers like `ConvLayer` introduce new sizes / dim tags in their `__init__`, but this logic should be moved to the `get_out_type_from_opts`. Like `EditDistanceTableLayer` in this PR here.

PR #578, opened 2021-08-16
Dim tag more

albertz:
Some follow up from #577.


albertz:
(Actually I wanted to push all unrelated commits to master. As I pushed just all, GitHub marked this PR as merged... I will open another PR...)

PR #579, opened 2021-08-18
Explicit Data dimension tags

albertz:
Follow-up from #577. Needed for #391.

This introduces the `dim_tags` argument and attribute for `Data`.

This disallows `Data.shape` and `Data.batch_dim_axis` assignments.

Much `Data` util code got adapted to consistently make use of dim tags.
Many layers got adapted to not operate on the shape but on dim tags instead.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -281,7 +281,7 @@ class LayerBase(object):
     # You are supposed to set self.output.{batch_dim_axis,time_dim_axis} explicitly,
     # as well as check the inputs if they are as you would suggest.
     # However, a good default is often to use the same as the input.
-    if all([k not in out_type for k in Data.SpecialAxesNames]):
+    if all([k not in out_type for k in Data.SpecialAxesNames]) and "dim_tags" not in out_type:
       if sources_data:
         out_type.setdefault("batch_dim_axis", sources_data.batch_dim_axis)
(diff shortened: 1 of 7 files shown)


albertz:
Some intermediate comments:

There are some tricky parts on this. It's not so much a problem for all RETURNN code itself, as we can just transfer directly to the new-style of defining `Data`.

However, the user defines `Data` also in various ways:

- `extern_data`
- `out_type` in layers
- Maybe custom layer, or other custom code

`Data.get_kwargs()` changed to not include `shape` anymore but `dim_tags` instead. But there is a lot of code (also mixing with `out_type`) which assumes `shape` in it, or which would additionally put `shape` into it and ignore `dim_tags`. This leads to inconsistencies or other bugs.

`Data.shape` or other assignments to a `Data` instance is not really supported anymore. We might be able to have some wrappers (as we also have already for `size_placeholder`). But this would not work correct in all cases, and be extremely ugly, and complicated. I'm not sure how much external code does this.

It is a high priority for RETURNN to not break user configs. At least for `extern_data` and `out_type`. Custom layers is another thing, which was never really recommended anyway. And fixes should be simple.

albertz:
Wow, finally all tests are passing. This still needs some cleanup though...

albertz:
Note that I don't expect you to do a full review but you might want to take a look in any case on this, as this does some deep changes to RETURNN (on `Data`).

Issue #569, opened 2021-08-09
`DotLayer` behavior unintuitive with rec automatic optimization


albertz:
(*Just a suggestion for now.*)
(*Also see [wiki on rec automatic optimization, special layer behavior](https://github.com/rwth-i6/returnn/wiki/Rec-automatic-optimization,-special-behavior-of-layers).*)

When `DotLayer` operates on values inside a rec loop (`RecLayer`), the `RecLayer` automatic optimization could move it + its inputs out of the loop, which would introduce a time axis on one or both of the inputs. To have the behavior correct, this extra axis must be added to the corresponding `var` option. The current solution is to use `T?` to have this working for both cases.

This is against the principle that the user should not need to think about the automatic optimization, and unintuive.

Possible solutions:

- Instead of specifying `var1`/`var2` for `DotLayer`, the user could specify the common shared axes (`common` or `share` or so). Currently this is implied/inferred. But when this `common` would be specified, we could infer `var1`/`var2`.
  One reason I did not add `common` originally because I found it slightly annoying that `B` would need to be specified then all the time. Or maybe `B` could always be included?
  I'm not sure if there are other problems.

- Somehow add special logic for the automatic optimization for `DotLayer`, to extend `var` of corresponding input?

Zettelkasten:
> - Instead of specifying `var1`/`var2` for `DotLayer`, the user could specify the common shared axes (`common` or `share` or so). Currently this is implied/inferred. But when this `common` would be specified, we could infer `var1`/`var2`.

If it was like this, then the user would instead need to specify `T?` as common axes when using a DotLayer inside a RecLayer which operates independent of the rec time axis.
This is different to the behavior of all other layers, like e.g. LinearLayer or GatherLayer which just assume that all not explicitly specified dims are "batch" dims.
I feel like specifying it differently here for DotLayer would hurt readability of configs because it's different from what all other layers expect you to specify.

albertz:
Ah yes, you are right. But my suggestion was not to remove the `var` option. It was just an additional way. The user either would provide `var` or `common`.

Btw, the use case I have in mind here are e.g.:
```
d[output + '_att_energy'] = {"class": "dot", "red1": -1, "red2": -1, "var1": "T", "var2": "T?",
               "from": ['base:' + output + '_att_key', output + '_att_query']}  # (B, H, enc-T, 1)
```
Or also (via #391):
```
[...]

PR #590, opened 2021-08-26
Tests for self attention using CumConcatLayer

Zettelkasten:
Some tests for #391.
I didn't run any of this yet (so most likely buggy) because there is no implementation.
It still needs `CumConcatLayer` from #589, but also the `common` arguments for `DotLayer` from #569.
Also, some way to set dim tags (e.g. `set_dim_tags: Dict[str, DimensionTag]` for `ReinterpretDataLayer`. I'm not sure, didn't we add this already somewhere?)

We might also want more tests than this: e.g. some with positional encoding.

### DIFF
--- a/tests/test_TFNetworkRecLayer.py
+++ b/tests/test_TFNetworkRecLayer.py
@@ -6837,6 +6837,148 @@ def test_RelativePositionalEncodingLayer():
     print(out)  # random...
 
 
+def _build_self_attention_layer(d, input, output, inside_rec_layer, query_axis, num_heads=8, key_dim=64,
+                                value_dim=64, dropout=0.0):
+  """
+  Essentially this does
+    d[output + '_att'] = {"class": "self_attention", "num_heads": num_heads,
+      "total_key_dim": num_heads * key_dim,
+      "n_out": num_heads * value_dim, "from": [input],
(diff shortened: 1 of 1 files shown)


Zettelkasten:
This should be ready now.

I left in `test_self_attention_optimize_out`, which essentially does the same as `test_reclayer_optimize_out_cum_concat_gen_self_att`. Should we remove that?

albertz:
There is also `test_generalized_non_rec_self_attention` already now.

I did not check in detail, is this really exactly the same, or just similar? What is the difference?
If there is some difference which makes sense for another test, then leave it. Otherwise remove it.

Zettelkasten:
> There is also `test_generalized_non_rec_self_attention` already now.

Oh okay, I missed that.

The test case in this PR also checks that `SelfAttentionLayer` does the same as the generalized self attention for the recurrent case (not just non-rec). Maybe we want to check that too.
But maybe we can just close this PR, it's a lot new code which originally was meant to be much more general. I think the test cases we already have are nicer because they use an explicit net dict and seem easier to debug if something goes wrong.

albertz:
Hm, I don't know. It might be nice as an example. I would say leave it.

albertz (approved):

PR #626, opened 2021-09-03
Dataset, make it possible to use a custom dataset object in the config

jvhoffbauer:
Make it possible to define a custom dataset object in the returnn config.

### DIFF
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -227,11 +227,9 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
-  elif config.is_typed(files_config_key) and isinstance(config.typed_value(files_config_key), Dataset):
-    # Config contains dataset object, load it
-    data = config.typed_value(files_config_key)
-    data.initialize()
-
+  elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
+    # Config contains callable, returning dataset object, load it
+    data = init_dataset(config.typed_value(files_config_key))
   else:
     config_str = config.value(files_config_key, "")
     data = init_dataset_via_str(config_str, config=config, cache_byte_size=cache_byte_size, **kwargs)
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -228,7 +228,6 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
   elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
-    # Config contains callable, returning dataset object, load it
     data = init_dataset(config.typed_value(files_config_key))
   else:
     config_str = config.value(files_config_key, "")
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -227,6 +227,11 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
+  elif config.is_typed(files_config_key) and isinstance(config.typed_value(files_config_key), Dataset):
+    # Config contains dataset object, load it
+    data = config.typed_value(files_config_key)
+    data.initialize()
+
   else:
     config_str = config.value(files_config_key, "")
     data = init_dataset_via_str(config_str, config=config, cache_byte_size=cache_byte_size, **kwargs)
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -228,7 +228,7 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
   elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
-    data = init_dataset(config.typed_value(files_config_key))
+    data = init_dataset(config.typed_value(files_config_key), default_kwargs=kwargs)
   else:
     config_str = config.value(files_config_key, "")
     data = init_dataset_via_str(config_str, config=config, cache_byte_size=cache_byte_size, **kwargs)
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1261,8 +1261,9 @@ def init_dataset(kwargs, extra_kwargs=None, default_kwargs=None):
   """
   assert kwargs
   if isinstance(kwargs, Dataset):
-    kwargs.initialize()
-    return kwargs
+    data = kwargs
+    data.initialize()
+    return data
   if callable(kwargs):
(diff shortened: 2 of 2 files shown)


returnn/datasets/basic.py
```diff
@@ -1261,6 +1261,7 @@ def init_dataset(kwargs, extra_kwargs=None, default_kwargs=None):
   """
   assert kwargs
   if isinstance(kwargs, Dataset):
+    kwargs.initialize()
```
jvhoffbauer, line 1264: I think, this is required for search to work properly. But I think someone with better knowledge of returnn should check. 
@albertz @patrick-wilken @dthulke  @JackTemaki

dthulke: Why is this only an issue for search? Afaik initialize isn't called somewhere else, so this should always cause an issue if the dataset passed to this method is not already initialised.

jvhoffbauer: It only crashed in search. But I do not know exactly, that is why I am asking.

albertz: What is the crash/error?

returnn/__main__.py
```diff
@@ -231,6 +231,11 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
+  elif config.is_typed(files_config_key) and isinstance(config.typed_value(files_config_key), Dataset):
```
dthulke, line 234: Maybe directly add support for callables, i.e. add:
```
or callable(config.typed_value(files_config_key))
```

returnn/__main__.py
```diff
@@ -231,6 +231,11 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
...
+    # Config contains dataset object, load it
+    data = config.typed_value(files_config_key)
+    data.initialize()
+
```
albertz, line 238: Why do you add the empty line? The code should be consistent. Either empty line before all `if`/`elif`, or none.

returnn/__main__.py
```diff
@@ -231,6 +231,11 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
+  elif config.is_typed(files_config_key) and isinstance(config.typed_value(files_config_key), Dataset):
+    # Config contains dataset object, load it
+    data = config.typed_value(files_config_key)
+    data.initialize()
```
albertz, line 237: Why do you need this here? I thought you would have a callable here.
So:
```suggestion
  elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
    data = init_dataset(config.typed_value(files_config_key))
```

albertz: Also, I would very much recommend to not directly have a `Dataset` instance inside your config. This could lead to strange problems.

returnn/datasets/basic.py
albertz: But also, I think this is probably just consistent to call this here. I think it also depends on the `Dataset` class. For many classes, this might also work without.

patrick-wilken: The datasets defined in the config via `train`, `dev` and `eval` are initialized globally in `__main__().load_data()`. The search dataset set via `search_data` is not. That's the difference. So yes, this is needed here. (Assuming we don't want to require the dataset in the config to already be initialized.) However, for consistency we should initialize all datasets with this line here, also "train" etc. See my other comment...

returnn/__main__.py
patrick-wilken, line 237: Using `init_dataset(data)` (from datasets.basic) would be a bit more consistent.

albertz: But as said, I think we should not allow to have a `Dataset` instance directly here. It should be a `callable`.

returnn/datasets/basic.py
```diff
@@ -1260,8 +1260,11 @@ def init_dataset(kwargs, extra_kwargs=None, default_kwargs=None):
   :rtype: Dataset
   """
   assert kwargs
-  if isinstance(kwargs, Dataset):
-    return kwargs
+  if callable(kwargs):
+    # Init dataset if it was a callable/lambda returning a dataset
+    data = kwargs()
+    data.initialize()
+    return data
```
albertz, line 1267: I don't understand. You have that `if callable(kwargs)` checks already here (just directly below). Why do you add it again? This doesn't make sense.

Also, why do you remove the `if isinstance(kwargs, Dataset):`? That might break some other code which made use of this. Don't remove that.

returnn/__main__.py
```diff
@@ -227,6 +227,9 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
+  elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
+    # Config contains callable, returning dataset object, load it
```
albertz, line 231: I find this comment to be quite redundant (and maybe even confusing).

jvhoffbauer:
Sorry, it should be correct now.

returnn/__main__.py
albertz:
```suggestion
```

returnn/__main__.py
```diff
@@ -227,6 +227,8 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
         kwargs["cache_byte_size"] = cache_byte_size
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
+  elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
+    data = init_dataset(config.typed_value(files_config_key))
```
albertz, line 231: This is missing `kwargs`.

returnn/__main__.py
```diff
@@ -228,7 +228,7 @@ def load_data(config, cache_byte_size, files_config_key, **kwargs):
     Dataset.kwargs_update_from_config(config, kwargs)
     data = init_dataset(kwargs)
   elif config.is_typed(files_config_key) and callable(config.typed_value(files_config_key)):
-    data = init_dataset(config.typed_value(files_config_key))
+    data = init_dataset(config.typed_value(files_config_key), default_kwargs=kwargs)
```
albertz, line 231: Doesn't `extra_kwargs` make more sense? What do you have in `kwargs` here?

dthulke: `kwargs` in this case are the globally defined dataset arguments from `Dataset.get_default_kwargs_eval(config=config)`.

In the case that the output of the callable is a Dataset, both `default_kwargs` and `extra_kwargs` are currently not used.

If the output of the callable is a dict, `extra_kwargs` would overwrite the values in the dict with these global default values. So `default_kwargs` should be correct here.

albertz: I just notice that there are more problems here. In case that the callable returns a dict, I think we want that it applies the logic from above, esp the logic involving `kwargs`.

albertz: Maybe it's easier if I just fix this in a follow-up PR.

albertz: Although, maybe it's not so important... What you miss here is the `Dataset.kwargs_update_from_config`. Not sure...

Issue #629, opened 2021-09-03
`DotLayer` reduce over dynamic axis should respect the seq mask


albertz:
`DotLayer` should respect the dyn size (or seq mask) when you reduce dynamic axes, just like `ReduceLayer` etc would do over dynamic axes. The `DotLayer` currently completely ignores this. This is fine when the user explicitly performs masking beforehand. E.g. `SoftmaxOverSpatialLayer` will do the right thing. However, otherwise it will be wrong in general. This is esp also a problem for the concept of dynamic axes with extended dynamic sizes.

Implementing this for `DotLayer` is not so nice, though, because this additional masking is not needed for the common case where there was a `SoftmaxOverSpatialLayer` before. The additional masking is not wrong, but it would make it a bit slower. So we definitely want to avoid this. But how can we know when this can be skipped? It could explicitly check whether one of the inputs comes from `SoftmaxOverSpatialLayer` but this would be very ugly, hacky, and also fail in some cases (e.g. by any automatic internal wrapping layers such as `SelectSearchSourcesLayer`). But is there a better more generic way? Somehow some way that the input can say "I'm already masked, padded values are 0, no need to mask again". Other layers like `ReduceLayer` could also use this information. The padded values are also relevant. For `DotLayer` we want 0. For `ReduceLayer` with `sum` or `avg` we want 0. For `ReduceLayer` with `max` or also `SoftmaxOverSpatialLayer` we want -inf. But then, I'm not sure if we are maybe making it too complicated in end.

_Originally posted by @albertz in https://github.com/rwth-i6/returnn/issues/391#issuecomment-906772624_

Zettelkasten:
Related to this and also to #627, I also just now noticed that it is a little weird/inconsistent that we have `red1` and `red2` as arguments. These axes really need to be the same, and also share the same dyn lengths.
But if I understood correctly it's also possible to have dim tags that are marked the same, but have different descriptions. Then that logic here would fail. I'm not sure if that's a real issue or if that's something we should even allow. The user could always just rename the axes before.

albertz:
> Related to this and also to #627, I also just now noticed that it is a little weird/inconsistent that we have `red1` and `red2` as arguments. These axes really need to be the same, and also share the same dyn lengths.

This is only very recent that we allow to specify dim tags. At least directly. A bit longer that we allow it by name. But then we also still do not guarantee to have them unique (which is another thing we want at some point).

[...]

PR #631, opened 2021-09-04
DotLayer, mask for dynamic axis

albertz:
Fix/implement #629.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1635,7 +1635,7 @@ class SoftmaxOverSpatialLayer(_ConcatInputLayer):
       By default, if dyn seq len exists, it uses it.
     :param bool log_space: if True, returns in log space (i.e. uses log_softmax)
     """
-    from returnn.tf.util.basic import where_bc
+    from returnn.tf.util.basic import where_bc, set_padding_info
     super(SoftmaxOverSpatialLayer, self).__init__(**kwargs)
     self.start = start
     self.window_start = window_start
@@ -1665,6 +1665,8 @@ class SoftmaxOverSpatialLayer(_ConcatInputLayer):
     self.output_before_activation = OutputWithActivation(
       energy, act_func=tf.nn.log_softmax if log_space else tf.nn.softmax)  # (...,T)
     self.output.placeholder = self.output_before_activation.y
+    if use_time_mask:
+      set_padding_info(self.output.placeholder, dim=self.output.dim_tags[axis], pad_value=0.)
     # Never allow inf in output, as softmax should remove all -inf values used for masking
     self.allow_inf_in_output = False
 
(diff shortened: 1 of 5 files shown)


albertz:
I was planning to add `padding_value` to `Data` (as discussed in #391). However, then I realized that the padding might be per dynamic dim, so rather a dict dim tag -> value.

Also, `padding_value` in `Data` has the problem that most existing code will just leave it as-is (via `copy` or `get_kwargs`) which is then incorrect. So it should better be the other way around, i.e. explicit. But still e.g. in `copy`? But even there, this might break the logic when `placeholder` gets reassigned. Except maybe when we reset it on every `placeholder` assignment. I'm not sure.

Or, we attach it to the `tf.Tensor` object directly. Like we do for other things. This is probably much easier and also more safe w.r.t. the behavior we want.

Zettelkasten:
> Or, we attach it to the `tf.Tensor` object directly. Like we do for other things. This is probably much easier and also more safe w.r.t. the behavior we want.

I was about to argue for something entirely different, but I think this is the cleanest solution overall.
I think it's important to distinguish the tasks between `get_out_data_from_opts` and `__init__` of a layer.
`get_out_data_from_opts` should really construct the template only (perhaps `get_out_template_from_opts` would be a better name?), and `__init__` then the placeholder (and stuff that is related to the placeholder).
If `padding_values` was a direct property of `Data`, it would not be directly clear if should be set in `get_out_data_from_opts` or in `__init__`. Conceptually, `__init__` makes more sense I think because this is really not part of the template, it's part of the placeholder itself. So making this a property of the tensor makes that super clear.
Another option would be to add this property to `LayerBase` instead, just like we do for right now for `allow_inf_in_output`. But that's inconvenient (all utility methods in `Data` couldn't automatically set this). I'd also argue that `allow_inf_in_output` should better be attached to the `placeholder` tensor itself for the same reasons.

One problem maybe that we have when we attach this to the tensor itself is that two `Data` objects might want to have the same placeholder, but different `padding_values`. For example if the user themselves is sure that some layer output is masked but RETURNN cannot figure this out, they might want to introduce a layer (or update `ReinterpretDataLayer`) to update this `padding_values` info. This layer would need a `tf.identity` op then. But that's fine I think because this is all after template construction within `__init__`.

albertz:
> One problem maybe that we have when we attach this to the tensor itself is that two `Data` objects might want to have the same placeholder, but different `padding_values`. For example if the user themselves is sure that some layer output is masked but RETURNN cannot figure this out, they might want to introduce a layer (or update `ReinterpretDataLayer`) to update this `padding_values` info.

I don't think it make sense that a tensor can have different padding values (per dyn dim tag).
(For simplicity, let's assume there is only a single dyn dim tag. Like in a standard [B,T,D] tensor.)
Either the padding value is undefined, unknown, or known to be a specific value. It cannot really be multiple different values.
It might make sense that the user can explicitly set this information (as you say, maybe via `ReinterpretDataLayer` or another new layer). But the logic should be: When it is unset, set it. When it is already set, assert that it is the same. Again with the reasoning that it cannot be sth different.
Or a layer like `SeqLenMaskLayer` will anyway explicitly overwrite the padding, and thus also set the padding value.

albertz:
Already merged now. But that doesn't mean we might want this to change in some way.

PR #640, opened 2021-09-10
Fix #637

### LINKED ISSUE #637: RecStepInfoLayer (`:i`) problem in unbroadcasting
It seems like the commit 7e3b939237e69a396a017e88bf836972f0d63e79 broke part of my config.
The config looks like this:

```
network = {
  "output": {
    "class": "rec", "from": "data",  # time-sync
    "include_eos": True, "back_prop": (task == "train") and train, "unit": {
      "segment_starts": { 
        "class": "switch", "condition": "prev:output_is_not_blank", "true_from": ":i",
        "false_from": "prev:segment_starts", "initial_output": 0
      },
      "segment_lens0": {"class": "combine", "kind": "sub", "from": [":i", "segment_starts"], "is_output_layer": True},
      "output_prob": {
        "class": "softmax", "from": "data:source", "target": "classes", "loss": "ce"
      },
      'output': {
        'class': 'choice', 'target': "classes", 'beam_size': 3, 'from': "output_prob", "input_type": "prob",
        "initial_output": 0,
      }, "output_is_not_blank": {
        "class": "compare", "from": "output", "value": 0, "kind": "not_equal", "initial_output": True
      },
    },
  }
}
```

When I try to run the config after commit 7e3b939237e69a396a017e88bf836972f0d63e79, I get the following error:

[error.txt](https://github.com/rwth-i6/returnn/files/7135243/error.txt). Before that commit, it works. The error seems to be caused by `segment_lens0`. I don't know if the way I am doing it is wrong or if the commit broke something which should work. @albertz

albertz:
Fix #637

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -135,6 +135,8 @@ class DimensionTag(object):
       # If static dim, no effect.
       assert not self.batch
       return self
+    if batch.is_broadcast():
+      return self  # just leave as-is. should not matter.
     same_base = self.get_same_base()
     # Might be uninitialized in some cases. Assume batch is global.
     if not same_base.batch:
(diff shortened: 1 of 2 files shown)


albertz:
@robin-p-schmitt I'm waiting for the tests. But your test (the first one) seems to pass, so I guess this is fixed. It's a bit ugly but I want to keep it simpe for now.

PR #644, opened 2021-09-11
Data description/repr shorter

albertz:


albertz:
Whoops, this first needs some new changes (via #589). So waiting for them, then this will be rebased.

PR #649, opened 2021-09-14
Prefix decoding

patrick-wilken:
This implements enforcing target tokens in ChoiceLayer provided by a separate data stream until it is "used up". After that (different lengths per sequence are possible) search continues as usual. The correct scores are assigned to the prefixes.

Useful to continue a target sequence provided by a user, or to enforce target-side prefix tokens that provide meta-information, e.g. the dialect to translate to.

Getting additional data streams into the recurrent layer that are on the same time axis but have a shorter length is somewhat tricky. I implemented that you have to mark them explicitly via a `padded_data_keys` parameter. The check for equal length is disabled for those data keys, instead they are padded if necessary.


albertz:
We should not introduce another separate padding concept. This was already discussed as part of #391, where we agreed to add sth like `dyn_mask_ext` (earlier `seq_mask_ext` in that discussion) to the `DimensionTag`.

albertz:
Btw, before you implement some bigger change like this, it would be good to open an issue first where the implementation details are being discussed.

albertz:
One way this would already work without these changes (at least conceptually):

You could do one pass over the given prefix with search disabled (so `ChoiceLayer` uses the prefix). Then you do a second pass with search enabled, where you init all the hidden states by the final states of the first pass.

So sth (conceptually) like:
```
_, hidden = decoder(targets=prefix, search=False)
out, _ = decoder(hidden=hidden, search=True)
```

I write "conceptually" because there are likely a couple of details why this would not work yet. But then we should fix those issues.

I think this is easier and more straight-forward than adding another new option to `ChoiceLayer`.

returnn/tf/layers/rec.py
```diff
@@ -4986,6 +5016,59 @@ def _get_cheating_targets_and_src_beam_idxs(self, scores):
...
+      """
+      assert self.prefix_target
+
+      # Get the labels of the prefixes which should be enforced. They are padded with zeros, therefore we will
+      # get zeros for those sequences where the current timestep is beyond the length of the prefix.
+      target_prefix_labels = self._get_target_value(
+        target=self.prefix_target).get_placeholder_as_batch_major()  # (batch * beam,), int32
+
+      # Get prefixes that have already ended (i.e. have a smaller length than the current time step).
+      target_prefix_ended = tf.equal(target_prefix_labels, 0)
```
albertz, line 5042: This has the hardcoded assumption that EOS=0. I would avoid that. Better rely on the target seq len (using rec step info to check the current pos).

returnn/tf/layers/rec.py
```diff
@@ -5041,8 +5126,8 @@ def _create_search_beam(cls, name, beam_size, sources, network):
       name="%s%s" % (network.get_absolute_name_prefix(), name))
 
   @classmethod
-  def get_out_data_from_opts(cls, name, sources, target, network,
-                             beam_size, search=NotSpecified, scheduled_sampling=False, cheating=False, **kwargs):
+  def get_out_data_from_opts(cls, name, sources, target, network, beam_size, search=NotSpecified,
+                             scheduled_sampling=False, cheating=False, prefix_target=None, **kwargs):
```
albertz, line 5130: Instead of `prefix_target`, this could be a layer, like `prefix`. And you could directly refer to the whole sequence. So you would have `prefix="base:data:prefix"` or so in your config. This would avoid the whole padding logic.

albertz:
Btw, also conceptually, what we want in the future is that such things can be written more directly in the config, in a flexible way, without the need to modify RETURNN. One approach would be what I suggested before. But another option is when you write sth like this in the config:

```
output = cond(
  cond=i < seq_len(prefix),
  true_layer=iter(prefix),
  false_layer=choice(prob))
```
(I'm using the new proposed syntax from returnn-common here, but just for readability. You can write this in the old-style dict syntax as well if you prefer.)

This uses the `IterLayer` from #552, which is basically the `TensorArray` logic but wrapped in a layer, not directly in the `RecLayer`.

This assumes that `CondLayer` supports broadcasting, which it does not. But this is needed here to work correctly, because the `IterLayer` (tensor array `read`) should only happen if we are still within the prefix. This resolves the padding logic then.

Basically, `CondLayer` for `cond` which is not a scalar would behave like:
```
tf.case(
  [(tf.reduce_all(cond), true_layer),
   (tf.reduce_all(tf.logical_not(cond)), false_layer)],
  default=where_bc(...))
```

patrick-wilken:
Yep, #552 sounds very related and could improve my padding approach.

I agree, that my implementation is not "modular".

About two decoding passes: Yes, I was aware of that concept, but then I would need to take the graph apart and write a custom search, with two session run calls and feeding/fetching states. I want to avoid that if there is a way around it. Or do you mean having two decoders in the usual search graph? That would mean copying the decoder definition in the config and sharing all the parameters...

About using `CondLayer`: Yes, this could be a better alternative. As you described, selecting the prefix using layers is possible, even if it is a bit complicated. I wonder about the other things ChoiceLayer does though: Would setting up the search choice dependencies work if the `ChoiceLayer` output is used only after going through a `CondLayer`? I guess yes. Would the search scores and beams be set correctly for the prefix timesteps? I think no, currently they would correspond to the n-best hyps and not to the forced prefix. Beams does not matter much, because prefix has no real beam, but we want the correct score. So it is a bit tricky.

Conceptually, what I'm doing is switching between the two "modes" of `ChoiceLayer` in training vs. search, either provide the target or the predicted label. So maybe it is acceptable to have that behaviour implemented in there, even though it's a separate implementation...

returnn/tf/layers/rec.py
patrick-wilken: I don't understand. You mean `ChoiceLayer` gets the whole prefix and selects the label of the current timestep itself? Or would a layer somehow handle the padding for me?

albertz:
> About two decoding passes: Yes, I was aware of that concept, but then I would need to take the graph apart and write a custom search, with two session run calls and feeding/fetching states.

No. My example above is just a single session run. Why should it be two session runs? This here would be your normal net dict definition:

```
_, hidden = decoder(targets=prefix, search=False)
out, _ = decoder(hidden=hidden, search=True)
```

> Or do you mean having two decoders in the usual search graph? That would mean copying the decoder definition in the config and sharing all the parameters...

Yes, this is what I mean. What's the problem with that? It's just a reformulation in the end of what you are doing.

> About using `CondLayer`: Yes, this could be a better alternative. As you described, selecting the prefix using layers is possible, even if it is a bit complicated. I wonder about the other things ChoiceLayer does though: Would setting up the search choice dependencies work if the `ChoiceLayer` output is used only after going through a `CondLayer`? I guess yes. 

Yes sure, why not? This is really independent of `ChoiceLayer` also but just how `CondLayer` works in general.

> Would the search scores and beams be set correctly for the prefix timesteps? I think no, currently they would correspond to the n-best hyps and not to the forced prefix. Beams does not matter much, because prefix has no real beam, but we want the correct score. So it is a bit tricky.

Ah yes, you are right. But also this can be solved.

> Conceptually, what I'm doing is switching between the two "modes" of `ChoiceLayer` in training vs. search, either provide the target or the predicted label. So maybe it is acceptable to have that behaviour implemented in there, even though it's a separate implementation...

I don't exactly understand the reasoning. We have three proposed approaches here so far:

- Two `decoder` calls. Logic implemented in user config.
- Using `ConditionLayer`, etc. Logic implemented in user config.
- Extending `ChoiceLayer` by further logic. Logic implemented in RETURNN.

What you want to do can be done in these three ways (among maybe other ways).

In general, we always prefer if we have RETURNN flexible enough such that the user can implement things in the config. We want to keep RETURNN simple, and not extend more and more. (See e.g. `SelfAttentionLayer` for another such bad example, and #391 how this was resolved.)

PR #650, opened 2021-09-14
FeedDictDataProvider: fixed no advance in single threaded case

patrick-wilken:
Was not noticed because the tests that use "single_threaded" create only one batch.


albertz:
> Was not noticed because the tests that use "single_threaded" create only one batch.

Maybe we should extend the tests then as well?

patrick-wilken:
Added a test.

albertz (approved):

PR #653, opened 2021-09-14
Cleanups and improved performance

albertz:


returnn/tf/util/data.py
```diff
@@ -34,17 +34,17 @@ class Types:
     Defines possible values for ``kind``.
     """
     Unspecified = None
-    Batch = "batch"
-    Spatial = "spatial"  # also time
-    Time = "spatial"  # we don't treat this as different
-    Feature = "feature"
+    Batch = Entity("batch")
+    Spatial = Entity("spatial")  # also time
+    Time = Entity("spatial")  # we don't treat this as different
```
Zettelkasten, line 39: This comment is no longer true now, is it?

albertz: Whoops, good point!

albertz: Ok fixed. But strangely all the tests also passed without this fix. Maybe `Time` is not often (or not at all) used.

PR #635, opened 2021-09-07
SliceNdLayer now uses GatherLayer to get the slices.

robin-p-schmitt:
The start argument of SliceNdLayer can now be of any shape and not only of shape (B,) as before. I am not 100% sure if we need any asserts to check for shapes or so. But I think it should work for any non-exotic shapes.

@Zettelkasten #625

PR #657, opened 2021-09-15
Accumulate loop dyn size

albertz:
Support of accumulation of dynamic sizes (`DimensionTag.dyn_size_ext`) within a recurrent context (`control_flow_context`). E.g. when the dyn size (seq lengths) is of shape [B] inside the loop, and this dyn size changes in every iteration (i.e. `control_flow_context` is set to the rec loop), then the accumulated dyn size outside the loop would have shape [T,B].

Further, when some layer inside the loop has such a dynamic shape [B,T_per_frame], where T_per_frame is the dyn size as mentioned before, this means that we cannot use the current `TensorArray` logic, which expects that elements always have the same shape in every frame. This PR also extends that logic to support that.

This was a request by #635 and also makes sense for #589, #579 (and other related commits for #391).

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -1473,6 +1473,36 @@ class _SubnetworkRecCell(object):
         print(s)
       raise
 
+  def _add_template_layer(self, layer_name, layer_dict):
+    """
+    Use this for simple helpers, after the main template net is already created.
+    This does not expect layer creation exceptions,
+    and expects that all dependencies are already created,
+    and all dependencies are other layers inside our subnet.
+
+    :param str layer_name:
+    :param dict[str] layer_dict: not yet transformed
+    :rtype: _TemplateLayer
+    """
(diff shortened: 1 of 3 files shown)


albertz:
Btw, also check PyCharm warnings (and corresponding test).
```
  returnn/tf/layers/basic.py:899: WARNING PyUnboundLocalVariableInspection: Local variable 'size_t' might be referenced before assignment
```

albertz:
Finally. Again more effort than expected. This still needs to be cleaned up. Maybe some aspects will be separated. Like the `DimensionTag.declare_same_as` changes.

JackTemaki:
> Finally. Again more effort than expected. This still needs to be cleaned up. Maybe some aspects will be separated. Like the `DimensionTag.declare_same_as` changes.

Can you explain shortly what this is about?

albertz:
> Can you explain shortly what this is about?

I wrote some description.

PR #658, opened 2021-09-16
Data get_common_data cleanup for new consistent dim tags

albertz:


albertz:
Changes are ready now. But I still need to clean up the commits.

This was once again more complicated than expected. But `get_common_data` had some nasty side effects and now not anymore, so these changes definitely improve the overall quality and robustness.

albertz:
@Zettelkasten @robin-p-schmitt You don't really need to do a review. I just thought these changes might be interesting/relevant to you.

PR #659, opened 2021-09-16
HDFDataset: search without reference

patrick-wilken:
I want to do search on a HDFDataset without having a reference, and without creating a dummy target.
This is possible using `++search_do_eval False`, but I needed these two fixes:

1. Make it possible to provide the target labels in the HDF file without having target data. For this, simply don't include the additional labels in `self.target_keys`. (`self.target_keys` always has to correspond to the existing data, e.g. because of number of `seqLengths` entries)
2. Don't check the missing targets in `check_matched_dataset()` because the dataset might not be able to provide the shape and dtype of missing data. (For HDFDataset, `get_data_dtype()` fails.)

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -113,7 +113,6 @@ class HDFDataset(CachedDataset):
       prev_target_keys = self.target_keys
     if 'targets' in fin:
       self.target_keys = sorted(
-        set(fin['targets/labels'].keys()) |
         set(fin['targets/data'].keys()) |
         set(fin['targets/size'].attrs.keys()))
     else:
--- a/returnn/datasets/cached.py
+++ b/returnn/datasets/cached.py
@@ -43,7 +43,7 @@ class CachedDataset(Dataset):
     self._index_map = range(len(self._seq_index))  # sorted seq idx -> seq_index idx
     self._tag_idx = {}  # type: typing.Dict[str,int]  # map of tag -> real-seq-idx. call _update_tag_idx
     self.targets = {}
-    self.target_keys = []
+    self.target_keys = []  # the keys for which we provide data; we may have labels for additional keys in self.labels
     self.timestamps = None
 
   def initialize(self):


returnn/datasets/hdf.py
```diff
@@ -113,7 +113,6 @@ def add_file(self, filename):
       prev_target_keys = self.target_keys
     if 'targets' in fin:
       self.target_keys = sorted(
-        set(fin['targets/labels'].keys()) |
```
JackTemaki, line 116: Why is this change relevant?

albertz:
But `check_matched_dataset` is not about checking whether the data is available. That is what the `available_for_inference` flag is for (and it depends on the eval flag).

`check_matched_dataset` just is a sanity check for the data keys, whether their shape / dim and other meta information matches.

`check_matched_dataset` also only checks the **used** data keys. Used means that somewhere in the network the actual value is used (so this would actual be an error when `available_for_inference=False` and `eval=False`).

So I wonder more, why does this target end up as a used data key? This should not really be the case. And if this would not be the case, `check_matched_dataset` would also ignore it. So this seems to be the actual issue here. Which might be due to some bug.

returnn/datasets/hdf.py
albertz: Yep, I also wonder. When reading the title of this PR, my first thought was, why should this be in any way specific about `HDFDataset`?

patrick-wilken: Yep, title is probably not optimal, but it is what connects the two commits 😅 Could have also opened two PRs. Commit messages are clear, I hope.

Simplest example:
You have a trained net, one input, one sparse output, i.e. with vocabulary/labels. To do search you need inputs and output labels. So a _minimal_ HDF would contain the fields: `inputs`, `seqLengths` and `targets/labels/classes`. It does not contain `targets/data/classes`.
Now the question is whether `self.target_keys` refers to any keys the dataset knows of or whether it refers to the data  provided by the dataset. It has to be the latter one. For example the second dim of `seqLengths` is `len(self.target_keys)` and we only have sequence lengths for `data`, not for `classes`.

patrick-wilken:
> `check_matched_dataset` also only checks the **used** data keys. Used means that somewhere in the network the actual value is used (so this would actual be an error when `available_for_inference=False` and `eval=False`).

Oh, uhm, but the network makes use of the data template configured in `extern_data`. So this is not "using" the data? Ok, then I will debug where it is marked as used and take this back.^^

returnn/datasets/hdf.py
albertz: You should add some comment to clarify what `target_keys` is used for. So you say it has to be the latter one because of `seqLengths`. But maybe this is used inconsistently also for other things where the first case makes more sense? Anyway, add a comment about `seqLengths`, and maybe some more.

albertz:
> > `check_matched_dataset` also only checks the **used** data keys. Used means that somewhere in the network the actual value is used (so this would actual be an error when `available_for_inference=False` and `eval=False`).
> 
> Oh, uhm, but the network makes use of the data template configured in `extern_data`. So this is not "using" the data? Ok, then I will debug where it is marked as used and take this back.^^

Yes. There is `TFNetwork.get_extern_data(self, key, mark_data_key_as_used=True)`. You can still use the template, which is often used just to define shapes/dimensions. But this will not mark it as used then. If it is marked as used, it means that the values itself are used. And only then it will end up in `used_data_keys`.

At least that is what it is supposed to do. If it does not do that correctly, that would be a bug.

patrick-wilken:
Oops, sorry, seems I got confused during debugging. This already works! When setting `++search_do_eval False` keys with `available_for_inference` are not checked in `check_matched_dataset()` (because they are not used).

returnn/datasets/hdf.py
patrick-wilken: I think this comment I have now is enough. `self.labels` is independent from anything else, it is only used in `serialize_data()`. So including its keys in `self.target_keys` was kind of wrong in the first place. If anything, it was an assertion that the keys match, but I want to relax that.

albertz:
Sorry, the failing test seems to be some unrelated thing. I fixed this via PR #665, and already merged to master.

Can you rebase to current master such that we have all tests passing?

PR #660, opened 2021-09-16
HDFDataset: support 'data' as target key

patrick-wilken:
@albertz You were right, allowing the `"data"` stream to be provided via a target field in the HDF file is actually useful. And easy to do.

### DIFF
--- a/returnn/datasets/cached.py
+++ b/returnn/datasets/cached.py
@@ -517,7 +517,7 @@ class CachedDataset(Dataset):
     return alloc_data[o:o + l]
 
   def get_data_dim(self, key):
-    if key == "data":
+    if key == "data" and self.num_inputs > 0:  # if num_inputs == 0, we allow "data" as a target key
       return self.num_inputs * self.window
     return self.num_outputs[key][0]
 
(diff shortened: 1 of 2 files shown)


albertz:
I don't exactly understand, why is this needed?

patrick-wilken:
See #438 and https://github.com/rwth-i6/returnn/pull/438#discussion_r574292823. We don't use `inputs` in the HDF file anymore, because we have multiple. Using `targets` for everything is more consistent. But in all our configs one input is still called `"data"`, because previously there was no way around that.

albertz:
The suggestion there was to use `if key == "data" and 'targets/data/data' not in fin:`. So I assume you assume that `num_inputs > 0` implies `'targets/data/data' not in fin`? Shouldn't this at least be checked somewhere?

> We don't use `inputs` in the HDF file anymore, because we have multiple.

I don't understand. You can still use `inputs` and then the others (in `targets/data/*`) for remaining inputs.

> But in all our configs one input is still called `"data"`

I don't understand. I thought this is not needed anymore with #438?

> Using `targets` for everything is more consistent.

Yea, that might be true.

patrick-wilken:
> I don't understand. I thought this is not needed anymore with #438?

True. So currently this is about backwards compatibility. But also there should be no restrictions about the key if not necessary.

> So I assume you assume that num_inputs > 0 implies 'targets/data/data' not in fin? Shouldn't this at least be checked somewhere?

Right, this way around it was not checked, I added it.

albertz:
Does our HDF writer code already handle this?

albertz (approved):

PR #670, opened 2021-09-20
Extend SliceNdLayer to use a layer as the size argument

robin-p-schmitt:
Right now, `start` and `size` need to have the same shape when they are given as layers. This suffices for my use cases but we can also add support for different shapes if you think we need them.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -859,7 +859,8 @@ class SliceNdLayer(_ConcatInputLayer):
   def __init__(self, start, size, min_size=None, **kwargs):
     """
     :param LayerBase start: (B,...)
-    :param int|None size: if None, it uses the max possible size, and it becomes a dynamic axis
+    :param int|LayerBase|None size: if None, it uses the max possible size,
+    and it becomes a dynamic axis. If LayerBase, it needs to have the same shape as start
     :param int|None min_size: if size is None, but we want to have a min-size
     """
     super(SliceNdLayer, self).__init__(**kwargs)
@@ -867,6 +868,7 @@ class SliceNdLayer(_ConcatInputLayer):
     x = self.input_data.copy_as_batch_major()
     seq_lens = x.get_sequence_lengths() if x.is_time_axis_dynamic() else None  # (B,) or None
     self.start = start
+    self.size = size
     start_data = start.output.copy_as_batch_major()  # e.g. (B,) or (B,T)
     start_t = start_data.placeholder
     if size is None:
@@ -874,13 +876,19 @@ class SliceNdLayer(_ConcatInputLayer):
         min_size = 0
       if seq_lens is None:
         assert isinstance(x.batch_shape[x.time_dim_axis], int)
-        size = tf.maximum(tf.reduce_max(x.batch_shape[x.time_dim_axis] - start_t), min_size)  # scalar
+        size_t = x.batch_shape[x.time_dim_axis] - start_t
       else:
         # make seq_lens compatible with start_t
         seq_lens = expand_multiple_dims(  # e.g. (B,) or (B,1)
           x=seq_lens,
           axes=[-1] * (len(start_t.shape) - len(seq_lens.shape)))
-        size = tf.maximum(tf.reduce_max(seq_lens - start_t), min_size)  # scalar
+        size_t = seq_lens - start_t
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+    elif isinstance(size, LayerBase):
+      assert size.output.batch_ndim == start_data.batch_ndim
+      min_size = 0
+      size_t = size.output.get_placeholder_as_batch_major()
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
     # for each start index in start_data, we want to gather a slice
     # therefore, the output's first axes are the same as the ones from start_data
     # and the next axis will therefore be the slice axis
@@ -888,10 +896,7 @@ class SliceNdLayer(_ConcatInputLayer):
     assert slice_tag.description.startswith("sliced-time:")
     if not isinstance(size, int):
       # in this case, size is not known before runtime and becomes dynamic and we need to set dyn_size
-      if seq_lens is None:
-        dyn_size = tf.maximum(x.batch_shape[x.time_dim_axis] - start_t, min_size)  # (B,) or (B,T)
-      else:
-        dyn_size = tf.maximum(seq_lens - start_t, min_size)  # (B,) or (B,T)
+      dyn_size = tf.maximum(size_t, min_size)  # (B,) or (B,T)
       dyn_size_ext = Data(
         name=("%s:dyn_size" % slice_tag.description),
         dtype=Data.size_dtype,
@@ -923,6 +928,8 @@ class SliceNdLayer(_ConcatInputLayer):
         tf.greater(gather_positions, x.batch_shape[1] - 1),
         tf.less(gather_positions, 0))
       gather_positions = tf.clip_by_value(gather_positions, 0, x.batch_shape[1] - 1)
+    if isinstance(self.size, LayerBase):
+      pad_mask = tf.logical_or(tf.greater(gather_positions, tf.expand_dims(start_t + size_t - 1, -1)), pad_mask)
     gather_positions_data.placeholder = gather_positions
     position = InternalLayer(
       network=self.network,
@@ -951,7 +958,10 @@ class SliceNdLayer(_ConcatInputLayer):
     """
     :rtype: list[LayerBase]
     """
-    return super(SliceNdLayer, self).get_dep_layers() + [self.start]
+    dep_layers = super(SliceNdLayer, self).get_dep_layers() + [self.start]
+    if isinstance(self.size, LayerBase):
+      dep_layers += [self.size]
+    return dep_layers
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources=(), start=None, size=None, **kwargs):
@@ -966,6 +976,8 @@ class SliceNdLayer(_ConcatInputLayer):
     start_data = start.output.copy_as_batch_major()
     input_data = sources[0].output.copy_as_batch_major()
     gather_positions_data = start_data.copy_template(name="%s_gather_positions" % name)
+    if isinstance(size, LayerBase):
+      size = None
     # size might be None here in which case we set the dyn_size in __init__
     tag = DimensionTag(
       kind=DimensionTag.Types.Spatial,
@@ -991,6 +1003,8 @@ class SliceNdLayer(_ConcatInputLayer):
     """
     super(SliceNdLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
     d["start"] = get_layer(d["start"])
+    if isinstance(d["size"], str):
+      d["size"] = get_layer(d["size"])
 
 
 class GatherLayer(_ConcatInputLayer):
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,44 @@ class SliceNdLayer(_ConcatInputLayer):
   def __init__(self, start, size, min_size=None, **kwargs):
     """
     :param LayerBase start: (B,...)
-    :param int|None size: if None, it uses the max possible size, and it becomes a dynamic axis
+    :param int|LayerBase|None size: if None, it uses the max possible size,
+    and it becomes a dynamic axis. If LayerBase, it needs to have the same shape as start
```
albertz, line 863:
```suggestion
      and it becomes a dynamic axis
```
Indent by two spaces.

Remove the layer comment. This should not be the case. Use `get_common_data` on `start`/`size` and then `copy_compatible_to`.

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,44 @@ class SliceNdLayer(_ConcatInputLayer):
...
-        size = tf.maximum(tf.reduce_max(x.batch_shape[x.time_dim_axis] - start_t), min_size)  # scalar
+        size_t = x.batch_shape[x.time_dim_axis] - start_t
       else:
         # make seq_lens compatible with start_t
         seq_lens = expand_multiple_dims(  # e.g. (B,) or (B,1)
           x=seq_lens,
           axes=[-1] * (len(start_t.shape) - len(seq_lens.shape)))
-        size = tf.maximum(tf.reduce_max(seq_lens - start_t), min_size)  # scalar
+        size_t = seq_lens - start_t
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
```
albertz, line 886: I don't understand your naming. `_t` stands for tensor. So `size_t` is a tensor (not e.g. `Data`). But `size` is what then? This is also a tensor. This doesn't make sense.

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,44 @@ class SliceNdLayer(_ConcatInputLayer):
...
     :param LayerBase start: (B,...)
-    :param int|None size: if None, it uses the max possible size, and it becomes a dynamic axis
+    :param int|LayerBase|None size: if None, it uses the max possible size,
+    and it becomes a dynamic axis. If LayerBase, it needs to have the same shape as start
     :param int|None min_size: if size is None, but we want to have a min-size
     """
     super(SliceNdLayer, self).__init__(**kwargs)
     from returnn.tf.util.basic import where_bc, expand_multiple_dims
     x = self.input_data.copy_as_batch_major()
     seq_lens = x.get_sequence_lengths() if x.is_time_axis_dynamic() else None  # (B,) or None
```
albertz, line 869: Note, instead of using `seq_lens` directly here, you could also get the dim tag, then use `dyn_size_ext`, and then functions like `copy_compatible_to`, to simplify the `expand_multiple_dims` below, and also have it more generic.

returnn/tf/layers/basic.py
```diff
@@ -951,7 +958,10 @@ def get_dep_layers(self):
...
     :rtype: list[LayerBase]
     """
-    return super(SliceNdLayer, self).get_dep_layers() + [self.start]
+    dep_layers = super(SliceNdLayer, self).get_dep_layers() + [self.start]
+    if isinstance(self.size, LayerBase):
+      dep_layers += [self.size]
+    return dep_layers
 
   @classmethod
   def get_out_data_from_opts(cls, name, sources=(), start=None, size=None, **kwargs):
```
albertz, line 967: `size` doc should be updated here as `:param int|LayerBase|None size:`.

returnn/tf/layers/basic.py
robin-p-schmitt: Size is either a tensor (in case we use `tf.maximum`) or an integer (if the user provides a static size). Should we maybe rename `size` to `max_size`?

albertz: Ah sorry, yes. Or also a `LayerBase`. No, leave it like it is.

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,44 @@ class SliceNdLayer(_ConcatInputLayer):
...
+    and it becomes a dynamic axis. If LayerBase, it needs to have the same shape as start
     :param int|None min_size: if size is None, but we want to have a min-size
     """
     super(SliceNdLayer, self).__init__(**kwargs)
     from returnn.tf.util.basic import where_bc, expand_multiple_dims
     x = self.input_data.copy_as_batch_major()
     seq_lens = x.get_sequence_lengths() if x.is_time_axis_dynamic() else None  # (B,) or None
     self.start = start
+    self.size = size
     start_data = start.output.copy_as_batch_major()  # e.g. (B,) or (B,T)
```
albertz, line 872: You use `copy_as_batch_major` here and also for `size.output` below.
Do we really need it as batch major? Why? (Please add such a comment in the code.)
Or do we just need it compatible to the input and to `size`? Then we better should use the `get_common_data` and `copy_compatible_to` logic.

This should also be fixed/clarified accordingly in `get_out_data_from_opts`.

albertz:
Btw, also check PyCharm warnings (and corresponding test).
```
  returnn/tf/layers/basic.py:899: WARNING PyUnboundLocalVariableInspection: Local variable 'size_t' might be referenced before assignment
```

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,44 @@ class SliceNdLayer(_ConcatInputLayer):
...
           x=seq_lens,
           axes=[-1] * (len(start_t.shape) - len(seq_lens.shape)))
-        size = tf.maximum(tf.reduce_max(seq_lens - start_t), min_size)  # scalar
+        size_t = seq_lens - start_t
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+    elif isinstance(size, LayerBase):
+      assert size.output.batch_ndim == start_data.batch_ndim
+      min_size = 0
+      size_t = size.output.get_placeholder_as_batch_major()
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
```
albertz, line 891: There is some `else:` missing which assigns `size_t`, right? (See also warning)

albertz: Ok, maybe `size_t` is never used in this case. But anyway assign `None`.

returnn/tf/layers/basic.py
```diff
@@ -859,70 +859,82 @@ class SliceNdLayer(_ConcatInputLayer):
...
         placeholder=dyn_size,
-        dim_tags=start_data.dim_tags,
+        dim_tags=start_data.copy_as_batch_major().dim_tags,  # as batch major because the placeholder is too
         batch=slice_tag.batch,
         beam=slice_tag.batch.beam if slice_tag.batch else self.output.beam,
         control_flow_ctx=slice_tag.control_flow_ctx)
       slice_tag.dyn_size_ext = dyn_size_ext
       slice_tag.set_tag_on_size_tensor(dyn_size)
-    gather_positions_data = start_data.copy_template(name="%s_gather_positions" % self.name)
+    # get as batch major, because the placeholder later will also be batch major
```
albertz, line 915: Why is the placeholder batch major? The placeholder is the same as `start` (`start_t`), or not?

returnn/tf/layers/basic.py
```diff
@@ -859,70 +859,82 @@ class SliceNdLayer(_ConcatInputLayer):
...
+    gather_positions_data = start_data.copy_template(name="%s_gather_positions" % self.name).copy_as_batch_major()
     gather_positions_data = gather_positions_data.copy_add_dim_by_tag(
       slice_tag,
       unbroadcast=True,
       axis=start_data.batch_ndim)
     # [start+0, start+1, ...]
     gather_positions = tf.expand_dims(start_t, -1) + tf.range(0, size)  # e.g. (B, size) or (B, T, size)
```
albertz, line 922: If `start` is e.g. [T,B], then this is [T,B,size], i.e. not batch major. Right?

returnn/tf/layers/basic.py
```diff
@@ -859,70 +859,82 @@ class SliceNdLayer(_ConcatInputLayer):
...
+      size_data = size.output.copy()
+      common_data = Data.get_common_data([start_data, size_data])
+      size_data = size_data.copy_compatible_to(common_data)
+      size_t = size_data.get_placeholder_as_batch_major()
+      start_data = start_data.copy_compatible_to(common_data)
+      start_t = start_data.get_placeholder_as_batch_major()
+      min_size = 0
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+    else:
+      # in this case, size_t is never used but needs to be set to avoid a PyCharm warning
```
albertz, line 895: Do not add such comment.

returnn/tf/layers/basic.py
robin-p-schmitt: But in the beginning, I set `start_t = start_data.get_placeholder_as_batch_major()`. Should I change that?

albertz: Why do you use `get_placeholder_as_batch_major` there. If there is no reason for it, yes, remove that. Otherwise comment why it is needed.

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,48 @@ class SliceNdLayer(_ConcatInputLayer):
...
     from returnn.tf.util.basic import where_bc, expand_multiple_dims
-    x = self.input_data.copy_as_batch_major()
-    seq_lens = x.get_sequence_lengths() if x.is_time_axis_dynamic() else None  # (B,) or None
+    from returnn.tf.util.data import Data
+    x = self.input_data.copy()
+    seq_lens_data = x.get_time_dim_tag().dyn_size_ext  # (B,) or None
     self.start = start
-    start_data = start.output.copy_as_batch_major()  # e.g. (B,) or (B,T)
+    self.size = size
+    start_data = start.output.copy()  # e.g. (B,) or (B,T)
```
albertz, line 873: We potentially have 3 sources here for the shape of the positions:
`seq_lens_data`, `start` and `size`.
I think we should use `get_common_data` on all three, and then store sth like `common_data = Data.get_common_data([seq_lens_data, start, size.output if isinstance(size, LayerBase) else None])`, and below, use `copy_compatible_to(common_data, ...)`.

returnn/tf/layers/basic.py
```diff
@@ -859,39 +859,48 @@ class SliceNdLayer(_ConcatInputLayer):
...
+      min_size = 0
+      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+    else:
+      size_t = None
     # for each start index in start_data, we want to gather a slice
     # therefore, the output's first axes are the same as the ones from start_data
     # and the next axis will therefore be the slice axis
     slice_tag = self.output.dim_tags[start_data.batch_ndim]
     assert slice_tag.description.startswith("sliced-time:")
     if not isinstance(size, int):
```
albertz, line 901: It might be cleaner to check `if size_t is not None:` and then add `assert not isinstance(size, int)`.

returnn/tf/layers/basic.py
robin-p-schmitt: So `get_common_data` does not allow the list items to be `None`. Can I instead do:

```
data_objs = [start_data]
data_objs += [size.output] if isinstance(size, LayerBase) else []
data_objs += [seq_lens_data] if isinstance(seq_lens_data, Data) else []
common_data = Data.get_common_data(data_objs)?
```

albertz: Ah yes.

albertz:
Note again the code warning:
```
  returnn/tf/layers/basic.py:867: WARNING PyUnresolvedReferencesInspection: Unused import statement <code>expand_multiple_dims</code>
```

albertz (approved):
Looks good now

albertz:
Just for reference: Related was #661 and #635

PR #671, opened 2021-09-20
consistent checkpoint filepattern

JackTemaki:
The allowed checkpoint paths were not consistent across the different parameters, so I tried to clean this up a little bit.

### DIFF
--- a/returnn/engine/base.py
+++ b/returnn/engine/base.py
@@ -8,9 +8,9 @@ from __future__ import print_function
 import os
 import sys
 import typing
-from returnn.util.basic import BackendEngine, model_epoch_from_filename, get_model_filename_postfix
 from returnn.log import log
 from returnn.pretrain import Pretrain
+from returnn.util import basic as util
 
 
 class EngineBase(object):
@@ -52,7 +52,7 @@ class EngineBase(object):
         if os.path.exists(fn):
           file_list[epoch] = fn
           break
-        if BackendEngine.is_tensorflow_selected():
+        if util.BackendEngine.is_tensorflow_selected():
           if os.path.exists(fn + ".index"):
             file_list[epoch] = fn
             break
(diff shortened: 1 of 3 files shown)


returnn/tf/network.py
```diff
@@ -15,7 +15,7 @@
 from returnn.tf.layers.basic import LayerBase, get_layer_class
 import returnn.tf.compat as tf_compat
 import returnn.tf.util.basic as tf_util
-from returnn.tf.util.basic import Data, DimensionTag, reuse_name_scope, VariableAssigner
+from returnn.tf.util.basic import Data, DimensionTag, reuse_name_scope, VariableAssigner, get_checkpoint_filepattern
```
albertz, line 18: I prefer to not add too much global imports except for the very common things like `Data`, esp not for rarely used functions.
Rather, use `tf_util.`.
(Maybe the other imports here should be cleaned up as well, like `VariableAssigner`. But this is off-topic here.)

returnn/tf/network.py
```diff
@@ -3486,7 +3486,7 @@ def __init__(self, filename, saveable_params, params_prefix="", load_if_prefix="
       renamed vars in the checkpoint
     :param TFNetwork network:
     """
-    self.filename = filename
+    self.filepattern = get_checkpoint_filepattern(filename)
```
albertz, line 3489:
```suggestion
    self.filepattern = tf_util.get_checkpoint_filepattern(filename)
```

returnn/engine/base.py
```diff
@@ -11,6 +11,7 @@
 from returnn.util.basic import BackendEngine, model_epoch_from_filename, get_model_filename_postfix
 from returnn.log import log
 from returnn.pretrain import Pretrain
+from returnn.tf.util.basic import get_checkpoint_filepattern
```
albertz, line 14: Same comment as before, avoid the global import.

returnn/engine/base.py
```diff
@@ -72,19 +73,11 @@ def get_epoch_model(cls, config):
...
+    load_model_epoch_filename = get_checkpoint_filepattern(config.value('load', ''))
     if load_model_epoch_filename:
       assert os.path.exists(load_model_epoch_filename + get_model_filename_postfix())
 
-    import_model_train_epoch1 = config.value('import_model_train_epoch1', '')
-    if import_model_train_epoch1.endswith(".meta"):
-      import_model_train_epoch1 = import_model_train_epoch1[:-len(".meta")]
-    elif import_model_train_epoch1.endswith(".index"):
-      import_model_train_epoch1 = import_model_train_epoch1[:-len(".index")]
+    import_model_train_epoch1 = get_checkpoint_filepattern(config.value('import_model_train_epoch1', ''))
```
albertz, line 80: This code here is intended to work in all cases, also when TF is not available. So no TF import should be done here (also not indirectly). So you cannot use things from `tf_util`.

returnn/tf/util/basic.py
```diff
@@ -59,6 +59,21 @@ def have_min_tf_version(version):
...
+
+  :param str filepath:
+  :return: CheckpointLoader compatible filepattern
+  :rtype: str
+  """
+  if filepath.endswith(".meta"):
+    return filepath[:-len(".meta")]
+  elif filepath.endswith(".index"):
+    return filepath[:-len(".index")]
+  return filepath
```
albertz, line 74: As this is also used in engine base logic, it needs to be in the base utils, not TF specific utils.

See also function `get_model_filename_postfix`, which is related. Maybe put it below.

albertz (approved):

PR #678, opened 2021-09-21
allow MetaDataset without 'data' key

JackTemaki:
This annoyed me for a long time now, and now I had a case again where this crashed, so I tried to fix it.


albertz (approved):

albertz:
I wonder whether we also maybe should make `self.target_list` configurable. Or what effect this actually has.

Does this even matter for you? In your cases, all keys are targets, and no key is not a target? Being in the targets was meant to imply that `available_for_inference=False` but I think we don't really enforce nor check this.

PR #685, opened 2021-09-22
cleanup AllophoneStateIdxParserLayer

### LINKED ISSUE #584: Cleanup AllophoneStateIdxParserLayer
I think this layer is not used by anyone.
It should also be obsolete as it should be possible to reimplement this using other base layers.

Remove?

albertz:
Fix #584.


Marvin84:
I am using eval layers for getting the phoneme labels out of the dense state-tying and never used this layer.

Issue #689, opened 2021-09-23
Cleanup Datasets structure


JackTemaki:
I would like to refactor the datasets code in a way, that the package/module structure is actually more reasonable.
This is especially about cleaning up `generating.py`, as more than 50% of the code has nothing to do with the "generating" datasets.

This would include in my opinion:

 - create `util` folder inside `datasets`
 - move all classes related to `Vocabulary` to a `util/vocabulary.py`
 - move `ExtractAudioFeatures` and related code to `util/feature_extraction`
 - move OggZipDataset to an `audio.py`
 - move `LibrispeechDataset` also there, or even to a `corpus_specific.py` or something like this.
 
 And of course we can keep imports inside the `generating.py` to not break the existing API.

PR #690, opened 2021-09-23
separate vocabulary and feature extraction code

JackTemaki:
solves a first part of #689


returnn/datasets/generating.py
```diff
@@ -5,6 +5,7 @@
 
 from __future__ import print_function
 
+from returnn.datasets.util.feature_extraction import ExtractAudioFeatures
```
albertz, line 8: I would prefer a relative import here.

returnn/datasets/generating.py
```diff
@@ -13,6 +14,8 @@
 import sys
 import typing
 
+from returnn.datasets.util.vocabulary import *
```
albertz, line 17: This import code is inconsistent.
Why is there an empty line before?
And why is it not next to the other dataset related imports?
Also, again I would make it a relative import.

returnn/datasets/util/feature_extraction.py
```diff
@@ -0,0 +1,452 @@
+"""
+ExtractAudioFeatures class and related helpers
+"""
+import numpy
```
albertz, line 4: I would add an empty line before the first comment, after the module docstring, for consistency.

returnn/datasets/util/vocabulary.py
```diff
@@ -0,0 +1,366 @@
+"""
+Vocabulary related classes for targets such as BPE, SentencePieces etc...
+"""
+from __future__ import print_function
```
albertz, line 4: Also here, add an empty line before the first import.

JackTemaki:
A general question here, If I resolve this kind of changes, do you prefer that I rebase this into the original commit `separate vocabulary and feature extraction code` or should I just add it as e.g `resolve comments` commit ? (With maybe a squash and merge instead of rebase and merge)

albertz:
> A general question here, If I resolve this kind of changes, do you prefer that I rebase this into the original commit `separate vocabulary and feature extraction code` or should I just add it as e.g `resolve comments` commit ? (With maybe a squash and merge instead of rebase and merge)

I want that a code move is an individual commit (which does not change the code), such that the Git history stays valid if possible, such that changes can be tracked back before the code move.

You already mix up several things here in this PR (which is in general not optimal and should be avoided, but maybe not too big a problem).

The first commit already also does more than just moving code. It also adopts some other code (by changing imports etc). Maybe this is a problem. Maybe every single moved block of code must be an individual commit such that Git can track the history properly.

I'm not sure whether this here already breaks the Git history or not. If it does, we must not do it this way, and it must be further split up. It basically depends on this question. If it is still fine, maybe all can be squashed together anyway.

JackTemaki:
> I want that a code move is an individual commit (which does not change the code), such that the Git history stays valid if possible, such that changes can be tracked back before the code move.

With `git blame -C`, the history is preserved. I do not think there is any way to do it in a way that PyCharm will show this. I tested this even with only moving code (no changes) in a single commit, it does not work.


> You already mix up several things here in this PR
You mean I should have split the vocabulary move and the feature extraction move into separate PRs? 

The `TimitDataset remove dead code` was necessary in this PR because I would not pass the tests otherwise, as there were incorrect dependencies on private code.

albertz:
> > I want that a code move is an individual commit (which does not change the code), such that the Git history stays valid if possible, such that changes can be tracked back before the code move.
> 
> With `git blame -C`, the history is preserved. I do not think there is any way to do it in a way that PyCharm will show this. I tested this even with only moving code (no changes) in a single commit, it does not work.

I don't understand what you mean with PyCharm. It does correctly for me when I use `git blame` in PyCharm:

image

I also think that PyCharm just uses `git blame` under the hood and does not have own custom logic here.
Maybe you just miss the corresponding options in PyCharm (as you see in my screenshot).

Also, I think the command should be `git blame -CM` (`-M` is for code moves, which is maybe more relevant here).

But anyway, so when `git blame -CM` shows it correctly, then this is fine, this is what I want.

> > You already mix up several things here in this PR
> 
> You mean I should have split the vocabulary move and the feature extraction move into separate PRs?
> 
> The `TimitDataset remove dead code` was necessary in this PR because I would not pass the tests otherwise, as there were incorrect dependencies on private code.

I mostly mean the latter.

But this is ok then.

JackTemaki:
> Maybe you just miss the corresponding options in PyCharm

Indeed I am missing those options... I do not even have that options sub-menu!!!

But then everything is fine with that regards.

Still, now we discussed everything but my actual question, I will just add it as commit.

albertz:
> Still, now we discussed everything but my actual question, I will just add it as commit.

I would have squashed it into your first commit.

But if you think squashing everything together is fine I can also squash the whole PR. But I assume you did not test this, as you have it separate?

JackTemaki:
> > Still, now we discussed everything but my actual question, I will just add it as commit.
> 
> I would have squashed it into your first commit.
> 
> But if you think squashing everything together is fine I can also squash the whole PR. But I assume you did not test this, as you have it separate?

Let me test this

albertz:
> > Maybe you just miss the corresponding options in PyCharm
> 
> Indeed I am missing those options... I do not even have that options sub-menu!!!

Maybe your PyCharm version is not up-to-date. I use PyCharm 2021.2.2.

JackTemaki:
> > > Still, now we discussed everything but my actual question, I will just add it as commit.
> > 
> > 
> > I would have squashed it into your first commit.
> > But if you think squashing everything together is fine I can also squash the whole PR. But I assume you did not test this, as you have it separate?
> 
> Let me test this

After squashing the three commits into one with `rebase -i`, the `git blame -CM` command still showed the correct history.

albertz:
> After squashing the three commits into one with `rebase -i`, the `git blame -CM` command still showed the correct history.

Ok, then I will just squash all together here. So you can leave the commits separate, or just as you want.

(In general, if it is in some way helpful to keep them separate, I keep them separate. But that requires that it is cleaned up, which is not the case now. Or otherwise I squash them all together. Then it doesn't matter whether the individual commit history is cleaned up.)

albertz (approved):
Still waiting for tests. But otherwise ok.

JackTemaki:
> > After squashing the three commits into one with `rebase -i`, the `git blame -CM` command still showed the correct history.
> 
> Ok, then I will just squash all together here. So you can leave the commits separate, or just as you want.
> 
> (In general, if it is in some way helpful to keep them separate, I keep them separate. But that requires that it is cleaned up, which is not the case now. Or otherwise I squash them all together. Then it doesn't matter whether the individual commit history is cleaned up.)

This is exactly why I asked the question. E.g. Eugen does not like to squash only fixing commits back into the original commit for cleaning, because then you do not see the update history properly in the PR. But then I will keep doing it in this way for RETURNN PRs.

albertz:
> > > After squashing the three commits into one with `rebase -i`, the `git blame -CM` command still showed the correct history.
> > 
> > 
> > Ok, then I will just squash all together here. So you can leave the commits separate, or just as you want.
> > (In general, if it is in some way helpful to keep them separate, I keep them separate. But that requires that it is cleaned up, which is not the case now. Or otherwise I squash them all together. Then it doesn't matter whether the individual commit history is cleaned up.)
> 
> This is exactly why I asked the question. E.g. Eugen does not like to squash only fixing commits back into the original commit for cleaning, because then you do not see the update history properly in the PR. But then I will keep doing it in this way for RETURNN PRs.

But Eugen squashes always when merging as far as I remember. So this is what I described. In this case it does not matter.

JackTemaki:
> > > > After squashing the three commits into one with `rebase -i`, the `git blame -CM` command still showed the correct history.
> > > 
> > > 
> > > Ok, then I will just squash all together here. So you can leave the commits separate, or just as you want.
> > > (In general, if it is in some way helpful to keep them separate, I keep them separate. But that requires that it is cleaned up, which is not the case now. Or otherwise I squash them all together. Then it doesn't matter whether the individual commit history is cleaned up.)
> > 
> > 
> > This is exactly why I asked the question. E.g. Eugen does not like to squash only fixing commits back into the original commit for cleaning, because then you do not see the update history properly in the PR. But then I will keep doing it in this way for RETURNN PRs.
> 
> But Eugen squashes always when merging as far as I remember. So this is what I described. In this case it does not matter.

Yes exactly. I was confirming your statement.

PR #697, opened 2021-09-28
Profiling with the TF profiler

Zettelkasten:
Adds an option `store_tf_profile` to record data for the [TF profiler here](https://www.tensorflow.org/guide/profiler#profiler_tools).
I found this extremely useful.

### DIFF
--- a/docs/advanced/profiling.rst
+++ b/docs/advanced/profiling.rst
@@ -9,7 +9,21 @@ You should profile where the computing time or memory is spend.
 This is less specific about RETURNN but more about TensorFlow,
 so please refer to the TensorFlow documentation for more recent details.
 
-In RETURNN, there is the option ``store_metadata_mod_step`` which has the effect that
+Since Tensorflow 2.2, you can use the TensorFlow Profiler integrated in
+TensorBoard.
+Set the option ``store_tf_profile=True`` to record performance data for all `session.run` calls.
+Then open TensorBoard in the model directory, where a "Profile" tab should appear.
+The `official guide <https://www.tensorflow.org/guide/profiler#profiler_tools>`__
+describes all available tools that are available here.
+
+.. note::
+    There is
+    `a bug in TF 2.3 and earlier <https://github.com/tensorflow/tensorflow/issues/42123#issuecomment-675047711>`__
+    where the "Memory Profile" view does not show a breakdown of tensors at peak memory usage.
+    This was fixed in TF 2.4.
+
+
+There is also the option ``store_metadata_mod_step`` which has the effect that
 every Nth step, it will do the ``session.run`` with these additional options::
 
     run_metadata = tf.RunMetadata()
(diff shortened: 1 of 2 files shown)


docs/advanced/profiling.rst
```diff
@@ -25,7 +39,6 @@ Also, it will write a timeline in Google Chrome trace format
 
 See also this for further information:
 
+* `Optimize TensorFlow performance using the Profiler <https://www.tensorflow.org/guide/profiler#profiler_tools>`__
 * `TensorFlow Profiler and Advisor <https://github.com/tensorflow/tensorflow/blob/b2edbd5a640fb2f50989c5579a4cfe87d1fc675e/tensorflow/core/profiler/README.md>`__
-* `tfprof: TensorFlow Profiler and Beyond <https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/tfprof/README.md>`__
```
Zettelkasten, line 29: This link and the last one are dead. Does anyone care to fix them / should I just remove them?

albertz: Yea ok, this can be removed.

albertz (approved):

albertz:
Looks good. Did you try? What's your experience with it?

Zettelkasten:
> Looks good. Did you try? What's your experience with it?

So far very good (apart from that we had to build TF2.4 from source because we use an outdated Cuda and there is this bug I mentioned - but that's now also available internally at `
/work/tools/asr/python
`).
I mainly used this to profile the peak memory usage, which was super simple to use and gave me exactly what I needed (a list of tensors at peak time + their shapes and memory usages).
But there's so many features more that are available which I'll definitely also test out.
I can highly recommend to try it yourself when you need any sort of profiling.

albertz:
> TF2.4

Why not a more recent TF version?

> outdated Cuda

Why not a more recent CUDA version?

> there is this bug I mentioned

Which bug?

> I can highly recommend to try it yourself

I definitely will.

I think we should maybe also do that more frequently, to see if there is any unexpected bottleneck in some cases. There is lots of code which is not really tested with profiling, where we can potentially do things different, but we just don't really know what's faster.

Esp also compare to alternative implementations. E.g. for Conformer, there are multiple implementations, also in PyTorch or in TensorFlow. It would be good to compare them under a profiler.

Zettelkasten:
> Why not a more recent TF version?

> Why not a more recent CUDA version?

Hm yea, would've been nicer to just use the latest versions.
I mainly did this because I wanted to debug something entirely else, and with the ancient tools that we at i6 use (TF 2.3, Cuda 10, GCC 5.4.0), this seemed like the least effort. I already had some issues because of the old GCC version when compiling TF2.4 (see https://github.com/tensorflow/tensorflow/issues/41926 if you are interested), not sure how it is with TF2.6. Maybe it works with clang better.
I talked to the admins, and they said installing Cuda 11 was on their list for quite some time already, but it just didn't happen yet. I don't know how much effort it is to build this s.t. it works on all cluster nodes and so.

> > there is this bug I mentioned
> 
> Which bug?

In the docs, I added a note: In TF 2.3 and before, [the memory view does not show a breakdown of the tensors at peak memory usage when on GPU](https://github.com/tensorflow/tensorflow/issues/42123#issuecomment-675047711).

> I think we should maybe also do that more frequently, to see if there is any unexpected bottleneck in some cases. There is lots of code which is not really tested with profiling, where we can potentially do things different, but we just don't really know what's faster.

Yes, that would be very cool.

albertz:
> installing Cuda 11

You can also always install it by yourself, by simply putting it into your home in some directory.

Despite, at least on my machine (indigo), I already have it in `/usr/local/cuda-11.0`.

Issue #283, opened 2020-04-01
TensorFlow 2 compatibility: draft


albertz:
I want to list all possible options we have to implement this.

First some references:
* [Official TF1 to TF2 migration guide](https://www.tensorflow.org/guide/migrate).
* [My StackOverflow question: Best practice to write code compatible to both TensorFlow 1 and 2
](https://stackoverflow.com/questions/60966376/best-practice-to-write-code-compatible-to-both-tensorflow-1-and-2)
* [StackOverflow: Implicitly enable TensorFlow v2 behaviour in TensorFlow v1](https://stackoverflow.com/questions/58625549/implicitly-enable-tensorflow-v2-behaviour-in-tensorflow-v1). In TF 1, use `enable_v2_behavior` and `tensorflow.compat.v2` (since TF 1.14).
* Lingvo [compat.py](https://github.com/tensorflow/lingvo/blob/master/lingvo/compat.py). The code then mostly uses `import lingvo.compat as tf`.

In any case, also for TF 2, we would use graph-mode, not eager mode.

The current goal:
* A transition phase for a while (a year or so):
  * The code should run fine on both TF 1 (including earlier versions, e.g. 1.8) and TF 2.
  * We should be able to *optionally* use some new TF 2 features.
* We slowly could drop the TF 1 support later.

I don't really see this question answered in the migration guide. Maybe the answer is that there simply is no "best practice" for this.

We could maybe do sth like this:

    try:
      # https://www.tensorflow.org/guide/migrate
      import tensorflow.compat.v1 as tf
      tf.disable_v2_behavior()
    except ImportError:
      import tensorflow as tf

However, that has some drawbacks:

* We would need to place this code snippet in every module where I use TF.
  * We could put this in some own module, like `TFUtil.py`, and then simply do `from TFUtil import tf` everywhere. But this is somewhat ugly.
* This might make code complicated which actually wants to use some new TF 2 feature.

We could maybe do sth like this:

    try:
      # https://www.tensorflow.org/guide/migrate
      import tensorflow.compat.v1 as tf1
      tf1.disable_v2_behavior()
      import tensorflow as tf2
    except ImportError:
      import tensorflow as tf1
      tf2 = None

This might be more clean. But I'm not sure if this is the way to go. Or how other people do it.

Also, instead of calling `disable_v2_behavior`, we could maybe just call `disable_eager_execution`. But that might make some other code more difficult to make compatible for both TF 1 and TF 2.

[...]

PR #700, opened 2021-09-30
remove disable_control_flow_v2

albertz:


albertz:
I'm actually not really sure why I put this at all. I assume something breaks with this. But let's see for the tests.

albertz:
In `test_FetchHelper_loop`, there is the comment:
```
# Note: This only works correct with tf.compat.v1.disable_control_flow_v2() currently...
# Maybe we need to extend graph_editor?
```
Although I'm not sure if the logic around the fetch helper is anyway partly broken? I remember that it does not work properly in some (many? all?) cases.

albertz:
Note, this was introduced via #283, and then 07dc3ca19807e2ab.

albertz:
Ok, as expected, some tests are failing. Actually a lot. I need to look into that in more depth.

Some few things related to the fetch helper logic, as expected.

Then some more on probably some confusion on the control flow context. Does this concept not work anymore with control flow V2? Or is different in some way?

albertz:
This is continued in #961.

PR #701, opened 2021-09-30
added warning when extra config parameter can't be evaluated

tbscode:
### Why:

Overwriting a config param "++param" with a function tag from the current config (e.g.: `++search_data "config:get_sprint_dataset('dev')"` ) fails if the param was already set in the config, and had a different type.

### What:

Adding a warning; so user knows that the evaluation failed because the param was previously set in the config.
Also if the evaluation fails, it will use the param as-is. ( Optional )

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -190,7 +190,11 @@ class Config:
       if value_type == str:
         pass  # keep as-is
       else:
-        value = eval(value)
+        try:
+          value = eval(value)
+        except:
+          from returnn.log import log
+          print("WARNING: can't evaluate config param '{}' to previous type: {}. Keeping as string.".format(value, value_type),file=log.v1)
       self.typed_dict[key] = value
       return
     if value.find(',') > 0:
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -192,9 +192,10 @@ class Config:
       else:
         try:
           value = eval(value)
-        except:
+        except SyntaxError:
           from returnn.log import log
-          print("WARNING: can't evaluate config param '{}' to previous type: {}. Keeping as string.".format(value, value_type),file=log.v1)
+          print("WARNING: can't evaluate config param '%s' to previous type: %s. Keeping as string."
+                % (value, value_type), file=log.v1)
       self.typed_dict[key] = value
       return
     if value.find(',') > 0:
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -194,8 +194,9 @@ class Config:
           value = eval(value)
         except SyntaxError:
           from returnn.log import log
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -190,7 +190,11 @@ def add_line(self, key, value):
       if value_type == str:
         pass  # keep as-is
       else:
-        value = eval(value)
+        try:
+          value = eval(value)
+        except:
```
albertz, line 195: You should never catch all exceptions (including system exceptions) but always only those derived from `Exception` max, or better only those which you really expect here.

As I understood you, probably it should be:
```suggestion
        except SyntaxError:
```

Btw, PyCharm should have told you exactly about this. Please see our contribution rules.

returnn/config.py
```diff
@@ -190,7 +190,11 @@ def add_line(self, key, value):
...
+          from returnn.log import log
+          print("WARNING: can't evaluate config param '{}' to previous type: {}. Keeping as string.".format(value, value_type),file=log.v1)
```
albertz, line 197: You should keep the code consistent to our other code. We use the `%` formatting for strings.

Also, you should follow PEP8. There is a missing space. But PyCharm also should tell you about this. See our contribution rules.

albertz:
I see that you made your own fork and you pushed your changes directly to your master (tbscode:master).
You should always make a custom branch for a PR.
Also, you could have created that branch also directly here in this repo, and prefix it by your name, e.g. call it like `schupp-config-eval-fix` or so.
This is for the future, not really needed for this PR now (you anyway cannot change this anymore for this PR).

Also, in general, read the [contribution rules](https://github.com/rwth-i6/returnn/blob/master/CONTRIBUTING.md).

tbscode:
Ok I've fixed for pep8 & our guideline < 120 chars pl.
Using "%s" for the should be correct?

returnn/config.py
```diff
@@ -190,7 +190,12 @@ def add_line(self, key, value):
       if value_type == str:
         pass  # keep as-is
       else:
-        value = eval(value)
+        try:
+          value = eval(value)
+        except SyntaxError:
+          from returnn.log import log
+          print("WARNING: can't evaluate config param '%s' to previous type: %s. Keeping as string."
+                % (value, value_type), file=log.v1)
```
albertz, line 198: Do not use a hanging indent like that.
```suggestion
          print(
            "WARNING: can't evaluate config param '%s' to previous type: %s. Keeping as string." % (value, value_type),
            file=log.v1)
```

tbscode: Ok, else the line is over 120 chars, should I change the message?

tbscode: ah i see

Issue #694, opened 2021-09-27
NotFoundError: No registered 'Const' OpKernel for 'GPU' devices


robin-p-schmitt:
When using the code from https://github.com/rwth-i6/returnn_common/blob/main/models/transducer/recomb_recog.py inside of a transducer experiment (e.g. https://github.com/rwth-i6/returnn-experiments/blob/master/2020-rnn-transducer/configs/rna3c-lm4a.convtrain.switchout6.l2a_1e_4.nohdf.encbottle256.attwb5_am.dec1la-n128.decdrop03.decwdrop03.pretrain_less2_rep6.mlr50.emit2.fl2.fixmask.rna-align-blank0-scratch-swap.encctc.devtrain.config), there is the following error:

```
NotFoundError: No registered 'Const' OpKernel for 'GPU' devices compatible with node {{node ConstantFolding/output/rec/out_str/global_tensor_shared_vocab_0x479c99809419f4b4/Const_enter}}
	 (OpKernel was found, but attributes didn't match) Requested Attributes: _XlaHasReferenceVars=false, dtype=DT_STRING, value=Tensor<type: string shape: [1031] values: <s>  UNK  i ...>, _device="/job:localhost/replica:0/task:0/device:GPU:0"
	.  Registered:  device='XLA_CPU_JIT'; dtype in [DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8, DT_INT16, ..., DT_COMPLEX128, DT_HALF, DT_UINT32, DT_UINT64, DT_STRING]
  device='XLA_GPU_JIT'; dtype in [DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8, DT_INT16, ..., DT_COMPLEX128, DT_HALF, DT_UINT32, DT_UINT64, DT_STRING]
  device='CPU'
  device='GPU'; dtype in [DT_VARIANT]
  device='GPU'; dtype in [DT_BOOL]
  device='GPU'; dtype in [DT_COMPLEX128]
  device='GPU'; dtype in [DT_COMPLEX64]
  device='GPU'; dtype in [DT_UINT64]
  device='GPU'; dtype in [DT_INT64]
  device='GPU'; dtype in [DT_QINT32]
  device='GPU'; dtype in [DT_UINT32]
  device='GPU'; dtype in [DT_QUINT16]
  device='GPU'; dtype in [DT_QINT16]
  device='GPU'; dtype in [DT_INT16]
[...]

PR #702, opened 2021-10-01
Explicitly execute function on CPU

robin-p-schmitt:
This is related to #694.
@albertz

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -3868,6 +3868,11 @@ def get_shared_vocab(vocab_strings):
   :return: shape (len(vocab_strings),), tf.string
   :rtype: tf.Tensor
   """
+  # Usually TF automatically places the const strings where appropriate.
+  # This can fail in some cases:
+  # https://github.com/rwth-i6/returnn/issues/694
+  # https://github.com/tensorflow/tensorflow/issues/52200
+  # So we enforce it to be on CPU, which always should work.
   with tf.device("/cpu:0"):
     return global_tensor(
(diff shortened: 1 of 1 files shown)


albertz (approved):

returnn/tf/util/basic.py
```diff
@@ -3868,9 +3868,10 @@ def get_shared_vocab(vocab_strings):
   :return: shape (len(vocab_strings),), tf.string
   :rtype: tf.Tensor
   """
-  return global_tensor(
-    lambda: tf.convert_to_tensor(vocab_strings),
-    name="shared_vocab_%s" % hex(hash(tuple(vocab_strings))).replace("-", "_"))
+  with tf.device("/cpu:0"):
```
albertz, line 3871:
```suggestion
  # Usually TF automatically places the const strings where appropriate.
  # This can fail in some cases:
  # https://github.com/rwth-i6/returnn/issues/694
  # https://github.com/tensorflow/tensorflow/issues/52200
  # So we enforce it to be on CPU, which always should work.
  with tf.device("/cpu:0"):
```

albertz (approved):

albertz:
I don't see a better solution for now to fix #694.

However, I'm also not perfectly happy with it. Maybe at some point in the future, TF supports const strings on GPU. And we did not really consider other alternative hardware here like TPUs, or whatever else. For all those cases, this solution will make it suboptimal now, maybe even cause slowdowns.

That is why I would have preferred to leave it to TF to automatically handle it. I hope we can get rid of this at some point, either when tensorflow/tensorflow#52200 is fixed, or when we use control flow V2 (#700).

Issue #596, opened 2021-08-31
`SplitDimsLayer` wrong feature axis


albertz:
Via https://github.com/rwth-i6/pytorch-to-returnn/pull/54.

```
# data: [B,F|F'feature:data'(512),T|'time:var:extern_data:data'[B]]
"Unflatten": {'class': 'split_dims', 'from': 'data', 'axis': 'T', 'dims': [-1, 1]},
# unflatten: [B,F'feature:data'(512),T|'time:var:extern_data:data'[B],F|'Unflatten_split_dims1'(1)]
```

I would expect that the original feature dim axis stays as before.

This is due to the default behavior of `NotSpecified`.
As it is not specified (`NotSpecified`) before, it stays not specified after.
But that changes the feature dim axis (1 before, then -1 after).

albertz:
@Zettelkasten what do you think? This should be changed, right?

Zettelkasten:
Ouh yes, this is against all principles that the order of axes should not matter.
In this case, maybe even in general, I am also quite sure that nobody relied on this behavior.
From my POV, passing `feature_dim_axis=NotSpecified` to the `Data` should be resolved to some concrete `int|None`, just like it is now for `time_dim_axis`. But that leaves some questions open, e.g. what should happen if you take a `[B,T]` tensor with `feature_dim_axis=NotSpecified`, and than expand dims by 1. Should the new dim be the feature dim or not? And so on ...

But then again I avoided using `NotSpecified` in all my configs, I never found it useful, only non-transparent. Maybe people that use this have other opinions, and use/like this "feature"?

albertz:
@dierkes-j Note that this was fixed now.

PR #705, opened 2021-10-01
SplitDimsLayer fix feature_dim_axis on feature-dim split

### LINKED ISSUE #704: `SplitDimsLayer` wrong out feature axis when splitting feature axis
`data` has shape [B,T,40]

`SplitDimsLayer` config:
```
"source0": {"class": "split_dims", "axis": "F", "dims": (-1, 1), "from": "data"},  # (T,40,1)
```

Results in:
```
layer root/'source0' output: Data{'source0_output', [B,T|'time'[B],F|F'feature:data'(40),'source0_split_dims1'(1)]}
```

Earlier, the feature dim axis was the last dim (i.e. the `source0_split_dims1` here).

Related is #596. Maybe actually the fix for that in #683 introduces this new bug.

(I found this bug in #703 but this is probably otherwise unrelated.)

albertz:
Fix #704.

WARNING: This potentially changed the behavior of configs.
E.g. the one reported in #703.
Most of the attention and transducer encoders which used initial convolutional layers.

Related is #596. Maybe actually the fix for that in #683 introduces this new bug.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3128,6 +3128,9 @@ class SplitDimsLayer(_ConcatInputLayer):
     out = data.copy_template_new_dim_tags(new_dim_tags)
     if data.time_dim_axis is None:
       out.time_dim_axis = None
+    else:
+      out.time_dim_axis = cls._map_old_axis_to_new_axis(
+        split_axis=axis, dims=dims, old_axis=data.time_dim_axis, use_remaining=True, split_offset=0)
     if data.feature_dim_axis is not None:
       expected_out_feature_dim_axis = cls._map_old_axis_to_new_axis(
         split_axis=axis, dims=dims, old_axis=data.feature_dim_axis, use_remaining=False, split_offset=-1)
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3058,11 +3058,12 @@ class SplitDimsLayer(_ConcatInputLayer):
     return tuple(new_dims_resolved)
 
   @classmethod
-  def _map_old_axis_to_new_axis(cls, split_axis, dims, old_axis, split_offset=None):
+  def _map_old_axis_to_new_axis(cls, split_axis, dims, old_axis, use_remaining=True, split_offset=None):
     """
     :param int split_axis:
     :param tuple[int] dims: might include -1
     :param int old_axis:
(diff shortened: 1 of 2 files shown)


albertz:
So, Git bisecting:

It is bad in current master (ea3e91f6e5c89d653a02939d21106945a39d61f6).
It was still good in 6f7f6227e8b6c1cad2bcd1894557dd7421df1f5a (10 days ago).
2ff056de31838b6425333a882759df0093b5bb63 is the first bad commit.
```
commit 2ff056de31838b6425333a882759df0093b5bb63
Author: Albert Zeyer <albert.zeyer@rwth-aachen.de>
Date:   Wed Sep 22 14:45:18 2021 +0200

    SplitDimsLayer, fix feature_dim_axis for batch-feature-major input
    
    Fix #596

 returnn/tf/layers/basic.py | 20 +++++++++++++++-----
 1 file changed, 15 insertions(+), 5 deletions(-)
```
So yes, as expected #683 introduced this bug.

albertz:
# Note on the potential wrong behavior

**This did not cause an error, just wrong behavior!**
When you used split dims like this:
```
"source0": {"class": "split_dims", "axis": "F", "dims": (-1, 1), "from": "data"},  # (B,T,40,1)
```
And then 2D convolution, it would operate on the two expected spatial dims. The spatial dims are those which are not the feature dim and the batch dim. So it matters, from these splitted dims 40 -> (40,1), which of these are actually the feature dim.

Earlier, the last axis (with dim 1) was the new feature dim.
#683 changed this, and dim 40 here kept being the feature dim, as you can see in this output:
```
layer root/'source0' output: Data{'source0_output', [B,T|'time'[B],F|F'feature:data'(40),'source0_split_dims1'(1)]}
```
In some follow up 2D conv with options:
```
"conv0": {
  "class": "conv", "from": "source0",
  "padding": "same", "filter_size": (3, 3),
  "n_out": 32, "activation": None, "with_bias": True},  # (B,T,40,32)
```
you then see:
```
layer root/'conv0' output: Data{'conv0_output', [B,T|'time'[B],'source0_split_dims1'(1),F|F'conv0:channel'(32)]}
```
So the second spatial dim this operated on was `source0_split_dims1` from above.
This was not the intended and old behavior.

albertz:
Note that with #706 we could have detected such problem more directly.

PR #707, opened 2021-10-05
replace http by https

albertz:


JackTemaki:
There seems to be an error in the diff for the c++ code related files, as every line got edited. Is this maybe a tabs vs. spaces problem?

albertz:
> There seems to be an error in the diff for the c++ code related files, as every line got edited. Is this maybe a tabs vs. spaces problem?

I assume wrong line ending (before the change), and then PyCharm fixed line endings.
Do you want that as separate commit or PR?

PR #712, opened 2021-10-11
Partition epoch as a multi-GPU dataset distribution method

patrick-wilken:
The main commit is 704ee2b. I propose to reuse the partition epoch logic to distribute different parts of the data to different GPUs. The reason I needed an alternative to the `random_seed_offset` method is that this one does not work in cases where the sequence ordering is not, or at least in some aspects not random. In my case this was true for the following setup: several `HDFDatasets` in default sequence ordering (too costly to shuffle at run-time because of huge number of sequences), combined with `CombinedDataset` with `laplace` sequence ordering, using the `sampling_sizes` parameter, i.e. taking a fixed number of sequences from the HDFDatasets per epoch. `random_seed_offset` has the desired effect on the `CombinedDataset` level, however the sequences that are sampled from the HDFDatasets are the same for all GPUs, which is bad.
With the new `partition` method, the sampling and shuffling will be done identical for all GPUs, but then a different partition is selected per GPU. This is done using the sequence ordering, so no data is loaded and then thrown away as with the `shard` method. An additional advantage over `random_seed_offset` is that the original meaning of an epoch is preserved.

Implementation-wise this is not 100% optimal, because it only works for datasets using `get_seq_order_for_epoch()`.
Also, I had to add a nasty `disable_horovod_partition` attribute to only apply the partitioning on `CombinedDataset` level and not in the sub-datasets. But this is a very special case for this sampling sequence ordering, I think all other current meta dataset configurations should work with partitioning being done in the sub-datasets.

### DIFF
--- a/tests/test_Dataset.py
+++ b/tests/test_Dataset.py
@@ -357,6 +357,49 @@ def test_horovod_partition():
     assert set(data_out) == set(range(num_seqs))
 
 
+def test_horovod_partition_combined_dataset_sampling():
+    num_seqs = 10
+    sampling_size = 12
+    dummy_data = [{"data": numpy.array([i])} for i in range(num_seqs)]
+    from returnn.datasets.meta import CombinedDataset
+    dataset = MapDatasetWrapper(FromListDataset(data_list=dummy_data))
+    combined_dataset = CombinedDataset(
+      datasets={"dataset": dataset}, data_map={("dataset", "data"): "data"}, sampling_sizes={"dataset": sampling_size},
+      data_dims={"data": (1, 1)}, seq_ordering="random")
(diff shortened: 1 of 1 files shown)


albertz:
> if you have a huge dataset and you want to shuffle it you will always need to store an `O(total_num_seq)` list of integers.

But you always have this `list(range(num_seqs))`, also with your proposed extensions.

Also, what is `num_seqs` in your case? Even if this is very large, this should still not be too much a problem.

And then shuffling this also not.

albertz:
Btw, I just saw [this Twitter post on large data shuffling behavior](https://twitter.com/borisdayma/status/1447939363296489473), which might be relevant for you.

Issue #719, opened 2021-10-19
CombineLayer mismatches feature dims


vieting:
I have a case where the inputs to the `CombineLayer` are of shape `(B, F, T)` and `(1, F)`. In this case, it does currently not work to find the matching dims as the `1` axis is mapped to `F` of input 1 and then, the `F` axis cannot be mapped anymore.

albertz:
> My issue was resolved with [rwth-i6/pytorch-to-returnn#58](https://github.com/rwth-i6/pytorch-to-returnn/pull/58). Should we close the issue or do you want to keep it open since it should still work with RETURNN?

No, this problem is not solved on RETURNN. You just avoid the RETURNN problem (which still exists) in pytorch-to-returnn now, so your linked issue is not relevant here anymore.

PR #720, opened 2021-10-19
add test_CombineLayer_feature_broadcast_swapped

vieting:
Added test case for #719


vieting:
I accidentally force pushed here, didn't know this branch was still used for a PR. I'm closing this due to inactivity.

albertz:
You did not add much details to the corresponding issue in #719. Is this still an issue? If so, it would be good if you could add the test case directly there.

vieting:
> You did not add this test case to the corresponding issue in #719. It would be good if you could add the test case directly there.

Done.

vieting:
Can I somehow unlink the branch from this PR? Otherwise I will always push changes to this PR if I use a branch `dev` on my fork.

Issue #723, opened 2021-10-21
TFUtil naming in comments


Atticus1806:
When looking through the code, I realized that often in the comments there is a reference for functions in `TFUtil` for more details, e.g. `:param str|float gamma_init: see :func:`TFUtil.get_initializer`, for the scale`.

If I understand it correctly, this e.g. is now in `tf.util.base`. I can start making some changes to the comments in a branch, but maybe an open issue would be nice, since most likely some parts in the comments will be missed.

PR #724, opened 2021-10-21
replace TFUtil by tf.util.basic or .data if appropiate

Atticus1806:
updated comments as mentioned in #723


Atticus1806:
> I just notice, `tf.util.basic` is wrong, it should be `returnn.tf.util.basic`.

Oh yes, you are right, I will change that. But still Pycharm does realize that it is a function but can not find the declarations which is not optimal I guess, but I am not sure if we can change that.

PR #747, opened 2021-11-10
Add UnflattenBatchLayer

vieting:
This PR adds `UnflattenBatchLayer` which is the inverse of the `FlattenBatchLayer`. It should undo the flattening and restore the original data in batch major/time major format as it was input to the `FlattenBatchLayer`.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3331,6 +3331,38 @@ class FlattenBatchLayer(_ConcatInputLayer):
     return out
 
 
+class UnflattenBatchLayer(_ConcatInputLayer):
+  """
+  Inverse of :class:`FlattenBatchLayer`, so recovers an axis previously merged into the batch axis
+
+  This basically wraps :func:`unflatten_with_seq_len_mask`.
+  """
+  layer_class = "unflatten_batch"
+
+  def __init__(self, **kwargs):
+    super(UnflattenBatchLayer, self).__init__(**kwargs)
+    x = self.input_data.copy_as_batch_major()
+    from ..util.data import BatchInfo
+    batch_info = x.get_batch_dim_tag().batch
+    seq_lens = [v.sizes for v in batch_info.virtual_dims if isinstance(v, BatchInfo.PackedDim)]
+    assert len(seq_lens) == 1, "only implemented for single packed dim"
+    seq_lens = seq_lens[0]
+    self.output.placeholder = tf_util.unflatten_with_seq_len_mask(
+      x.placeholder, seq_lens=seq_lens, batch_major=self.output.is_batch_major)
+
+  @classmethod
+  def get_out_data_from_opts(cls, sources, name, **kwargs):
+    """
+    :param list[LayerBase] sources:
+    :param str name:
+    :rtype: Data
+    """
+    out = get_concat_sources_data_template(sources, name="%s_output" % name)
+    out = out.copy_as_batch_major()
+    out = out.copy_template_unpack_batch()
+    return out
+
+
 class UnflattenNdLayer(_ConcatInputLayer):
   """
   This keeps the batch axis as-is, i.e. the flattening/unflattening did not happen on the batch axis.
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -1229,6 +1229,16 @@ class BatchInfo:
     dims_wo_beam = [dim for dim in self.virtual_dims if not isinstance(dim, BatchInfo.BeamDim)]
     return root._global_descendants_by_virtual_dims[tuple(dims_wo_beam)]  # must exist
 
+  def copy_remove_dim(self, remove_dim):
+    """
+    :param VirtualDimBase remove_dim:
+    :rtype: BatchInfo
+    """
+    assert self.virtual_dims
+    root = self.get_global_base()
+    dims_wo_dim = [dim for dim in self.virtual_dims if dim != remove_dim]
+    return root._global_descendants_by_virtual_dims[tuple(dims_wo_dim)]  # must exist
+
   def copy_set_beam(self, beam):
     """
     :param SearchBeam|None beam:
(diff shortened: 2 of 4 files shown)


vieting:
I'm not sure if `Data.copy_unpack_batch()` is already generic enough, especially the handling of virtual dims which are not a `PackedDim`.

returnn/tf/util/data.py
```diff
@@ -2337,6 +2337,32 @@ def copy_extend_batch(self, batch):
...
+    :rtype: Data
+    """
+    assert self.have_batch_axis()
+    assert self.batch, "%s: batch unset" % self
+    data = self.copy()
+    kwargs = self.get_kwargs(include_special_axes=False)
+
+    dim_tags = []
+    for dim_tag in data.dim_tags:
+      if dim_tag.is_batch_dim() and hasattr(dim_tag.batch, "virtual_dims"):
```
albertz, line 2353: You should not use `hasattr`. Why is that needed?
Maybe you had some cases where `dim_tag.batch` was `None`. Then check for that.
```suggestion
      if dim_tag.is_batch_dim() and dim_tag.batch:
```

returnn/tf/util/data.py
```diff
@@ -2337,6 +2337,32 @@ def copy_extend_batch(self, batch):
...
+        for virtual_dim in dim_tag.batch.virtual_dims:
+          if isinstance(virtual_dim, BatchInfo.PackedDim):
+            dim_tags.append(virtual_dim.dim_tag)
+            kwargs["batch"] = virtual_dim.dim_tag.batch
+          else:
+            dim_tags.append(DimensionTag(kind=DimensionTag.Types.Batch))
```
albertz, line 2359: This doesn't really make sense.
You should create a new function in `BatchInfo` similar to `copy_remove_beam`, like `copy_remove_dim` or so.
And then the code above should be like:
```suggestion
        batch = dim_tag.batch
        new_batch_dim_tag = None
        for virtual_dim in batch.virtual_dims:
          if isinstance(virtual_dim, BatchInfo.PackedDim):
            dim_tags.append(virtual_dim.dim_tag)
            batch = batch.copy_remove_dim(virtual_dim)
          elif not new_batch_dim_tag:
            new_batch_dim_tag = DimensionTag(kind=DimensionTag.Types.Batch)
            dim_tags.append(new_batch_dim_tag)
        assert new_batch_dim_tag
        new_batch_dim_tag.batch = batch
        kwargs["batch"] = batch
```

returnn/tf/util/data.py
vieting: I don't think I had cases with `dim_tag.batch=None` but was thinking of the case that there are no virtual dims, but of course the attribute then exists anyway and contains an empty list.

returnn/tf/util/data.py
```diff
@@ -2337,6 +2337,32 @@ def copy_extend_batch(self, batch):
         data.placeholder = x
     return data
 
+  def copy_unpack_batch(self):
```
albertz, line 2340: This copies only the template. The function name should reflect that.
```suggestion
  def copy_template_unpack_batch(self):
```

albertz (changes requested):
Conceptional looks good! Some smaller problems though. But should be simple to fix.

returnn/tf/util/data.py
albertz: The batch usually always has virtual dims, specifically at least the `GlobalBatchDim`. (Although no virtual dims at all are allowed, and means that this is a batch with dim 1 then, which is useful for broadcasting.)

albertz: > I don't think I had cases with `dim_tag.batch=None`

Ok, then don't include that in the check. It should be invalid. You can add `assert dim_tag.batch` if you want.

returnn/tf/util/data.py
```diff
@@ -2337,6 +2337,32 @@ def copy_extend_batch(self, batch):
...
+    :rtype: Data
+    """
+    assert self.have_batch_axis()
+    assert self.batch, "%s: batch unset" % self
+    data = self.copy()
+    kwargs = self.get_kwargs(include_special_axes=False)
+
+    dim_tags = []
+    for dim_tag in data.dim_tags:
+      if dim_tag.is_batch_dim() and dim_tag.batch and len(dim_tag.batch.virtual_dims) > 0:
```
albertz, line 2353: See my previous comment.
```suggestion
      if dim_tag.is_batch_dim() and len(dim_tag.batch.virtual_dims) > 0:
```
Ok, you can leave the `len(dim_tag.batch.virtual_dims) > 0` though. In that case, it makes sense to just leave it as-is.

returnn/tf/layers/basic.py
```diff
@@ -3331,6 +3331,37 @@ def get_out_data_from_opts(cls, sources, name, axis="T", batch_major=True, **kwa
...
+  def __init__(self, **kwargs):
+    super(UnflattenBatchLayer, self).__init__(**kwargs)
+    x = self.input_data
+    from ..util.data import BatchInfo
+    batch_info = x.get_batch_dim_tag().batch
+    seq_lens = [v.sizes for v in batch_info.virtual_dims if isinstance(v, BatchInfo.PackedDim)]
+    assert len(seq_lens) == 1, "only implemented for single packed dim"
+    seq_lens = seq_lens[0]
+    self.output.placeholder = tf_util.unflatten_with_seq_len_mask(
+      x.placeholder, seq_lens=seq_lens, batch_major=x.copy_unpack_batch().is_batch_major)
```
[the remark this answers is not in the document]
vieting: Shouldn't this be reflected here already? If the packed dim comes first, the output of `x.copy_unpack_batch()` should be batch major, otherwise not.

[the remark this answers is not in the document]
albertz: No. Batch major means that batch is the very first dim in `x`.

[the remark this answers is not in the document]
albertz: Actually, `unflatten_with_seq_len_mask` requires `x` to be batch major, and would fail when it is not.

returnn/tf/layers/basic.py
```diff
@@ -3331,6 +3331,37 @@ def get_out_data_from_opts(cls, sources, name, axis="T", batch_major=True, **kwa
...
+  """
+  Inverse of :class:`FlattenBatchLayer`, so recovers an axis previously merged into the batch axis
+
+  This basically wraps :func:`unflatten_with_seq_len_mask`.
+  """
+  layer_class = "unflatten_batch"
+
+  def __init__(self, **kwargs):
+    super(UnflattenBatchLayer, self).__init__(**kwargs)
+    x = self.input_data
```
albertz, line 3344:
```suggestion
    x = self.input_data.copy_as_batch_major()
```

returnn/tf/layers/basic.py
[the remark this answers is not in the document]
albertz: So with my proposed fix above, it should be fine.

albertz (changes requested):

returnn/tf/layers/basic.py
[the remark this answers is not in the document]
albertz: However, better use:
```suggestion
      x.placeholder, seq_lens=seq_lens, batch_major=self.output.is_batch_major)
```

albertz (approved):

Issue #597, opened 2021-08-31
Specify dim tags for layers that create new axes


Zettelkasten:
E.g. `CumConcatLayer` from #589 allows (even enforces) you to specify `dim_tag: DimensionTag` for the output axis.

I propose something similar for all other layers that introduce new dims. Basically for [all layers introducing a new dynamic seq length here](https://github.com/rwth-i6/returnn/wiki/Layers-which-introduce-new-dynamic-sequence-lengths) that could be useful.

This is already possible to do using an additional `ReinterpretDataLayer` (or so, something that lets you specify new dim tags. I don't think this exists yet, but is easy to add). Then you would do:
```
'x_': {'class': 'split_dims',  ...},
'x': {'class': 'reinterpret_data', 'from': 'x_', 'set_dim_tags': {
  'stag:split_dims0': DimensionTag(...), 'stag:split_dims1': DimensionTag(...)}}
```
Here, `split_dims0` and so on are the tag descriptions that `SplitDimsLayer` internally sets.
Not only is this solution pretty verbose, it also relies on these internal tag descriptions. As a user, you would need to look them up frequently, and we would need to make sure that they don't change internally.
I would like to avoid this, proposing to add a `dim_tags: List[DimensionTag]` option directly to `SplitDimsLayer`.
With other layers, its the same idea.
This also promotes the best-practice that you should always name your axes properly.

Maybe, for some layers, we would even enforce the use of this. Especially for layers that really introduce an entirely new dim (like `SplitDimsLayer`). But also for others. When enforced, this would need a new behavior version (#508).

(Related to that I would also allow to specify just a string with the description as `dim_tag` to make this easier to use. Calling the `DimensionTag` constructor is not that nice in Sisyphus configs.)

---

# List of layers which need this

Also see: [Operations on dimension tags](https://github.com/rwth-i6/returnn/wiki/Operations-on-dimension-tags)

## Layers (potentially) introducing a new dynamic seq length

[List](https://github.com/rwth-i6/returnn/wiki/Layers-which-introduce-new-dynamic-sequence-lengths)

* `ConvLayer` (done, `out_spatial_dims`, `out_dim`, `in_spatial_dims`, `in_dim`, #789)
* `PoolLayer` (done, `out_spatial_dims`, `out_dim`, `in_spatial_dims`, `in_dim`, #789)
* `TransposedConvLayer` (done, `out_spatial_dims`, `out_dim`, `in_spatial_dims`, `in_dim`, #791)
[...]

PR #759, opened 2021-11-23
allow_broadcast_all_sources, changed default, new behavior version

### LINKED ISSUE #691: Implicit broadcasting for all inputs might be bad
Example for `CombineLayer` (also `EvalLayer`) where you have some input [B,D] and [B,T,D]. Via the current logic, it uses `Data.get_common_data`, and that results in [B,T,D], and then `Data.copy_compatible_to` extends [B,D] to [B,1,D] such that it can be broadcasted. I think this is still fine.

But now consider inputs [B,T,D1] and [B,T,D2]. `get_common_data` would return [B,T,D1,D2], so both (all) inputs would broadcast.
Or inputs [B,T1,D] and [B,T2,D], which would result in [B,T1,T2,D].

Consider also the case where we make the distinction based on dim tags, where D1 and D2 could be different dim tags but have the same dimension.

There are surely valid use cases for this. But I question whether this should be implicit, or whether broadcasting for all inputs should only be allowed in an explicit way. Broadcasting for all inputs means that `get_common_data` returns a strict superset of dimensions over all inputs. Otherwise, there would be at least one input which already contains all dimensions. I think for this case, implicit broadcasting is probably always fine.

For broadcasting of all inputs, this could maybe not what the user intended (maybe he/she expected D1 == D2), and only be a consequence of non-matching dim tags. And the error might only be in later layers, or maybe even no error at all, which might be confusing.

Related is also #597, #634, #632, #706

Maybe it could be allowed only if the user specifies `out_dims` (or `out_shape`). Or more specific for this issue, sth like `broadcast_dims`. The most obvious example is `CombineLayer` (and `EvalLayer`) but there are probably other layers as well. This might also be an issue for all `_ConcatInputLayer` derived layers.

In all cases, it probably needs a new behavior version (#508).

albertz:
Fix #691.

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -22,6 +22,19 @@ and not listing legacy/deprecated parameters.
 Version History
 ---------------
 
+Behavior version 4 (2021-11-23)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Broadcasting in all inputs simultaneously in layers and other ops
+is not allowed anymore by default.
+In all inputs simultaneously means that there is no input which has all common dimensions.
+
+Layers can explicitly allow this by specifying ``out_shape``.
+In case you stumble upon this, specify ``out_shape`` in the layer.
+
+See :func:`validate_broadcast_all_sources`
+and issue `#691 <https://github.com/rwth-i6/returnn/issues/691>`__.
+
 Behavior version 3 (2021-11-08)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -280,7 +280,7 @@ class LayerBase(object):
     return cls._base_get_out_data_from_opts(**kwargs)
 
   @classmethod
-  def _base_get_out_data_from_opts(cls, network, name, out_type=None, n_out=NotSpecified,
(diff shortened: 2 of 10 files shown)


Zettelkasten (approved):
Cool, looks good! Also the `out_spatial_dim` options.
So, we just entirely rely on `out_shape` for broadcasting implicit dims and don't introduce a special option for this (e.g. `allow_broadcast_all_sources: bool`)? Yeah, sounds good, I like that.
I presume you needed to change a few tests for this to specify `out_shape` (with quite some changes sometimes because they didn't use dim tags at all before)? That's okay though, maybe this is a little annoying at first for users to adapt their config, but it should be fine.

albertz:
> Also the `out_spatial_dim` options.

This was just to be able to fix some of the test cases in a clean way using `out_shape` because I required some ref to a newly created dim.

I will later do a separate PR for more on #597.

> So, we just entirely rely on `out_shape` for broadcasting implicit dims and don't introduce a special option for this (e.g. `allow_broadcast_all_sources: bool`)?

So far yes, also as discussed in the issue #691. Let's just see how that goes. I don't really like such a special option, and also I don't know why you would not just use `out_shape`. Except maybe where it requires to change an existing config too much.

Maybe we could also loosen `out_shape` and allow things like `"B"`, `"T"` and `"F"`, which would solve the issue of maybe too much needed changes in a config. Not sure on this. But this should be discussed in a separate issue.

> I presume you needed to change a few tests for this to specify `out_shape` (with quite some changes sometimes because they didn't use dim tags at all before)?

Yea, you see all needed changes here in this PR.

PR #771, opened 2021-11-25
SliceNdLayer, size can be DimensionTag

albertz:
#597

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -978,8 +978,9 @@ class SliceNdLayer(_ConcatInputLayer):
   def __init__(self, start, size, min_size=None, **kwargs):
     """
     :param LayerBase start: (B,...)
-    :param int|LayerBase|None size: if None, it uses the max possible size,
-      and it becomes a dynamic axis.
+    :param int|LayerBase|DimensionTag|None size:
+      We assume that this is >=0. If this might not be the case, use ``min_size=0``.
+      If None, it uses the max possible size, and it becomes a dynamic axis.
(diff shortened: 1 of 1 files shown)


returnn/tf/layers/basic.py
```diff
@@ -997,21 +997,29 @@ def __init__(self, start, size, min_size=None, **kwargs):
...
         size_t = seq_lens_t - start_t
-      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+      size_t = tf.maximum(size_t, min_size or 0)  # must make sure >=0 in any case
+      size = tf.reduce_max(size_t)  # scalar
     elif isinstance(size, LayerBase):
       size_data = size.output.copy_compatible_to(common_data, check_sparse=False)
-      size_t = size_data.placeholder
-      min_size = 0
-      size = tf.maximum(tf.reduce_max(size_t), min_size)  # scalar
+      size_t = size_data.placeholder  # assume already >=0
```
robin-p-schmitt, line 1010: Why can we assume here that `size_t >= 0`?

robin-p-schmitt:
We also need to add the `DimensionTag` option in the docstring of the `__init__` function, this is already done in `get_out_data_from_opts`. Apart from that and my other comment, I think it looks good

returnn/tf/layers/basic.py
albertz: I mean that we just do this assumption here. So it is up to the user to not give negative sizes. The user should know about this. When the user knows that they might be negative and should be clipped to 0, `min_size=0` can be explicitly provided.

PR #779, opened 2021-11-26
PadLayer, rename out_spatial_dims to out_dims

albertz:
#597

This is more consistent because the new output dim(s) could also be a feature dim.


Zettelkasten (approved):
:+1:  Makes sense to make this more consistent.

PR #780, opened 2021-11-26
WindowLayer/CumsumLayer: Extend check for when recurrent

### LINKED ISSUE #781: Better way to specify axis for layers operating on rec time axis
This is a special case of #597.

I noticed that for #776 in `WindowLayer`, but also for e.g. `CumsumLayer` that operate on the rec time axis, we have a check like this:
```
    if axis == "T" and data.time_dim_axis is None:
      # Assume inside RecLayer.
      # (then the code for a single step)
    else:
      axis = data.get_axis_from_description(axis)
      # (then the code for all steps at once, operating on `axis`)
```

This is flawed as soon as the input has two dynamic dims: Then the input might have a `time_dim_axis` even in the recurrent case, and the layer will silently behave very wrongly.

Instead, I would propose to allow `axis` to be a DimensionTag. Then, we can just check `axis == network.get_rec_parent_layer(inside_loop=False).time_dim_tag`.
Would we also want to allow the `stag:description` syntax here?
Also, in #391 for `CumConcatLayer`, we briefly thought of allowing `:i` as axis name. Maybe allow this too?

I worked on this a month ago, for `CumsumLayer`, see #780. But I never tested this, it was just something a colleague of mine stumbled across but then gave up on. Now with #776, and dim tags becoming popular, I remembered to raise this issue here.

Zettelkasten:
Fix #781 for `CumsumLayer`.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2543,8 +2543,9 @@ class WindowLayer(_ConcatInputLayer):
       assert window_dim and window_dim.dimension
       window_size = window_dim.dimension
     data = self.input_data.copy_as_batch_major()
-    if axis == "T" and data.time_dim_axis is None:
-      # Assume inside RecLayer.
+    from returnn.tf.util.basic import is_axis_from_description_recurrent
+    if is_axis_from_description_recurrent(axis=axis, network=self.network, data=data):
+      # Inside RecLayer.
       assert self._rec_previous_layer, "%s: expected to be used inside a RecLayer" % self
       assert padding == "same"
       assert window_right is not None or window_left is not None, (
@@ -2590,11 +2591,12 @@ class WindowLayer(_ConcatInputLayer):
     self.output.placeholder.set_shape(tf.TensorShape(self.output.batch_shape))
 
   @classmethod
-  def get_out_data_from_opts(cls, name, sources, window_size=None, window_dim=None,
+  def get_out_data_from_opts(cls, name, network, sources, window_size=None, window_dim=None,
                              axis="T", out_spatial_dim=None, padding="same", stride=1,
                              **kwargs):
     """
     :param str name:
+    :param returnn.tf.network.TFNetwork network:
     :param list[LayerBase] sources:
     :param int|None window_size:
     :param DimensionTag|None window_dim:
@@ -2610,8 +2612,9 @@ class WindowLayer(_ConcatInputLayer):
     data = get_concat_sources_data_template(sources)
     data = data.copy_template(name="%s_output" % name)
     data = data.copy_as_batch_major()
-    if axis == "T" and data.time_dim_axis is None:
-      # Assume inside RecLayer.
+    from returnn.tf.util.basic import is_axis_from_description_recurrent
+    if is_axis_from_description_recurrent(axis=axis, network=network, data=data):
+      # Inside RecLayer.
       assert not out_spatial_dim
       new_dim_axis = 1  # after batch
     else:
@@ -2641,9 +2644,10 @@ class WindowLayer(_ConcatInputLayer):
 
   # noinspection PyMethodOverriding
   @classmethod
-  def get_rec_initial_extra_outputs(cls, batch_dim, rec_layer, window_size=None, window_dim=None,
+  def get_rec_initial_extra_outputs(cls, network, batch_dim, rec_layer, window_size=None, window_dim=None,
                                     axis="T", sources=(), **kwargs):
     """
+    :param returnn.tf.network.TFNetwork network:
     :param tf.Tensor batch_dim:
     :param returnn.tf.layers.rec.RecLayer|LayerBase rec_layer:
     :param int|None window_size:
@@ -2657,8 +2661,9 @@ class WindowLayer(_ConcatInputLayer):
       window_size = window_dim.dimension
     data = get_concat_sources_data_template(sources)
     data = data.copy_as_batch_major()
-    if axis == "T" and data.time_dim_axis is None:
-      # Assume inside RecLayer.
+    from returnn.tf.util.basic import is_axis_from_description_recurrent
+    if is_axis_from_description_recurrent(axis=axis, network=network, data=data):
+      # Inside RecLayer.
       shape = list(data.batch_shape)
       shape[0] = batch_dim
       shape.insert(1, window_size)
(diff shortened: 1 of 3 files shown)


returnn/tf/layers/basic.py
```diff
@@ -2695,22 +2693,41 @@ def get_out_data_from_opts(cls, name, sources, axis="T", **kwargs):
...
 
+  @classmethod
+  def _is_recurrent(cls, axis, network, data):
+    """
+    :param str|DimensionTag axis:
+    :param returnn.tf.network.TFNetwork network:
+    :param Data data:
+    :rtype: bool
+    """
+    if not network.is_inside_rec_layer(inside_loop=False):
```
albertz, line 2721: `inside_loop=True`, right? When it is moved out, you want to operate on some normal axis.

But also, this is not the check you want to make. The layer should operate the same inside a rec layer, when you do cum_sum over some other unrelated axis.

returnn/tf/layers/basic.py
```diff
@@ -2695,22 +2693,41 @@ def get_out_data_from_opts(cls, name, sources, axis="T", **kwargs):
...
+      return False
+    if axis.lower() == "t" and data.time_dim_axis is None:  # legacy fallback
+      return True
+    if axis.lower() == ":i":
+      return True
+    if not isinstance(axis, DimensionTag):
+      return False
+    return axis == network.get_rec_parent_layer(inside_loop=False).time_dim_tag
```
albertz, line 2729: This misses some check if a rec parent layer exists. Otherwise this returns `None` and you get an attribute error here.

But some code like this would be better anyway:
```suggestion
    return axis == network.get_inside_rec_time_dim()
```

returnn/tf/layers/basic.py
```diff
@@ -2695,22 +2693,41 @@ def get_out_data_from_opts(cls, name, sources, axis="T", **kwargs):
...
+    :param str|DimensionTag axis:
+    :param returnn.tf.network.TFNetwork network:
+    :param Data data:
+    :rtype: bool
+    """
+    if not network.is_inside_rec_layer(inside_loop=False):
+      return False
+    if axis.lower() == "t" and data.time_dim_axis is None:  # legacy fallback
+      return True
+    if axis.lower() == ":i":
```
albertz, line 2725: I'm not sure that we really need this. I remember this was some suggestion in the self-attention generalization discussion but we actually did not implement anything like that.

And when you just use dim tags consistently (e.g. specify `axis` of the outer `RecLayer`), you don't need this special string logic here, you can directly pass the dim tag of the outer rec layer.

Zettelkasten: Okay yeah, I'll just use the dim tags and remove the `:i` case.

returnn/tf/layers/basic.py
Zettelkasten: Ah yes, I was confused about `inside_loop`. You are right.

> But also, this is not the check you want to make. The layer should operate the same inside a rec layer, when you do cum_sum over some other unrelated axis.

But in case I do cumsum over some unrelated axis, then `is_recurrent` should always return `False`. Which is what this check returns, so that should be fine, right?
There were some other issues though, is it correct now?

albertz: > But in case I do cumsum over some unrelated axis, then `is_recurrent` should always return `False`.

No? Why? `is_recurrent` (I guess you mean `is_inside_rec_layer` and related?) returns whether the layer (network) is inside a rec layer. It is not at all about what axis the `CumSumLayer` would operate on. When you call this function `is_inside_rec_layer`, you do not pass any axis or dim tag to it.

Zettelkasten: With "`_is_recurrent`" (sorry, with `_` before) I meant the function I wrote.
Now I moved this to `tf/util/basic`, and renamed it to `is_axis_from_description_recurrent`.
Maybe the name is not ideal still.

This function is simply meant to replace the old, flawed check `axis == T and data.time_dim_axis is None`, i.e. it should simply say whether `CumsumLayer` should operate step-by-step, or on all steps at once.

albertz:
Looks good.

Did you test this now? Maybe write a very trivial test case. Or two, one for the case of some unrelated dim, another for the case of the rec dim. Maybe in both cases similar to `test_reclayer_optimize_out_linear`, i.e. using `check_reclayer_optimize_out` such that it will check whether it is behaving correctly inside a rec layer and outside.

Zettelkasten:
Yeah okay, I'll add some tests later today

tests/test_TFNetworkRecLayer.py
```diff
@@ -4609,6 +4609,26 @@ def make_extern_data():
     assert "encoder_int" in cell.input_layers_moved_out
 
 
+def test_reclayer_optimize_out_cumsum_step_by_step():
+  from returnn.tf.util.data import BatchDim, DimensionTag
+  time_dim = DimensionTag(kind=DimensionTag.Types.Spatial, description="time")
+  feat_dim = DimensionTag(kind=DimensionTag.Types.Feature, description="feat", dimension=11)
+  check_reclayer_optimize_out(
+    subnet_layer_dict={
+      "class": "cumsum", "axis": time_dim, "out_shape": {BatchDim, feat_dim}, "n_out": 11},
```
albertz, line 4618: Why do you need `n_out`?

This looks inconsistent. If at all, I would use `out_dim` instead of `n_out`. But then I don't think you need it at all here, right?

tests/test_TFNetworkRecLayer.py
```diff
@@ -4609,6 +4609,26 @@ def make_extern_data():
...
+    feat_dim=feat_dim, time_dim=time_dim)
+
+
+def test_reclayer_optimize_out_cumsum_unrelated_axis():
+  from returnn.tf.util.data import BatchDim, DimensionTag
+  time_dim = DimensionTag(kind=DimensionTag.Types.Spatial, description="time")
+  feat_dim = DimensionTag(kind=DimensionTag.Types.Feature, description="feat", dimension=11)
+  check_reclayer_optimize_out(
+    subnet_layer_dict={
+      "class": "cumsum", "axis": feat_dim, "out_shape": {BatchDim, feat_dim}, "n_out": 11},
```
albertz, line 4628: Same here on `n_out`.

tests/test_TFNetworkRecLayer.py
Zettelkasten: I need `n_in = n_out` in `check_reclayer_optimize_out`, which has this logic currently:
```
  n_out = subnet_layer_dict.get("n_out", 17)
  # ...
  rec_layer_dict = {
    "class": "rec", ... "n_out": n_out
  }
```
I could also set the my input dim (my `feat_dim`) to 17 to match this default.

Zettelkasten: Maybe it's nicer how I have it now?

albertz: Ah this is just because of `check_reclayer_optimize_out`. Well yea, it doesn't really matter much then.

albertz (approved):

albertz:
This probably needs some further extensions:

- Also for `RecLayer`.
- It should also be possible to operate on a single step without any outer rec layer (e.g. [here](https://github.com/rwth-i6/returnn_common/blob/6d263503989777157fd40409ca2f4492949411cb/nn/rec.py#L37)).
  - Specifying `axis=None` would maybe be the natural thing for this. Maybe just extending `is_axis_from_description_recurrent` for that case is enough?
  - Or otherwise I thought about a special placeholder object like `NoAxis`, and then `axis=NoAxis`.
  - Or otherwise maybe an extension (or similar mechanism) of what I introduced in #846.

PR #784, opened 2021-11-27
MergeDimsLayer, only allow keep_order=True

### LINKED ISSUE #654: `MergeDimsLayer` should use `keep_order=True` (and disallow `keep_order=False`)
This is to follow the principle that the order of axes should never matter.
(`keep_order` is about using the user specified order in the layer `axes` argument, and ignoring the incoming data axes order.)

This probably needs a new behavior version (#508).

albertz:
Fix #654.

Also introduce `Data.get_axes_from_description` `"dim:%i"` variant. This is mostly as a simple way to fix many of the test cases. But I guess it could also be useful for the user in general. At least as a simple intermediate solution, once people start using dim tags (#597).

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -32,6 +32,8 @@ If that causes any problems, there is probably some other issue in your config.
 
 See issue `#654 <https://github.com/rwth-i6/returnn/issues/654>`__.
 
+Further, :class:`GenericAttentionLayer` uses ``auto_squeeze=False`` by default.
+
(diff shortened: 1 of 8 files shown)


albertz:
You see in the individual commits that I also changed `GenericAttentionLayer`, the default `auto_squeeze` to `False`, because I stumbled upon this, and the behavior was very unexpected.

However, it turned out that this breaks many other configs (esp single headed MLP attention configs). So I reverted this again.

We might want to think about this maybe later. But this should not be in this PR here.

Issue #594, opened 2021-08-29
2D/3D convolution/pooling (`ConvLayer`/`PoolLayer`) depends on order of axes


albertz:
We want to avoid anything which depends on the order of axes.

albertz:
Some related discussion is in #597. E.g. there was the proposal of `in_to_out_dims`. E.g. for `ConvLayer` with 2D conv, it would have 3 entries: 2 for the two spatial dims, and then one for the linear feature transformation. For `PoolLayer` and others which do not do a linear transformation, it would contain only entries for the spatial dims.

albertz:
The `filter_size` and related options would not be a list/tuple but instead a `dict[DimensionTag, int]` which also specifies the spatial axes it operates on.

albertz:
One open question is the native format of the `tf.Variable`. For that, the order of spatial axes matters.

Also, the native order of the input could change for various reasons, so we must not make it dependent on that.

JackTemaki:
This would be a very welcome addition, I remember I had some cases where I had to play around a lot to make Returnn run the filters over the axes I wanted... Isn't it fine to have a logic that moves the axes in the optimal order? Just like the [T,B,F] transformation for the nativelstm? I am not sure what is happening right now, but I think a re-ordering based in the selected Dimension tags is possible, and the tf.Variable can be defined accordingly.

albertz:
One core principle of RETURNN is that the order of the axes should never ever matter in any way. This core principles allows layers to reorder axes for efficiency (e.g. like `RecLayer`). And this is exactly what this issue here is about, because this is violated by `ConvLayer` for 2D/3D convolution, and similarly for `PoolLayer`, and some other similar layers.

The proposal here would also not have any order. It would just be like `filter_size={dimtag1: 3, dimtag2: 5}` or so.

But then, the order of axes for the underlying `tf.Variable` is ambiguous.

albertz:
> Isn't it fine to have a logic that moves the axes in the optimal order

We already do that. But this should not matter to the user.

> I had to play around a lot to make Returnn run the filters over the axes I wanted

This is not so much this issue here. This issue here is about the order of spatial axes, which matters.

I assume in your case, you had some mixup between spatial and feature dim. You do not want to reorder the axes. You simply need to redefine which axis is the feature dim (channel dim). You can use `ReinterpretDataLayer` with `set_axes: {F: ...}` for that.

JackTemaki:
[...]

Issue #573, opened 2021-08-11
Some layers (in certain configurations) do not behave correctly with rec automatic optimization


albertz:
What we want is: The user should never need to think about automatic optimization. At least for the behavior. The behavior should always be just like without the optimization. If there are any examples where this is not the case, this is a bug.
(The user might want to think about automatic optimization for performance reasons though. Maybe the user can write it in a way that allows for better optimization.)

This issue here is to collect all potential cases. Maybe directly create a separate issue for each specific case, and add it to the list here. Or if unclear, just make a comment first.

List of specific cases:

* #569
* `ConvLayer`, `PoolLayer`, `TransposedConvLayer`. Because the number of axis define whether it is 1D, 2D or 3D conv. Fixed via #789, #790, #791.
* `RecLayer` operating on an inner time axis (i.e. inside the loop you have sth like [B,W,D], then outside it would be [T,B,W,D]). Fixed via #793.
* Any layer which gets an `axis` as argument, where you specify `T`. When this gets optimized out, the meaning of `T` could change.
* Any layer which makes use of `time_dim_axis`, similar as the point before, similar also like `RecLayer`.

(*Also see [wiki on rec automatic optimization, special layer behavior](https://github.com/rwth-i6/returnn/wiki/Rec-automatic-optimization,-special-behavior-of-layers).*)

---

Why do we want this?

To simplify the mental model for the user. To make it straight-forward to write model definitions, and not run in unexpected bugs. E.g. when training works, recognition should work too, and vice versa.

But also when users write more complex code inside a loop, e.g. via https://github.com/rwth-i6/returnn_common/issues/16, it becomes really difficult to debug if automatic optimization could somehow break things.

albertz:
All layers in principle should be fixed now.

When dim tags are consistently used, together with #780, this is also not a problem.

PR #789, opened 2021-11-27
ConvLayer and PoolLayer, in_dim, in_spatial_dims, out_dim, out_spatial_dims

albertz:
#597

Fix 2D/3D convolution/pooling (`ConvLayer`/`PoolLayer`) depends on order of axes (#594), when `in_spatial_dims` is used. There should be a follow-up maybe where this is required (for 2D/3D), with new behavior version.

Fix handling of unrelated dims, and also rec automatic optimization (#573).


albertz:
I might also directly do the same for `PoolLayer` here.

Issue #508, opened 2021-05-19
Ideas for RETURNN behaviour changes


JackTemaki:
As there are now many different ways (as in parameter names, syntax styles etc...) to make RETURNN do the same things, it might be a good idea to add a `behaviour_version` parameter to prohibit using "deprecated" parameters and suboptimal behaviour of RETURNN. It can also be used to change the default behaviour of some layers which is currently suboptimal, such as the batch norm layers. With each desired change (e.g. commit or PR) to the behaviour, the version number would have to be increased by one.

Here are some things we already recommend to avoid according to the documentation, e.g.:
 - `num_inputs` and `num_outputs` should not be used in favor of `extern_data`
 - using the "dict" syntax for `extern_data` instead of the tuple syntax
 - prohibit using old theano parameters such as "multiprocessing" when working with tensorflow
 - prohibit direct optimizer parameters, so writing `optimizer = {'class': "adam"}` or maybe `optimizer="adam"` instead of `adam=True`
 - always use explicit input and target definitions (no default fallback to "data" or "classes")
 
 Then there are other redundant parameters which I can not name explicitely, but I think we have:
  -  overlapping/redundant parameters names for the dataset (e.g. `eval` versus `search_data` or so)
  - learning_rate vs learning_rates
 
 And then we have sub-optimal default parameters, that should be changed or not have a default in the first place, e.g.:
 - batch norm layer parameters
 - `ExtractAudioFeatures` should have no defaults, so that the feature settings are clearly visible in the config
 - no default `lstm` unit for rec layer
 
Other things I would personally recommend doing in configs, e.g.:
 - always provide an explicit `available_for_inference` for the `extern_data` entries
 - always use `seq_order_control_dataset` within a `MetaDataset`
 - not using `out_type` for any layer except maybe eval layer. Although here I am not sure if this has any major implications. Maybe there are some networks with rather instable construction and you want to have this fixed by hand.
 
There is probably a lot more which needs to be collected/discussed, so this is just intended to give some initial ideas.

albertz:
As discussed, this is just an example list. For every individual potential change, we should make a separate issue, and discuss things there. Every such issue should use the potential-new-behavior tag (and maybe reference this issue here).

This issue here probably can be closed once we have a first version of `behaviour_version`.
The setup of `behaviour_version` can probably be done in a similar way as `use_tensorflow`.
(Btw, `use_tensorflow` is also a default we can change by this.)

JackTemaki:
[...]

PR #792, opened 2021-11-29
Conv/Pool, require in_spatial_dims for more than one spatial dims

### LINKED ISSUE #594: 2D/3D convolution/pooling (`ConvLayer`/`PoolLayer`) depends on order of axes

albertz:
Fix #594.

This introduces a new behavior version (#508).

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -22,6 +22,30 @@ and not listing legacy/deprecated parameters.
 Version History
 ---------------
 
+Behavior version 8 (2021-11-30)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


albertz:
Ok, some further thought is needed here.

In principle, it should be done like that. And that requires a new behavior version.

However, we have the rule for new behavior version that a transition is very easy (#508). Requiring `DimensionTag` instances (`in_spatial_dims`), while somewhat straight-forward, still requires some non-trivial changes, and somewhat invalidates this rule.

So, to allow for a somewhat simpler solution, my idea was to allow `in_spatial_dims` to also contain `str` to refer to axes via descriptions (`"F"`, `"stag:..."` etc). However, descriptions like `"static:..."` or so again have the fundamental problem of depending on the order. For that reason, I created #795 to also disallow this, which is anyway a good idea.

PR #795, opened 2021-11-29
Enforce axis description order independent

albertz:
This is required to ensure that layers can reorder axes in whatever way to allow for potential optimizations. Nothing should depend on the order of axes. See the [RETURNN principles](https://github.com/rwth-i6/returnn/wiki/RETURNN-principles).

This is also for #792 to allow for an easier transition.

This introduces a new behavior version (#508).

While it probably requires changes for many configs, the changes should still be quite simple.


Zettelkasten (commented):
Looks good.
I don't like the use of `dim:123` so much: I don't like that my config could break if I choose an unfortunate combination of hyperparameters, or - much worse - my config could do silently do something entirely else if I specify `axes: "dim:8"` somewhere...
But probably these cases where it fails silently are rare (most layers expect a single input axis, in which case you get an explicit error if multiple match), and I agree it is easier to adapt than to enforce dim tags. I don't really know what is better.

albertz:
`"dim:N"` would never silently fail though. At least that was my intention. I just fixed this in d08d0e0. It should only be allowed when unique.

Zettelkasten (approved):
Okay yes, then I am happy :)

PR #821, opened 2021-12-03
fix for Data.copy_template_unpack_batch

vieting:
I have a case in the PyTorch-to-RETURNN converter, where `Data.get_axes_by_tag_name` runs into a problem when checking `tag.description.lower()` because the `DimensionTag` that is created in `Data.copy_template_unpack_batch` does not have a `description`. This fixes the error for me, but I'm not sure if it should be fixed in a more generic way since creating a `DimensionTag` with `description=None` is allowed in general and could maybe lead to this error in other cases as well.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -2956,7 +2956,7 @@ class Data(object):
             dim_tags.append(virtual_dim.dim_tag)
             batch = batch.copy_remove_dim(virtual_dim)
           elif not new_batch_dim_tag:
-            new_batch_dim_tag = Dim(kind=Dim.Types.Batch)
+            new_batch_dim_tag = batch_dim
             dim_tags.append(new_batch_dim_tag)
         assert new_batch_dim_tag
         new_batch_dim_tag.batch = batch


returnn/tf/util/data.py
```diff
@@ -2928,7 +2928,7 @@ def copy_template_unpack_batch(self):
             dim_tags.append(virtual_dim.dim_tag)
             batch = batch.copy_remove_dim(virtual_dim)
           elif not new_batch_dim_tag:
-            new_batch_dim_tag = DimensionTag(kind=DimensionTag.Types.Batch)
+            new_batch_dim_tag = DimensionTag(kind=DimensionTag.Types.Batch, description="unpack_batch_dim")
             dim_tags.append(new_batch_dim_tag)
```
albertz, line 2932: Wait for #822. And then just do:

```suggestion
            dim_tags.append(batch_dim)
```

albertz:
I'm also wondering whether we really should allow `DimensionTag` with `description=None`.

returnn/tf/util/data.py
vieting: Alright, thanks :+1:

albertz: #822 is merged now. Can you rebase and do this change?

vieting: Yes, done. I just need to keep track of the `new_batch_dim_tag`.

returnn/tf/util/data.py
```diff
@@ -2956,8 +2956,8 @@ def copy_template_unpack_batch(self):
             dim_tags.append(virtual_dim.dim_tag)
             batch = batch.copy_remove_dim(virtual_dim)
           elif not new_batch_dim_tag:
-            new_batch_dim_tag = Dim(kind=Dim.Types.Batch)
-            dim_tags.append(new_batch_dim_tag)
+            new_batch_dim_tag = dim_tag
+            dim_tags.append(dim_tag)
```
albertz, line 2960: But it should be `batch_dim`, right?
```suggestion
            new_batch_dim_tag = batch_dim
            dim_tags.append(new_batch_dim_tag)
```

vieting: `batch_dim` is not an existing variable, not sure what you mean here?

albertz:
You need to rebase.


vieting ***@***.***> schrieb am Fr., 3. Dez. 2021, 17:54:

returnn/tf/util/data.py
albertz: You need to rebase.

vieting: I already did, now again with the current master. Still not seeing `batch_dim`.

albertz: [Here](https://github.com/rwth-i6/returnn/blob/4195c3fe3b2bf7871cdb33779df5fbf0792bbd0b/returnn/tf/util/data.py#L813) is `batch_dim`. Why don't you see that?

vieting: I expected it to be a local variable. Sorry. Added your suggestion now.

PR #827, opened 2021-12-06
Allow search output layers without target or with dense outputs 

dthulke:
This change mainly allows to set layers with dense outputs as `search_output_layer` so that their output in included in the search output file. My main use case for this is a simple way to output confidence scores calculated in the network, but this also supports to output any other type of dense outputs like logits, hidden states, etc.

The main changes in `search` are:
- the target key does not default to `default_target` for dense outputs
- outputs without target key or that are not serialisable by the dataset are directly printed as output

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -883,6 +883,8 @@ class Dataset(object):
 
   def serialize_data(self, key, data):
     """
+    Warning: this is deprecated. Directly use the corresponding Vocabulary to serialize the data
+
     :param str key: e.g. "classes". self.labels[key] should be set
     :param numpy.ndarray data: 0D or 1D
     :rtype: str
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -24,7 +24,8 @@ import typing
 
 from returnn.log import log
 from returnn.engine.batch import Batch, BatchSetGenerator
-from returnn.util.basic import PY3, try_run, NumbersDict, unicode, OptionalNotImplementedError
+from returnn.datasets.util.vocabulary import get_seq_labels_from_labels
+from returnn.util.basic import try_run, NumbersDict, unicode, OptionalNotImplementedError
 
 
 class Dataset(object):
@@ -876,6 +877,8 @@ class Dataset(object):
 
   def can_serialize_data(self, key):
     """
+    Warning: Deprecated, use the corresponding vocabulary to serialize the data
+
     :param str key: e.g. "classes"
     :rtype: bool
     """
@@ -883,29 +886,13 @@ class Dataset(object):
 
   def serialize_data(self, key, data):
     """
-    Warning: this is deprecated. Directly use the corresponding Vocabulary to serialize the data
+    Warning: Deprecated, use the corresponding vocabulary to serialize the data
 
     :param str key: e.g. "classes". self.labels[key] should be set
     :param numpy.ndarray data: 0D or 1D
     :rtype: str
     """
-    labels = self.labels[key]
-    if data.ndim == 0:
-      data = numpy.expand_dims(data, axis=0)
-    assert data.ndim == 1
-    if len(labels) < 1000 and all([len(label) == 1 for label in labels]):
-      # are these actually raw bytes? -> assume utf8
-      if all([ord(label) <= 255 for label in labels]):
-        try:
-          if PY3:
-            return bytes([ord(labels[c]) for c in data]).decode("utf8")
-          else:
-            return b"".join([bytes(labels[c]) for c in data]).decode("utf8")
-        except UnicodeDecodeError:
-          pass  # pass on to default case
-      return "".join(map(labels.__getitem__, data))
-    else:
-      return " ".join(map(labels.__getitem__, data))
+    return get_seq_labels_from_labels(data, labels=self.labels[key])
 
   def calculate_priori(self, target="classes"):
     """
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -24,7 +24,7 @@ import typing
 
 from returnn.log import log
 from returnn.engine.batch import Batch, BatchSetGenerator
-from returnn.datasets.util.vocabulary import get_seq_labels_from_labels
+from returnn.datasets.util.vocabulary import Vocabulary
 from returnn.util.basic import try_run, NumbersDict, unicode, OptionalNotImplementedError
 
 
@@ -877,8 +877,6 @@ class Dataset(object):
 
   def can_serialize_data(self, key):
     """
-    Warning: Deprecated, use the corresponding vocabulary to serialize the data
-
     :param str key: e.g. "classes"
     :rtype: bool
     """
@@ -886,13 +884,16 @@ class Dataset(object):
 
   def serialize_data(self, key, data):
     """
-    Warning: Deprecated, use the corresponding vocabulary to serialize the data
+    In case you have a :class:`Vocabulary`, just use :func:`Vocabulary.get_seq_labels`.
 
     :param str key: e.g. "classes". self.labels[key] should be set
     :param numpy.ndarray data: 0D or 1D
     :rtype: str
     """
-    return get_seq_labels_from_labels(data, labels=self.labels[key])
+    vocab = Vocabulary.create_vocab_from_labels(self.labels[key])
+    if data.ndim == 0:
+      return vocab.labels[data]
+    return vocab.get_seq_labels(data)
 
   def calculate_priori(self, target="classes"):
     """
--- a/returnn/datasets/util/vocabulary.py
+++ b/returnn/datasets/util/vocabulary.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 
 __all__ = [
   "Vocabulary",
+  "get_seq_labels_from_labels",
   "BytePairEncoding",
   "SamplingBytePairEncoding",
   "SentencePieces",
@@ -162,10 +163,41 @@ class Vocabulary(object):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int] seq:
+    :param list[int]|int|numpy.ndarray seq: 0D or 1D
     :rtype: str
     """
+    if isinstance(seq, int):
+      seq = [seq]
+    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
+      seq = numpy.expand_dims(seq, axis=0)
     return " ".join(map(self.labels.__getitem__, seq))
+
+
+def get_seq_labels_from_labels(seq, labels):
+  """
+  Warning: Deprecated, use the corresponding vocabulary to serialize the data
+
+  :param list[int]|int|numpy.ndarray seq: 0D or 1D
+  :param list[str] labels:
+  :rtype: str
+  """
+  if isinstance(seq, int):
+    seq = [seq]
+  elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
+    seq = numpy.expand_dims(seq, axis=0)
+  if len(labels) < 1000 and all([len(label) == 1 for label in labels]):
+    # are these actually raw bytes? -> assume utf8
+    if all([ord(label) <= 255 for label in labels]):
+      try:
+        if PY3:
+          return bytes([ord(labels[c]) for c in seq]).decode("utf8")
+        else:
+          return b"".join([bytes(labels[c]) for c in seq]).decode("utf8")
+      except UnicodeDecodeError:
+        pass  # pass on to default case
+    return "".join(map(labels.__getitem__, seq))
+  else:
+    return " ".join(map(labels.__getitem__, seq))
 
 
 class BytePairEncoding(Vocabulary):
@@ -335,6 +367,14 @@ class CharacterTargets(Vocabulary):
     return seq + self.seq_postfix
 
   def get_seq_labels(self, seq):
+    """
+    :param list[int]|int|numpy.ndarray seq: 0D or 1D
+    :rtype: str
+    """
+    if isinstance(seq, int):
+      seq = [seq]
+    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
+      seq = numpy.expand_dims(seq, axis=0)
     return "".join(map(self.labels.__getitem__, seq))
 
 
@@ -371,7 +411,11 @@ class Utf8ByteTargets(Vocabulary):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int] seq:
+    :param list[int]|int|numpy.ndarray seq: 0D or 1D
     :rtype: str
     """
+    if isinstance(seq, int):
+      seq = [seq]
+    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
+      seq = numpy.expand_dims(seq, axis=0)
     return bytearray(seq).decode(encoding="utf8")
--- a/returnn/datasets/util/vocabulary.py
+++ b/returnn/datasets/util/vocabulary.py
@@ -334,6 +334,9 @@ class CharacterTargets(Vocabulary):
       seq = [self.vocab[k] for k in sentence]
     return seq + self.seq_postfix
 
+  def get_seq_labels(self, seq):
+    return "".join(map(self.labels.__getitem__, seq))
+
 
 class Utf8ByteTargets(Vocabulary):
   """
@@ -365,3 +368,10 @@ class Utf8ByteTargets(Vocabulary):
     else:
       seq = list(bytearray(sentence.encode("utf8")))
     return seq + self.seq_postfix
+
+  def get_seq_labels(self, seq):
+    """
+    :param list[int] seq:
+    :rtype: str
+    """
+    return bytearray(seq).decode(encoding="utf8")
--- a/returnn/datasets/util/vocabulary.py
+++ b/returnn/datasets/util/vocabulary.py
@@ -6,7 +6,6 @@ from __future__ import print_function
 
 __all__ = [
   "Vocabulary",
-  "get_seq_labels_from_labels",
   "BytePairEncoding",
   "SamplingBytePairEncoding",
   "SentencePieces",
@@ -47,18 +46,20 @@ class Vocabulary(object):
       clz = BytePairEncoding
     return clz(**opts)
 
-  def __init__(self, vocab_file, seq_postfix=None, unknown_label="UNK", num_labels=None):
+  def __init__(self, vocab_file, seq_postfix=None, unknown_label="UNK", num_labels=None, labels=None):
     """
     :param str|None vocab_file:
     :param str|None unknown_label:
     :param int num_labels: just for verification
     :param list[int]|None seq_postfix: labels will be added to the seq in self.get_seq
+    :param list[str]|None labels:
     """
     self.vocab_file = vocab_file
     self.unknown_label = unknown_label
     self.num_labels = None  # type: typing.Optional[int]  # will be set by _parse_vocab
     self.vocab = None  # type: typing.Optional[typing.Dict[str,int]]  # label->idx
-    self.labels = None  # type: typing.Optional[typing.List[str]]
+    self.labels = labels
+
     self._parse_vocab()
     if num_labels is not None:
       assert self.num_labels == num_labels
@@ -84,7 +85,12 @@ class Vocabulary(object):
     """
     filename = self.vocab_file
     import pickle
-    if filename in self._cache:
+
+    if self.labels is not None:
+      self.vocab = {label: i for i, label in enumerate(self.labels)}
+      assert self.unknown_label is None or self.unknown_label in self.vocab
+      self.num_labels = len(self.labels)
+    elif filename in self._cache:
       self.vocab, self.labels = self._cache[filename]
       assert self.unknown_label is None or self.unknown_label in self.vocab
       self.num_labels = len(self.labels)
@@ -123,6 +129,22 @@ class Vocabulary(object):
     d = {label: idx for (idx, label) in enumerate(labels)}
     assert len(d) == len(labels), "some labels are provided multiple times"
     return d
+
+  @classmethod
+  def create_vocab_from_labels(cls, labels):
+    """
+    Creates a `Vocabulary` from the given labels. Depending on whether the labels are identified as
+    bytes, characters or words a `Utf8ByteTargets`, `CharacterTargets` or `Vocabulary` vocab is created.
+
+    :param list[str] labels:
+    :rtype: Vocabulary
+    """
+    if len(labels) < 1000 and all([len(label) == 1 for label in labels]):
+      # are these actually ordered raw bytes? -> assume utf8
+      if all([ord(label) <= 255 and ord(label) == idx for idx, label in enumerate(labels)]):
+        return Utf8ByteTargets()
+      return CharacterTargets(vocab_file=None, labels=labels, unknown_label=None)
+    return Vocabulary(vocab_file=None, labels=labels, unknown_label=None)
 
   def tf_get_init_variable_func(self, var):
     """
@@ -163,41 +185,10 @@ class Vocabulary(object):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int]|int|numpy.ndarray seq: 0D or 1D
+    :param list[int]|numpy.ndarray seq: 1D sequence
     :rtype: str
     """
-    if isinstance(seq, int):
-      seq = [seq]
-    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
-      seq = numpy.expand_dims(seq, axis=0)
     return " ".join(map(self.labels.__getitem__, seq))
-
-
-def get_seq_labels_from_labels(seq, labels):
-  """
-  Warning: Deprecated, use the corresponding vocabulary to serialize the data
-
-  :param list[int]|int|numpy.ndarray seq: 0D or 1D
-  :param list[str] labels:
-  :rtype: str
-  """
-  if isinstance(seq, int):
-    seq = [seq]
-  elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
-    seq = numpy.expand_dims(seq, axis=0)
-  if len(labels) < 1000 and all([len(label) == 1 for label in labels]):
-    # are these actually raw bytes? -> assume utf8
-    if all([ord(label) <= 255 for label in labels]):
-      try:
-        if PY3:
-          return bytes([ord(labels[c]) for c in seq]).decode("utf8")
-        else:
-          return b"".join([bytes(labels[c]) for c in seq]).decode("utf8")
-      except UnicodeDecodeError:
-        pass  # pass on to default case
-    return "".join(map(labels.__getitem__, seq))
-  else:
-    return " ".join(map(labels.__getitem__, seq))
 
 
 class BytePairEncoding(Vocabulary):
@@ -347,13 +338,15 @@ class CharacterTargets(Vocabulary):
   Also see :class:`Utf8ByteTargets`.
   """
 
-  def __init__(self, vocab_file, seq_postfix=None, unknown_label="@"):
+  def __init__(self, vocab_file, seq_postfix=None, unknown_label="@", labels=None):
     """
-    :param str vocab_file:
+    :param str|None vocab_file:
     :param list[int]|None seq_postfix: labels will be added to the seq in self.get_seq
     :param str|None unknown_label:
+    :param list[str]|None labels:
     """
-    super(CharacterTargets, self).__init__(vocab_file=vocab_file, seq_postfix=seq_postfix, unknown_label=unknown_label)
+    super(CharacterTargets, self).__init__(
+      vocab_file=vocab_file, seq_postfix=seq_postfix, unknown_label=unknown_label, labels=labels)
 
   def get_seq(self, sentence):
     """
@@ -368,13 +361,9 @@ class CharacterTargets(Vocabulary):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int]|int|numpy.ndarray seq: 0D or 1D
+    :param list[int]|numpy.ndarray seq: 1D sequence
     :rtype: str
     """
-    if isinstance(seq, int):
-      seq = [seq]
-    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
-      seq = numpy.expand_dims(seq, axis=0)
     return "".join(map(self.labels.__getitem__, seq))
 
 
@@ -411,11 +400,7 @@ class Utf8ByteTargets(Vocabulary):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int]|int|numpy.ndarray seq: 0D or 1D
+    :param list[int]|numpy.ndarray seq: 1D sequence
     :rtype: str
     """
-    if isinstance(seq, int):
-      seq = [seq]
-    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
-      seq = numpy.expand_dims(seq, axis=0)
     return bytearray(seq).decode(encoding="utf8")
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -29,6 +29,7 @@ from tensorflow.python.client import timeline
 
 from returnn.engine.base import EngineBase
 from returnn.datasets.basic import Dataset, Batch, BatchSetGenerator, init_dataset
+from returnn.datasets.util.vocabulary import get_seq_labels_from_labels
 from returnn.learning_rate_control import load_learning_rate_control_from_config, LearningRateControl
 from returnn.log import log
 from returnn.pretrain import pretrain_from_config
@@ -2268,7 +2269,7 @@ class Engine(EngineBase):
       output_layer_names = [output_layer_names]
     num_output_layers = len(output_layer_names)
 
-    # Create lists with information about the output layers. All of length num_targets.
+    # Create lists with information about the output layers. All of length num_output_layers.
     output_layers = []  # type: typing.List[LayerBase]
     out_beam_sizes = []  # type: typing.List[typing.Optional[int]]
     output_layer_beam_scores = []  # type: typing.List[typing.Optional[tf.Tensor]]
@@ -2343,7 +2344,9 @@ class Engine(EngineBase):
           # Interpret output as bytes/utf8-string.
           outputs[output_layer_idx] = bytearray(outputs[output_layer_idx]).decode("utf8")
 
-      serialized_output = []
+      # Create lists with serialized data. All of length num_output_layers.
+      serialized_outputs = []  # type: typing.List[typing.Optional[typing.Union[str,numpy.ndarray]]]
+      serialized_targets = []  # type: typing.List[typing.Optional[typing.Union[str,numpy.ndarray]]]
       # noinspection PyShadowingNames
       for output_layer_idx in range(num_output_layers):
         if output_layers[output_layer_idx].output.sparse:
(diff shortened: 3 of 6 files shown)


returnn/tf/engine.py
```diff
@@ -2285,15 +2285,18 @@ def search(self, dataset, do_eval=True, output_layer_names="output", output_file
         print("Given output %r has beam %s." % (output_layer, out_beam), file=log.v1)
         output_layer_beam_scores.append(output_layer.get_search_choices().beam_scores)
       out_beam_sizes.append(out_beam.beam_size if out_beam else None)
-      target_keys.append(output_layer.target or self.network.extern_data.default_target)
+      target_key = output_layer.target
+      if output_layer.output.sparse and not target_key:
+        # Use the default target key for sparse outputs
+        target_key = self.network.extern_data.default_target
+      target_keys.append(target_key)
```
albertz, line 2292: `target_key` can be `None` here. In that case, it should not be added to `target_keys`.

returnn/tf/engine.py
```diff
@@ -2338,6 +2343,17 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
           # Interpret output as bytes/utf8-string.
           outputs[target_idx] = bytearray(outputs[target_idx]).decode("utf8")
 
+      serialized_output = []
+      for target_idx in range(num_targets):
+        if output_layers[target_idx].output.sparse and \
```
albertz, line 2348: Do not use `\` at line end. Rather use brackets (`(...)`) when you need to use line breaks.

returnn/tf/engine.py
```diff
@@ -2338,6 +2343,17 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+            target_keys[target_idx] and dataset.can_serialize_data(target_keys[target_idx]):
+          serialized_output.append(
+            [dataset.serialize_data(key=target_keys[target_idx], data=output) for output in outputs[target_idx]]
+          )
```
albertz, line 2352: Do not  put a closing bracket on its own line.

returnn/util/basic.py
```diff
@@ -965,6 +965,8 @@ def better_repr(o):
     return "set([\n%s])" % "".join(map(lambda v: better_repr(v) + ",\n", o))
   if isinstance(o, float) and (math.isnan(o) or math.isinf(o)):
     return "float('%s')" % repr(o)
+  if isinstance(o, np.ndarray):
+    return better_repr(o.tolist())
```
albertz, line 969: This is not exactly correct. It should have `numpy.array(...)` around. The fallback code below would already handle it correctly. Why do you need this?

returnn/tf/engine.py
```diff
@@ -2285,15 +2285,18 @@ def search(self, dataset, do_eval=True, output_layer_names="output", output_file
...
+        target_key = self.network.extern_data.default_target
+      target_keys.append(target_key)
 
     out_cache = None
     seq_idx_to_tag = {}
     if output_file:
       assert output_file_format in {"txt", "py"}
       if output_is_dict:
         assert output_file_format == "py", "Text format not supported in the case of multiple output layers."
-      assert all(dataset.can_serialize_data(target_key) for target_key in target_keys)
```
albertz, line 2296: Why is this removed?

returnn/tf/engine.py
albertz: Ah sorry, I realize that `target_keys` needs to match all the other lists...

returnn/tf/engine.py
albertz: If this is for the case of `target_key is None`, then just add `if target_key`.

returnn/tf/engine.py
```diff
@@ -2360,37 +2377,41 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
             out_idx = batch_idx * out_beam_sizes[target_idx]
-          if target_keys[target_idx] and dataset.can_serialize_data(target_keys[target_idx]):
+          if target_keys[target_idx]:
             if do_eval:
-              print("  ref:", dataset.serialize_data(key=target_keys[target_idx], data=targets[target_idx][batch_idx]),
-                    file=log.v4)
+              if dataset.can_serialize_data(target_keys[target_idx]):
+                print("  ref:",
+                      dataset.serialize_data(key=target_keys[target_idx], data=targets[target_idx][batch_idx]),
+                      file=log.v4)
```
albertz, line 2385: Never indent like that. Always indent with 2 spaces.

returnn/tf/engine.py
dthulke: It probably makes sense to rewrite this part of the code a bit. Later in the code I still iterate over `range(num_targets)` and there I assume that the length of target_keys is equal to the number of output layers.

returnn/util/basic.py
dthulke: The default `repr` doesn't add numpy in front:
```
>>> repr(numpy.array(0.0, dtype=numpy.float32))
'array(0., dtype=float32)'
```
I.e. you need to `from numpy import array, float32` to eval the expression. What makes parsing the output even more error prone.
But I agree that the current solution is not correct in `better_repr`, so I would remove it.

dthulke: The better solution would be to add another output file format (e.g. json) that does not rely on `repr` and would be more suitable to be parsed by other applications.

returnn/tf/engine.py
dthulke: Why do we wan't to disallow this? I.e. if the dataset cannot serialise a target key, the corresponding layer output is just directly printed / included in the output file. Not sure though whether there is any case in which this is really the desired behaviour.

returnn/util/basic.py
albertz: Well, `better_repr` is to create a repr in Python format. For any other format, you would not use `better_repr`.

returnn/tf/engine.py
albertz: I think I mostly did this to avoid that the user can maybe have mistake. Now, any such mistake (wrong target key which is not serializable) would not throw an error anymore but silently just alter the behavior.

So, the question is, is this ok or not? Some others should give some opinion on this. (@patrick-wilken  or @JackTemaki maybe?)

returnn/tf/engine.py
albertz: Yea...

returnn/util/basic.py
patrick-wilken: Also note that the representation of a numpy array does not contain all the data if the array is large.
For the eval task I added `with numpy.printoptions(threshold=sys.maxsize):` to print all scores.

returnn/tf/engine.py
dthulke: Another point: in the case of e.g. regression tasks we have dense outputs and targets that are not serialisable by the corresponding dataset. But we still might want to output the result.

patrick-wilken: Well, if we want to support printing dense outputs this check is obsolete.
And the new check in the next line conceptually is also not quite correct. It assumes that if the output layer has a target, it should output sparse data. But any layer can have a target, most notably SoftmaxLayer.
The logic should be: if the output is sparse and the layer has a target and the dataset can serialize this target, do the serialization; else output raw data (which might be dense or sparse).
Then also it would be guaranteed that you get an output for all the keys defined by the user.

dthulke: > The logic should be: if the output is sparse and the layer has a target and the dataset can serialize this target, do the serialization; else output raw data (which might be dense or sparse).

This is the current behaviour.

I removed the check in the next line again to avoid confusion.

albertz:
Note on the serialization (@dthulke @patrick-wilken @JackTemaki ):

As for many things, the logic is a bit historical.

But in any case, we don't need to stick to it but we can clean it up. It should be logical.

I think it is a mistake that the serialization is a function of the dataset, and that it only works when you set a target on some layer with sparse output. Back then, I think only the dataset knew the labels, so it was kind of an easy way to get it serialized via the dataset.

This changed in many ways. Now every layer, or ever `Data` object, or specifically even every `Dim` (`DimensionTag`) can know about the labels, via the `vocab: Vocabulary` attribute. When a `Data` is sparse, there is `Data.sparse_dim: Dim`.

I think to clean this up, I would suggest this logic:

- If the layer output is sparse:
  - Try to use `Data.sparse_dim.vocab` when available.
  - If the vocab is not set, fallback to a target it if it set on the layer.
    - Via `ExternData`, see if the vocab is set on the target. (Again independent from the dataset.)
    - If that is not the case, then fallback to `Dataset.labels` and use that as vocab.
  - If that fails as well, error (but do not just fallback to the case below).
- If the layer output is not sparse, output as-is.

We can still use the `Dataset.serialize_data` logic for the sparse case. Although this code should be moved. Maybe to `Vocabulary` or so.

I think this extended cleaned up logic would not break any existing setups while providing sane error behavior. It would avoid unexpected behavior because some target or vocab is not correctly set. And it would still allow for any flexibility you need.

patrick-wilken:
Yes, using the vocabulary from the `Data` and making this independent from the dataset would be nice. For example, I remember having a case where I shifted a data stream with a layer and then used `register_as_extern_data` to be able to use it as a target of the output layer. I could not serialize it out of the box because the dataset did not know the new key. Now, it is already possible to set the vocabulary via the `out_type` of the layer, right? And then with your proposed changes this would work.

patrick-wilken:
> If that fails as well, error (but do not just fallback to the case below).

So never output sparse data as integers? That's ok I think to keep the behaviour simple. On the other hand, to always create a string with all the output tokens is already somewhat application specific. Maybe you don't even need a vocabulary, e.g. if the outputs are just cluster ids. But yeah, you can always just make your vocabulary a mapping from `int(id)` to `str(id)`. So I don't really mind.

albertz:
> > If that fails as well, error (but do not just fallback to the case below).
> 
> So never output sparse data as integers? That's ok I think to keep the behaviour simple. On the other hand, to always create a string with all the output tokens is already somewhat application specific. Maybe you don't even need a vocabulary, e.g. if the outputs are just cluster ids. But yeah, you can always just make your vocabulary a mapping from `int(id)` to `str(id)`. So I don't really mind.

Yea, I thought about this. But it's probably a rare case, and you already provided one simple solution for this. Another option would be to use `reinterpret_data` and set `sparse=False`. It's still int32 then and you get the indices.

I think this is fine in order to catch some other mistakes early on. And also I think the behavior is more predictable and sane.

albertz:
> what is the best way to set the vocab of the `sparse_dim`?

Usually this comes in the config in `extern_data`.

Either like:
```
from returnn.tf.util.data import batch_dim, SpatialDim, FeatureDim
time_dim = SpatialDim("time")
classes_dim = FeatureDim("classes", dimension=<num_classes>, vocab=<vocab>)

extern_data = {
  ...,
  "classes": {"dim_tags": [batch_dim, time_dim], "sparse_dim": classes_dim}
}
```
or (old way):
```
extern_data = {
  ...,
  "classes": {"sparse": True, "vocab":  <vocab>}
}
```
That should be correct for any layers which set the target.

In general (via the first way), you could set any sparse dim tags.

There is also `ReinterpretDataLayer` to do that.

Maybe some of these things are slightly incomplete, as some of this is only very recent. But then we can very simply fix or extend it now.

> `out_type` is currently ignored by some layers. But also for the layers where it is used (e.g. `CopyLayer`) it does not get set since the `sparse_dim` is overwritten by the version from the source layer

`out_type` is not really intended for that.

albertz:
> Also I implemented `get_seq_labels` for CharacterTargets and Utf8ByteTargets vocabs to get the same behaviour as in Dataset.serialize_data (i.e. no space between characters, utf8 bytes are properly decoded).

Why is that needed? Why not just call `get_seq_labels_from_labels` in all cases? I think it would be more safe to always rely on the same code.

tools/get-attention-weights.py
```diff
@@ -290,12 +291,13 @@ def fetch_callback(seq_idx, seq_tag, target_data, target_classes, output, output
...
-          if dataset.can_serialize_data(network.extern_data.default_target):
-            title += "\n" + dataset.serialize_data(
-              network.extern_data.default_target, target_classes[i][:output_len[i]])
+          if network.extern_data.default_target in dataset.labels:
+            title += "\n" + get_seq_labels_from_labels(
+              target_classes[i][:output_len[i]].tolist(), labels=dataset.labels[network.extern_data.default_target])
             ax = plt.gca()
             tick_labels = [
-              dataset.serialize_data(network.extern_data.default_target, np.array([x], dtype=target_classes[i].dtype))
+              get_seq_labels_from_labels(np.array([x], dtype=target_classes[i].dtype).tolist(),
```
albertz, line 299: Same here, avoid the `tolist()`. I think it's not really needed.

tests/test_GeneratingDataset.py
```diff
@@ -103,8 +103,9 @@ def test_StaticDataset_utf8():
   data.load_seqs(0, 1)
   raw_ = data.get_data(seq_idx=0, key="data")
   assert_equal(raw.tolist(), raw_.tolist())
-  assert data.can_serialize_data(key="data")
-  s_serialized = data.serialize_data(key="data", data=raw)
+  assert 'data' in data.labels
+  from returnn.datasets.util.vocabulary import get_seq_labels_from_labels
+  s_serialized = get_seq_labels_from_labels(raw.tolist(), labels=data.labels['data'])
```
albertz, line 108: Same here, avoid the `tolist()`.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,76 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+          # Try to get the vocab corresponding to the sparse dim
+          vocab = None
+          if output_layers[output_layer_idx].output.sparse_dim.vocab is not None:
+            vocab = output_layers[output_layer_idx].output.sparse_dim.vocab
+          elif target_keys[output_layer_idx] and self.network.extern_data.get_data(target_keys[output_layer_idx]).vocab is not None:
+            vocab = self.network.get_extern_data(target_keys[output_layer_idx]).vocab
+
+          if vocab is not None:
+            layer_outputs = outputs[output_layer_idx]
+            serialized_output = [vocab.get_seq_labels(output.tolist()) for output in layer_outputs]
```
albertz, line 2361: Use `get_seq_labels_from_labels` here instead.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,76 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+          elif target_keys[output_layer_idx] and target_keys[output_layer_idx] in dataset.labels:
+            # Fallback to the serialisation provided by the dataset
+            dataset_labels = dataset.labels[target_keys[output_layer_idx]]
+            serialized_output = [get_seq_labels_from_labels(output.tolist(), dataset_labels)
```
albertz, line 2365: Avoid the `tolist()`.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,76 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+                                 for output in outputs[output_layer_idx]]
```
albertz, line 2366: Do not indent like that. No hanging indents.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,76 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+          assert n_batch == len(targets[output_layer_idx])
 
-        if output_layers[target_idx].output.dim == 256 and output_layers[target_idx].output.sparse:
+        if output_layers[output_layer_idx].output.dim == 256 and output_layers[output_layer_idx].output.sparse:
           # Interpret output as bytes/utf8-string.
-          outputs[target_idx] = bytearray(outputs[target_idx]).decode("utf8")
+          outputs[output_layer_idx] = bytearray(outputs[output_layer_idx]).decode("utf8")
+
+      serialized_outputs = []
+      serialized_targets = []
```
albertz, line 2348: Type of these two lists should be defined, like above.
Also, some comment should be added on the expected length. I assume also `num_targets`?

returnn/tf/engine.py
```diff
@@ -2266,13 +2267,13 @@ def search(self, dataset, do_eval=True, output_layer_names="output", output_file
     output_is_dict = isinstance(output_layer_names, list)
     if not output_is_dict:
       output_layer_names = [output_layer_names]
-    num_targets = len(output_layer_names)
+    num_output_layers = len(output_layer_names)
 
     # Create lists with information about the output layers. All of length num_targets.
```
albertz, line 2272: Here it is not `num_targets` but `num_output_layers` instead now.

returnn/tf/engine.py
albertz: Or rather `num_output_layers` now.

returnn/datasets/basic.py
```diff
@@ -874,37 +874,6 @@ def is_less_than_num_seqs(self, n):
...
-          if PY3:
-            return bytes([ord(labels[c]) for c in data]).decode("utf8")
-          else:
-            return b"".join([bytes(labels[c]) for c in data]).decode("utf8")
-        except UnicodeDecodeError:
-          pass  # pass on to default case
-      return "".join(map(labels.__getitem__, data))
-    else:
-      return " ".join(map(labels.__getitem__, data))
-
```
albertz, line 907: I think I would not remove these functions, in case some people have some scripts which rely on these.
Rather, `serialize_data` can simply call to `get_seq_labels_from_labels`.

returnn/datasets/util/vocabulary.py
```diff
@@ -165,9 +166,34 @@ def get_seq_labels(self, seq):
     :param list[int] seq:
     :rtype: str
     """
+    if isinstance(seq, int):
```
albertz, line 169: Why is this needed? The documentation above says that it only accepts `list[int]`. Sth is wrong then.

But anyway, as commented elsewhere, I don't think you need any of the changes to `get_seq_labels`.

albertz:
Despite my comments, this is ready, right? If it is ready from your point of you, i.e. it should be reviewed by us, then you should click "Ready for review". But I anyway did some review already now.

albertz:
Btw, also see failing test. I guess some code style problem.

dthulke:
>> Also I implemented get_seq_labels for CharacterTargets and Utf8ByteTargets vocabs to get the same behaviour as in Dataset.serialize_data (i.e. no space between characters, utf8 bytes are properly decoded).

> Why is that needed? Why not just call get_seq_labels_from_labels in all cases? I think it would be more safe to always rely on the same code.

I don't like that get_seq_labels_from_labels uses heuristics to detect the type of labels to support the serialisation of three different vocabularies (the default word based vocabulary, CharacterTargets, Utf8ByteTargets).

I think it's nicer if each vocabulary directly does the correct serialisation for it's type of labels/data.
I.e. if you use characters as targets you should use the CharacterTargets vocabulary. If you are still the default vocabulary you should expect that your characters are handled as words (i.e. whitespaces are added / expected in serialisation and deserialisation).

I would suggest to use get_seq_labels_from_labels only as a fallback and maybe even disallow it in a future behaviour version (to force users to explicitly set a vocabulary).

albertz:
> I don't like that get_seq_labels_from_labels uses heuristics to detect the type of labels to support the serialisation of three different vocabularies (the default word based vocabulary, CharacterTargets, Utf8ByteTargets).

Yes right. But in any case, we should not use two different functions which are similar but not exactly the same, and now call in some cases one and in other cases the other. That could lead to very unexpected behavior in corner cases.

So either always call `get_seq_labels_from_labels` or always call `get_seq_labels`.

> I think it's nicer if each vocabulary directly does the correct serialisation for it's type of labels/data. ... CharacterTargets ...

Yes right.

> I would suggest to use get_seq_labels_from_labels only as a fallback and maybe even disallow it in a future behaviour version (to force users to explicitly set a vocabulary).

Why is this fallback needed? Just always use `get_seq_labels`.

In case of `Dataset.labels`, you can simply wrap that up in a `Vocabulary`.

albertz:
Note that you changed the behavior of `get_seq_labels` for some of the derived `Vocabulary` classes. I'm not sure if there are other cases in the code or in scripts where this would be unexpected and could break things.

There is also the function `get_seq`, which is defined in derived classes already. Which kind of matches the behavior of the additional `get_seq_labels` functions (just being the inverse of it). I think it should match exactly for consistency. Would that be the case?

returnn/datasets/util/vocabulary.py
```diff
@@ -162,12 +163,43 @@ def get_seq_indices(self, seq):
 
   def get_seq_labels(self, seq):
     """
-    :param list[int] seq:
+    :param list[int]|int|numpy.ndarray seq: 0D or 1D
```
albertz, line 166: Why is `int` or 0D needed?

dthulke:
> So either always call get_seq_labels_from_labels or always call get_seq_labels.

I agree

> In case of Dataset.labels, you can simply wrap that up in a Vocabulary.

The reason I didn't do it directly this way is that the vocabulary currently always requires a vocab_file (except the Utf8ByteTargets). I would suggest then to change the logic in the vocabulary constructor to allow directly passing the vocab /labels.

returnn/datasets/util/vocabulary.py
```diff
@@ -162,12 +163,43 @@ def get_seq_indices(self, seq):
...
+      try:
+        if PY3:
+          return bytes([ord(labels[c]) for c in seq]).decode("utf8")
+        else:
+          return b"".join([bytes(labels[c]) for c in seq]).decode("utf8")
+      except UnicodeDecodeError:
+        pass  # pass on to default case
+    return "".join(map(labels.__getitem__, seq))
+  else:
+    return " ".join(map(labels.__getitem__, seq))
```
albertz, line 200: This logic should be removed, and you should instead wrap it up in a `Vocabulary` and then call to `get_seq_labels`. Maybe it could use a similar heuristic on what type of vocab class you use here.

returnn/tf/engine.py
albertz: Do not use `tolist()`.

returnn/datasets/util/vocabulary.py
dthulke: The layer output can be 0D (i.e. if you don't have a time dim). We could either directly map this to 1D when calling this function (e.g. in search) or do it here. I think it's more robust to do it here.

returnn/datasets/util/vocabulary.py
albertz: Maybe you should make a separate function `create_vocab_from_labels`.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,79 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+            vocab = self.network.get_extern_data(target_keys[output_layer_idx]).vocab
+
+          if vocab is not None:
+            layer_outputs = outputs[output_layer_idx]
+            serialized_output = [vocab.get_seq_labels(output.tolist()) for output in layer_outputs]
+          elif target_keys[output_layer_idx] and target_keys[output_layer_idx] in dataset.labels:
+            # Fallback to the default serialization
+            dataset_labels = dataset.labels[target_keys[output_layer_idx]]
+            serialized_output = [
+              get_seq_labels_from_labels(output, dataset_labels) for output in outputs[output_layer_idx]]
```
albertz, line 2368: I think you also should use `get_seq_labels` here. Use `create_vocab_from_labels` on `dataset_labels`.

tools/dump-dataset.py
```diff
@@ -160,8 +161,7 @@ def dump_dataset(options):
         elif options.type == "stdout":
           extra = ""
           if target in dataset.labels and len(dataset.labels[target]) > 1:
-            assert dataset.can_serialize_data(target)
-            extra += " (%r)" % dataset.serialize_data(key=target, data=targets)
+            extra += " (%r)" % get_seq_labels_from_labels(targets, labels=dataset.labels[target])
```
albertz, line 164: This change is not needed anymore. (It doesn't really make sense to adapt code to replace one deprecated function by another deprecated function.)

Also, you miss the `can_serialize_data` check here.

tools/get-attention-weights.py
```diff
@@ -290,12 +291,13 @@ def fetch_callback(seq_idx, seq_tag, target_data, target_classes, output, output
...
-            title += "\n" + dataset.serialize_data(
-              network.extern_data.default_target, target_classes[i][:output_len[i]])
+          if network.extern_data.default_target in dataset.labels:
+            title += "\n" + get_seq_labels_from_labels(
+              target_classes[i][:output_len[i]].tolist(), labels=dataset.labels[network.extern_data.default_target])
             ax = plt.gca()
             tick_labels = [
-              dataset.serialize_data(network.extern_data.default_target, np.array([x], dtype=target_classes[i].dtype))
+              get_seq_labels_from_labels(np.array([x], dtype=target_classes[i].dtype),
+                                         dataset.labels[network.extern_data.default_target])
```
albertz, line 300: I would also just leave the old code here.

dthulke:
> Note that you changed the behavior of get_seq_labels for some of the derived Vocabulary classes. I'm not sure if there are other cases in the code or in scripts where this would be unexpected and could break things.

Yes, that could be an issue. But I would argue that the old behaviour of get_seq_labels for the character targets was wrong, so not sure whether anyone relied on that.

> There is also the function get_seq, which is defined in derived classes already. Which kind of matches the behavior of the additional get_seq_labels functions (just being the inverse of it). I think it should match exactly for consistency. Would that be the case?

Yes, now both functions match exactly.

returnn/datasets/util/vocabulary.py
```diff
@@ -162,12 +163,43 @@ def get_seq_indices(self, seq):
...
+    if isinstance(seq, int):
+      seq = [seq]
+    elif isinstance(seq, numpy.ndarray) and seq.ndim == 0:
+      seq = numpy.expand_dims(seq, axis=0)
     return " ".join(map(self.labels.__getitem__, seq))
 
 
+def get_seq_labels_from_labels(seq, labels):
+  """
+  Warning: Deprecated, use the corresponding vocabulary to serialize the data
```
albertz, line 178: I would remove this warning. When it is used (directly or indirectly) and there is no real alternative (such as for `Dataset.labels`, `Dataset.serialize_data`), then it doesn't make sense to declare it as deprecated.

You could instead write sth like:
```suggestion
  In case you have a :class:`Vocabulary`, just use :func:`Vocabulary.get_seq_labels`.
```

returnn/datasets/basic.py
```diff
@@ -876,34 +877,22 @@ def is_less_than_num_seqs(self, n):
 
   def can_serialize_data(self, key):
     """
+    Warning: Deprecated, use the corresponding vocabulary to serialize the data
```
albertz, line 880: Remove that.

returnn/datasets/basic.py
```diff
@@ -876,34 +877,22 @@ def is_less_than_num_seqs(self, n):
...
+
     :param str key: e.g. "classes"
     :rtype: bool
     """
     return key in self.labels
 
   def serialize_data(self, key, data):
     """
+    Warning: Deprecated, use the corresponding vocabulary to serialize the data
```
albertz, line 889: Remove that.

albertz:
> The reason I didn't do it directly this way is that the vocabulary currently always requires a vocab_file (except the Utf8ByteTargets). I would suggest then to change the logic in the vocabulary constructor to allow directly passing the vocab /labels.

Yes sure.

albertz:
> > Note that you changed the behavior of get_seq_labels for some of the derived Vocabulary classes. I'm not sure if there are other cases in the code or in scripts where this would be unexpected and could break things.
> 
> Yes, that could be an issue. But I would argue that the old behaviour of get_seq_labels for the character targets was wrong, so not sure whether anyone relied on that.
> 
> > There is also the function get_seq, which is defined in derived classes already. Which kind of matches the behavior of the additional get_seq_labels functions (just being the inverse of it). I think it should match exactly for consistency. Would that be the case?
> 
> Yes, now both functions match exactly.

Ok, so I would say, do that then.

returnn/datasets/util/vocabulary.py
albertz: I'm not sure this is confusing. The function name mentions `seq`.

I'm also not sure this would complicate the logic here and for all derived functions where you always repeat the same code.

When you just do this when you call the function, I think it is easier.

But maybe also for the case that it is not a sequence (the layer output does not have a time dim), you would not call this function at all, but just do a direct lookup in `labels`?

Similarly, when the layer output has other unrelated dims (maybe a second spatial dim), you would also not directly call this function. I'm not really sure what you would do. It's ok to just throw an exception then.

returnn/datasets/util/vocabulary.py
```diff
@@ -123,6 +130,15 @@ def create_vocab_dict_from_labels(cls, labels):
     assert len(d) == len(labels), "some labels are provided multiple times"
     return d
 
+  @classmethod
+  def create_vocab_from_labels(cls, labels):
+    if len(labels) < 1000 and all([len(label) == 1 for label in labels]):
+      # are these actually ordered raw bytes? -> assume utf8
+      if all([ord(label) <= 255 and ord(label) == idx for idx, label in enumerate(labels)]):
+        return Utf8ByteTargets()
```
dthulke, line 138: One potential issue is that this does not support datasets anymore that use utf8 bytes that are not ordered. We could fix that by changing Utf8ByteTargets to use explicit labels/vocab.

But I don't think that anybody is using something like that, so it may be fine as it currently is.

albertz: But why would it not be ordered, or why would you use `Utf8ByteTargets` then? `Utf8ByteTargets` is intended to represents the bytes, so it is by definition exactly the byte.

If we ever would have or need such a thing, I think we should just create a separate vocab class.

And the user would just not use this fallback code for that but instead properly define the vocab.

So I don't think this is a problem.

returnn/datasets/util/vocabulary.py
```diff
@@ -315,13 +338,14 @@ class CharacterTargets(Vocabulary):
   Also see :class:`Utf8ByteTargets`.
   """
 
-  def __init__(self, vocab_file, seq_postfix=None, unknown_label="@"):
+  def __init__(self, vocab_file, seq_postfix=None, unknown_label="@", labels=None):
     """
-    :param str vocab_file:
+    :param str|None vocab_file:
     :param list[int]|None seq_postfix: labels will be added to the seq in self.get_seq
     :param str|None unknown_label:
```
albertz, line 345: `labels` should be documented.

returnn/datasets/util/vocabulary.py
```diff
@@ -46,18 +46,20 @@ def create_vocab(cls, **opts):
...
     return clz(**opts)
 
-  def __init__(self, vocab_file, seq_postfix=None, unknown_label="UNK", num_labels=None):
+  def __init__(self, vocab_file, seq_postfix=None, unknown_label="UNK", num_labels=None, labels=None):
     """
     :param str|None vocab_file:
     :param str|None unknown_label:
     :param int num_labels: just for verification
     :param list[int]|None seq_postfix: labels will be added to the seq in self.get_seq
+    :param list[str] labels:
```
albertz, line 55: The type is wrong.
```suggestion
    :param list[str]|None labels:
```

returnn/datasets/util/vocabulary.py
```diff
@@ -46,18 +46,20 @@ def create_vocab(cls, **opts):
...
     """
     self.vocab_file = vocab_file
     self.unknown_label = unknown_label
     self.num_labels = None  # type: typing.Optional[int]  # will be set by _parse_vocab
     self.vocab = None  # type: typing.Optional[typing.Dict[str,int]]  # label->idx
-    self.labels = None  # type: typing.Optional[typing.List[str]]
+    self.labels = labels  # type: typing.Optional[typing.List[str]]
```
albertz, line 61: You don't really need to define the type if it is already defined.

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,80 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+        if output_layers[output_layer_idx].output.sparse:
+          # Try to get the vocab corresponding to the sparse dim
+          vocab = None
+          if output_layers[output_layer_idx].output.sparse_dim.vocab is not None:
+            vocab = output_layers[output_layer_idx].output.sparse_dim.vocab
+          elif target_keys[output_layer_idx] and (
+            self.network.extern_data.get_data(target_keys[output_layer_idx]).vocab is not None):
+            vocab = self.network.get_extern_data(target_keys[output_layer_idx]).vocab
+          elif target_keys[output_layer_idx] and target_keys[output_layer_idx] in dataset.labels:
+            vocab = Vocabulary.create_vocab_from_labels(dataset.labels[target_keys[output_layer_idx]])
```
albertz, line 2361: I think you can simplify the code.
```suggestion
          vocab = output_layers[output_layer_idx].output.sparse_dim.vocab
          if not vocab and target_keys[output_layer_idx]:
            vocab = self.network.get_extern_data(target_keys[output_layer_idx]).vocab
            if not vocab:
              vocab = Vocabulary.create_vocab_from_labels(dataset.labels[target_keys[output_layer_idx]])
```

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,80 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+            assert not output_file, "Unable to serialize sparse output of layer '%s'." % (
+              output_layer_names[output_layer_idx])
+        else:
+          # Output dense layers as-is
+          serialized_output = outputs[output_layer_idx]
+        serialized_outputs.append(serialized_output)
+
+        if target_keys[output_layer_idx] and do_eval:
+          target_extern_data = self.network.get_extern_data(target_keys[output_layer_idx])
+          if not target_extern_data.sparse:
```
albertz, line 2378: Wrong way around, right?
```suggestion
          if target_extern_data.sparse:
```

returnn/tf/engine.py
```diff
@@ -2316,27 +2320,80 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+            vocab = None
+            if target_extern_data.vocab is not None:
+              vocab = target_extern_data.vocab
+            elif target_keys[output_layer_idx] in dataset.labels:
+              vocab = Vocabulary.create_vocab_from_labels(dataset.labels[target_keys[output_layer_idx]])
```
albertz, line 2383: Simplify:
```suggestion
            vocab = target_extern_data.vocab
            if not vocab and target_keys[output_layer_idx] in dataset.labels:
              vocab = Vocabulary.create_vocab_from_labels(dataset.labels[target_keys[output_layer_idx]])
```

albertz (approved):
Everything looks good now to me.

albertz:
Maybe someone else should also take a look. @patrick-wilken or @JackTemaki ?

returnn/tf/engine.py
```diff
@@ -2348,64 +2400,64 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
+              seq_idx[batch_idx], seq_tag[batch_idx],
+              output_layer_names[output_layer_idx], outputs[output_layer_idx][batch_idx]), file=log.v4)
             out_idx = batch_idx
           else:
             print("seq_idx: %i, seq_tag: %r, outputs %r: %r" % (
-              seq_idx[batch_idx], seq_tag[batch_idx], output_layer_names[target_idx],
-              outputs[target_idx][batch_idx * out_beam_sizes[target_idx]:(batch_idx + 1)*out_beam_sizes[target_idx]]),
+              seq_idx[batch_idx], seq_tag[batch_idx], output_layer_names[output_layer_idx],
+              outputs[output_layer_idx]
+              [batch_idx * out_beam_sizes[output_layer_idx]:(batch_idx + 1)*out_beam_sizes[output_layer_idx]]),
```
patrick-wilken, line 2413: This is hard to read. Maybe calculate the indices in separate lines. Also, spacing around `*` is inconsistent (which is was before).

returnn/tf/engine.py
```diff
@@ -2348,64 +2400,64 @@ def extra_fetches_callback(seq_idx, seq_tag, **kwargs):
...
                   file=log.v4)
-            out_idx = batch_idx * out_beam_sizes[target_idx]
-          if target_keys[target_idx] and dataset.can_serialize_data(target_keys[target_idx]):
-            if do_eval:
-              print("  ref:", dataset.serialize_data(key=target_keys[target_idx], data=targets[target_idx][batch_idx]),
-                    file=log.v4)
-            if out_beam_sizes[target_idx] is None:
-              print("  hyp:", dataset.serialize_data(key=target_keys[target_idx], data=outputs[target_idx][out_idx]),
+            out_idx = batch_idx * out_beam_sizes[output_layer_idx]
+          if target_keys[output_layer_idx] and serialized_outputs[output_layer_idx]:
```
patrick-wilken, line 2416: Why only print if there is a target?

patrick-wilken:
Looks nice overall!

returnn/tf/engine.py
dthulke: Mainly because that was the behaviour before.

But I don't see any reason to not always print the hyp.

albertz:
Thanks! Merged now.

PR #830, opened 2021-12-06
Fixes the shape of output_flat in loss with a static time dimension

dthulke:
Currently, the shape of output_flat is set to the original data shape (that does not include the batch dim), i.e. the length of the new batch x time dim is set to the length of the time dim (even if it is static).

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -2434,7 +2434,7 @@ class Loss(object):
             output_with_activation.x, self.output_seq_lens, time_major=output.is_time_major)
         else:
           self.output_flat = flatten_or_merge(output.placeholder, self.output_seq_lens, time_major=output.is_time_major)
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/base.py
```diff
@@ -2434,7 +2434,7 @@ def init(self, output, output_with_activation=None, target=None, layer=None):
             output_with_activation.x, self.output_seq_lens, time_major=output.is_time_major)
         else:
           self.output_flat = flatten_or_merge(output.placeholder, self.output_seq_lens, time_major=output.is_time_major)
-          self.output_flat.set_shape(tf.TensorShape(output.shape))
+          self.output_flat.set_shape(tf.TensorShape((None,) + output.shape[1:]))
```
albertz, line 2437: It might be wrong when time dim or batch dim >= 2. But this is anyway not currently supported here...

PR #845, opened 2021-12-07
SplitDimsLayer, fixes for dim tags

Zettelkasten:


albertz (approved):

Zettelkasten:
@albertz Wait, this isn't done yet :D

Zettelkasten:
Okay, the test case should be fixed now.

I changed it now so that `SplitDimsLayer` will complete the "-1" dim tag ("`rem_dim`") if you supply it to add `dyn_size_ext` and `dimension` if it missed those attributes before.
I am not sure if this is so nice, because now passing a dim tag has a side effect, I am not sure if any other layer does this right now.
But it is also not really an option that the user needs to set `dyn_size_ext` of the -1` dim tag themselves.

returnn/tf/layers/basic.py
```diff
@@ -3414,35 +3406,47 @@ def get_out_data_from_opts(cls, name, axis, dims, pad_to_multiples=None, sources
...
-    else:
-      rem_dim = Dim(
-        kind=axis_dim_tag.kind,
-        description="%s_split_dims%i_rem" % (name, rem_dim_idx),
-        dimension=resolved_shape_dims[rem_dim_idx])
     if not resolved_dims:
       resolved_dims = tuple(
         Dim(
-          kind=Dim.Types.Spatial,
+          kind=Dim.Types.Spatial if shape_dim is None else Dim.Types.Feature,
```
albertz, line 3420: The logic for `kind` is a bit incorrect or weird. I think there should not be multiple feature dims.

Why did you remove the `rem_dim` logic? I think you separately need to create a `rem_dim` which copies `axis_dim_tag.kind`, and here use it for `rem_dim_idx`.

tests/test_TFNetworkLayer.py
```diff
@@ -1867,7 +1865,8 @@ def test_SplitDimsLayer_batch_feature_major_keep_feature():
       "output": {'class': 'split_dims', 'from': 'data', 'axis': 'T', 'dims': [-1, 1]}  # [B,D,T,1]
     })
     out = net.get_default_output_layer().output
-    assert_equal(out.get_dim_tag(2), net.extern_data.get_default_input_data().get_time_dim_tag())
+    assert out.get_dim_tag(2).is_equal(
+      net.extern_data.get_default_input_data().get_time_dim_tag(), derived_matches=True)
```
albertz, line 1869: `derived_matches=True` should not be needed here. The splitting with `[-1, 1]` should yield the same dim tag (under standard equality).

tests/test_TFNetworkLayer.py
```diff
@@ -3896,7 +3912,7 @@ def test_TransposedConvLayer_2d_simple():
     })
     out = net.get_default_output_layer().output.copy_as_batch_feature_major()
     assert out.batch_shape == (None, 13, None, 2)
-    assert_equal(out.get_dim_tag(2), net.extern_data.get_default_input_data().get_time_dim_tag())
+    assert out.dim_tags[2].is_equal(net.extern_data.get_default_input_data().get_time_dim_tag(), derived_matches=True)
```
albertz, line 3915: `derived_matches=True` should also not be needed here. You have filter size 1 and strides 1 in the T axis, and the split dims before also should leave it like it is.

tests/test_TFNetworkLayer.py
```diff
@@ -1824,9 +1823,8 @@ def test_SplitDimsLayer_simple_time2():
     net = TFNetwork(config=config)
     net.construct_from_dict({
       "output": {"class": "split_dims", "axis": "t", "dims": (1, -1), "from": "data:data"}})
-    assert_equal(
-      net.get_default_output_layer().output.get_dim_tag(2),
-      net.extern_data.get_default_input_data().get_dim_tag(1))
+    assert net.get_default_output_layer().output.get_dim_tag(2).is_equal(
+      net.extern_data.get_default_input_data().get_dim_tag(1), derived_matches=True)
```
albertz, line 1827: `derived_matches=True` is wrong here.

tests/test_TFNetworkLayer.py
```diff
@@ -1802,9 +1802,8 @@ def test_SplitDimsLayer_simple_time():
     net = TFNetwork(config=config)
     net.construct_from_dict({
       "output": {"class": "split_dims", "axis": "t", "dims": (-1, 1), "from": "data:data"}})
-    assert_equal(
-      net.get_default_output_layer().output.get_dim_tag(1),
-      net.extern_data.get_default_input_data().get_dim_tag(1))
+    assert net.get_default_output_layer().output.get_dim_tag(1).is_equal(
+      net.extern_data.get_default_input_data().get_dim_tag(1), derived_matches=True)
```
albertz, line 1806: `derived_matches=True` is wrong here.

tests/test_TFNetworkLayer.py
Zettelkasten: Hm okay. Then I'll have to change some things around in the logic of `SplitDimsLayer`.

returnn/tf/layers/basic.py
```diff
@@ -3414,35 +3406,47 @@ def get_out_data_from_opts(cls, name, axis, dims, pad_to_multiples=None, sources
...
+      if rem_dim.dimension is None:  # need to set dyn_size
+        if rem_dim.dyn_size_ext is not None:
+          pass  # assume it is set up correctly
+        if axis_dim_tag.dimension is not None:  # out dim stays static if in dim was static
+          assert axis_dim_tag.dimension % n == 0
+          rem_dim.dimension = axis_dim_tag.dimension // n
+        elif n == 1:  # can overtake old dyn size
+          assert axis_dim_tag.dyn_size_ext is not None
+          rem_dim.dyn_size_ext = axis_dim_tag.dyn_size_ext
+          rem_dim.derived_from_tag = axis_dim_tag
```
albertz, line 3443: `derived_from_tag` is used incorrectly. It means that it is derived from some other tag, but different to the other tag.

But here we have the case that it is the same. So you could use `declare_same_as`. Or why not directly?
```
rem_dim = axis_dim_tag
```

As it was also in the old code.

Zettelkasten: This is not needed if the user doesn't specify dims as a tag list. I will change it so that then it actually just re-uses the dim tag of the input.
However, if you e.g. use `dims = [SpatialDim("rem"), FeatureDim("broadcast", 1)]` or so, like I have it in my test, then I need to properly assign `dyn_size_ext` to the new `rem` dim tag.

Zettelkasten: In that case, should the user-specified "rem" dim tag set derived_from to the input axis?

albertz: No, if there is an existing object already, you should use `declare_same_as`. I think we have some similar logic in other layers.

albertz: For this case where it is the same. If it is not the same (because some other dims > 1), then you can set `derived_from`.

returnn/tf/layers/basic.py
```diff
@@ -3414,35 +3406,48 @@ def get_out_data_from_opts(cls, name, axis, dims, pad_to_multiples=None, sources
...
-      rem_dim = axis_dim_tag  # we can overtake the existing dim tag
+    if len(rem_dim_indices) == 1:
+      import numpy
+      n = int(numpy.prod([dim for dim in dims if dim > 0]))
+      rem_dim_idx = rem_dim_indices[0]
+      if n == 1:
+        rem_dim = axis_dim_tag
+      else:  # need to create a new rem_dim
+        rem_dim = Dim(
+          kind=Dim.Types.Spatial,
```
albertz, line 3428: Not this?
```suggestion
          kind=axis_dim_tag.kind,
```

albertz:
Some failing test.

returnn/tf/layers/basic.py
Zettelkasten: Ah yeah, I missed that .. this is also the reason for the failing test.

returnn/tf/util/data.py
```diff
@@ -234,7 +234,7 @@ def get_for_batch_ctx(self, batch, ctx, allow_none=False):
       if same_base.dyn_size_ext:
         assert batch == batch_base
         same_base.batch = batch
-        assert not same_base.dyn_size_ext.batch
+        assert not same_base.dyn_size_ext.batch or same_base.dyn_size_ext.batch == batch
```
Zettelkasten, line 237: Is this okay?
Otherwise, `rem_dim.declare_same_as(resolved_dims[rem_dim_idx])` failed for me when `rem_dim` was just copied from the input (in the `[-1,1]` split).

albertz: Yea I guess ok.
The whole code is a bit messy...

albertz (approved):

PR #862, opened 2021-12-16
Dim __copy__ and __deepcopy__ 

### LINKED ISSUE #860: Wanted behavior for `deepcopy(dim_tag)`?
Do we want:
```
tag == deepcopy(tag)
```
or not?

In returnn_common (and similarly in PyTorch), it is common do do sth like this:
```
class MultiLayer(Module):
  def __init__(self, out_dim: nn.Dim, num_layers: int, layer: Optional[Module] = None):
    if not layer:
      layer = nn.Linear(out_dim)
    self.layers = nn.Sequential(deepcopy(layer) for _ in range(num_layers)
```

That `deepcopy` on such layer has a reference to the `out_dim`.

The current behavior has `tag != deepcopy(tag)`, although I think this is actually buggy for other reasons.

Related is also how we actually define the equality of dim tags (#634). Although, maybe not so much, as we can keep the `deepcopy` logic as we want it.

I'm somewhat undecided about what behavior we want, or we expect. I tend to prefer `tag == deepcopy(tag)`, also because this simplifies the code in returnn_common.

albertz:
Fix #860.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -178,6 +178,8 @@ class Dim(object):
     :return: copy, maybe as new kind. setting same_as to self
     :rtype: Dim
     """
+    if not same_as_self:
+      assert description is not None, "%s copy with not same_as_self should have a new description" % self
     tag = Dim(
       kind=kind or self.kind, description=description or self.description,
       dimension=self.dimension, dyn_size_ext=self.dyn_size_ext,
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -145,19 +145,47 @@ class Dim(object):
         desc += "{ctx=%s}" % self.control_flow_ctx.repr_inner()
     return desc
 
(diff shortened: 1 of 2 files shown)


returnn/tf/util/data.py
```diff
@@ -145,19 +145,47 @@ def short_repr(self):
...
+      kind=kind or self.kind, description=description or self.description,
       dimension=self.dimension, dyn_size_ext=self.dyn_size_ext,
       batch=self.batch,
       src_data=self.src_data, src_axis=self.src_axis)
-    tag.same_as = self  # not declare_same_as, none of the extra checks needed
-    tag._same_as_tb = traceback.extract_stack()
+    if same_as_self:
+      tag.same_as = self  # not declare_same_as, none of the extra checks needed
+      tag._same_as_tb = traceback.extract_stack()
     return tag
```
Zettelkasten, line 161: I'd add `else: assert description is not None` here.

PR #863, opened 2021-12-16
GatherLayer, cleanup matching dims logic

Zettelkasten:
Some cleanups for GatherLayer.


albertz (approved):

albertz:
This is now very similar as I did it in `DotLayer`, right?

Does this change anything in behavior? Or is this just cleanup? If this also fixes some cases, some test case would be nice.

But anyway, merged now. Not so important.

Zettelkasten:
It also fixes some cases that were super weird before where one axes would match multiple ones.
I'll make a test and directly push it to master later.

PR #864, opened 2021-12-16
Fix #666, broadcast no longer matches dims

### LINKED ISSUE #666: `CombineLayer` and others (`Data.get_common_data`) should use `broadcast_matches=False`
`CombineLayer` and others rely on `Data.get_common_data`.
 
`Data.get_common_data` uses `broadcast_matches=True` for the `DimensionTag.is_equal` opts. Such that you can match a tensor [B,1,T] with [B,4,T].

However, this relies on heuristics, and is not really needed. There is never a reason in RETURNN to create such broadcast dim [1] in the first place. So instead of [B,1,T] you would just have a tensor [B,T]. The broadcasting in RETURNN would automatically happen on non-existing dims. (`Data.copy_compatible_to` usually does that internally.) This would be much more consistent and reliable.

This needs a new behavior version (#508) because older configs rely on this.

Zettelkasten:
Needs #863 first (which should be ready now).

Fix #666.

Introduces a new behavior version (#508) because it would break some older configs.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3774,7 +3774,10 @@ class ExpandDimsLayer(_ConcatInputLayer):
       data = data.copy_as_batch_major()
     axis = cls._get_axis(data=data, axis=axis)
 
-    new_dim = SpatialDim("%s_expand_dims" % name, dim)
+    new_dim = Dim(
+      kind=Dim.Types.Feature if init_axis.lower() == "f" else Dim.Types.Spatial,
+      description="%s_expand_dims" % name,
+      dimension=dim)
     data = data.copy_template(name="%s_output" % name)
     data = data.copy_add_dim_by_tag(new_dim, unbroadcast=True, axis=axis)
     if isinstance(init_axis, str):
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1282,7 +1282,10 @@ class GatherLayer(_ConcatInputLayer):
     :return: (common_axes_input, common_axes_position, specific_input_axes, specific_position_axes), all counted with
     batch dim.
     """
-    is_equal_opts = dict(allow_same_spatial_dim=True, broadcast_matches=True)
+    from returnn.util import BehaviorVersion
+    is_equal_opts = dict(allow_same_spatial_dim=True)
+    if BehaviorVersion.get() < 11:
+      is_equal_opts["broadcast_matches"] = True
     all_dim_tags, tags_dict = Dim.get_all_dimension_tags([input_data, position_data], is_equal_opts=is_equal_opts)
     input_tags, pos_tags = tags_dict[input_data], tags_dict[position_data]
     specific_input_axes = [i for i, tag in enumerate(input_tags) if tag not in pos_tags and i != old_gather_axis]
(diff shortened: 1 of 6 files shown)


albertz:
Why do you say it needs #863 first but all tests are anyway passing without #863?

Anyway, I merged #863 now. So you should rebase here and resolve the conflict.

Zettelkasten:
Ok, try now.
Sorry, I wasn't so clear about this, but the branch for this PR branched off #683, that is why the tests here also worked (and okay, we could've also closed #683 and just merged this directly).

tests/test_TFNetworkLayer.py
```diff
@@ -3376,27 +3377,22 @@ def test_GatherLayer_constant_position():
           np.testing.assert_almost_equal(out_seqs[b, f1, f2], values_seqs[b, f1, position, f2])
 
 
-def test_GatherLayer_search_beam():
```
albertz, line 3379: Why did you remove this test?

returnn/tf/util/data.py
```diff
@@ -5039,16 +5039,18 @@ def get_batch_shape_dim_tags(self):
     return self.dim_tags
 
   @classmethod
-  def get_common_data(cls, sources, ignore_feature_dim=False, allow_broadcast_all_sources=NotSpecified):
+  def get_common_data(cls, sources, ignore_feature_dim=False, allow_broadcast_all_sources=NotSpecified, name=None):
```
albertz, line 5042: Is there any specific reason this was added, or just to make it nicer?

tests/test_TFNetworkLayer.py
```diff
@@ -920,7 +920,8 @@ def test_CombineLayer_broadcast():
     net_dict = {
       "lin1": {"class": "linear", "activation": "sigmoid", "n_out": 5, "from": "data:data"},
       "lin2": {"class": "linear", "activation": "sigmoid", "n_out": 1, "from": "data:data"},
-      "combine": {"class": "combine", "kind": "add", "from": ["lin1", "lin2"]},
+      "lin2_squeeze": {"class": "squeeze", "from": "lin2", "axis": "f"},
```
albertz, line 923: I wonder whether we should somehow simplify this, and unify into a single layer. But not sure. Maybe not.

returnn/tf/layers/basic.py
```diff
@@ -3771,7 +3774,7 @@ def get_out_data_from_opts(cls, name, axis, dim=1, sources=(), **kwargs):
       data = data.copy_as_batch_major()
     axis = cls._get_axis(data=data, axis=axis)
 
-    new_dim = SpatialDim("%s_expand_dims" % name, dim)
+    new_dim = (FeatureDim if init_axis.lower() == "f" else SpatialDim)("%s_expand_dims" % name, dim)
```
albertz, line 3777: I personally would have preferred to use `Dim(kind=... if ... else ..., description=..., dimension=dim)`. But maybe that's just personal preference.

albertz:
Code style errors:
```
returnn/tf/layers/basic.py:5670: WARNING PyArgumentListInspection: Parameter 'name' unfilled
returnn/tf/layers/basic.py:6956: WEAK WARNING Pep8CodeStyle: E131: continuation line unaligned for hanging indent
returnn/tf/layers/basic.py:7207: WEAK WARNING Pep8CodeStyle: E131: continuation line unaligned for hanging indent
```

albertz:
Isn't there some overlap with #866?

returnn/tf/util/data.py
Zettelkasten: Ah yeah, my rebase messed this up. This is #866. I'll remove it from here again.

returnn/tf/layers/basic.py
Zettelkasten: I agree it's a bit nicer what you have, changed it.

tests/test_TFNetworkLayer.py
Zettelkasten: True. It is weird to think of `ExpandDimsLayer` as deprecated, but at the same time have valid use cases for `SqueezeLayer`.

Maybe via a special `out_dim` or so. But not too important.

albertz (approved):

PR #866, opened 2021-12-16
get_common_data, add name parameter

Zettelkasten:
Before, the output name was just the first input's name.
This was very confusing when get_common_data errored because of implicit broadcasting:
Then you would think the error was in the first input,
while it was in the layer that combined the dims instead.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1051,7 +1051,7 @@ class SliceNdLayer(_ConcatInputLayer):
     data_objs = [start_data]
     data_objs += [size.output] if isinstance(size, LayerBase) else []
     data_objs += [seq_lens_data] if isinstance(seq_lens_data, Data) else []
-    common_data = Data.get_common_data(data_objs)
+    common_data = Data.get_common_data(data_objs, name="%s_inputs")
     start_data = start_data.copy_compatible_to(common_data, check_sparse=False)
     start_t = start_data.placeholder
     if size is None:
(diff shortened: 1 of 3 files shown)


albertz (approved):

albertz:
Failing test:
```
returnn/tf/layers/basic.py:6951: WEAK WARNING Pep8CodeStyle: E131: continuation line unaligned for hanging indent
returnn/tf/layers/basic.py:7203: WEAK WARNING Pep8CodeStyle: E131: continuation line unaligned for hanging indent
```

Zettelkasten:
Sorry, I don't get it. What's wrong here?
```
    if n_out is not NotSpecified:
      out_type_["dim"] = n_out
    elif out_type_.get("sparse", False):  # E131: continuation line unaligned for hanging indent
      out_type_["dim"] = 2
    out_type_["dtype"] = "bool"
```

returnn/tf/layers/basic.py
```diff
@@ -6942,10 +6943,11 @@ def get_out_data_from_opts(cls, eval_locals=None, n_out=NotSpecified, sources=()
         allow_broadcast_all_sources = True
       out_type_.update(
         Data.get_common_data(
-          [s.output for s in sources], allow_broadcast_all_sources=allow_broadcast_all_sources).get_kwargs())
+          [s.output for s in sources],
+          allow_broadcast_all_sources=allow_broadcast_all_sources).get_kwargs(),
+          name="%s_output" % kwargs["name"])
```
albertz, line 6948: This indent is wrong, and also `name` should not be an argument to `update`, but to `get_common_data` instead, right?

returnn/tf/layers/basic.py
```diff
@@ -7193,14 +7195,15 @@ def get_out_data_from_opts(cls, n_out=NotSpecified, out_type=None, out_shape=Non
...
```
albertz, line 7200: Same here, this indent is wrong, and the `name` argument also wrong.

returnn/tf/layers/basic.py
Zettelkasten: Urgh yeah I was so confused ... thanks!

albertz (approved):

PR #871, opened 2021-12-19
Solution for ambiguous dim tags

albertz:
Consider the example (also in the test) to use `DotLayer` using a square matrix, so its two dimensions are the same.

```
from returnn.tf.util.data import batch_dim
time_dim = SpatialDim("time")
feat_dim = FeatureDim("feature", dimension=3)
config = Config({
  "extern_data": {
    "data": {"dim_tags": [batch_dim, time_dim, feat_dim]},
    "matrix_ambiguous": {"dim_tags": [feat_dim, feat_dim], "available_for_inference": True},
  },
})
```

Here, `{"class": "dot", "from": ["data:data", "data:matrix_ambiguous"], "reduce": feat_dim}` does not work because it is ambiguous.

Introducing the `Dim.match_priority` solves this problem, by having:
```
"matrix_non_ambiguous": {
  "dim_tags": [feat_dim.copy(match_priority=1), feat_dim], "available_for_inference": True},
```

This was suggested here: https://github.com/rwth-i6/returnn_common/issues/17#issuecomment-997463222

While maybe not the nicest solution, I don't really see any better solution at the moment.

This PR is also a test if anything else breaks.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -57,6 +57,7 @@ class Dim(object):
                vocab=None,
                dyn_size=None, dyn_size_ext=None,
                undefined=False, generic=False, special=False,
+               match_priority=0,
                derived_from_tag=None, derived_from_op=None,
                batch=None, control_flow_ctx=None,
                src_data=None, src_axis=None):
@@ -79,6 +80,8 @@ class Dim(object):
       the behavior is to consider them as equal,
       and assume that the chain of operations (e.g. padding + valid conv) results in the same dim.
     :param Dim.Op|None derived_from_op:
+    :param int match_priority: when there is ambiguity between multiple dim tags, this can be used to resolve it.
+      E.g. for a square matrix used for a linear transformation, the reduce dim tag should have a higher priority.
     :param BatchInfo|None batch: for batch-dim, or dynamic dims per batch
     :param ControlFlowContext|None control_flow_ctx:
     :param Data|None src_data:
@@ -95,6 +98,7 @@ class Dim(object):
     self.derived_from_op = derived_from_op
     if derived_from_op and not derived_from_op.output:
       derived_from_op.output = self
+    self.match_priority = match_priority
     if src_data:
       assert isinstance(src_data, Data) and isinstance(src_axis, int)
     if not batch and dyn_size_ext:
@@ -170,11 +174,12 @@ class Dim(object):
     """
     return self
 
-  def copy(self, same_as_self, description=None, kind=None):
+  def copy(self, same_as_self=True, description=None, kind=None, match_priority=None):
     """
     :param bool same_as_self:
     :param str|None description: new description
     :param Entity|None kind: if set, overwrites self.kind
+    :param int|None match_priority:
     :return: copy, maybe as new kind. setting same_as to self
     :rtype: Dim
     """
@@ -182,6 +187,7 @@ class Dim(object):
       assert description is not None, "%s copy with not same_as_self should have a new description" % self
     tag = Dim(
       kind=kind or self.kind, description=description or self.description,
+      match_priority=match_priority if match_priority is not None else self.match_priority,
       dimension=self.dimension, dyn_size_ext=self.dyn_size_ext,
       batch=self.batch,
       src_data=self.src_data, src_axis=self.src_axis)
@@ -4366,6 +4372,9 @@ class Data(object):
       # Once we have not guaranteed unique dim tags, multiple axes could match.
       # https://github.com/rwth-i6/returnn/issues/632
       dims = [i for (i, tag) in enumerate(self.dim_tags) if tag == axes]
+      if len(dims) > 1:
+        max_match_priority = max(self.dim_tags[i].match_priority for i in dims)
+        dims = [i for i in dims if self.dim_tags[i].match_priority == max_match_priority]
       assert len(dims) <= 1, "%s: matching dim %s must be unique" % (self, axes)
       return dims
     if isinstance(axes, int):
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -80,7 +80,9 @@ class Dim(object):
       the behavior is to consider them as equal,
       and assume that the chain of operations (e.g. padding + valid conv) results in the same dim.
     :param Dim.Op|None derived_from_op:
-    :param int match_priority: when there is ambiguity between multiple dim tags, this can be used to resolve it.
+    :param int match_priority: when there is ambiguity between multiple dim tags, this value defines the order
+      in which the dimension are assigned to their matching counterparts.
+      A dimension tag with a higher priority value is assigned first.
       E.g. for a square matrix used for a linear transformation, the reduce dim tag should have a higher priority.
     :param BatchInfo|None batch: for batch-dim, or dynamic dims per batch
     :param ControlFlowContext|None control_flow_ctx:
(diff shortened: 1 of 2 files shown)


Zettelkasten:
Technically this PR looks fine, but I'm not sure if this is really a clean solution in general.
In particular it is not so clean that you have to specify a matching priority when you define your weight variable, which is not really specific to the variable itself but rather specific to the operation you use it as argument to afterwards (in your case, dot layer).
For example, in our MT configs we often use the transposed target embeddings as projection matrix. In this case I need my dot operation to reduce the axis exactly on the other axis then before, i.e. the one with the lower matching priority.
Sure, there are solutions to this (e.g. to reinterpret the variable output to change the matching priorities), but I personally think that this is an indicator that this concept in general is not that complete and perhaps more confusing/error prone.

I think the code you posted for your Linear module implementation is not too bad. I like about it that the variable has unique dim tags, I think that makes it pretty clear what is going on. I think I would prefer it to the somewhat less meaningful matching prios. Note that it would probably also be nice if it exposed the inner out dim tag (so that some other layer could e.g. transpose the variable).

I do agree though that it is not that nice and that it would be very annoying if RETURNN configs would often need to do such extra handling. I don't really know a better solution now either. Maybe just adding utility functions to make the code you already had with the inner dim tag less verbose would help. Maybe you have new thoughts since you opened this, too?

How often do you think does one need these extra checks? Only for layers introducing variables like Linear, probably also for a few others?
Would this also be an issue for RETURNN code that the user usually writes? Or could the user just call what you have in your Linear module (or some related, some what elementary building block)?

albertz:
> In particular it is not so clean that you have to specify a matching priority when you define your weight variable, which is not really specific to the variable itself but rather specific to the operation you use it as argument to afterwards (in your case, dot layer).

It would work for all operations which do some reduce, e.g. also `gather`, and that will always be the main purpose of this variable in this case.

> For example, in our MT configs we often use the transposed target embeddings as projection matrix. In this case I need my dot operation to reduce the axis exactly on the other axis then before, i.e. the one with the lower matching priority.

There is no problem on this with the matching priority because that will be two different dims anyway - the target classes dim and some intermediate hidden dim, which would be almost always something different.

Also, when dim orders do not matter, there is nothing like a transpose.

> I think the code you posted for your Linear module implementation is not too bad. I like about it that the variable has unique dim tags, I think that makes it pretty clear what is going on. I think I would prefer it to the somewhat less meaningful matching prios. Note that it would probably also be nice if it exposed the inner out dim tag (so that some other layer could e.g. transpose the variable).

The code has several issues:

- It is too complicated.
- It is not straightforward. I only stumbled upon it because I got it wrong first. This is a bad sign. Ideally it should be designed such that the obvious and straightforward solution works. (Although you could argue, the matching priority is not straightforward either.)
- Now using it for the transposed linear operation as you mentioned before would actually not work anymore, or only with further complicated code.
- I was also not sure, should I create this dummy dim always or only when `in_dim == out_dim`? In both cases, it's not so nice. Creating it always is maybe easier to reason about because there will never be two cases, only one case, which always makes it simpler and better.
- Some other code which creates the variable or any tensor which would be used inplace of the variable would also need to have this extra logic. When assigning `linear.weights = ...`, it would additionally also need to assign `linear.out_dim_inner = ...`.

> Maybe you have new thoughts since you opened this, too?

Not really.

> How often do you think does one need these extra checks? Only for layers introducing variables like Linear, probably also for a few others? Would this also be an issue for RETURNN code that the user usually writes? Or could the user just call what you have in your Linear module (or some related, some what elementary building block)?

Depends maybe on your research focus. But when you are writing new models or model parts (deriving new classes from `nn.Module`), you need to create variables all the time. Well, in many cases, you might reuse existing modules to define the new model, but often enough it could be also more custom. It's hard to tell.

Now when you want to do e.g. param sharing between two different modules (e.g. `Linear` and `LSTM`), this is easy and straightforward to do in principle (just `linear.weights = lstm.weights`), but it would be complicated if you also need to deal with sth like `out_dims_inner`.

I think this case that we have `in_dim == out_dim` is not so rare actually.

The potential cases where such matching priorities could go wrong are probably very rare. Although, when this happens, this will be also very annoying to detect even as a bug (because no error will occur, it will just do sth) and to debug.

albertz:
@Zettelkasten Have you further thought on this? I really don't like the solution with `out_dims_inner`. It should be possible to write generic code which would copy params around in whatever way. Sth like `out_dims_inner` would make this impossible because you never know if there is some `xy_inner_temp_dummy_dim` for some weight which also needs to be copied alongside.

While I totally see your concerns with the proposed solution here, I currently don't see a better solution. Any suggestions are very welcome.

Zettelkasten:
Hm no, not really, I thought about this some time, but didn't come up with something else really.

> Now when you want to do e.g. param sharing between two different modules (e.g. `Linear` and `LSTM`), this is easy and straightforward to do in principle (just `linear.weights = lstm.weights`), but it would be complicated if you also need to deal with sth like `out_dims_inner`.

For this concrete example with LSTM and Linear param sharing its not a problem btw, because LSTM has no square params.
But I agree this is a problem in general when you write your own modules and re-use parameters.

albertz:
This is still an open issue. We need some solution here.

albertz:
I just noticed that this does not cover what we are actually interested in. E.g. the file `better_exchook.py` is indented using 4 spaces (this is also configured in `.editorconfig` and PyCharm correctly uses those settings). However, this change now uses `--indent-size=2` for all files including `better_exchook.py`. But then you would expect that it should throw a lot of warnings for this file but it does not! It only prints those two warnings:
```
returnn/util/better_exchook.py:1338:13: E127 continuation line over-indented for visual indent
returnn/util/better_exchook.py:1339:13: E127 continuation line over-indented for visual indent
```
So this means, `pycodestyle` still does not cover what we are really interested in.

So, now the question is, is this change then really worth it? It anyway does not cover what we want, i.e. the check for correct indentation.

JackTemaki:
I do not see any better solution here, in the end it is always the question: what does the user actually want? So for the regular network construction it is fine to throw an error and let the user set the match_priority, and I guess for returnn_common this will be set automatically in many cases. But the docstring and the error message definitely need to be more accurate. I will add some suggestions.

returnn/tf/util/data.py
```diff
@@ -79,6 +80,8 @@ def __init__(self, kind=Types.Unspecified, description=None,
       the behavior is to consider them as equal,
       and assume that the chain of operations (e.g. padding + valid conv) results in the same dim.
     :param Dim.Op|None derived_from_op:
+    :param int match_priority: when there is ambiguity between multiple dim tags, this can be used to resolve it.
```
JackTemaki, line 83:
```suggestion
    :param int match_priority: when there is ambiguity between multiple dim tags, this value defines the order in which the dimension are assigned to their matching counterparts. A dimension tag with a higher priority value is assigned first.
```

returnn/tf/util/data.py
```diff
@@ -4366,6 +4372,9 @@ def get_axes_from_description(self, axes, allow_int=NotSpecified):
       # Once we have not guaranteed unique dim tags, multiple axes could match.
       # https://github.com/rwth-i6/returnn/issues/632
       dims = [i for (i, tag) in enumerate(self.dim_tags) if tag == axes]
+      if len(dims) > 1:
+        max_match_priority = max(self.dim_tags[i].match_priority for i in dims)
+        dims = [i for i in dims if self.dim_tags[i].match_priority == max_match_priority]
       assert len(dims) <= 1, "%s: matching dim %s must be unique" % (self, axes)
```
JackTemaki, line 4378:
```suggestion
      assert len(dims) <= 1, "%s: matching dim %s must be unique, use `match_priority` to resolve the matching order of ambiguous dimensions" % (self, axes)
```

albertz:
Ok, due to the lack of any better solution, I'm merging this now.

PR #875, opened 2021-12-21
Use -std=c++11 for gcc v4 and lower

patrick-wilken:
To keep support for TF <=1.14, which use gcc 4.8.

### DIFF
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3563,7 +3563,13 @@ class NativeCodeCompiler(object):
       f.write(self.code)
     common_opts = ["-shared", "-O2"]
     if self.is_cpp:
-      common_opts += ["-std=c++14"]
+      cpp_version_opt = "-std=c++14"
+      if BackendEngine.is_tensorflow_selected():
+        from returnn.tf.util.basic import get_tf_gcc_version
+        tf_gcc_version = get_tf_gcc_version()
+        if tf_gcc_version and int(tf_gcc_version[0]) <= 4:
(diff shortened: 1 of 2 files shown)


returnn/util/basic.py
```diff
@@ -3563,7 +3563,13 @@ def _maybe_compile_inner(self):
       f.write(self.code)
     common_opts = ["-shared", "-O2"]
     if self.is_cpp:
-      common_opts += ["-std=c++14"]
+      cpp_version_opt = "-std=c++14"
+      if BackendEngine.is_tensorflow_selected():
```
albertz, line 3567: This code here should be generic and not TF dependent. It should not matter here whether TF is selected, and what GCC was used for TF. This `NativeCodeCompiler` could be used for anything, not just for compiling TF ops.

Maybe we don't add `-std=...` here at all and only add that in `OpCodeCompiler` and `TFNativeUtilCompiler` (which are both TF specific, so there this logic is fine). For that, we might need to extend `common_opts`, e.g. like:
```
common_opts += self._extra_common_opts()
```
And `_extra_common_opts` can be derived and then return the `-std=...` flag.

Alternatively, you might also check `_get_compiler_bin()` here and try to parse the version from it. It is likely sth like `/usr/bin/gcc-4.8` or so.

albertz:
Unfortunately you created this PR in your own repo. Next time, please create it here in this repo such that we can easily work on this together.

I opened a new PR now with some updates: #896

So we can close this one here.

PR #887, opened 2022-01-03
ReinterpretDataLayer, complete incomplete dim tags

Zettelkasten:
I think in most cases if the user does something like this:
```
  other_time_dim = SpatialDim("other-time")
  ..
  net_dict = { # ..
    'reinterpret': {'from': 'input', 'set_dim_tags': {time_dim: other_time_dim}}
  .. }
```
they'd want `other_time_dim.dyn_size_ext == time_dim.dyn_size_ext`.
This PR adds this.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4333,6 +4333,9 @@ class ReinterpretDataLayer(_ConcatInputLayer):
     if set_dim_tags:
       for axis, tag in set_dim_tags.items():
         axis_int = out.get_axis_from_description(axis)
+        if tag.dimension is None and tag.dyn_size_ext is None:
+          # new, user-specified tag is incomplete, maybe copy its dynamic sizes.
+          tag.dyn_size_ext = out.dim_tags[axis_int].dyn_size_ext
         out = out.copy_template_replace_dim_tag(axis=axis_int, new_dim_tag=tag)
     if set_sparse is not None:
       assert isinstance(set_sparse, bool)
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -2016,6 +2016,23 @@ def test_out_shape():
       raise Exception("Expected an exception but did not get any.")
 
 
+def test_ReinterpretDataLayer_complete_incomplete_dim_tags():
+  from returnn.tf.util.data import batch_dim
+  with make_scope():
+    time_dim = SpatialDim("time")
+    other_time_dim = SpatialDim("other-time")
+    feat_dim = FeatureDim("feature", dimension=10)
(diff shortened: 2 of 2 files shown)


Zettelkasten:
Note that currently, `dim_tag.copy(..)` will not copy over the `dyn_size_ext` properly because usually it is `None` when this is called by the user.
`Dim.dyn_size_ext` is only completed via `extern_data` or by some layer later during the net dict construction.
This is somewhat non intuitive.

albertz:
> I think in most cases if the user does something like this:
> ...
> they'd want `other_time_dim.dyn_size_ext == time_dim.dyn_size_ext`. This PR adds this.

Why? The user probably never should access `dyn_size_ext` directly. So I don't exactly understand why this is relevant?

albertz:
Note that having `other_time_dim.dyn_size_ext == time_dim.dyn_size_ext` but `other_time_dim != time_dim` could be problematic with some legacy code which identifies the dim tag via size placeholder. For that reason, we still have a `tf.identity` in some cases to make really a new dim tag.

albertz:
> Note that currently, `dim_tag.copy(..)` will not copy over the `dyn_size_ext` properly because usually it is `None` when this is called by the user. `Dim.dyn_size_ext` is only completed via `extern_data` or by some layer later during the net dict construction. This is somewhat non intuitive.

But when is this an issue?

With `same_as_self=True`, this never should be an issue I think.

With `same_as_self=False`, I only really see use cases for static dims.

albertz:
What is the behavior of `ReinterpretDataLayer` before this PR?

Usually there is the `Data.sanity_check` which should check that `dyn_size` (and thus `dyn_size_ext`) is set for all dim tags, after the layer is constructed. Why is this not triggered?

Zettelkasten:
Okay, you are right, I had the wrong impression on how dimension tags completed their `dyn_size_ext`.

What I was confused about is that `mask` and the other layers do not have `other_dim_tag` itself, but rather a copy via `get_for_batch_ctx`. This all happens implicitly in the constructor of `Data` via `_adapt_batch_consistent_dim_tags`, which I only realized now.
Then, in `__init__` of `ReinterpretDataLayer`, `dyn_size_ext` of the `Dim` with batch is set properly, while `dyn_size_ext` of the original `other_time_dim` always remains `None`.

I guess this behavior is okay. I was confused by it at first.
I thought that `other_time_dim.dyn_size_ext is None` implied that layers after `ReinterpretDataLayer` didn't perform any masking via the seq lengths, but they do.

So I'll close this PR.

albertz:
Yea, the whole logic around `get_for_batch_ctx(...) -> Dim` is unfortunately a bit messy and un-intuitive. Better would maybe be some function `get_dyn_size_for_batch_ctx(...) -> Data`.

PR #892, opened 2022-01-04
remove eval check in transform_config_dict

JackTemaki:
Draft to check if the code for target dependency resolve can be simplified.


albertz:
Ok, the last commit makes problems. Can you revert that again? I can later check that separately.

albertz (approved):

JackTemaki:
Done!

PR #893, opened 2022-01-04
MergeDimsLayer, set default out_dim to prod of merged dims

Zettelkasten:


albertz (approved):

albertz (changes requested):
Whoops, failing tests.

returnn/tf/layers/basic.py
```diff
@@ -3107,29 +3118,24 @@ def get_out_data_from_opts(cls, name, axes, keep_order=NotSpecified,
...
       res_dim = int(numpy.prod([data.batch_shape[i] for i in axes]))
-    merge_dim_tags = [tag for (i, tag) in enumerate(data.dim_tags) if i in axes]
+    merge_dim_tags = [data.dim_tags[axis] for axis in axes]
     merge_target_axis = cls._get_target_axis(input_data=data, merge_axes=axes)
-    if any(tag.is_batch_dim() for tag in merge_dim_tags):
-      res_dim_tag_kind = Dim.Types.Batch
-    elif any(tag.is_feature_dim() for tag in merge_dim_tags):
-      res_dim_tag_kind = Dim.Types.Feature
-    else:
-      res_dim_tag_kind = Dim.Types.Spatial
```
albertz, line 3126: We might need to keep the `res_dim_tag_kind` logic.

returnn/tf/layers/basic.py
```diff
@@ -3107,29 +3118,24 @@ def get_out_data_from_opts(cls, name, axes, keep_order=NotSpecified,
...
     if out_dim:
       assert out_dim.dimension == res_dim
     else:
-      out_dim = Dim(
-        kind=res_dim_tag_kind, description="%s_merge_dims" % name,
-        dimension=res_dim)
+      from numpy import prod
+      out_dim = prod(merge_dim_tags)
+      assert isinstance(out_dim, Dim)
```
albertz, line 3137: Maybe add here:
```
out_dim = out_dim.copy(kind=res_dim_tag_kind)
```

returnn/tf/layers/basic.py
Zettelkasten: Either this, or we extend the dim arithmetic in general via `_OpMultTerm/_OpLinearTerm` to set the dim tag kind like here. What do you think?

albertz: Currently `_OpMultTerm.as_dim()` uses `self.base_term().kind`, where `base_term()` returns `self.terms[-1]`. So this is kept simple by intention.

But yea, maybe it makes sense to redefine this via the logic from here.

Zettelkasten: I implemented this in `_Op*Term.as_tag()` now.

returnn/tf/util/data.py
```diff
@@ -5603,6 +5603,20 @@ def _default_feature_dim_axis(batch_dim_axis, time_dim_axis, batch_shape, sparse
   return axes[-1]
 
 
+def _get_most_general_dim_kind(dim_tags):
```
albertz, line 5606: I think the name is misleading. It's actually the most specific dim kind.

But I would better name it for what it is used for, ie like _get_merged_dim_kind.

Zettelkasten: Yeah okay, that's better.

albertz (approved):

Issue #890, opened 2022-01-03
Compiling NativeOps does not work anymore with GCC-5


JackTemaki:
When compiling `NativeLstm2`I get the following error:
```
/work/tools/asr/python/3.8.0_tf_2.3.4-haswell+cuda10.1+mkl/lib/python3.8/site-packages/tensorflow/include/absl/time/internal/cctz/include/cctz/civil_time_detail.h:587:20: error: call to non-constexpr function ‘absl::lts_2020_02_25::time_internal::cctz::detail::civil_time<absl::lts_2020_02_25::time_internal::cctz::detail::day_tag> absl::lts_2020_02_25::time_internal::cctz::detail::operator-(absl::lts_2020_02_25::time_internal::cctz::detail::civil_time<absl::lts_2020_02_25::time_internal::cctz::detail::day_tag>, absl::lts_2020_02_25::time_internal::cctz::diff_t)’ 
           return cd - (j - i); 
                    ^ 
```

This is a known bug because GCC-5 is not correctly (at least not in the sense of how the `absl` library wants it) handling the c++14 mode. In c++11 constexpr are disabled, but we added this specific hack to disable it for c++14 as well:

```
diff --git a/absl/time/internal/cctz/include/cctz/civil_time_detail.h b/absl/time/internal/cctz/include/cctz/civil_time_detail.h
--- a/absl/time/internal/cctz/include/cctz/civil_time_detail.h
+++ b/absl/time/internal/cctz/include/cctz/civil_time_detail.h
@@ -23,7 +23,7 @@
 #include "absl/base/config.h"
 
 // Disable constexpr support unless we are in C++14 mode.
-#if __cpp_constexpr >= 201304 || (defined(_MSC_VER) && _MSC_VER >= 1910)
+#if (!defined(NO_CONSTEXPR_FOR_YOU) && __cpp_constexpr >= 201304) || (defined(_MSC_VER) && _MSC_VER >= 1910)
 #define CONSTEXPR_D constexpr  // data
 #define CONSTEXPR_F constexpr  // function
 #define CONSTEXPR_M constexpr  // member
```

So what we (at i6) could do now is:
 - add this `NO_CONSTEXPR_FOR_YOU` flag to the compile opts (very specific, would only work for custom compiled TF versions but not for default pip installed ones)
 - default to c++11 also for GCC-5 (see #875 )
[...]

PR #896, opened 2022-01-04
Use -std=c++11 for GCC 5 and lower

albertz:
Updated PR #875, resolving the concerns, and also fixing #890.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -2786,6 +2786,19 @@ class OpCodeCompiler(NativeCodeCompiler):
       return nvcc_opts
     return super(OpCodeCompiler, self)._transform_compiler_opts(opts)
 
+  def _extra_common_opts(self):
+    if self.is_cpp:
+      tf_gcc_version = get_tf_gcc_version()
+      if tf_gcc_version and int(tf_gcc_version[0]) <= 5:
+        # GCC4 does not support c++14, needed to support TF 1.14 and earlier
+        #   https://github.com/rwth-i6/returnn/pull/875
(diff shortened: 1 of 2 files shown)


JackTemaki:
I tried and still got `c++14` in the compile command, I will debug this now.

JackTemaki:
Ah right, I was not so smart... This works of course for our custom compiled Tensorflow versions, but not for pip installed ones, as there the GCC version is 7.3.1, so not reflecting the actual GCC version that can be found.

Well, for me this would be good enough.

albertz:
> Ah right, I was not so smart... This works of course for our custom compiled Tensorflow versions, but not for pip installed ones, as there the GCC version is 7.3.1, so not reflecting the actual GCC version that can be found.
> 
> Well, for me this would be good enough.

Yes it just looks at the `tf.__compiler_version__` currently.

But this should work in your case when you use your custom compiled TF (compiled with GCC 5) (using `-std=c++11`).

When you use the pip package (compiled with GCC 7), and you have GCC 7 (or newer) available, it should also be correct (using `-std=c++14`).

When you use the pip package and you do not have a matching GCC version installed, it would anyway not work.

albertz:
@JackTemaki You did not respond anymore. I assume everything is fine? So I merge now.

albertz:
Small follow-up in commit 72cc8d15f72caefd55bb5eb87eaf4a629dfee09f.

JackTemaki:
> @JackTemaki You did not respond anymore. I assume everything is fine? So I merge now.

I could not validate yet the "it would anyway not work" statement, because I am sure this worked in some cases. But we can just leave it like it is until someone complains.

PR #904, opened 2022-01-13
add force_no_blas flag to compile_native_op.py

JackTemaki:
When embedding a compiled native op into a host software that is linked against Tensorflow (e.g. RASR), you want that the compiled native op uses exactly the same BLAS implementation as Tensorflow itself. But currently the tool searches automatically for a blas implementation, and might use the wrong one then.

When adding this flag no specific blas library will be linked.

### DIFF
--- a/tools/compile_native_op.py
+++ b/tools/compile_native_op.py
@@ -76,6 +76,9 @@ def main(argv):
                          help="search for blas inside numpys .libs folder")
   argparser.add_argument('--no_search_for_numpy_blas', dest='search_for_numpy_blas', action='store_false',
                          help="do not search for blas inside numpys .libs folder")
+  argparser.add_argument('--force_no_blas', action='store_true',
+                         help="do not link any blas lib into the native op binary "
+                              "(e.g. use the BLAS library from the host application)")
   argparser.add_argument("--verbosity", default=4, type=int, help="5 for all seqs (default: 4)")
   argparser.add_argument("--output_file", help='if given, will write the list of libs to this file')
   args = argparser.parse_args(argv[1:])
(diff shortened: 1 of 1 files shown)


albertz:
Why add a workaround instead of fixing the issue, i.e. namely:

> currently the tool searches automatically for a blas implementation, and might use the wrong one then.

JackTemaki:
It can be that it is required, because it is not part if the host software. I am not in the position to tell what is needed and what is not. I just do not want to mess with existing tool options, so adding a new flag is the only option that does not cause issues.

albertz:
> It can be that it is required, because it is not part if the host software.

I don't understand this. What host software? You mean TF? But TF always has BLAS?

> I am not in the position to tell what is needed and what is not. I just do not want to mess with existing tool options, so adding a new flag is the only option that does not cause issues.

But this doesn't make sense. Only because you don't know what is needed does not mean you should add an option to workaround the fact that you don't know and you want to ignore if it is really needed.

There should not be any complex logic, like having a new flag. What do you want to write in the documentation when this flag should be used? "I don't know. Try it out. Maybe it helps."?

We should simply figure out when linking BLAS is needed. Maybe it is never needed. Then we should simply remove this all together.

Maybe we made an error and incorrectly assumed it is never needed, and only later we figure out that there are some rare cases where it is needed. Then whatever the condition is, we can automatically put it in under this condition.

But there should not be a flag for this.

Esp, the flag should not be sth you would always need to use. The default behavior should be the correct behavior. A flag should do something which is special. So maybe the flag would be the opposite here, like `--link-blas`.

JackTemaki:
> What do you want to write in the documentation when this flag should be used?

Well, clearly in the case where you do not want RETURNN to find BLAS with its own heuristics.

> Esp, the flag should not be sth you would always need to use.

This is also not the case. I thought I need it, maybe nobody else needs this. But I will just use `--blas_lib` then. This is slightly more overhead in my pipeline, but maybe that is anyway better because it is more verbose to someone reading the pipeline what is going on.

albertz:
> > What do you want to write in the documentation when this flag should be used?
> 
> Well, clearly in the case where you do not want RETURNN to find BLAS with its own heuristics.

This does not say anything. A documentation "--force_no_blas: force no BLAS" does not help. I was asking about *why*, why do you want that?

If you say there are cases when RETURNN selects the wrong BLAS (I still did not understand this. is this the case?) then we should just fix that. (There is no issue describing this problem?)

If you say there is no BLAS linking needed at all (I still did not understand this. is this the case?), then why do we need a flag for that? Why don't we simply don't do that then? (There is no issue describing this problem?)

> > Esp, the flag should not be sth you would always need to use.
> 
> This is also not the case. I thought I need it, maybe nobody else needs this.

Why do you need it? That's what I asked before. That also should be said in the documentation. And why do you think you are the only one needing it?

> But I will just use `--blas_lib` then.

Why do you need that? That should not be needed.

PR #906, opened 2022-01-14
Auto-flatten optimization

albertz:
A (frame-wise) computation like `loss_per_frame = CE(softmax(y), targets)` would in the end only consider the non-padded frames for `loss = sum(loss_per_frame)`. Let's assume `loss_per_frame` is of shape [B,T]. This can be achieved in multiple ways:

* The `sum` could apply masking (like our `ReduceLayer`).
* We could flatten `loss_per_frame`, i.e. take only the non-padded frames, i.e. of shape [B'] where B' <= B*T.
* The flattening can be pushed upwards through the computation. We could already flatten `y` and `targets` instead. This would be more efficient, as we don't waste computation on softmax. But it can be further pushed upwards. E.g. `y` is likely the result of a matmul. The flattening can happen before that. Etc.

This PR adds code which automatically pushes the flattening upwards from a (frame-wise) loss as much as possible.

There are a couple of heuristics to figure out when this is safe to do. E.g. it does not push the flattening through a layer with `recurrent` flag set. It also doesn't apply the optimization when there are other dynamic spatial dims despite T.

---

Note that in the old way of having `"loss": "ce"` in a layer, this would try to figure out the input to the softmax (`y` in the example above) and then do the flattening on the `y` and on the `targets`. So right now, we already have that in a simple form but not really in a generic way.

Now when some layer output would be used for other purpose, or you would use `"loss": "as_is"`, this old way would not work. Of course we could similarly also add a number of checks to cover e.g. specifically `sparse_softmax_cross_entropy_with_logits` or so. But this would always be restricted.

For: https://github.com/rwth-i6/returnn_common/issues/68

---

We still need to think about cases when some layer output would be used elsewhere. Then it probably should not be flattened.

Or multiple losses.

See discussion below.

### DIFF
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -1218,6 +1218,7 @@ class TFNetwork(object):
     return losses_dict, total_loss, total_constraints
 
   def _construct_objective(self):
+    self._flatten_layer_with_losses()
     with tf.name_scope("objective"):
       losses_dict, total_loss, total_constraints = self.get_losses_initialized(with_total=True)
       self.losses_dict.clear()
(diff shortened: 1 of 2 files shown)


albertz:
Note what we further need to do here (copied TODO from code):

This code works for the simple test case, but likely it breaks for some nasty edge cases.
Also, in case of the test case and in many other examples, it actually pushes the flattening through most (all) of the network.
In principle this is fine (if all is correct) but we need to be careful.
Also, when there are multiple losses, or some of the outputs are used in some way for other purpose, even just for a summary, it would be wasted computation.
This should be avoided. So we need to check for any of the tensors if they end up to the other fetches, and if so, we should not flatten.
Or for multiple losses, every loss should not use its own flattening but this should be shared.

albertz:
How can we know when layer outputs are used elsewhere in the network, or the output is used directly (e.g. in forward or search task)?

The `LayerBase.is_output_layer()` logic is not exactly describing the case for outputs used directly. Maybe we further need to check the task but that behavior (dependence on task) might be unintuitive. I don't think it's a good idea that such optimization is triggered by the type of task.

When we can figure out all the output layers needed, we can go recursively and flag those layers to not be optimized for flattening.

When we can figure out all the endpoints in the network (computation graph), which are layer outputs or losses, we can jointly do this flattening logic, and for the output layers, provide unflattening again such that the layer output can be used.

In case the batch size is 1 (which is common for non-batched decoding e.g. in RASR), the flattening and unflattening however would unnecessarily slow down the computation because it implies unnecessary copies. This must be avoided.

Or maybe we introduce a global config option which enables or disables this optimization. And this option could be True by default in case of the train task, and False otherwise.

albertz:
@Zettelkasten @JackTemaki Do you have some opinion? I guess you agree that this is a good thing to have. The question is more about how to introduce it. In any case, this is a must-have for https://github.com/rwth-i6/returnn_common/issues/68.

Zettelkasten:
Why would you disable this option by default for the eval task etc.? Do you expect this optimization there to make things worse usually? Perhaps if everything intermediate is flattened, but then some output should not be flattened, you spend some time un-flattening again. But I would think in most cases, auto flattening should not decrease performance, or not?
If it almost always actually makes performance better, I think auto-flattening should just be enabled always by default.

Also, without thinking about this too much, wouldn't auto-flattening hide possible bugs if the padding is not handled correctly? In this case it might be that a config works well during training (where auto-flattening is enabled), but then, if it is not enabled anymore during eval or so, it is buggy because padding is not ignored properly somewhere by accident.

We definitely need an explicit "switch off" though anyway for debugging, I think it is more straight-forward to look at padded tensors than flattened ones.

JackTemaki:
I would actually vote against having heuristics to do this. While it is good to have this efficiency increase in many cases, it adds additional complexity and introduces a risk of not working in some special cases. I would prefer if this flattening stays verbose to the user, so that this is always done explicitly. For me a consistent behavior would be definitely preferable over some performance improvements.

albertz:
> Why would you disable this option by default for the eval task etc.?

I did not think about the eval task. Probably we can enable it by default for that as well.

I was mostly thinking about the forward task, or any task where a layer output would be used externally, e.g. by RASR. RASR would not expect a flattened tensor. I don't want to break anything. See the `compile_tf_graph.py` script. This doesn't even set any of the standard `task` values. I think it is better to be restrictive for now. If the environment or settings are somewhat unusual, we probably should not do the optimization.

> Do you expect this optimization there to make things worse usually?

No, I think it should always help. My concern was just about other external software which would not expect flattened tensors.

> Perhaps if everything intermediate is flattened, but then some output should not be flattened, you spend some time un-flattening again.

Yea, unflattening is also an option to do at the end for the output layers. But this again adds a bit of computation. E.g. if this is just an elem-wise operation at the end, like applying sigmoid or so, first flattening and then unflattening would be definitely way too costly. Even if it is a linear layer, I'm not sure if just doing that directly is worth it.

So I would mostly thinking about the case where we don't need the unflattened tensor at all anymore. Like for losses.

> Also, without thinking about this too much, wouldn't auto-flattening hide possible bugs if the padding is not handled correctly? In this case it might be that a config works well during training (where auto-flattening is enabled), but then, if it is not enabled anymore during eval or so, it is buggy because padding is not ignored properly somewhere by accident.

This can anyway only be applied for operations and layers which do not make use of the dynamic spatial dim. Otherwise they cannot work on the flattened tensor. And when they don't make use of it, I don't see how it could get wrong. But you are right that consistency would be better. But efficiency is even more important.

albertz:
> I would actually vote against having heuristics to do this.

The heuristics are pessimistic. Maybe also "heuristic" is the wrong word. There are a couple of rules to exclude a lot of cases where to apply the optimization. For the remaining cases, I think it should definitely be safe (unless there are other bugs elsewhere but as I excluded so much, I don't think so). E.g. I exclude layers which have the `recurrent` flag set. This is e.g. `RecLayer`, `ConvLayer`, `PoolLayer`, many others. This makes sense. Then I also exclude layers which have parameters. This is mostly just because I was too lazy to implement that case yet (the recreated layer would need to share params) but also because usually those layers are more complex and maybe that would reveal some hidden bugs. Then I also exclude layers where there are any other dynamic unrelated dims. Etc. I think the couple of remaining cases should be safe.

> While it is good to have this efficiency increase in many cases, it adds additional complexity and introduces a risk of not working in some special cases. I would prefer if this flattening stays verbose to the user, so that this is always done explicitly. For me a consistent behavior would be definitely preferable over some performance improvements.

I would say that efficiency is a higher priority for RETURNN. Also, as said, we already have this optimization right now, since a long time, even since Theano. This PR here just makes it a bit less heuristic and more generic. And it is required that we also have the optimization for returnn-common, which is a requirement I think (people would not want to use returnn-common when it is slower).

But maybe you argue it should just be done for the losses, as it was done before?

albertz:
Some technical details:

Where should the optimization be implemented?

(The current draft code triggers it in `Loss._flatten_or_merge` and then recreates corresponding layers, such as the `SparseSoftmaxCrossEntropyWithLogitsLayer`. But that is problematic when there are multiple losses or when the output would be used elsewhere in the network.)

We could do it at the end after the network was constructed, and then define the end points of the network as all the losses and all the output layers, and apply the optimization, i.e. go backwards from the end points as far as possible. One downside to this is maybe that it would recreate layers and add unnecessary paths in the computation graph, and also might be slightly hacky.

Maybe we can already do it earlier at construction time. To know whether it can be flattened, we need to know all the shape information. For that, we would first need a template construction of everything. We don't have that yet. I'm not sure how easy it is to add that.

JackTemaki:
> But maybe you argue it should just be done for the losses, as it was done before?

I argue that it should be explicit. So that this would be a part of the modules that people write and use, so that this is verbose to the user, and not something that is called afterwards and does "hidden" changes. Especially with the problems you are describing (multiple losses, construction overhead) I would prefer to not do it automatically but have this explicit. And this does not need to be losses only, but can be more in the scenarios where it makes sense.

albertz:
> > But maybe you argue it should just be done for the losses, as it was done before?
> 
> I argue that it should be explicit. So that this would be a part of the modules that people write and use, so that this is verbose to the user, and not something that is called afterwards and does "hidden" changes. Especially with the problems you are describing (multiple losses, construction overhead) I would prefer to not do it automatically but have this explicit. And this does not need to be losses only, but can be more in the scenarios where it makes sense.

But then in practice it will not be used. This is already available since a long time. No-one uses it except where it is already done automatically. So this is not an option.

Also, the things I'm describing are not really problems. They can easily be solved. Also in a way that they work 100% correct. I think when we can make sure that it is correct (and we can do this here), such optimizations are fine. In the same way, we do also a lot of other optimizations, like reordering of axes, or moving layers out of a recurrent loop. These are very crucial features of RETURNN. Otherwise it would be quite slow.

This is not really the problem. I already outlined solutions for all of that. My current draft implementation here is also already working fine, and in a safe way. All tests are passing.

The only concerns I have is for external software which maybe fetches from a RETURNN computation graph and which would not expect a flattened tensor. But this is also simple to avoid with my suggestion above. Another concern I had is for some rarely used buggy layers which might behave incorrect. But also this can be avoided by being very restrictive. But actually I would also say that we should fix those layers then. We should not try to hide or work around other bugs.

albertz:
Maybe we can stay conservative for now and just enable it for the losses during training.

In any case, we need some solution here. We must have this for returnn-common such that models defined in returnn-common would have the same efficiency optimizations we usually have.

albertz:
I extracted some somewhat unrelated changes into their own PRs: #942, #943, #944. #944 must be merged before this here. I will rebase once they are all merged.

The main logic was moved now to `TFNetwork._flatten_layer_with_losses`, which is called via `TFNetwork._construct_objective`, so only when we ask for the losses specifically, which is only in training or evaluation. There is no config option.

As said before, all the checks when the flattening optimization is applied are very restrictive. So I think this should be safe. All tests run through.

So, the PR is ready now. Feel free to review.

PR #907, opened 2022-01-17
Issues in SplitDimsLayer

vieting:
https://github.com/rwth-i6/returnn/commit/dbaa97a9d9cfd90d073d743786b6a233cdb5eeeb from #845 introduces an issue when unflattening the batch dim axis. All resulting axes now get a batch dim `dim_tag`, which was previously not the case.

I added a test case to demonstrate the issue.

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -2066,6 +2066,19 @@ def test_SplitDimsLayer_dim_tags():
     })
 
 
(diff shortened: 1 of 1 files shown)


albertz:
(Btw, it is better to first open an issue describing a problem. You can simply put the test case into the description of the issue. A PR is about fixing this then. If you want, you can additionally create a draft PR where you start working on it, e.g. by adding the test case. But this is not really needed.)

vieting:
Yes, sorry. I'll do it like this.

PR #909, opened 2022-01-17
Fix issues in SplitDimsLayer

### LINKED ISSUE #908: Issues in SplitDimsLayer
dbaa97a from #845 introduces an issue when unflattening the batch dim axis. All resulting axes now get a batch dim dim_tag, which was previously not the case.

vieting:
Fix #908.

Follow-up to #845.

Related: https://github.com/rwth-i6/pytorch-to-returnn/pull/78

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3487,7 +3487,7 @@ class SplitDimsLayer(_ConcatInputLayer):
     if not resolved_dims:
       resolved_dims = tuple(
         Dim(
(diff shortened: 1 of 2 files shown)


albertz:
(When you post a (draft) PR but do not intend to work on it further yourself, please tell so here. Otherwise it's not clear whether others should work on it. Potential code conflicts should be avoided.)

albertz:
(Btw, about branch names: It would be good to prefix it with your name. Further, it would be good to have the issue number in it when the PR is about fixing some issue.)

PR #911, opened 2022-01-19
VariableLayer, init_by_layer option

albertz:
Related:

https://github.com/rwth-i6/returnn/wiki/Parameter-initialization
https://github.com/rwth-i6/returnn_common/issues/95
https://github.com/rwth-i6/returnn_common/issues/92#issuecomment-1014403308


albertz:
Unfortunately, the code does not seem to work with TF 2.4.2 from our CI.

I'm not really sure whether it is worth to have extra code only for such older TF version.

I tried updating TF in the CI but that causes other problems, so I will do that in a separate PR.

Zettelkasten:
> Unfortunately, the code does not seem to work with TF 2.4.2 from our CI.
> 
> I'm not really sure whether it is worth to have extra code only for such older TF version.

Sure TF 2.4 is quite old, but from what I know most people here still use it or even something much earlier. There is not even a newer TF version in our "official" shared folder `/work/tools/asr/python` than TF 2.4.

PR #918, opened 2022-01-29
Contrastive loss demo

albertz:
@mmz33 @vieting

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -6417,6 +6417,7 @@ def test_contrastive_loss():
   enc_feat_dim = FeatureDim("encoder_dim", 20)
 
   dim_neg_samples = SpatialDim("neg_samples", 10)  # 100
+  dim_expand = SpatialDim("expand_dim", 1)
   contrastive_loss_temp = 0.1
   contrastive_loss_factor = 1.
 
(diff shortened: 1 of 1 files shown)


albertz:
@mmz33 Ok, now I have a version which compiles but produces nan loss. I have not looked further into it yet. It might be just some stupid thing that the masking masks everything and that causes nan, or sth like that. Or maybe still some bug.

albertz:
I added the `debug_print_layer_output` option, which is quite insightful here. I see multiple problems.

* ```[input_mask]
  [[False False False False False False  True]
   [False  True  True False False False False]
   [False False False False False False  True]]
  ```
  I guess the mask is ok. Just to verify: True means to zero out, False means to keep, right?

* ```[samples_rand_indices_]
  [[1 1 0 0 1 1 0 0 0 2]
   [1 0 0 2 1 0 0 0 1 1]
   [0 0 0 2 1 1 0 2 0 1]]
  ```
  This looks like it took the wrong max val. I assume it took the wrong batch len.
  Actually now I see much above:

* `[input_flat_len][3]`: **This is the wrong len.**

* For reference: `[input_flat][shape:][Data{'input_flat_output', [B&Packed{'time:var:extern_data:data'},F|F'encoder_dim'(20)]}][16 20]`. I.e. `input_flat_len` should have been 16.

* For reference (looks fine):
  ```
  [sampled_frames_][shape:][Data{'sampled_frames__output', [B&Packed{'input_masked_frames:masked:time'},'neg_samples'(10),F|F'encoder_dim'(20)]}][4 10 20]
  [sampled_frames][shape:][Data{'sampled_frames_output', [B&Packed{'input_masked_frames:masked:time'},T|'expand_dim+neg_samples'(11),F|F'encoder_dim'(20)]}][4 11 20]
  ```

* ```[enc_masked_frames_]
  [[[0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
    [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
    [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]]
  
   [[0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
    [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
    [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]]]
  ```
  **This looks wrong.** It should not be all 0.

* `[norm_b][0. 0. 0. 0.]` Yea, this is correct for the 0.
  **We see another problem here that we miss an eps.**

albertz:
@mmz33 Except of the first point about potentially masked frames outside the seq length, all other main issues are fixed, and I guess this looks ready.

PR #919, opened 2022-01-29
RandIntLayer: allow shape from layer

vieting:
For https://github.com/rwth-i6/pytorch-to-returnn/pull/98, we need the `RandIntLayer` to allow shapes based on other layers. In the added test case, we e.g. create rand ints with the shape `(B, T)`. The question is how to obtain the `dim_tag` if the input layer is not a `LengthLayer`, e.g. for the time dim in the test cast as the output of the `LengthLayer` is reduced here.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2248,7 +2248,7 @@ class RandIntLayer(LayerBase):
   # noinspection PyUnusedLocal
   def __init__(self, shape, maxval, minval=0, dtype="int32", sparse_dim=None, seed=None, **kwargs):
     """
-    :param tuple[Dim|int]|list[Dim|int] shape: desired shape of output tensor
+    :param tuple[Dim|int|LayerBase]|list[Dim|int|LayerBase] shape: desired shape of output tensor
     :param int|LayerBase maxval: upper bound (exclusive) on range of random values
     :param int|LayerBase minval: lower bound (inclusive) on range of random values
     :param str dtype: type of the output. For random ints, int32 and int64 make sense, but could also be floats
@@ -2257,7 +2257,7 @@ class RandIntLayer(LayerBase):
     """
     super(RandIntLayer, self).__init__(**kwargs)
     seed = seed if seed is not None else self.network.random.randint(2 ** 31)
-    shape_ = [d.get_dim_value() for d in self.output.dim_tags]
+    self.shape = [d.get_dim_value() for d in self.output.dim_tags]
     self.minval = minval
     self.maxval = maxval
     if isinstance(minval, LayerBase):
@@ -2265,7 +2265,7 @@ class RandIntLayer(LayerBase):
     if isinstance(maxval, LayerBase):
       assert maxval.output.batch_shape == ()  # only scalars supported
     self.output.placeholder = tf.random.uniform(
-      shape_,
+      self.shape,
       minval=minval.output.placeholder if isinstance(minval, LayerBase) else minval,
       maxval=maxval.output.placeholder if isinstance(maxval, LayerBase) else maxval,
       dtype=dtype, seed=seed)
@@ -2279,6 +2279,9 @@ class RandIntLayer(LayerBase):
       deps.append(self.minval)
     if isinstance(self.maxval, LayerBase):
       deps.append(self.maxval)
+    for dim in self.shape:
+      if isinstance(dim, LayerBase):
+        deps.append(dim)
     return deps
 
   @classmethod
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -2290,16 +2293,17 @@ def transform_config_dict(cls, d, network, get_layer):
     """
     d.setdefault("from", [])
     super(RandIntLayer, cls).transform_config_dict(d, network=network, get_layer=get_layer)
-    if isinstance(d.get("maxval", None), str):
-      d["maxval"] = get_layer(d["maxval"])
-    if isinstance(d.get("minval", None), str):
-      d["minval"] = get_layer(d["minval"])
+
+    d["maxval"] = get_layer(d["maxval"]) if isinstance(d["maxval"], str) else d["maxval"]
+    d["minval"] = get_layer(d["minval"]) if isinstance(d["minval"], str) else d["minval"]
```
albertz, line 2298: Why this change for maxval/minval?

returnn/tf/layers/basic.py
```diff
@@ -2290,16 +2293,17 @@ def transform_config_dict(cls, d, network, get_layer):
...
+    d["shape"] = tuple(get_layer(dim) if isinstance(dim, str) else dim for dim in d["shape"])
+
```
albertz, line 2300: One empty line too much.

albertz:
> The question is how to obtain the `dim_tag` if the input layer is not a `LengthLayer`, e.g. for the time dim in the test cast as the output of the `LengthLayer` is reduced here.

Not all layer outputs correspond to an existing dim tag.

Via `Dim.get_tag_from_size_tensor` on the `layer.output.placeholder`, you have a generic way to find any corresponding existing dim tag.

If it does not exist, you could also just create a new dim tag.

I think we should not try to be too clever on RETURNN side on this. I think we can much easier handle this on PyTorch-to-RETURNN side. I actually wonder, why do you need to pass layers at all? You can simply pass dim tags directly (which you get from `SizeValue`). What's the problem with that?

albertz:
Btw, somewhat related might be #920 and #918.

vieting:
> Via `Dim.get_tag_from_size_tensor` on the `layer.output.placeholder`, you have a generic way to find any corresponding existing dim tag.

Thanks for the hint. In the test case here, this returns `None` in both cases (even for the `LengthsLayer`)

> I think we should not try to be too clever on RETURNN side on this. I think we can much easier handle this on PyTorch-to-RETURNN side. I actually wonder, why do you need to pass layers at all? You can simply pass dim tags directly (which you get from SizeValue). What's the problem with that?

Currently we always do `x = x.as_tensor()` if `isinstance(x, SizeValue) and x.dim_tag.dimension is None` [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/878f62db88bbb0ed2c362d38f834082f4454dae7/pytorch_to_returnn/naming/naming.py#L205) in `self._make_tensor`. We could redefine this logic to use the dim tags directly for `RandInt`. Would that work also for the printed network dict?

albertz:
> > Via `Dim.get_tag_from_size_tensor` on the `layer.output.placeholder`, you have a generic way to find any corresponding existing dim tag.
> 
> Thanks for the hint. In the test case here, this returns `None` in both cases (even for the `LengthsLayer`)

This should not be. This might be another bug then.

> > I think we should not try to be too clever on RETURNN side on this. I think we can much easier handle this on PyTorch-to-RETURNN side. I actually wonder, why do you need to pass layers at all? You can simply pass dim tags directly (which you get from SizeValue). What's the problem with that?
> 
> Currently we always do `x = x.as_tensor()` if `isinstance(x, SizeValue) and x.dim_tag.dimension is None` [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/878f62db88bbb0ed2c362d38f834082f4454dae7/pytorch_to_returnn/naming/naming.py#L205) in `self._make_tensor`. We could redefine this logic to use the dim tags directly for `RandInt`.

Ah yes. No, this should not be changed. But you can get back the dim tag via `tensor_entry.is_dim`.

> Would that work also for the printed network dict?

We anyway make use of dim tags in other parts, and then also here. For the serialization of dim tags, we already spoke about that, this can be done. We did not implement it yet because it was not needed for the tests. Once we need that, we can follow a very similar implementation as in returnn-common.

One other small problem we might get here: Via dim tags, we will not properly derive dependencies. E.g. when the dim tag is some dynamic length originating from some other layer, when this dim tag is just used elsewhere, it might not be defined yet. To make sure it gets defined, we however can add it as explicit dependency. E.g. the `RandIntLayer` could just use `from`. It will ignore `from` otherwise. See here for example: https://github.com/rwth-i6/returnn/blob/0b49ab4c95911c040453e00f8358805e7560096a/tests/test_TFNetworkLayer.py#L6510

vieting:
> One other small problem we might get here: Via dim tags, we will not properly derive dependencies. E.g. when the dim tag is some dynamic length originating from some other layer, when this dim tag is just used elsewhere, it might not be defined yet. To make sure it gets defined, we however can add it as explicit dependency. E.g. the RandIntLayer could just use from. It will ignore from otherwise. See here for example: 

Yes I noticed this problem already. Is there a general answer to what should be the layer we use as "from" here?

albertz:
> > One other small problem we might get here: Via dim tags, we will not properly derive dependencies. E.g. when the dim tag is some dynamic length originating from some other layer, when this dim tag is just used elsewhere, it might not be defined yet. To make sure it gets defined, we however can add it as explicit dependency. E.g. the RandIntLayer could just use from. It will ignore from otherwise. See here for example:
> 
> Yes I noticed this problem already. Is there a general answer to what should be the layer we use as "from" here?

The `CopyLayer` has `extra_deps`. So you could add a `copy` on any of the inputs.

This would only work for layers with inputs however but not e.g. for `RandIntLayer`. For `RandIntLayer` however, we could use `from`. But still there are maybe others, e.g. `ConstantLayer` or so.

Maybe we should just add `extra_deps` to `LayerBase`. There was some technical reason why I did not do this but I don't exactly remember.

vieting:
My question was regarding what layer I should use here for `RandIntLayer` in `from` or if there is some specific layer that should be added and used as `from`.

albertz:
> My question was regarding what layer I should use here for `RandIntLayer` in `from` or if there is some specific layer that should be added and used as `from`.

The `SizeValue.originating_tensor` should be the right one.

albertz:
I guess this PR can be closed, right? From the discussion, I think we agreed that we should just use dim tags instead.

vieting:
> I guess this PR can be closed, right? From the discussion, I think we agreed that we should just use dim tags instead.

Yes, I agree

> > > Via `Dim.get_tag_from_size_tensor` on the `layer.output.placeholder`, you have a generic way to find any corresponding existing dim tag.
> > 
> > 
> > Thanks for the hint. In the test case here, this returns `None` in both cases (even for the `LengthsLayer`)
> 
> This should not be. This might be another bug then.

I could add an issue for this if you like.

albertz:
> > > > Via `Dim.get_tag_from_size_tensor` on the `layer.output.placeholder`, you have a generic way to find any corresponding existing dim tag.
> > > 
> > > Thanks for the hint. In the test case here, this returns `None` in both cases (even for the `LengthsLayer`)
> > 
> > This should not be. This might be another bug then.
> 
> I could add an issue for this if you like.

I think not so important, unless we run in it at some point. In general I think we should not use `Dim.get_tag_from_size_tensor` too much.

PR #928, opened 2022-01-31
fix num_outputs for speaker_name

### LINKED ISSUE #927: Speaker tags in SprintDataset
Currently there is no support to use speaker labels via the `ExternSprintDataset`.

In combination with 80d0a3845ca60ad78d49b25beeb5e22342b44ca7 I figured out that replacing
``` python
    self.num_outputs["speaker_name"] = (256, 1)
```
by
```python
    if "speaker_name" in self.target_maps:
      self.num_outputs["speaker_name"] = (len(self.target_maps["speaker_name"]), 1)
    else:
      self.num_outputs["speaker_name"] = (256, 1)
```
in https://github.com/rwth-i6/returnn/blob/master/returnn/datasets/sprint.py#L140
would be sufficient to get working embeddings.

However there are some issues here:
 - `speaker_name` is then containing indices instead of the utf8 string, which is counter-intuitive
 - It seems that `target_maps` can only be used for `speaker_name` and for nothing else, as the corresponding entry in `self.num_outputs` will always be wrong, thus causing a crash at some point. Also, there is no test case using `target_maps`.
 
 So my question would here be: should I simply proceed with the small fix (takes no time, would work for my use case) or should we refactor this to be better designed, meaning removing the possibly useless/broken `target_maps` parameter and code, and introducing a consistent and documented `speaker_map` parameter that will create a new output called `speaker_classes` (consistent naming with with `orth_classes`) based on the map file.

JackTemaki:
Fix #927.

### DIFF
--- a/returnn/datasets/sprint.py
+++ b/returnn/datasets/sprint.py
@@ -55,7 +55,9 @@ class SprintDatasetBase(Dataset):
                suppress_load_seqs_print=False,
                **kwargs):
     """
-    :param dict[str,str|dict] target_maps: e.g. {"speaker": "speaker_map.txt"}
+    :param dict[str,str|dict] target_maps: e.g. {"speaker_name": "speaker_map.txt"}, with "speaker_map.txt" containing
+      a line for each expected speaker. The indices will be given by the line index. Note that scalar
+      content (e.g. single index) will automatically get a time axis added with the length of the audio frames.
     :param bool str_add_final_zero: adds e.g. "orth0" with '\0'-ending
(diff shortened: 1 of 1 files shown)


albertz:
(Meta comment: You always should refer to the actual issue this is fixing in a PR or commit.)

albertz (approved):

JackTemaki:
> (Meta comment: You always should refer to the actual issue this is fixing in a PR or commit.

I got an "you can not edit this message because the content already changed", just be 30 seconds more patient next time :)

PR #929, opened 2022-02-01
CI, update TF to 2.6.2

albertz:


albertz:
Obsolete via TF 2.10, #1160.

PR #930, opened 2022-02-01
CachedDataset: fixed check whether sequence order changed

patrick-wilken:
Fixup for 39d5c6f.

@mattiadg, can you check whether this works for you?

### DIFF
--- a/returnn/datasets/cached.py
+++ b/returnn/datasets/cached.py
@@ -86,8 +86,12 @@ class CachedDataset(Dataset):
     old_index_map = self._index_map[:]
     self._index_map = range(len(seq_index))  # sorted seq idx -> seq_index idx
 
-    if (isinstance(seq_index, numpy.ndarray) and numpy.array_equal(self._seq_index, seq_index)
-        or self._seq_index == seq_index) and self.start_cache_initialized:
+    if isinstance(seq_index, numpy.ndarray):
+      seq_index_unchanged = numpy.array_equal(self._seq_index, seq_index)
+    else:
+      seq_index_unchanged = (self._seq_index == seq_index)
+
+    if seq_index_unchanged and self.start_cache_initialized:
       return False
 
     if epoch is not None:


mattiadg:
running it

patrick-wilken:
Sorry, updated to use a slightly better variable name.

JackTemaki (approved):

returnn/datasets/cached.py
```diff
@@ -86,8 +86,12 @@ def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
     old_index_map = self._index_map[:]
     self._index_map = range(len(seq_index))  # sorted seq idx -> seq_index idx
 
-    if (isinstance(seq_index, numpy.ndarray) and numpy.array_equal(self._seq_index, seq_index)
-        or self._seq_index == seq_index) and self.start_cache_initialized:
+    if isinstance(seq_index, numpy.ndarray):
+        seq_index_unchanged = numpy.array_equal(self._seq_index, seq_index)  # equal operator would compare element-wise
```
albertz, line 90: Indentation is wrong.

albertz: Do you know how? I wonder that PyCharm does not warn about this. Or maybe it just needs to be enabled?

JackTemaki: I just checked why it does not do that, but the PyCharm inspector for `inconsistent indentation` only checks tabs vs spaces, but not the indentation level. If I move something by 4 I get no PyCharm warning.

JackTemaki: And the PEP8 inspector can only check for 4-space indents, we disabled this specifically to allow for a 2 space indentation.

JackTemaki: The indentation check is disabled here: https://github.com/rwth-i6/returnn/blob/master/tests/PyCharm.idea/inspectionProfiles/Project_Default.xml#L45

patrick-wilken: Oops, fixed.
I also have checking for indentation disabled in my editor because of RETURNN. 😬

albertz (approved):

returnn/datasets/cached.py
albertz: @JackTemaki This is actually not used by our test case. It runs the PEP8 check independently. See here: https://github.com/rwth-i6/returnn/blob/0c6d863f4a4013acdd933ae21ae00cbbfd7309cd/tests/pycharm-inspect.py#L492

albertz: Maybe when we add `--indent-size=2`, we don't need to exclude `E111,E114` and it works?

albertz: I started draft PR #931 to test this.

mattiadg:
It didn't crash anymore, then I'm not sure about the behavior it should have but the training speed is comparable to the previous runs, and in the epochs after first one is slightly lower ~1:02h vs 1:09. Maybe it's due to random factors though

albertz:
I don't understand. This change here is simply about fixing a bug (crash), not about improving the speed of anything? Or did I misunderstand that? @patrick-wilken @mattiadg?

albertz:
I assume the PR here is anyway fine then and we can merge this?

patrick-wilken:
I don't understand either. I guess in your setup you always have the same sequence order and therefore always `seq_index_unchanged=True` except for the very first dataset initialization. Then it is expected that the first epoch takes a bit longer. And this is happening, right?
Anyway, let's merge.

PR #931, opened 2022-02-02
PyCharm inspect, check indentation

albertz:
https://github.com/rwth-i6/returnn/pull/930#discussion_r797587361

### DIFF
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -65,10 +65,10 @@ class OggZipDataset(CachedDataset2):
     from .meta import EpochWiseFilter
     self._separate_txt_files = {}  # name -> filename
     if (
-      isinstance(path, str)
-      and os.path.splitext(path)[1] != ".zip"
-      and os.path.isdir(path)
-      and os.path.isfile(path + ".txt")):
+          isinstance(path, str)
+          and os.path.splitext(path)[1] != ".zip"
+          and os.path.isdir(path)
+          and os.path.isfile(path + ".txt")):
       # Special case (mostly for debugging) to directly access the filesystem, not via zip-file.
       self.paths = [os.path.dirname(path)]
       self._names = [os.path.basename(path)]
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1107,8 +1107,9 @@ class Dataset(object):
             [length, batch_size.copy_like(length) - batch.get_all_slices_num_frames().copy_like(length)])
           assert num_frames.max_value() > 0
           batch.add_frames(seq_idx=seq_idx, seq_start_frame=t_start, length=num_frames)
-          if (batch.get_all_slices_num_frames().any_compare(batch_size, (lambda a, b: a >= b))
-                  or batch.get_num_seqs() > max_seqs):
+          if (
+                batch.get_all_slices_num_frames().any_compare(batch_size, (lambda a, b: a >= b))
+                or batch.get_num_seqs() > max_seqs):
             yield batch
             batch = Batch()
           t_start += num_frames
--- a/returnn/datasets/sprint.py
+++ b/returnn/datasets/sprint.py
@@ -981,9 +981,9 @@ class ExternSprintDataset(SprintDatasetBase):
       epoch = 1
     with self.lock:
       if (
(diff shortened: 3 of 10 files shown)


albertz:
```
  AssertionError: unexpected line 'Usage: pycodestyle [options] input ...'
777
```
I guess the PyCharm bundled pycodestyle does not support this option. Maybe it's too old. We could maybe install some own instead.

JackTemaki:
The results look reasonable, I would say we can go with this!

albertz:
Yes, looks good. We should fix those warnings before we merge this.

albertz:
I just noticed that this does not cover what we are actually interested in. E.g. the file `better_exchook.py` is indented using 4 spaces (this is also configured in `.editorconfig` and PyCharm correctly uses those settings). However, this change now uses `--indent-size=2` for all files including `better_exchook.py`. But then you would expect that it should throw a lot of warnings for this file but it does not! It only prints those two warnings:
```
returnn/util/better_exchook.py:1338:13: E127 continuation line over-indented for visual indent
returnn/util/better_exchook.py:1339:13: E127 continuation line over-indented for visual indent
```
So this means, `pycodestyle` still does not cover what we are really interested in.

So, now the question is, is this change then really worth it? It anyway does not cover what we want, i.e. the check for correct indentation.

JackTemaki:
> I just noticed that this does not cover what we are actually interested in. E.g. the file `better_exchook.py` is indented using 4 spaces (this is also configured in `.editorconfig` and PyCharm correctly uses those settings). However, this change now uses `--indent-size=2` for all files including `better_exchook.py`. But then you would expect that it should throw a lot of warnings for this file but it does not! It only prints those two warnings:
> 
> ```
> returnn/util/better_exchook.py:1338:13: E127 continuation line over-indented for visual indent
> returnn/util/better_exchook.py:1339:13: E127 continuation line over-indented for visual indent
> ```
> 
> So this means, `pycodestyle` still does not cover what we are really interested in.
> 
> So, now the question is, is this change then really worth it? It anyway does not cover what we want, i.e. the check for correct indentation.

It correctly pointed out a non 2-space indentation in another file though. So maybe it is considering it correctly, or it notices the "majority" in a file.

JackTemaki:
> > It correctly pointed out a non 2-space indentation in another file though. So maybe it is considering it correctly, or it notices the "majority" in a file.
> 
> Can you give an example?

https://github.com/rwth-i6/returnn/blob/master/returnn/sprint/interface.py#L697

albertz:
> > > It correctly pointed out a non 2-space indentation in another file though. So maybe it is considering it correctly, or it notices the "majority" in a file.
> > 
> > 
> > Can you give an example?
> 
> https://github.com/rwth-i6/returnn/blob/master/returnn/sprint/interface.py#L697

The warning is:

> indentation is not a multiple of 2

But this is not the only thing we are interested in. In #930, with indent by 4, this is a multiple of 2, so this would not have covered that.

We specifically want exactly an indent by 2 after `def`, `if`, `while`, `class` keywords.

albertz:
> > The warning is:
> > > indentation is not a multiple of 2
> > 
> > But this is not the only thing we are interested in. In #930, with indent by 4, this is a multiple of 2, so this would not have covered that.
> > We specifically want exactly an indent by 2 after `def`, `if`, `while`, `class` keywords.
> 
> Ah right, no then we can not test for this. Because in the default also 8 (multiple of 4) is allowed in some cases. I still think we can keep this here to have some more checks, but in the end we are working with a non-standard code format so I guess we just have to live with looking for this manually.

I think it's not too hard to write a custom script checking just for that. However, this is not really my priority currently. Also, I tend to detect such inconsistencies easily in the review.

Yes, maybe we can anyway go forward in fixing the remaining warnings and merge this in here anyway. I'm not sure it will cause some annoyances though because this can produce warnings which people would not see in PyCharm during development.

Also, see my recent commits here what I changed. For some of these changes, I'm not really sure if they are so helpful or clear.

JackTemaki:
Hmm ok, then it is your call if we leave this or not. It seems that it is not helpful in the way that we expected.

albertz:
Maybe let's merge this now and just collect a bit experience. I think it might still reasonable.

PR #935, opened 2022-02-04
LengthLayer: copy dim.dyn_size_ext before using as output

patrick-wilken:
I had the really peculiar issue that enabling `debug_print_layer_output_shape` would change the time dimension tag of my decoder 🙆‍♂️

That happened when using a LengthLayer (`"target_length": {"class": "length", "from": "data:classes"}`).

When `debug_print_layer_output_shape` the layer output placeholder is reset via:
```
layer.output.placeholder = py_print(
          layer.output.placeholder,
          ...)
```

But in case of LengthLayer, `layer.output is Dim.dyn_size_ext` and thus the placeholder can have an attribute `_is_size_of_dim_tag`. This attribute however does not survive the `tf.identity()` in `py_print()`. And in the end
```
        time_dim_tag = Dim.get_tag_from_size_tensor(fixed_seq_len)
        assert time_dim_tag == self.time_dim_tag
```
in `_SubnetworkRecCell.get_ouput()` failed because `time_dim_tag is None`.

Making `layer.output` a copy of `dyn_size_ext` solves the issue for me, but I'm not sure whether identity is needed for other reasons...

In general, adding new attributes to `tf.Tensor` is bad style in my opinion and will likely lead to more bugs like this.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1959,7 +1959,7 @@ class LengthLayer(LayerBase):
         name="%s_length" % name,
         shape=(), batch_dim_axis=0, time_dim_axis=None,
         dtype=dtype, sparse=sparse, dim=None if sparse else NotSpecified)
-    return dim.dyn_size_ext
+    return dim.dyn_size_ext.copy()
 
 
 class SoftmaxOverSpatialLayer(_ConcatInputLayer):


Zettelkasten:
> Making `layer.output` a copy of `dyn_size_ext` solves the issue for me

Why does this fix anything? This doesn't change the placeholder, so also `_is_size_of_dim_tag` should be just as before, what is different from before?

I agree that `_is_size_of_dim_tag` is bad. In general, the logic in `Dim.get_tag_from_size_placeholder` is outdated and from the days before we had proper dim tag support. Back then, dim tags were basically just the same as the tensor of the dyn sizes they represented.
Now this changed and we need to think about how dim tags should actually work. Especially how the conversion from `Data` representing some seq lengths to `Dim` should work.
Perhaps instead of attaching `_is_size_of_dim_tag` to the tensor, we can attach it to the `Data` instance and then handle it properly when the `Data` is copied. This all then would work completely independent of the `placeholder`, even when everything is a template.
I remember @albertz and I had some discussion somewhere about whether attributes should be part of `Data`, the `LayerBase`, or the `Tensor` before. But I can't find it anymore.

patrick-wilken:
> Why does this fix anything?

See the code related to `debug_print_layer_output_shape`. It re-assigns the placeholder of the layer output `Data` object. If this `Data` object is just used locally for this layer this might be okay. But if the `Data` object `is` (!) also a `dyn_size_ext` of some `Dim` this must not be allowed. Because, as far as I understand, `Dim` instances are somewhat global objects, they are shared and passed around, so changing them will have side effects everywhere. If we create a copy here, no `dyn_size_ext` is changed, only the placeholder attribute (not the placeholder itself) of a copy.

Zettelkasten:
> > Why does this fix anything?
> 
> See the code related to `debug_print_layer_output_shape`. It re-assigns the placeholder of the layer output `Data` object. If this `Data` object is just used locally for this layer this might be okay. But if the `Data` object `is` (!) also a `dyn_size_ext` of some `Dim` this must not be allowed. Because, as far as I understand, `Dim` instances are somewhat global objects, they are shared and passed around, so changing them will have side effects everywhere. If we create a copy here, no `dyn_size_ext` is changed, only the placeholder attribute (not the placeholder itself) of a copy.

Oh I see. So we could also fix this in the `debug_print_layer_output_shape` logic by simply making a copy of the output `Data` there before assigning the placeholder? Maybe that's more robust, I don't know whether other layers might also return some `dyn_size_ext` on accident.

The same problem also occurs with `debug_runtime_sanity_checks` and `debug_add_check_numerics_on_output` btw, who change the placeholder in a similar way.

Zettelkasten (approved):

albertz:
The change itself should be fine in any case. A `Data.copy` should always be fine.

Additionally, what @Zettelkasten said, the logic like `debug_print_layer_output_shape` and co should maybe also additionally create a copy of the `Data` object.

Additionally, whenever we introduce some identity (`tf.identity`, `tf.print` or whatever) of some placeholder which corresponds to a dim tag, to make sure some of the legacy code which uses `Dim.get_tag_from_size_tensor` still works, we must call `Dim.set_tag_on_size_tensor` accordingly.

Maybe those two additionally things can be in separate PRs though.

albertz (approved):

Issue #634, opened 2021-09-06
Definition of equality of dim tags


albertz:
I just want to raise the issue that the equality is not really well defined in all cases.

We implement `__eq__` and related functions. Which are somewhat restrictive currently, but less restrictive than `object.__eq__` (which is only True when `a is b`, i.e. identity).

We also have `is_equal`. `__eq__` reduces to `is_equal` with default options. The default options are the most restrictive equality. But the options allow to make it more loosely. E.g. such that a broadcast dim (dim = 1) can match to another dim. Or that static dims with the same static dim would match always.

There is also `same_as`, which is considered as always equal (recursively, so up to `same_base`).

The most restrictive variant currently falls back to the description when everything else (kind, dim) matches. Although I think this is mostly outdated and probably should be removed or deprecated. I'm not sure if this really makes sense. It does not check for the dynamic length tensor (`dyn_size` or `dyn_size_ext`) or identity.

Regarding dyn size, there is also the aspect of search beams or other different kind of batches. We consider dim tags as equal even when they belong to different batches / beams. So identity on a `dyn_size` tensor would also be wrong, and we have more loose equality. Also, we can have even the opposite, that two different dim tags make use of the same dyn size (at least in principle; while we still have `get_tag_from_size_tensor`, this is not really the case, but this will be resolved eventually).

Maybe the current logic is already all fine.

This might become more relevant when we rely more heavily on dim tags and their equality. E.g. for checking which axes should be shared (in `GatherLayer`, also see discussion in #625), or for a single reduce argument for `DotLayer`, or once we enforce unique dim tags (#632).

Issue #865, opened 2021-12-16
Dim tags matching should use `allow_same_spatial_dim=False`


Zettelkasten:
This is kind of an follow up to #666.

I have the problem currently, that `SpatialDim("a", 1)` matches `SpatialDim("b", 1)` in many cases, e.g. `get_common_data`, `copy_compatible_to` and `find_matching_dim_map` used by `GatherLayer`/`DotLayer`/etc.

Looking at the `Dim.is_equal` code, `allow_same_spatial_dim` matches exactly if
1. both tags are spatial dims (or feature if `treat_feature_as_spatial` which e.g. `get_common_data` sets)
2. the dimensions of `self` and `other` are equal, but not `None`

For my case, I definitely do not want these axes to match (they are not related in any way).

But this kind of matching is currently necessary for something like this:
```
      "lin1": {"class": "linear", "activation": "sigmoid", "n_out": 5, "from": "data:data"},
      "lin2": {"class": "linear", "activation": "sigmoid", "n_out": 5, "from": "data:data"},
      "output": {"class": "combine", "mode": "add", "from": ["lin1", "lin2"]},  # would give [B,T,F1,F2] with allow_same_spatial_dim=False!!
```

This is also because currently, `LinearLayer` does not set `derived_from_tag` on its output feature dim.
Then, `derived_matches` would catch this.
We can fix `LinearLayer` and to do this properly (currently this is via `_base_get_out_data_from_opts`, when creating a feature dim tag, I think it should just derive it from its input feature dim tag).
For user specified `out_dim`s however, we would not automatically set derived from to the input dim tag.

For other cases, this is more difficult maybe.
E.g. if you have two inputs with different time axes, and then split into the same dim size.
Then, the user would maybe want these axes to match, but they are in no way derived from the same thing.

Another solution is, similar to what @albertz wrote in #666, to separate user defined dim tags and automatically created ones. Then, user defined ones could have a more restrictive matching logic.

albertz:
`derived_from_tag`/`derived_matches` is not the right option here. `derived_from_tag` means that it is derived from this tag (via some op, like +- or so). We use it as a heuristic to assume in the rec layer, when some dim is derived (but unknown yet due to template construction, also this was before the dim math), that it is the same as the orig dim. E.g. we had the case where some people did some extra padding, then convolution with `padding="valid"` which removed the padding again such that it became the same, and that inside a rec layer, on the attention encoder axis.

[...]

PR #950, opened 2022-02-15
Dim auto_generated flag

albertz:
Allows for better Dim is_equal which does not rely on the description.

#634


Zettelkasten:
Looks good!
I wonder about #865: here we could also say that `same_spatial_matches` only matches two spatial dims if at least one is auto generated. Or maybe more general even, that two user defined dims only match when their identities are the same.
That would break transitivity though (e.g. an auto generated dims matches two different user defined dims, which themselves would not match). Not sure what the impact of that would be. This would be something for another PR.

PR #953, opened 2022-02-16
Add docs for debug_print_layer_output

mmz33:


albertz (approved):
Note that it can also be a dict with options.

PR #954, opened 2022-02-17
fix SyntaxWarning

vieting:
When running RETURNN, I get
```
/.../returnn/returnn/tf/network.py:1621: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if loss is 0:
/.../returnn/returnn/tf/util/basic.py:2024: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if dim is not 1:
/.../returnn/returnn/tf/util/basic.py:5550: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if start is 0 and stop is None:
/.../returnn/returnn/tf/layers/base.py:1383: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if c is 0:
```

I don't see a reason to do it the way it is currently done, so we could just fix it like proposed here.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -1380,7 +1380,7 @@ class LayerBase(object):
       c += self.spatial_smoothing * self.get_output_spatial_smoothing_energy()
     if self.darc1:
       c += self.darc1 * self.get_darc1()
-    if c is 0:
+    if c == 0:
       return None
     return c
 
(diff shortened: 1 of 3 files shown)


returnn/tf/layers/base.py
```diff
@@ -1380,7 +1380,7 @@ def get_constraints_value(self):
       c += self.spatial_smoothing * self.get_output_spatial_smoothing_energy()
     if self.darc1:
       c += self.darc1 * self.get_darc1()
-    if c is 0:
+    if c == 0:
```
albertz, line 1383: This is wrong in case `c` is a `tf.Tensor` or `numpy.ndarray`.

Same also for all the other changes.

PR #961, opened 2022-02-19
TF2, control flow v2 support

### LINKED ISSUE #962: `InvalidArgumentError: Retval[0] does not have value` in loss computation
I am getting a strange error related to loss computation i think. This also happen when computing loss scores on dev.

Example:
- Error with CTC loss: https://gist.github.com/mmz33/b40fb4099790f0048ece9553c1fab6ea
- Error with CL loss: https://gist.github.com/mmz33/96cb48cc034c1221add2aa4c542c6019

Config with CTC loss: https://gist.github.com/mmz33/17ff9794a98ab0fee02d341e48411cb1

Error:
```
InvalidArgumentError: Retval[0] does not have value
```
Looks related to this: https://github.com/tensorflow/tensorflow/issues/32106

If I checkout RETURNN with older version before this commit below, the issue does not occur:

```
commit e819b63be89fd6dae77e7d40950a8d990031b6a3
Author: Albert Zeyer <albzey@gmail.com>
Date:   Fri Feb 11 11:16:48 2022 +0100

    flatten_with_seq_len_mask cache (#944)
```

albertz:
This is an updated version of #700.

Might fix #962.

Related:

https://github.com/tensorflow/tensorflow/issues/32106

### DIFF
--- a/returnn/tf/compat.py
+++ b/returnn/tf/compat.py
@@ -11,6 +11,7 @@ https://github.com/tensorflow/lingvo/blob/master/lingvo/compat.py
 """
 
 import tensorflow as tf
+import os as _os
 
 if not getattr(tf, "compat", None) or not getattr(tf.compat, "v2", None):
   v1 = tf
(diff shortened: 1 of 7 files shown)


albertz:
Now I'm stumbling over this issue: https://github.com/tensorflow/tensorflow/issues/54458

albertz:
As there are still issues, maybe even in TF itself, and not sure if all of them can be resolved, and also not sure if this is actually better or faster than before, I think we better should make this optional for now. Maybe check `TF_ENABLE_COND_V2` and if this not set explicitly, disable control flow v2 by default for now.

albertz:
Note, the current issues are like:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Node 'optimize/gradients/output/rec/while_grad/output/rec/while_grad': Connecting to invalid output 122 of source node output/rec/while which has 122 outputs. Try using tf.compat.v1.experimental.output_all_intermediates(True).
```
We have `output_all_intermediates(True)` here already although that does not fix it and also not sure if this has other drawbacks.

This is basically: https://github.com/tensorflow/tensorflow/issues/54458

albertz:
This is still disabled by default, and can enabled via env `TF_ENABLE_COND_V2` (or via code).

So this can anyway be merged now as intermediate progress.

PR #963, opened 2022-02-20
flatten_with_seq_len_mask, fix cache and control flow context

### LINKED ISSUE #962: `InvalidArgumentError: Retval[0] does not have value` in loss computation

albertz:
Fix #962

Fix Retval does not have value


albertz:
@mmz33 I'm trying to reproduce the error you get in #962. For me locally, I don't get the error so far in the test case I added here. Maybe you can try to run this test? And maybe edit the test such that you get the error?

albertz:
Ok, now the test runs again, after copying the config from https://github.com/rwth-i6/returnn/issues/962 and then needing the fixes from #965 and #966, however, I still don't get the error. And also the error should be simplified now and make use of much smaller dimensions.

albertz:
Ok, the test don't work on TF 1 but this is because of batch norm:
```
TypeError: fused_batch_norm() got an unexpected keyword argument 'exponential_avg_factor'
```

PR #972, opened 2022-02-23
fix ResizeLayer for axis != 1

vieting:
It seems like there is a bug in the `ResizeLayer` if `axis` is not resolved to 1. 

In general, the code does not seem very flexible, so I added another assertion that the `batch_ndim` has to be 3 (the shape is extended by the `'x'` in  `dimshuffle` and then image shapes of `(B,H,W,C)` are assumed).

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -7261,6 +7261,7 @@ class ResizeLayer(_ConcatInputLayer):
     # self.output.shape and self.output.batch_dim_axis are already set here via self.get_out_data_from_opts().
     axis = self.output.get_axis_from_description(axis)
     assert axis > 0, "batch-dim resize not supported"
+    assert self.input_data.batch_ndim == 3, "not implemented otherwise"
     input_data = self.input_data.copy_as_batch_major()
     self.output.placeholder = input_data.placeholder
     out_dyn_size = input_data.dim_tags[axis].dyn_size
@@ -7320,7 +7321,7 @@ class ResizeLayer(_ConcatInputLayer):
         out_dyn_size = tf.reduce_sum(tf.cast(tf.logical_and(mask, orig_mask), tf.int32), axis=1)
     if axis != 1:
       perm = [0] + remaining_axes
-      perm.insert(axis, 1)
+      perm.insert(1, axis)
       x = tf.transpose(x, perm)
     if out_dyn_size is not None:
       self.output.dim_tags[axis].dyn_size = out_dyn_size
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -4495,6 +4495,27 @@ def test_ResizeLayer_fill_dropout():
       assert_equal([s for s in out[i] if s != fill_value], src_seqs[i])
 
 
+def test_ResizeLayer_BFT():
+  n_batch, n_time, n_in = 7, 3, 5
+  config = Config({
+    "extern_data": {
+      "data": {"dim": n_in, "shape": (n_in, None)}  # [B,F,T], i.e. batch-feature-major
+    }
+  })
+  with make_scope() as session:
+    net = TFNetwork(config=config)
+    assert net.extern_data.get_default_input_data().is_batch_feature_major
+    net.construct_from_dict({
+      "output": {"class": "resize", "axis": "T", "factor": 2, "kind": "nn", "from": "data"}
+    })
+    out = net.get_default_output_layer().output.copy_as_batch_feature_major()
+    in_v = numpy.arange(0, n_batch * n_time * n_in).astype("float32").reshape((n_batch, n_in, n_time))
+    session.run(tf_compat.v1.global_variables_initializer())
+    out_v = session.run(out.placeholder, feed_dict={net.extern_data.data["data"].placeholder: in_v})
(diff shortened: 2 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -7320,7 +7321,7 @@ def __init__(self, factor, axis, out_dim=None, kind="nn", fill_value=None, fill_
         out_dyn_size = tf.reduce_sum(tf.cast(tf.logical_and(mask, orig_mask), tf.int32), axis=1)
     if axis != 1:
       perm = [0] + remaining_axes
-      perm.insert(axis, 1)
+      perm.insert(1, axis)
```
albertz, line 7324: Right.

But then we can simplify this directly to:
```
x = tf.transpose(x, [0, axis] + remaining_axes)
```

returnn/tf/layers/basic.py
```diff
@@ -7261,6 +7261,7 @@ def __init__(self, factor, axis, out_dim=None, kind="nn", fill_value=None, fill_
     # self.output.shape and self.output.batch_dim_axis are already set here via self.get_out_data_from_opts().
     axis = self.output.get_axis_from_description(axis)
     assert axis > 0, "batch-dim resize not supported"
+    assert self.input_data.batch_ndim == 3, "not implemented otherwise"
```
albertz, line 7264: Why?

returnn/tf/layers/basic.py
albertz: But why actually? This `transpose` should revert the previous transpose above. The old code looks correct to me. Can you give an example?

albertz:
I think it actually makes more sense that we directly output as we reordered the dims.

This means, in `get_out_data_from_opts`, you move batch dim to front, and then the axis to the second.

returnn/tf/layers/basic.py
vieting: You can see an example in the test case that I added. The input is `(B,F,T)` and `T` is resized.

albertz: Ah right, but the wrong part here is `remaining_axes` and not this `insert`.
But anyway, see my suggestion, then we don't need the `transpose` at all.

vieting:
> Then in __init__ the same, and then we simply need an expand_dims and later a squeeze and no final transpose anymore.

You mean removing the `dimshuffle` and `tf.reshape` part and just `expand_dims` instead?

returnn/tf/layers/basic.py
```diff
@@ -7259,10 +7259,11 @@ def __init__(self, factor, axis, out_dim=None, kind="nn", fill_value=None, fill_
     out_dim  # noqa  # via get_out_data_from_opts
     super(ResizeLayer, self).__init__(**kwargs)
     # self.output.shape and self.output.batch_dim_axis are already set here via self.get_out_data_from_opts().
-    axis = self.output.get_axis_from_description(axis)
-    assert axis > 0, "batch-dim resize not supported"
-    assert self.input_data.batch_ndim == 3, "not implemented otherwise"
-    input_data = self.input_data.copy_as_batch_major()
+    axis_ = self.input_data.get_axis_from_description(axis)
```
albertz, line 7262: Wrong order. First you should do `input_data = self.input_data.copy_as_batch_major()`, then you can call `get_axis_from_description`.

returnn/tf/layers/basic.py
```diff
@@ -7338,8 +7335,11 @@ def get_out_data_from_opts(cls, factor, axis, sources, name, out_dim=None, **kwa
     :rtype: Data
     """
     out = get_concat_sources_data_template(sources).copy_as_batch_major()
+    axis_ = out.get_axis_from_description(axis)
+    out = out.copy_move_axis(old_axis=axis_, new_axis=1)
     out = out.copy_template(name="%s_output" % name)
     axis = out.get_axis_from_description(axis)
+    assert axis == 1
```
albertz, line 7342: The second call to `get_axis_from_description` is not needed, and you can just assign `axis = 1` here.

returnn/tf/layers/basic.py
```diff
@@ -7259,10 +7259,11 @@ def __init__(self, factor, axis, out_dim=None, kind="nn", fill_value=None, fill_
...
     # self.output.shape and self.output.batch_dim_axis are already set here via self.get_out_data_from_opts().
-    axis = self.output.get_axis_from_description(axis)
-    assert axis > 0, "batch-dim resize not supported"
-    assert self.input_data.batch_ndim == 3, "not implemented otherwise"
-    input_data = self.input_data.copy_as_batch_major()
+    axis_ = self.input_data.get_axis_from_description(axis)
+    assert axis_ > 0, "batch-dim resize not supported"
+    input_data = self.input_data.copy_as_batch_major().copy_move_axis(old_axis=axis_, new_axis=1)
+    axis = input_data.get_axis_from_description(axis)
+    assert axis == 1
```
albertz, line 7266: The second call to `get_axis_from_description` is not needed, and you can just assign `axis = 1` here.
```suggestion
    axis = 1
```

vieting:
> I think the `tf.reshape` is needed anyway in case batch_ndim > 3.

So should we just leave the rest as is?

albertz (approved):

PR #973, opened 2022-02-28
add support for librosa v0.9.1

### LINKED ISSUE #969: support new librosa version 0.9.1
Setting up a new environment I noticed that the latest version of librosa is not working currently with returnn.

One example (I did not check all functions) would be 
https://github.com/rwth-i6/returnn/blob/a86ba33e1ade77c4a4994b9b8a7cca529128d1a3/returnn/datasets/util/feature_extraction.py#L309-L312
where `audio` is given as a positional argument, but `librosa.feature.mfcc` expects keyword-only arguments now:
https://librosa.org/doc/0.9.1/generated/librosa.feature.mfcc.html

Alternatively I would also be happy if you add the min/max supported version to `requirements.txt`.

michelwi:
Fix #969.

as discussed in #969 here is a pr to make returnn compatible with the latest version of librosa.

All changes _should be_ backwards compatible based on the librosa documentation.

### DIFF
--- a/returnn/datasets/util/feature_extraction.py
+++ b/returnn/datasets/util/feature_extraction.py
@@ -283,7 +283,12 @@ def _get_audio_linear_spectrogram(audio, sample_rate, window_len=0.025, step_len
   assert num_feature_filters*2 >= min_n_fft
   assert num_feature_filters % 2 == 0
 
-  spectrogram = numpy.abs(librosa.core.stft(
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):
`librosa.core.stft` still works, but agree it is better to update this, as `librosa.stft` is the official API entry.

albertz (approved):

PR #976, opened 2022-03-04
Fix MaskedComputationLayer for RETURNN principles

### LINKED ISSUE #769: `MaskedComputationLayer` is violating the principle that the user should not need to think about rec automatic optimization
See [RETURNN principles](https://github.com/rwth-i6/returnn/wiki/RETURNN-principles).

When defining `MaskedComputationLayer` inside a rec loop and not thinking about rec automatic optimization, the behavior is clear. It copies the prev output (and prev state) whenever the mask is False, otherwise it uses the current output.

In this line of thought, there should be no need for `UnmaskLayer`.

However, the current situation is, when it is optimized out, a tensor of shape [B,D] inside would not just become [T,B,D] where T is the number of rec iterations, but instead it becomes [T',B,D] where T' = sum(mask). And some follow-up operation at some point requires the [T,B,D] shape (maybe because it combines it with other tensors from the rec loop), and thus the user explicitly must use the `UnmaskLayer` at some point after the `MaskedComputationLayer`.

This violates the RETURNN principle that the user should not need to think about rec automatic optimization.

---

What are possible solutions?

I don't really have a good idea so far. It should be opaque.

Maybe the new dim tag for T' could have some special flag that it is a masked subset of T, and whenever it is going to be combined with the other dim tag, it would automatically do the unmasking.

albertz:
Fix #769. Specifically, it fixes the problem that the behavior when optimized out of a rec loop is not consistent and opaque to the user.

The `masked_from` option still is an exception though.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3126,7 +3126,16 @@ class CumsumLayer(_ConcatInputLayer):
       self.output.placeholder = next_state
     else:
       axis = data.get_axis_from_description(axis)
-      self.output.placeholder = tf.cumsum(x, axis=axis, reverse=reverse)
+      y = tf.cumsum(x, axis=axis, reverse=reverse)
+      if self._initial_output is not None:
+        init_state = self.get_rec_initial_output(
+          name=self.name, output=self.output,
+          network=self.network, batch_dim=self.get_batch_dim(),
+          rec_layer=self.network.get_rec_parent_layer(inside_loop=False),  # should actually not matter
+          axis=axis, sources=self.sources,
+          initial_output=self._initial_output)
+        y = init_state + y
+      self.output.placeholder = y
     self.output.placeholder.set_shape(data.placeholder.get_shape())
     self.output.placeholder.set_shape(tf.TensorShape(self.output.batch_shape))
     self.output.size_placeholder = self.input_data.size_placeholder.copy()
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -3157,7 +3157,9 @@ class CumsumLayer(_ConcatInputLayer):
     from returnn.tf.util.basic import is_axis_from_description_recurrent
     if is_axis_from_description_recurrent(axis=axis, network=network, data=data):
       assert all(data.shape)
-      return {"state": tf.zeros(data.get_batch_shape(batch_dim=batch_dim), dtype=data.dtype)}
+      init_state = cls.get_rec_initial_output(
+        network=network, batch_dim=batch_dim, rec_layer=rec_layer, axis=axis, sources=sources, **kwargs)
+      return {"state": init_state}
     return {}
 
 
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -7578,6 +7578,7 @@ class MaskedComputationLayer(LayerBase):
 
   def __init__(self, mask, unit, masked_from,
                _layer_class, _layer_desc,
+               in_spatial_dim=None,
                out_spatial_dim=None,
                _parent_layer_cache=None,
                **kwargs):
@@ -7585,12 +7586,12 @@ class MaskedComputationLayer(LayerBase):
     :param LayerBase|None mask:
     :param dict[str] unit:
     :param LayerBase|None masked_from:
-    :param Dim|None out_spatial_dim:
+    :param Dim|None in_spatial_dim:
+    :param Dim|None out_spatial_dim: the masked dim
     :param type[LayerBase] _layer_class:
     :param dict[str] _layer_desc:
     :param dict[str,LayerBase]|None _parent_layer_cache:
     """
-    out_spatial_dim  # noqa  # handled in transform_config_dict
     from returnn.tf.network import get_layer_class
     from .base import WrappedInternalLayer
     from returnn.tf.util.basic import where_bc, get_shape, nd_indices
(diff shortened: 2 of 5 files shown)


albertz:
Remaining now is this TODO. Specifically, how to get the initial values in `Data.copy_replace_dim_tag`.

In `UnmaskLayer`, we use:
```python
rec_parent_layer = self.network.get_rec_parent_layer(inside_loop=False)
assert rec_parent_layer and isinstance(rec_parent_layer.cell, _SubnetworkRecCell)
initial = rec_parent_layer.cell._get_init_output(src_layer.name, batch_dim=batch_dim)  # [B,D']
```
However, this is already ugly, does not work in all cases (uses `src_layer.name` here, problematic for sub networks), and in our case, we would not even know `src_layer` in `Data.copy_replace_dim_tag`.

Attaching the layer to the `Data` instance came up sometime before already. I think this was discarded because any `Data.copy` and related would either drop this or it would not be handled correctly afterwards.

There was the other idea to attach it to the tensor. E.g. in `TFNetwork._create_layer`. Although a tensor can be the output of multiple layers (e.g. `CopyLayer`), so this is ambiguous. But maybe the first layer is fine.

albertz:
Extending the test case `test_MaskedComputationLayer_in_loop_auto_unmask` to cover the TODO (custom initial state/output) is actually not so simple.

The test currently uses `SimpleCumSumCell` (from the test cases). We could alternatively also use `CumsumLayer`. However, both have the issue that handling an initial state (`initial_state`) is only partially supported. And that is not really enough, it also needs `initial_output`. Although `initial_output` and `initial_state` should mean the same thing here. But effectively they do not. But this is somewhat out-of-the-scope of this test case.

Maybe we should use a different layer or rec unit for the test. But which? I think LSTM has a similar problem for `initial_state` vs `initial_output`.

Manual explicit logic using `prev:...` is not really supported in the `MaskedComputationLayer` (yet?).

albertz:
So, the remaining problem seem to be how to store and get the initial output when we do the automatic unmasking.

Another idea: Why do we actually use the masked_dim = sum(mask) when this is inside a rec loop? When outside a rec loop, and when the masking is actually explicit, this is the expected outcome. When inside a rec loop, and optimized out, this is all internal, and never should matter. So instead of masked_dim = sum(mask), we could do masked_dim = sum(mask) + 1, i.e. just extend the dim by one, and store the initial output in the first frame. Would this work?

One question would be, where do we do this extension? Already for the masked inputs? No, because any sub layer(s) might have own custom `initial_output` configuration, and this would not correctly be used then. So it should be done after the whole `MaskedComputationLayer` is done. But it should be done for all sub layers as well.

It is basically doing the logic of `UnmaskLayer` then:
```python
initial_wt = tf.expand_dims(initial, axis=0)  # add time axis
src_t = concat_with_opt_broadcast(
  [initial_wt, src.placeholder], allow_broadcast=[True, False], axis=0, name="concat_in_time")  # [T'+1,B,D']
```

albertz:
What actually would be the expected behavior for this inside a rec layer:
```
x_masked = masked_comp(x, mask=mask)
y = lstm(x_masked)
```
Well, the expected behavior is mostly clear for the case when it is inside a rec loop and not optimized.
And this should be the true behavior in all cases, per [RETURNN principles](https://github.com/rwth-i6/returnn/wiki/RETURNN-principles) because the user should not need to think about the automatic optimization (#769).

Currently, however, when this is optimized, the result would be wrong (or not match this expected behavior) because `lstm` would operated on the masked output and not the unmasked. (@robin-p-schmitt btw, do you maybe have this case?)

It would also still be wrong with all the new changes here because the automatic unmasking as implemented here happens only in `copy_compatible_to`.

The question is, what follow-up operations would we usually use actually?

In case of a transducer, it is common to have a joint network on top. We should actually carefully check current setups. I think current setups might actually really expect the current behavior, and not the "expected" behavior.

robin-p-schmitt:
> @robin-p-schmitt btw, do you maybe have this case?

Do you mean if this causes the tests to fail in my current PR regarding the MaskedComputationLayer?

albertz:
> > @robin-p-schmitt btw, do you maybe have this case?
> 
> Do you mean if this causes the tests to fail in my current PR regarding the MaskedComputationLayer?

No, I mean whether you have the case I was describing.

robin-p-schmitt:
> > > @robin-p-schmitt btw, do you maybe have this case?
> > 
> > 
> > Do you mean if this causes the tests to fail in my current PR regarding the MaskedComputationLayer?
> 
> No, I mean whether you have the case I was describing.

No, in my case I used a layer from outside the `MaskedComputationLayer` with `base:` and this layer also had a time dimension and therefore it was automatically masked. The problem was that I didn't notice that it was masked and therefore my network did something different than what I expected.

albertz:
The implication from my previous post is basically:

- If `masked_from` is set, do the current logic. I think the documentation needs to be improved.
- Else, if inside rec loop: Just as canonical, as we do now.
- Else, if inside rec layer but optimized out: unmask right away. This is a change from before. But this is required for any other recurrent follow-up layers to behave correctly.

It also means, this PR here changes a bit. We don't need such automatic broadcasting logic in `copy_compatible_to`.

albertz:
> > > > @robin-p-schmitt btw, do you maybe have this case?
> 
> No, in my case I used a layer from outside the `MaskedComputationLayer` with `base:` and this layer also had a time dimension and therefore it was automatically masked. The problem was that I didn't notice that it was masked and therefore my network did something different than what I expected.

But this is not what I mean, or what I described. I am referring to the case that I described. I.e. it is all about other layers after (not within) the `MaskedComputationLayer`. It is not about `MaskedComputationLayer`. The question was whether there are any other layers afterwards, and how they behave, and whether their behavior was correct. I remember you said that you have other layers afterwards, even other `MaskedComputationLayer`s.

albertz:
> ``



> > But this is not what I mean, or what I described. I am referring to the case that I described. I.e. it is all about other layers after (not within) the `MaskedComputationLayer`. It is not about `MaskedComputationLayer`. The question was whether there are any other layers afterwards, and how they behave, and whether their behavior was correct. I remember you said that you have other layers afterwards, even other `MaskedComputationLayer`s.
> 
> I didn't have any problems with any layers outside the `MaskedComputationLayer`

This is also not what I meant. I meant whether the behavior is correct. You will not see any problems or errors. It would produce sth in any case, without error. I'm speaking about the logical behavior. You cannot really tell about that by looking for problems. You need to think about whether your config matches the case that I described. In that case, there are no directly observable problems but it might just behave wrongly.

robin-p-schmitt:
> This is also not what I meant. I meant whether the behavior is correct. You will not see any problems or errors. It would produce sth in any case, without error. I'm speaking about the logical behavior. You cannot really tell about that by looking for problems. You need to think about whether your config matches the case that I described. In that case, there are no directly observable problems but it might just behave wrongly.

So I actually never have this case:
```
x_masked = masked_comp(x, mask=mask)
y = lstm(x_masked)
```
I always have smth like:
```
x_masked = masked_comp(x, mask=mask)
x_unmasked = unmask(x_masked, mask=mask)
y = lstm(x_unmasked)
```
In this case, there shouldn't be any unexpected behavior, right?

PR #981, opened 2022-03-07
HDFDataset: add get_data_by_seq_tag

vieting:
For a use case that I have in mind, I need to be able to get data from an `HDFDataset` based on the sequence tag. I added `HDFDataset.get_data_by_seq_tag()` for that purpose.

It is very similar to `HDFDataset.get_data()` which has a special case `if self.cache_byte_size_total_limit > 0`. I'm not familiar with that and not sure how to handle this here, that's why I mark this PR as draft.

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -48,6 +48,7 @@ class HDFDataset(CachedDataset):
     self.data_dtype = {}  # type: typing.Dict[str,str]
     self.data_sparse = {}  # type: typing.Dict[str,bool]
     self._num_codesteps = None  # type: typing.Optional[typing.List[int]]  # accumulated sequence length per target
+    self._tag_to_real_idx_map = None  # type: typing.Optional[typing.Dict[str,int]]
 
     if files:
       for fn in files:
@@ -273,6 +274,27 @@ class HDFDataset(CachedDataset):
 
     # Otherwise, directly read it from file now.
     real_seq_idx = self._seq_index[seq_idx]
+    return self._get_data_by_real_seq_idx(real_seq_idx, key)
+
+  def get_data_by_seq_tag(self, seq_tag, key):
+    """
+    :param str seq_tag:
+    :param str key:
+    :rtype: numpy.ndarray
+    """
+    if self.cache_byte_size_total_limit > 0:  # Use the cache?
+      assert False  # what should we do here?
+
+    # Otherwise, directly read it from file now.
+    real_seq_idx = self._get_real_idx_by_tag(seq_tag)
+    return self._get_data_by_real_seq_idx(real_seq_idx, key)
+
+  def _get_data_by_real_seq_idx(self, real_seq_idx, key):
+    """
+    :param int real_seq_idx:
+    :param str key:
+    :rtype: numpy.ndarray
+    """
     file_idx = self._get_file_index(real_seq_idx)
     fin = self.h5_files[file_idx]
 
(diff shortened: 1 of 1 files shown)


returnn/datasets/hdf.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, files=None, use_cache_manager=False, **kwargs):
     self.data_dtype = {}  # type: typing.Dict[str,str]
     self.data_sparse = {}  # type: typing.Dict[str,bool]
     self._num_codesteps = None  # type: typing.Optional[typing.List[int]]  # accumulated sequence length per target
+    self._tag_to_real_idx_map = None  # type: typing.Optional[typing.Dict[str,int]]
```
albertz, line 51: Why do you introduce that? You already have `_tag_idx` which is exactly the same?

returnn/datasets/hdf.py
```diff
@@ -355,6 +377,33 @@ def _get_tag_by_real_idx(self, real_seq_idx):
...
+    """
+    :param str seq_tag:
+    :rtype: int
+    """
+    if self._tag_to_real_idx_map is None:
+      self._build_tag_to_real_idx_map()
+
+    return self._tag_to_real_idx_map[seq_tag]
+
+  def _build_tag_to_real_idx_map(self):
```
albertz, line 390: You don't need that. Instead, there is already `_update_tag_idx`.

returnn/datasets/hdf.py
```diff
@@ -355,6 +377,33 @@ def _get_tag_by_real_idx(self, real_seq_idx):
     s = self._decode(s)
     return s
 
+  def _get_real_idx_by_tag(self, seq_tag):
```
albertz, line 380: I don't think such a helper function is needed. Just call `_update_tag_idx` and access `_tag_idx` directly.

returnn/datasets/hdf.py
```diff
@@ -273,6 +274,27 @@ def get_data(self, seq_idx, key):
...
+    return self._get_data_by_real_seq_idx(real_seq_idx, key)
+
+  def get_data_by_seq_tag(self, seq_tag, key):
+    """
+    :param str seq_tag:
+    :param str key:
+    :rtype: numpy.ndarray
+    """
+    if self.cache_byte_size_total_limit > 0:  # Use the cache?
+      assert False  # what should we do here?
```
albertz, line 286: Throw a better exception instead, like `raise Exception("%s: get_data_by_seq_tag not supported with cache" % self)` or so.

albertz (approved):
Looks ok to me. Maybe either @JackTemaki or @patrick-wilken should check as well.

JackTemaki (approved):

patrick-wilken:
Nothing wrong with the code. But if we in general want to support getting data via tag this maybe should become part of the Dataset base class at some point.
And just as a side note: I intentionally removed calls to `self._update_tag_idx()` in the past to only create `self._tag_idx` if really needed, because it can get large if the number of sequences is huge. But that does not seem to be the case for you.

albertz:
> Nothing wrong with the code. But if we in general want to support getting data via tag this maybe should become part of the Dataset base class at some point.

Yea, we thought about this. But basically almost every dataset would need a custom implementation, and this implementation can be non-straightforward, or even tricky (e.g. `ExternSprintDataset`), or also impossible.

Also, this is a function which would not be used by RETURNN itself but only by user code. In this case, it is used to implement a new custom dataset which wants random access by seq tag to another (HDF) dataset.

> And just as a side note: I intentionally removed calls to `self._update_tag_idx()` in the past to only create `self._tag_idx` if really needed

But we do the same here? It is only called when really needed.

PR #984, opened 2022-03-09
MaskedComputationLayer sub layers

albertz:


albertz:
This is not only about `MaskedComputationLayer` but about `SubnetworkLayer` and how sub layers are handled in general, specifically w.r.t. their hidden state (what should `get_rec_initial_extra_outputs` return), and maybe other logic. Related is also similar handling in `CondLayer` but we probably will not touch that here.

When should sub layers of `SubnetworkLayer` be created, esp those which are not used by the main `"output"` sub layer? Currently they are created dynamically and only when used by some other (maybe parent) layer, which is good.

There might be situations like in `CondLayer` when an access from outside is not easily possible, and we might want to create all in advance. Although for `MaskedComputationLayer`, this is maybe not needed. This is not clear yet. Maybe it is needed because `MaskedComputationLayer.get_rec_initial_extra_outputs` needs to return it for all layers.

The question is whether the layers are registered in the parent network. Then `MaskedComputationLayer.get_rec_initial_extra_outputs` should actually not return their state.

`MaskedComputationLayer.get_rec_initial_extra_outputs` actually needs more for those sub layers which are accessed outside. For those it also needs to store the last output, to be able to unmask. If we wrap all the sub layers and register them in the parent network, we don't need that in `MaskedComputationLayer.get_rec_initial_extra_outputs`.

We could maybe also implement `MaskedComputationLayer.cls_get_sub_network`. Or even we need to when sub layers should properly be registered in the parent net.

It's strange that `SubnetworkLayer.get_rec_initial_extra_outputs` returns all the sub layers state. Maybe this actually obsolete and duplicated. The `SubnetworkLayer` instance itself might not even be created when the layer is never accessed on its own and only its sub layers are used.

albertz:
@patrick-wilken Note that this PR also includes some changes to the `get_sub_layer_out_data_from_opts` API, and `_TemplateLayer.get_sub_layer`.

PR #994, opened 2022-03-15
add StftLayer

vieting:
I added the `StftLayer` which does an STFT on its input and is much simpler than the related existing `MultiChannelMultiResolutionStftLayer` and `MultiChannelStftLayer`.

### DIFF
--- a/returnn/tf/layers/signal_processing.py
+++ b/returnn/tf/layers/signal_processing.py
@@ -633,12 +633,6 @@ class StftLayer(_ConcatInputLayer):
     in_dim = self.input_data.feature_dim_or_sparse_dim
     input_data, num_batch_dims = ConvLayer.transform_input(
       self.input_data, network=self.network, in_dim=in_dim)
-    # We want to prepare the input data such that the batch-dim(s) is the very first,
-    # the feature-dim is right after batch-dim ("NCHW"), and the spatial dim is last.
-    if self.output.feature_dim_axis == num_batch_dims:
-      input_data = input_data.copy_with_feature_dim_axis(num_batch_dims)
-    else:
-      input_data = input_data.copy_with_feature_dim_axis(-1)
     x = input_data.placeholder
     # squeeze feature axis
     assert input_data.batch_shape[input_data.feature_dim_axis] == 1, "only implemented for single channel"
@@ -669,6 +663,7 @@ class StftLayer(_ConcatInputLayer):
     data = ConvLayer.get_out_data_from_opts(
       name=name, sources=sources, network=network, filter_size=[frame_size], strides=[frame_shift],
       padding="valid", n_out=fft_size // 2 + 1)
+    data = data.copy_with_feature_dim_axis(-1)
     data.dtype = "complex64"
     return data
 


albertz (approved):

returnn/tf/layers/signal_processing.py
```diff
@@ -607,6 +607,57 @@ def get_out_data_from_opts(cls, fft_size, use_rfft=True, nr_of_channels=1, **kwa
...
+    fft_size = fft_size or frame_size
+    assert "n_out" not in kwargs
+    assert "out_type" not in kwargs
+    super(StftLayer, self).__init__(**kwargs)
+    assert not self.input_data.sparse
+    assert self.input_data.have_batch_axis()
+    assert self.input_data.have_time_axis()
+    out_dim = fft_size // 2 + 1
+    in_dim = self.input_data.feature_dim_or_sparse_dim
+    input_data = self.input_data.copy_as_batch_feature_major()
```
albertz, line 629: Is this `copy_as_batch_feature_major` correct? I think you should handle it afterwards. See `PoolLayer` or `ConvLayer`.

albertz (changes requested):
See comments

tests/test_TFNetworkLayer.py
```diff
@@ -4069,6 +4069,37 @@ def test_ConvLayer_feature_dim_unspecified():
...
+    assert out_time != in_layer.output.dim_tags[1]
+    layer_desc["output"] = stft_out
+    with tf_compat.v1.variable_scope("stft"):
+      stft_layer = StftLayer(**layer_desc)
+    net.layers["stft"] = stft_layer
+    net.initialize_params(session)
+    v_out = session.run(
+      (stft_layer.output.placeholder, stft_layer.output.get_sequence_lengths()),
+      feed_dict=make_feed_dict(net.extern_data, n_time=1024))
+    import ipdb
```
albertz, line 4098: Remove

returnn/tf/layers/signal_processing.py
```diff
@@ -607,6 +607,72 @@ def get_out_data_from_opts(cls, fft_size, use_rfft=True, nr_of_channels=1, **kwa
...
+    :param returnn.tf.network.TFNetwork network:
+    :param int frame_shift: frame shift for STFT
+    :param int frame_size: frame size for STFT
+    :param Optional[int] fft_size: size of the FFT to apply
+    :rtype: Data
+    """
+    fft_size = fft_size or frame_size
+    data = ConvLayer.get_out_data_from_opts(
+      name=name, sources=sources, network=network, filter_size=[frame_size], strides=[frame_shift],
+      padding="valid", n_out=fft_size // 2 + 1)
```
albertz, line 671: According to [the docs](https://www.tensorflow.org/api_docs/python/tf/signal/stft), the output data would always have feature-last, right?

returnn/tf/layers/signal_processing.py
```diff
@@ -607,6 +607,72 @@ def get_out_data_from_opts(cls, fft_size, use_rfft=True, nr_of_channels=1, **kwa
...
+    out_dim = fft_size // 2 + 1
+    in_dim = self.input_data.feature_dim_or_sparse_dim
+    input_data, num_batch_dims = ConvLayer.transform_input(
+      self.input_data, network=self.network, in_dim=in_dim)
+    # We want to prepare the input data such that the batch-dim(s) is the very first,
+    # the feature-dim is right after batch-dim ("NCHW"), and the spatial dim is last.
+    if self.output.feature_dim_axis == num_batch_dims:
+      input_data = input_data.copy_with_feature_dim_axis(num_batch_dims)
+    else:
+      input_data = input_data.copy_with_feature_dim_axis(-1)
```
albertz, line 641: I just notice [from the doc](https://www.tensorflow.org/api_docs/python/tf/signal/stft) that this is anyway irrelevant because the dim must be 1 (or non-existent) and then you anyway squeeze it.

returnn/tf/layers/signal_processing.py
vieting: Yes that's right. Does that change anything here?

returnn/tf/layers/signal_processing.py
vieting: So we can just remove this part here and keep `input_data` as is, right?

albertz: Yes I think so.

albertz (approved):

PR #1004, opened 2022-03-17
DctLayer: fix for size_placeholder

vieting:
I had an issue with the `DctLayer` where the position of the time axis was not the same in input and output (we always enforce batch spatial major). This PR fixes the issue for me.


albertz (approved):
The code is written in a deprecated way (we should not read or write to `size_placeholder`).

But it anyway improves over the existing code, so anyway merged.

PR #1006, opened 2022-03-21
Add option to decide whether i should be updated in the graph

robin-p-schmitt:


tools/compile_tf_graph.py
```diff
@@ -832,10 +833,11 @@ class ConstructionState:
     InLoop = Entity("in_body")
 
   @classmethod
-  def prepare_compile(cls, rec_layer_name, net_dict):
+  def prepare_compile(cls, rec_layer_name, net_dict, opts):
     """
     :param str rec_layer_name:
     :param dict[str,dict[str]] net_dict:
+    :param dict[str,bool] opts:
```
albertz, line 840:
```suggestion
    :param dict[str] opts:
```

tools/compile_tf_graph.py
```diff
@@ -1070,7 +1073,7 @@ def assign(self, value):
         value = x.placeholder
       return tf_compat.v1.assign(self.var, value, name="final_state_var_%s" % self.name).op
 
-  def __init__(self, _orig_sources, sources, network, name, output, unit, axis,
```
albertz, line 1073: `rec_step_by_step_opts` missing in doc string.

tools/compile_tf_graph.py
```diff
@@ -1476,13 +1480,15 @@ def main(argv):
   argparser.add_argument("--output_file", help='allowed extensions: pb, pbtxt, meta, metatxt, logdir')
   argparser.add_argument("--output_file_model_params_list", help="line-based, names of model params")
   argparser.add_argument("--output_file_state_vars_list", help="line-based, name of state vars")
+  argparser.add_argument("--update_i_in_graph", action="store_true", help="whether to update i in the graph")
```
albertz, line 1483: You are changing the logic from before here. Before, it was always True. Now it is False by default and can optionally be set to True.

I wonder if the default should be as before.

albertz: Although, the default maybe should match the RASR logic. The question is more if we want to cleanup and generalize RASR.

albertz: Maybe this changed default is ok. I assume this is also how it was in @ZhouW321's fork.

albertz: Btw, there is also the `cond` global state var. I just noticed that we currently exclude this for some strange reason. Which is maybe again for RASR because RASR currently has this strange assumption that there is only exactly one single global state var which is `i`.

albertz (changes requested):
See comments.

tools/compile_tf_graph.py
albertz: Any comment?

robin-p-schmitt: Ah, sorry. I think because RASR currently has this assumption, it might be good if the default behavior of the compile script would match this assumption

albertz (approved):

PR #1011, opened 2022-03-22
Fix 'only_on_search' ignored if 'register_as_extern_data' set

patrick-wilken:
Layer used "only_on_search" are currently constructed also in training if "register_as_extern_data" is set. This is a bug or at least very counter-intuitive.

(Using those flags in combination might be questionable though... I'm abusing extern data lacking knowledge of a better way to make a layer input dependent on whether it is training or search. My layer with "only_on_search" overwrites a previous "register_as_extern_data". )

### DIFF
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -607,24 +607,32 @@ class TFNetwork(object):
     """
     self.layers_desc.update(net_dict)
 
+    def ignore_layer(name_, layer_desc_):
+      """
+      :param str name_:
+      :param dict layer_desc_:
+      :rtype: bool
+      """
+      assert isinstance(name_, str)
+      if name_.startswith("#"):
+        return True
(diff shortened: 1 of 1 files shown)


returnn/tf/network.py
```diff
@@ -607,24 +607,30 @@ def construct_from_dict(self, net_dict, get_layer=None):
...
+      :param dict layer_desc_:
+      :rtype: bool
+      """
+      assert isinstance(name_, str)
+      if name_.startswith("#"):
+        return True
+      assert isinstance(layer_desc_, dict)
+      return (
+        layer_desc_.get("only_on_search") and not self.search_flag
+        or layer_desc_.get("only_on_eval") and not self.eval_flag)
```
albertz, line 622: I would prefer to have this as two separate `if` statements, just as before. I think this is more readable.

albertz (approved):

PR #1019, opened 2022-04-06
Data seq mask, fix for extended dyn size

albertz:
This looks like a major bug. It could affect all more custom masking logic (`SoftmaxOverSpatialLayer`, `GenericAttentionLayer`/`DotLayer`) for custom dyn sizes (e.g. in some transducer or segmental models).

(@robin-p-schmitt But it doesn't seem to fix the problem you showed me with nan when you disable attention dropout, [config](https://gist.github.com/albertz/3de318fca54b76518e24cde024913585).)

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -5084,7 +5084,7 @@ class Data(object):
         max_idx = tf.reduce_max(tag.dyn_size)
         # We use the assumption that self.placeholder.shape[axis] == max_idx.
         idx_range = tf.range(max_idx)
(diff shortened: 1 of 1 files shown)


albertz:
Ok, I was a bit afraid first that this was seriously wrong before and many experiments were invalid, but now I realize that in most cases it was already fine. The earlier code would result in one dim less for `idx_range`, missing in front. Due to the broadcasting rules, when this is combined with another tensor with one more dim, a broadcast dim will automatically be added in front, so it would result in what we intended anyway. The only case where it was wrong is for `axis = 0`. In that case however, very likely it would have resulted in an error due to shape mismatch.

Issue #576, opened 2021-08-12
`SoftmaxOverSpatialLayer` might produce NaN gradients


albertz:
> > > Note that it's really important to never mask away all keys for a single query, even if the query itself is inside padding and might be masked away later.
> > > This will cause the softmax to return `NaN` for that query pos, causing the loss for that query also to be `NaN`. And even though it might be masked away when the total loss is accumulated, this can still cause the total loss to be `NaN`. I had this issue sometimes. That's not difficult to fix (e.g. `SoftmaxOverSpatialLayer` could never apply a softmax for masked away query positions), but maybe we should keep it in mind.
> > 
> > 
> > I think this is somewhat orthogonal to this issue, so we should open a separate issue and discuss over there.
> > As I understand you, this happened for you for frames which are inside the padding (i.e. outside of the sequence).
> > I'm not exactly sure how it can happen then that this ends up in the final loss, because for framewise losses (e.g. CE), it would cut out only the sequences, i.e. remove the padded frames, and then sum only over the real relevant frames.
> 
> The problem is that this "cutting to only the sequences" is implemented with `tf.where`, and that the gradient of `tf.where` is NaN as soon as one input is NaN, even if it is masked away. From the [tf.where](https://www.tensorflow.org/api_docs/python/tf/where) docs:
> 
> > Note that if the gradient of either branch of the tf.where generates a NaN, then the gradient of the entire tf.where will be NaN

We even have `test_where_nan` about that.

> > But anyway, let's discuss this in a separate issue.
> 
> I can make it an issue, but it's more a TF issue than a RETURNN issue (actually see [tensorflow/tensorflow#38349](https://github.com/tensorflow/tensorflow/issues/38349)). I just mentioned this because once/if we add some implicit "HistoryMaskLayer", then this is something we also should implicitly handle. Right now, you can just handle this all explicitly.

No, it is a RETURNN issue. There should be no NaN in the loss or gradient when it is incorrect. Then it just means we need to add some workaround. But anyway, this is independent of this issue here, so in any case, it should be discussed elsewhere, to not blow up this issue thread even more.

_Originally posted by @albertz / @Zettelkasten in https://github.com/rwth-i6/returnn/issues/391#issuecomment-896882118_

Issue #721, opened 2021-10-19
Layers applying a mask should expand input to all `dyn_size_ext` dimensions


Zettelkasten:
Consider I have some input of shape `[T]`, where `T` has dyn_sizes with some other axes, e.g. `[B]`.
I want to use `ReduceLayer` on this axis, which needs to apply a sequence mask.
Currently, this fails because we assert that all axes in `dyn_size_ext` need to be present in the input.

Instead, I would like to support this by first expanding the input to all dims from `dyn_size_ext` that are missing. So in my case, we would first broadcast `[T]` to `[B,T]`, then apply the sequence mask, reduce the axis, and get some output `[B,T]`.

Zettelkasten:
> I wonder whether this can be slow and suboptimal in some cases.
> 
> E.g. in `DotLayer`, you definitely would not want to do that, when the axis is present in one of the inputs.

For `DotLayer`, we would always prefer to mask an input that already has all needed dims, and only use this in the case where neither input has all required dims. Performance should be the same as before.

Maybe, if you think this can happen by accident, we can also make this explicit, similar to #691.

> What are the layers which would do this? `ReduceLayer`, `DotLayer`. Others?

Also `SeqLenMaskLayer`, and related to that `SoftmaxOverSpatialLayer`.
In general, all that use `Data.get_sequence_mask_broadcast` in some way.
Also `SearchSortedLayer` (but not so important, probably I'm the only one using that layer).

albertz:
I guess [the list of layers with special behavior on dynamic spatial axes](https://github.com/rwth-i6/returnn/wiki/Layers-with-special-behavior-on-dynamic-spatial-axes) is basically exactly this.

I'm not sure if we need to implement it for all of those now though. It should be fine when there is proper error handling (i.e. some exception is thrown, with some reference to this issue).

This is already the case for all layers using `get_sequence_mask_broadcast`. And I just added a reference in the comment to this issue.

So, what's left? Otherwise I think we can close this, and just implement it only when it is really needed.

You said you need it for `ReduceLayer`? Maybe you can just create a PR? I think you just need to call `copy_add_dim_by_tag` with `unbroadcast=True` for the missing dim tags. Maybe we can add a helper like `copy_add_missing_dim_tags` or so.

Zettelkasten:
Now that we disallow implicit broadcasting in #691, I think it makes sense to also enforce this here. I.e. only allow to broadcast to the implicit size dims if the user specifies `out_shape`.

[...]

Issue #717, opened 2021-10-18
DotLayer masking heuristic fails in some cases with `dyn_size_ext`


Zettelkasten:
DotLayer currently has this heuristic to determine on which input we apply masking:

```
        # We need to apply a mask.
        # We don't need it on both a and b. We can either apply it on a or on b.
        # Use some very simple heuristic where the mask is maybe cheaper.
        if len(a_shape) < len(b_shape):
          a = mask_dyn_seq_len_nd(a_out, pad_value=0, axes=a_reduce_dyn_axes)
          self._info_reduce_mask = "mask-source-0"
        else:
          b = mask_dyn_seq_len_nd(b_out, pad_value=0, axes=b_reduce_dyn_axes)
          self._info_reduce_mask = "mask-source-1"
```

However with `dyn_size_ext`, it can happen that the seq lengths depend on another axis which is only present in one of these inputs.
In our case:
```
a: Data{'att_weights', [B?,F|'heads'(8),T|'classes'[B?],'keys'['classes'[B?]]]}
b: Data{'att_value', [T|'keys'['classes'[B?]],B?,'heads'(8),F|F'feature'(64)]}
```
the seq lengths of `keys` depends on `classes`, so we have to mask input `a`.

PR #1020, opened 2022-04-07
Data ext seq mask, use non-zero sizes for padded ext dyn size

albertz:
size_ext might have invalid (zero) sizes when it itself has some padding, e.g. when its own shape is dynamic.

A zero size can lead to problems in some cases, e.g. in SoftmaxOverSpatialLayer, when everything is masked to -inf, it results in nan, and this likely produces nan in backprop or elsewhere.

Thus, mask size_ext itself, and set the padded values to 1.

---

This is somewhat an unrelated or orthogonal follow-up to #1019, resulting from [this config](https://gist.github.com/albertz/3de318fca54b76518e24cde024913585).

@robin-p-schmitt This probably should fix the problem you showed me with nan when you disable attention dropout.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -250,26 +250,44 @@ def set_padding_info(x, dim, pad_value):
 def mask_dyn_seq_len_nd(x, pad_value, axes):
   """
   :param Data x:
-  :param float|int pad_value:
+  :param float|int|tf.Tensor pad_value:
   :param list[int]|tuple[int] axes:
   :return: masked x
   :rtype: tf.Tensor
   """
-  x_ = x.placeholder
-  dim_tags = [x.dim_tags[i] for i in axes]
-  d = get_padding_info_dict_ref(x_)
-  existing_pad_values = [d.get(tag) for tag in dim_tags]
-  if set(existing_pad_values) == {pad_value}:
-    return x.placeholder  # nothing to do
+  # Filter out some axes which should not be used for masking.
+  axes_ = []
+  for axis in axes:
+    tag = x.dim_tags[axis]
+    assert tag.dyn_size_ext
+    # It only makes sense to apply for this axis if the dyn size dims are all existing in x itself.
+    # E.g. if the dyn_size_ext shape is [B] but the shape of x is just [T] (without B),
+    # then we do not need masking.
+    if set(tag.dyn_size_ext.dim_tags).issubset(x.dim_tags):
+      axes_.append(axis)
+  axes = axes_
 
-  mask = tf.ones([1] * x.batch_ndim, dtype=tf.bool)
+  x_ = x.placeholder
+  if not axes:
+    return x_
+
+  pad_value_is_const = isinstance(pad_value, (int, float))
+  if pad_value_is_const:
+    d = get_padding_info_dict_ref(x_)
+    existing_pad_values = [d.get(x.dim_tags[axis]) for axis in axes]
+    if set(existing_pad_values) == {pad_value}:
+      return x_  # nothing to do
+
+  mask = None
   for axis in axes:
     mask_ = x.get_sequence_mask_broadcast(axis=axis)
-    mask = tf.logical_and(mask, mask_)
-  x_ = where_bc(mask, x_, tf.cast(tf.constant(pad_value, name="pad_value"), dtype=x_.dtype))
-  d = get_padding_info_dict_ref(x_)
-  d.clear()
-  d.update({k: pad_value for k in dim_tags})
+    mask = tf.logical_and(mask, mask_) if mask is not None else mask_
+  assert isinstance(mask, tf.Tensor)
+  x_ = where_bc(mask, x_, tf.cast(tf.convert_to_tensor(pad_value, name="pad_value"), dtype=x_.dtype))
+  if pad_value_is_const:
+    d = get_padding_info_dict_ref(x_)
+    d.clear()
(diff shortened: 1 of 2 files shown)


albertz:
> I wonder if `1` is really the best pad value here. I think I would rather use the full unpadded seq length as mask value. `1` seems a bit arbitrary and may produce very unexpected results. E.g. if the first time frame is `-inf` in the data, and then the seq length is masked to `1`, this will also produce a `NaN` gradient after a softmax. So using the entire unpadded seq length in that case gives the highest chance to have at least one finite input frame.

Right, using `max_idx` is a better idea.

> How specific is this to `SoftmaxOverSpatialLayer`,

This is the case where I stumbled upon this problem.

This is classic attention (but over a dyn_size_ext of shape [B,S]), i.e. softmax over spatial, then dot layer. The softmax would produce nan for any seq size = 0 because softmax(-inf) = nan.

We had the strange case that we also had dropout on the attention weights, and with dropout, all was fine, but once we removed the dropout, we got nan.

I figured out that dropout resulted in removing the padding info (`set_padding_info`), so then our new `DotLayer` now again did masking, just to be sure, and replaced all the nans by 0, and then all was fine. But when there is no dropout, it would not do the masking again, and the nan will result in further problems in backprop or elsewhere.

> or are other layers maybe also a problem?

I could imagine but I didn't really check all the layers.

> We have #576 about NaN gradients after `SoftmaxOverSpatialLayer` in general.

Ah, I forgot about this issue. Yea, this is related to this PR.

> I think a nicer fix would be to change `SoftmaxOverSpatialLayer` to always mask away sequences for queries where all keys are invalid with something finite before applying the softmax. I.e. to fix #576 directly.

I don't remember the details of the issue anymore. Is this about invalid data in the padded area?

So this PR here fixes invalid seq lens in the padded area. So this is one part (or already the whole thing) to fix it, right? Is there anything missing then?

But anyway, I think this PR is also useful as-is, or not?

albertz:
@Zettelkasten I changed the implementation a bit. `mask_dyn_seq_len_nd` got extended a bit. Can you check if this looks ok?

returnn/tf/util/basic.py
```diff
@@ -250,26 +250,44 @@ def set_padding_info(x, dim, pad_value):
...
   :return: masked x
   :rtype: tf.Tensor
   """
+  # Filter out some axes which should not be used for masking.
+  axes_ = []
+  for axis in axes:
+    tag = x.dim_tags[axis]
+    assert tag.dyn_size_ext
+    # It only makes sense to apply for this axis if the dyn size dims are all existing in x itself.
+    # E.g. if the dyn_size_ext shape is [B] but the shape of x is just [T] (without B),
```
Zettelkasten, line 264: Wait, isn't this very dangerous / wrong?
In this case, you'd want the output of the layer to have this implicit dims (`B` in your example) as concrete dims in the output. As in https://github.com/rwth-i6/returnn/issues/721. Right?
We e.g. had a bug with this before in https://github.com/rwth-i6/returnn/issues/717.

Zettelkasten:
> > We have #576 about NaN gradients after `SoftmaxOverSpatialLayer` in general.
> 
> Ah, I forgot about this issue. Yea, this is related to this PR.
> 
> > I think a nicer fix would be to change `SoftmaxOverSpatialLayer` to always mask away sequences for queries where all keys are invalid with something finite before applying the softmax. I.e. to fix #576 directly.
> 
> I don't remember the details of the issue anymore. Is this about invalid data in the padded area?

Yes it is. It is actually kind of the same case as here, where because of masking, all keys for one specific query would be masked away with -inf, resulting in NaN after the softmax.
Here, this masking is because of the seq len mask. In #576, more generally, the masking was done by me explicitly using other layers because I wanted a more complex mask: E.g. I masked away all keys with key=query position. But I tried quite a few of these masks.
Just like in this issue, I didn't really think about what my mask did for invalid query positions, and if I was unlucky, all keys would be masked away to -inf for one key, resulting in the NaN softmax output + gradients.

> So this PR here fixes invalid seq lens in the padded area. So this is one part (or already the whole thing) to fix it, right? Is there anything missing then?

It's fixed for the case where this masking is performed implicitly via the seq len mask, but not when the input to the softmax layer already has -inf for all keys for some query.

> But anyway, I think this PR is also useful as-is, or not?

Yes, certainly.

albertz:
> > So this PR here fixes invalid seq lens in the padded area. So this is one part (or already the whole thing) to fix it, right? Is there anything missing then?
> 
> It's fixed for the case where this masking is performed implicitly via the seq len mask, but not when the input to the softmax layer already has -inf for all keys for some query.

I see.

But I would keep that for a separate PR / change. Also, masking the dyn seq size is cheap and easy to do (it should not really cost anything). Handling the other issue will be more involved and I cannot really imagine any solution which would not at least add some overhead.

returnn/tf/util/basic.py
albertz: The usage of `mask_dyn_seq_len_nd` is only in `DotLayer` and in `Data.copy_masked` which itself is only used for the ext dyn size in `Data.get_sequence_mask_broadcast`. 

In `DotLayer`, we already check for that.

In `Data.get_sequence_mask_broadcast` for the ext dyn size, it should not matter. We just want to make sure that there are no invalid sizes (zero sizes).

You are right, it could be wrong in other cases. Currently there are no other cases and I think the current cases are both correct.

So, what should we do? Maybe just add a big warning comment with this info there?

PR #1033, opened 2022-05-09
Reduce Layer documentation

Atticus1806:
The documentation of the reduce layer was outdated since more than just `sum` and `max` are allowed. Still kept them as examples, but maybe we can just drop the completely from there.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -5791,8 +5791,8 @@ class TransposedConvLayer(_ConcatInputLayer):
 
 class ReduceLayer(_ConcatInputLayer):
   """
-  This reduces some axis by using "sum" or "max".
-  It's basically a wrapper around tf.reduce_sum or tf.reduce_max.
+  This reduces some axis by using a method defined by the mode parameter, e.g. "sum" or "max".
+  It's basically a wrapper around the corresponding tensorflow functions like tf.reduce_sum or tf.reduce_max.
   """
   layer_class = "reduce"
 
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -5791,8 +5791,8 @@ class TransposedConvLayer(_ConcatInputLayer):
 
 class ReduceLayer(_ConcatInputLayer):
   """
-  This reduces some axis by using a method defined by the mode parameter, e.g. "sum" or "max".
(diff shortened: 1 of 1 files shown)


Atticus1806:
What I primarily added was the "generalization". Because from reading: 
`This reduces some axis by using "sum" or "max".` I would stop reading and think I can only use these two modes.
We can also just do like 
```
This reduces some axis by using a method defined by the mode parameter
It's basically a wrapper around the corresponding tensorflow functions
```

albertz:
As I said, I would keep it short, like: "This reduces some axis, e.g. by using "sum" or "max"."

albertz:
Also, I would have expected that it's already clear from the layer name (`ReduceLayer`) that sum and max are only examples.

albertz:
Btw, the documentation always anyway should also be read together with the arguments. When you look at the generated doc: https://returnn.readthedocs.io/en/latest/layer_reference/basic.html?highlight=reducelayer#returnn.tf.layers.basic.ReduceLayer

It already becomes pretty clear what it does. I wonder if adding more text really adds any value, or rather makes it just more convoluted. But this is just my opinion. Maybe someone else could also state his opinion.

Atticus1806:
>"This reduces some axis, e.g. by using "sum" or "max"

I can live with that too. Just thought it would be helpful to make clear sum and max are just examples. 

> Also, I would have expected that it's already clear from the layer name (ReduceLayer) that sum and max are only examples.

I am not sure if I agree. Even though different reductions are possible its not always sure they are implemented (yet)

In the end this is really minor, just thought it would be helpful.
If we just add the 'e.g." then im fine with the second line staying the same.

albertz:
> > "This reduces some axis, e.g. by using "sum" or "max"
> 
> I can live with that too. Just thought it would be helpful to make clear sum and max are just examples.

The "e.g." says exactly that, right?

> > Also, I would have expected that it's already clear from the layer name (ReduceLayer) that sum and max are only examples.
> 
> I am not sure if I agree. Even though different reductions are possible its not always sure they are implemented (yet)

But then this is not a problem of the documentation but about the implementation. If you find out that sth is missing, you should not edit the documentation but instead just add whatever is missing.

> In the end this is really minor, just thought it would be helpful. If we just add the 'e.g." then im fine with the second line staying as before.

Ok, at least I would prefer that.

albertz (approved):

PR #1046, opened 2022-05-19
set dyn_size_ext for out_dim

Atticus1806:
Fix rwth-i6/returnn_common#163

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4379,6 +4379,7 @@ class RepeatLayer(_ConcatInputLayer):
     original_axis = data.get_axis_from_description(axis, allow_int=False)
     tag = data.dim_tags[original_axis]
     data = data.copy_move_axis(original_axis, data.get_batch_axis(0))
+    out_dim.dyn_size_ext = tag.dyn_size_ext.copy_template()
     if isinstance(repetitions, int):
       out_dim_ = tag * repetitions
     else:


returnn/tf/layers/basic.py
```diff
@@ -4379,6 +4379,7 @@ def get_out_data_from_opts(cls, name, sources, axis, repetitions, out_dim=None,
     original_axis = data.get_axis_from_description(axis, allow_int=False)
     tag = data.dim_tags[original_axis]
     data = data.copy_move_axis(original_axis, data.get_batch_axis(0))
+    out_dim.dyn_size_ext = tag.dyn_size_ext.copy_template()
```
albertz, line 4382: This only works if `tag` has `dyn_size_ext`.

albertz: Also, I think it's more safe if you do `out_dim = out_dim.get_for_batch_ctx(data.batch, data.control_flow_ctx)` before.

Atticus1806: What would we do if it does not? So checking it with an if is no problem, but would we handle the other case aswell?

albertz: If the input dim is static (i.e. `dyn_size_ext` is not set), you also don't need to set `out_dim. dyn_size_ext`.

Atticus1806: > Also, only if out_dim is set here. I think better do it below for out_dim_.
I tried that first, to set it directly before the return (or after the `if else`), but this does not work for some reason.

albertz: Was this together with `get_for_batch_ctx`? But anyway, it has to be this way.

albertz: Yes.

albertz: I think it's already ok. The whole handling of `dyn_size_ext` and `get_for_batch_ctx` is unfortunately not so obvious or even well defined. (I think there is also another issue on that already...)

PR #1055, opened 2022-06-02
Fix dim tag derived_from_tag cycle

### LINKED ISSUE #1054: Dim tag derived_from_tag/same_as cycle due to conv+padding
I have the following exemplary config reproducing the problem:
```
#!returnn.py
# kate: syntax python;

use_tensorflow = True

task = "initialize_model"
train = {"class": "Task12AXDataset", "num_seqs": 1000}

network = {
        "Conv1": {'class': 'conv', 'from': 'data', 'activation': None, 'with_bias': True, 'n_out': 9, 'filter_size': (4,), 'padding': 'same', 'in_spatial_dims': ['T'], 'strides': (2,)},
        "ConstantPad": {'class': 'pad', 'mode': 'constant', 'axes': ['T'], 'padding': [(1, 1)], 'from': 'Conv1', 'value': 0},
        "Conv2": {'class': 'conv', 'from': 'ConstantPad', 'activation': None, 'with_bias': True, 'n_out': 9, 'filter_size': (3,), 'padding': 'valid', 'in_spatial_dims': ['T'], 'groups': 9},
        "add": {'class': 'combine', 'kind': 'add', 'from': ['Conv1', 'Conv2']},
        "output": {"class": "softmax", "loss": "ce", "from": ["add"]},
}

log_verbosity = 5
```

This config gives the error:
```
  File "/u/dierkes/returnn/returnn/tf/util/data.py", line 870, in Dim.undefined
    line: assert id(base) not in visited  # should not have cycles. normally this should never be triggered
    locals:
      id = <builtin> <built-in function id>
      base = <local> Dim{'Conv2:conv:s0'[B]}
      visited = <local> {140193235905552: Dim{'Conv2:conv:s0'[B]}, 140193235855968: Dim{'Conv2:conv:s0'[B]}, 140193235561152: Dim{'Conv2:conv:s0'[B]}, 140193235534224: Dim{'1+Conv1:conv:s0+1'[B]}, 140193235535712: Dim{'1+Conv1:conv:s0+1'[?]}}
AssertionError
```

Full stack-trace is posted below. 

Apparently the problem is that there is a circle in the dimension ancestry of the time output dimension of `Conv2`. This is problematic in `Dim.undefined` where the problem is detected. The circle is the following: `Conv2` time dimension links to `ConstantPad` time dimension which links back to `Conv2` time dimension. I think the problem is created in when for the layer `Conv2` the function `get_out_data_from_opts` is called:
In `Conv2.get_out_data_from_opts` the function `set_output_dim_tag` is called https://github.com/rwth-i6/returnn/blob/ae42d2473e44f8bce4fe04f88b021b4d0bffea7b/returnn/tf/layers/basic.py#L5300
In the function `set_output_dim_tag` we then calculate the out tag in https://github.com/rwth-i6/returnn/blob/ae42d2473e44f8bce4fe04f88b021b4d0bffea7b/returnn/tf/layers/basic.py#L5043
and finally set the calculated output tag to be `same_as` the `out_tag` in https://github.com/rwth-i6/returnn/blob/ae42d2473e44f8bce4fe04f88b021b4d0bffea7b/returnn/tf/layers/basic.py#L5048
Apparently because of the structure of our network the calculated output tag is the time dimension of `Conv1` and then appears to be overwritten to the out tag of `Conv2`. Thus the time dimension ancestor of `ConstantPad` is linking now to `Conv2` time dimension instead of `Conv1` creating the circle.

This problem is also relevant to the issue in the pytorch-to-returnn converter https://github.com/rwth-i6/pytorch-to-returnn/issues/124

A proposed solution to the problem you can find in the PR draft #1055

dierkes-j:
Fix #1054.
Fix https://github.com/rwth-i6/pytorch-to-returnn/issues/124.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -5260,7 +5260,9 @@ class ConvLayer(_ConcatInputLayer):
       dim_tags += out_spatial_dims
     else:
       for i in range(len(filter_size)):
-        old_tag = old_spatial_dim_tags[i] if i < len(old_spatial_dim_tags) else None
+        old_tag = old_spatial_dim_tags[i].copy(
+                    same_as_self=False, description=old_spatial_dim_tags[i].description
+                ) if i < len(old_spatial_dim_tags) else None
         if old_tag and (filter_size[i] == strides[i] == 1 or (strides[i] == 1 and padding == "SAME")):
           dim_tags.append(old_tag)  # identity in this axis
           continue
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -5260,9 +5260,7 @@ class ConvLayer(_ConcatInputLayer):
       dim_tags += out_spatial_dims
     else:
       for i in range(len(filter_size)):
-        old_tag = old_spatial_dim_tags[i].copy(
-                    same_as_self=False, description=old_spatial_dim_tags[i].description
-                ) if i < len(old_spatial_dim_tags) else None
+        old_tag = old_spatial_dim_tags[i] if i < len(old_spatial_dim_tags) else None
         if old_tag and (filter_size[i] == strides[i] == 1 or (strides[i] == 1 and padding == "SAME")):
           dim_tags.append(old_tag)  # identity in this axis
           continue
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -916,7 +916,6 @@ class Dim(object):
         self.dyn_size_ext = self_same_as.get_dyn_size_ext_for_batch_ctx(self.batch, self.control_flow_ctx)
     self_derived_bases = set(self.get_derived_bases_list())
     other_derived_bases = set(other.get_derived_bases_list())
-    assert self_derived_bases != other_derived_bases
     if self_derived_bases.issubset(other_derived_bases):
       # Avoid cycles on derived_from_tag. https://github.com/rwth-i6/returnn/issues/1054
       return other.declare_same_as(self)
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -914,6 +914,11 @@ class Dim(object):
       self_same_as.declare_same_as(other_same_base)
       if (self.dyn_size_ext is None or not self._validate_in_current_graph()) and self_same_as.dyn_size_ext:
         self.dyn_size_ext = self_same_as.get_dyn_size_ext_for_batch_ctx(self.batch, self.control_flow_ctx)
+    self_derived_bases = set(self.get_derived_bases_list())
+    other_derived_bases = set(other.get_derived_bases_list())
+    if self_derived_bases.issubset(other_derived_bases):
+      # Avoid cycles on derived_from_tag. https://github.com/rwth-i6/returnn/issues/1054
+      return other.declare_same_as(self)
     other_same_base._merge_same_for_batch_ctx_dict(self)
     other._maybe_update()
     self.same_as = other_same_base
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -907,6 +907,12 @@ class Dim(object):
       # We actually want it to be the other way around.
       other_same_base.declare_same_as(self_same_as)
       return
+    other_derived_bases = set(other.get_derived_bases_list())
+    self_derived_bases = set(self.get_derived_bases_list())
+    assert other_derived_bases != self_derived_bases
+    if self_derived_bases.issubset(other_derived_bases):
+      # Avoid cycles on derived_from_tag. https://github.com/rwth-i6/returnn/issues/1054
+      return other.declare_same_as(self)
     if self_same_as is not self:
       assert not self_same_as.same_as
       if self_same_as is other_same_base:
@@ -914,11 +920,6 @@ class Dim(object):
       self_same_as.declare_same_as(other_same_base)
       if (self.dyn_size_ext is None or not self._validate_in_current_graph()) and self_same_as.dyn_size_ext:
         self.dyn_size_ext = self_same_as.get_dyn_size_ext_for_batch_ctx(self.batch, self.control_flow_ctx)
-    self_derived_bases = set(self.get_derived_bases_list())
-    other_derived_bases = set(other.get_derived_bases_list())
-    if self_derived_bases.issubset(other_derived_bases):
-      # Avoid cycles on derived_from_tag. https://github.com/rwth-i6/returnn/issues/1054
-      return other.declare_same_as(self)
     other_same_base._merge_same_for_batch_ctx_dict(self)
     other._maybe_update()
     self.same_as = other_same_base
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -895,7 +895,7 @@ class Dim(object):
     assert self.can_be_used_as_dim() and other.can_be_used_as_dim()  # declare_same_as does not make sense otherwise
     self._maybe_update()
     self._validate_in_current_graph()
-    if self is other:
+    if self == other:
       return
     other_same_base = other.get_same_base()
     if self is other_same_base or self.same_as is other_same_base:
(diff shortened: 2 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -5260,7 +5260,9 @@ def get_out_data_from_opts(
       dim_tags += out_spatial_dims
     else:
       for i in range(len(filter_size)):
-        old_tag = old_spatial_dim_tags[i] if i < len(old_spatial_dim_tags) else None
+        old_tag = old_spatial_dim_tags[i].copy(
+                    same_as_self=False, description=old_spatial_dim_tags[i].description
```
albertz, line 5264: Formatting (indentation) is off.

But also, this does not make sense. You explicitly want that you have `same_as_self=True` here. Also, there should be no copy of the tag.

returnn/tf/util/data.py
```diff
@@ -914,6 +914,11 @@ def declare_same_as(self, other):
       self_same_as.declare_same_as(other_same_base)
       if (self.dyn_size_ext is None or not self._validate_in_current_graph()) and self_same_as.dyn_size_ext:
         self.dyn_size_ext = self_same_as.get_dyn_size_ext_for_batch_ctx(self.batch, self.control_flow_ctx)
+    self_derived_bases = set(self.get_derived_bases_list())
+    other_derived_bases = set(other.get_derived_bases_list())
+    if self_derived_bases.issubset(other_derived_bases):
+      # Avoid cycles on derived_from_tag. https://github.com/rwth-i6/returnn/issues/1054
+      return other.declare_same_as(self)
```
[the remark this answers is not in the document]
dierkes-j: I checked with this test case: 
```python
from returnn.tf.network import TFNetwork
from returnn.config import Config

n_in, n_out = 2, 3
config = Config({"extern_data": {"data": {"dim": n_in}}})
network = TFNetwork(config=config)
net_dict = {
  "output": {
    "class": "rec",
    "name_scope": "",
    "from": "data:data",
    "optimize_move_layers_out": False,
    "unit": {
      "layer1": {"class": "linear", "n_out": n_out, "from": "data:source", "is_output_layer": True},
      "layer2": {"class": "linear", "n_out": n_out, "from": "data:source", "is_output_layer": True,
                 "name_scope": "layer1"},
      "output": {"class": "combine", "kind": "add", "from": ["layer1", "layer2"]}
    }
  }
}
network.construct_from_dict(net_dict)
recusion_bug.py (END)
```

here the recursion is introduced for the `self` having tag `Dim{F'layer1:feature-dense'(3)}` and `other` having the same tag `Dim{F'layer1:feature-dense'(3)}` but the two dimensions are not the same object. The problem why this recursion is introduced is because `self is other` evaluates to `False` but `self == other` evaluates to `True` and thus the subset property is given. 

Is it required to have:
```python
if self is other:
  return
```
or can this be changed to:
```python
if self == other:
  return
```

Alternatively we could to the subset check with `is` instead of `==` like:
```python
other_derived_bases = other.get_derived_bases_list()
self_derived_bases = self.get_derived_bases_list()
subset = True
for self_dim in self_derived_bases:
  exists_in_base = False
  for other_dim in other_derived_bases:
    if self_dim is other_dim:
      exists_in_base = True 
      break
  if not exists_in_base:
    subset = False
    break
if subset:
  return other.declare_same_as(self)
```
This solves the recursion error and still solves the circular dependency problem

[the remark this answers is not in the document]
dierkes-j: Oh, right I see. But why do you introduce an extra line with `self == other`? Instead we could just change 'self is other` to `self == other` in line 898. If `self is other` is True than also `self == other`.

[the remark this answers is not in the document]
albertz: > But why do you introduce an extra line with `self == other`? Instead we could just change `self is other` to `self == other` in line 898.

I did not see that here on GitHub. Right, this makes sense.

returnn/tf/util/data.py
```diff
@@ -914,6 +914,11 @@ def declare_same_as(self, other):
       self_same_as.declare_same_as(other_same_base)
       if (self.dyn_size_ext is None or not self._validate_in_current_graph()) and self_same_as.dyn_size_ext:
         self.dyn_size_ext = self_same_as.get_dyn_size_ext_for_batch_ctx(self.batch, self.control_flow_ctx)
+    self_derived_bases = set(self.get_derived_bases_list())
+    other_derived_bases = set(other.get_derived_bases_list())
```
albertz, line 918: There should be an assert now like:
```python
assert self_derived_bases != other_derived_bases
```

albertz: I don't understand. Why can it be equal? Then we should have `self == other` or not?

dierkes-j: I don't know exactly. I will look into it a post a test case with explanation. However it is probably the case that `self` has an ancestor that is equal to `other` but not the same object.  Similarly `other` has an ancestor that is equal to `self` but is not the same object. Then the two sets are equal but not containing the same objects. Thus we are running into the assertion.

albertz: But equality (`==`, `!=`) of a set does check the equality of the set, i.e. equality of the objects (`==`), not identity (`is`).

dierkes-j: Right, that was what I was saying. The dimensions can be equal (`==`) without being identical (`is`). If the assertion would check for identity it would not fail (I believe).

albertz: But `assert self_derived_bases != other_derived_bases` uses `!=` (`==`) not `is`.

dierkes-j: Yes. Because it uses `==` and not `is` the assertion fails. 
> However it is probably the case that self has an ancestor that is equal to other but not the same object. Similarly other has an ancestor that is equal to self but is not the same object. Then the two sets are equal but not containing the same objects. Thus we are running into the assertion.

Here I tried to describe a case where the equality (`==`) of the two sets is given but the sets are not identical (`is`). So even if `self` and `other` are not equal (`!=`) the two sets can still be equal (`==`) in the end (because both have ancestors that are equal to the other). But the sets are not identical. 

Or am I getting something wrong here?

albertz: `self_derived_bases` also includes `self`, and `other_derived_bases` also includes `other`.

dierkes-j: > Why do you talk about identity (is) at all?

If the dimension tags are equal (as the assertion suggests) but we are not having the recursion problem that we are trying to solve, then the two sets must contain objects that are not identical. At least that's how I make sense of the situation. 

The problem can be reproduced like this:
```python
from returnn.tf.network import TFNetwork
from returnn.config import Config

import numpy as np
batch_size = 8
time_size = 20
feat_size = 10
shift_amount = 5  # right-shift of 5 elements
config = Config()
config.update({
  "num_outputs": feat_size,
  "num_inputs": feat_size,
  "network": {
    "output": {"class": "shift_axis", "from": ["data"], "amount": shift_amount, "pad": True, "axis": "T"}
  }
})
network = TFNetwork(config=config, train_flag=True)
network.construct_from_dict(config.typed_dict["network"])
```

The problematic line is https://github.com/rwth-i6/returnn/blob/ae0cc8c4f802013d69a86782b7dce30fb6db16f3/returnn/tf/util/data.py#L914

which is called after our check `self==other` but which might make `self` and `other` equal so that our two sets can then be equal again. We just have to do the subset check earlier in the code to avoid this problem I think.

albertz: Why did you mark this as resolved? Is it resolved?

dierkes-j: Yes, I added the assertion and it seems to work now. Do you disagree?

albertz: No I think it's fine.

Overall I'm not too happy with the whole function. It's way too complicated. But not sure how to simplify it. But anyway, this is not about this PR now.

albertz:
You can mark the PR as ready when all tests are passing.

albertz (approved):

PR #1056, opened 2022-06-08
Automatic flattening, fix for CopyLayer, comments

Atticus1806:
Since `CastLayer` is of instance `CopyLayer` it will be "skipped" during flattening which potentially causes dtype issues. This should fix the check, maybe its worth looking into the other subinstances of `CopyLayer` for other cases. 
From what I found these are: 
`DropoutLayer`, `ScaledGradientLayer`, `BatchNormLayer` and `TikhonovRegularizationLayer`. If you think only skipping `CopyLayer` would make sense I would change the whole check to:
`type(layer_) is CopyLayer`

Additionally I added some comments for some of the private functions of the flatten logic to make them more understandable, especially when seeing them first. 
Feel free to comment on them and make suggestions how to improve them, these are for now how I understood them and by far not perfect I think.
Also if you have a good suggestion for other functions in there we can add them too.

### DIFF
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -1444,7 +1444,7 @@ class TFNetwork(object):
         if isinstance(layer_, SubnetworkLayer):
           layer_ = layer_.subnetwork.layers["output"]
           continue
-        if isinstance(layer_, CopyLayer) and len(layer_.sources) == 1 and not isinstance(layer_, CastLayer):
+        if type(layer_) is CopyLayer and len(layer_.sources) == 1:
           layer_ = layer_.sources[0]
           continue
         return layer_
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -1485,13 +1485,13 @@ class TFNetwork(object):
       if layer in end_points:
         continue
       layer_queue.append(layer)
-
     while layer_queue:
       layer = layer_queue.pop(0)
       if layer in blacklist:
         continue
       blacklist.add(layer)
       layer_queue.extend(_layer_deps(layer))
+
     # Collect back refs, starting from end points.
     deps_used_by_end_points = {layer: {layer} for layer in end_points}  # dep -> set(end points), direct and indirect
     deps_used_by = {layer: set() for layer in end_points}  # dep -> set(used by), direct dependencies only
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -1437,7 +1437,7 @@ class TFNetwork(object):
       """
       Flattens the layer structure, removes irrelevant layers and returns next successor layer
       :param LayerBase layer_:
-      :return Next layer in succession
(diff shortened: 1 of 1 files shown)


returnn/tf/network.py
```diff
@@ -1480,7 +1486,6 @@ def _resolve_layer(layer_):
         continue
       blacklist.add(layer)
       layer_queue.extend(_layer_deps(layer))
-
```
albertz, line 1483: Why did you remove this empty line?

returnn/tf/network.py
```diff
@@ -1431,14 +1435,16 @@ def _check_push_flattening_to_inputs_for_layer(layer_):
 
     def _resolve_layer(layer_):
       """
+      Flattens the layer structure, removes irrelevant layers and returns next successor layer
       :param LayerBase layer_:
+      :return Next layer in succession
```
albertz, line 1440:
```suggestion
      :return: next layer in succession
```

returnn/tf/network.py
```diff
@@ -1337,6 +1337,7 @@ def _map_layer_dict_value(v):
 
     def _make_layer(layer_cls, layer_dict, map_opts=True):
       """
+      Creates the flattened layer
```
albertz, line 1340: There should be an empty line after the description before the params (when the description is non-empty).

(Here and for all further docstrings as well.)

returnn/tf/network.py
```diff
@@ -1431,14 +1435,16 @@ def _check_push_flattening_to_inputs_for_layer(layer_):
...
       :param LayerBase layer_:
+      :return Next layer in succession
       :rtype: LayerBase
       """
       while True:
         if isinstance(layer_, SubnetworkLayer):
           layer_ = layer_.subnetwork.layers["output"]
           continue
-        if isinstance(layer_, CopyLayer) and len(layer_.sources) == 1:
+        if isinstance(layer_, CopyLayer) and len(layer_.sources) == 1 and not isinstance(layer_, CastLayer):
```
albertz, line 1447: I think instead of having such explicit check for `CastLayer`, we better should do just this:
```suggestion
        if type(layer_) is CopyLayer and len(layer_.sources) == 1:
```

returnn/tf/network.py
```diff
@@ -1474,13 +1485,13 @@ def _resolve_layer(layer_):
       if layer in end_points:
         continue
       layer_queue.append(layer)
+
```
albertz, line 1488: Why did you add this empty line?

returnn/tf/network.py
albertz: Why was this removed?

returnn/tf/network.py
Atticus1806: "Restored" the wrong line, coffee not working yet :)

albertz (approved):

PR #1065, opened 2022-06-27
fix passing int axis to expand_dims

JackTemaki:
It seems there was a regression bug related to using integer axis values in `expand_dims` layers, so one of my old setups stopped working.

Not sure if my fix is correct for all cases, I first want to see the tests.


albertz:
Sorry, I think I messed up the tests. It should be fixed in master now (still waiting for tests). Can you rebase?

albertz (approved):

PR #1066, opened 2022-06-29
Refactoring code for better vectorization

maldil:
Thank you very much for your excellent work in rwth-i6/returnn.  

I am a graduate student at the University of Colorado, studying the best practices of evolving ML codes. From our research, one of the most common evolution best practice in ML code is the migration of loop-based computations, since it improves performance and code quality. We made the following changes in rwth-i6/returnn, which remove the FOR loop and use NumPy APIs. I carefully checked the modification to ensure that it does not break the code. I will gladly contribute. Please help me to merge this.

### DIFF
--- a/returnn/theano/layers/hidden.py
+++ b/returnn/theano/layers/hidden.py
@@ -2275,9 +2275,7 @@ class RandomRouteLayer(_NoOpLayer):
       if test_route >= 0:
         self.output = self.sources[test_route].output
       else:
-        output = numpy.float32(p[0]) * self.sources[0].output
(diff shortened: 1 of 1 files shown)


albertz:
The number of sources (`self.sources`) is low (usually 1) so the proposed solution is actually slower for that case, and I think the current code is also easier to read, so I would keep the current code.

Anyway, this code is very much deprecated. I would prefer to not touch it at all unless someone actively works on it.

But anyway, thanks for the contribution and trying to make RETURNN better!

maldil:
Thanks for the considering the PR

PR #1067, opened 2022-07-04
Check for tag.batch when creating size placeholder

Atticus1806:
Fix for https://github.com/rwth-i6/i6_experiments/issues/73

The check was only for batch info of the batch dim tag, but in cases where the tag has the info instead this was not working. 

I am not sure whether to actually prefer the tag batch info over the batch_dim info, but wanted to not change potential old behavior, thats why the check is for batch_dim first.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -5615,7 +5615,8 @@ def _create_size_placeholder(name, axis_wo_b, tag, batch_dim):
       batch = None
     dyn_size_ext = Data(
       "%s_dim%i_size" % (name, axis_wo_b), dtype=Data.size_dtype,
-      dim_tags=[batch_dim] if batch_dim else [], batch=batch)
+      dim_tags=[batch_dim] if batch_dim else [],
+      batch=batch if batch_dim else None)
     dyn_size = tf_compat.v1.placeholder(
       name=dyn_size_ext.name, dtype=dyn_size_ext.dtype, shape=dyn_size_ext.batch_shape)
     dyn_size_ext.placeholder = dyn_size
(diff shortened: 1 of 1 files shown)


Atticus1806:
I was not sure about that. The issue was caused by the `dyn_size_ext` batch info so I sticked with that, but probably it makes sense to change that aswell. 
change it to: `if tag.batch is not None and batch_dim and batch_dim.batch:` ?

albertz:
The sanity check checks that `tag.batch == dyn_size_ext.batch`, right?
Maybe we should simply do (without any
```python
if batch_dim:
  tag.batch = batch
```

returnn/tf/util/data.py
```diff
@@ -5607,9 +5607,15 @@ def _create_size_placeholder(name, axis_wo_b, tag, batch_dim):
...
+    if batch_dim is not None and batch_dim.batch is not None:
+      batch = batch_dim.batch
+    elif tag.batch is not None:
+      batch = tag.batch
+    else:
+      batch = None
     dyn_size_ext = Data(
       "%s_dim%i_size" % (name, axis_wo_b), dtype=Data.size_dtype,
-      dim_tags=[batch_dim] if batch_dim else [], batch=batch_dim.batch if batch_dim else None)
+      dim_tags=[batch_dim] if batch_dim else [], batch=batch)
```
albertz, line 5618: I wonder, wouldn't that make more sense:
```suggestion
      dim_tags=[batch_dim] if batch_dim else [],
      batch=batch if batch_dim else None)
```

Atticus1806:
>The sanity check checks that tag.batch == dyn_size_ext.batch, right?
Maybe we should simply do:
if batch_dim:
  tag.batch = batch

Thats how it is right now right? (Besides the None check)

albertz (approved):

albertz:
I think these changes were actually a bit problematic. I did some follow-up in #1068.

PR #1072, opened 2022-07-06
sprint.cache, more clear code

DanEnergetics:
This is a straight-forward fix for the ambiguous use of the variable "typ". It is overwritten multiple times and did not retain the desired information about the alignment type. This caused undesired behavior of the SprintCacheDataset which disregarded the substate.

### DIFF
--- a/returnn/sprint/cache.py
+++ b/returnn/sprint/cache.py
@@ -299,9 +299,9 @@ class FileArchive:
 
     elif typ == "feat":
       type_len = self.read_U32()
-      typ = self.read_str(type_len)
+      type_ = self.read_str(type_len)
       # print(typ)
-      assert typ == "vector-f32"
+      assert type_ == "vector-f32"
       count = self.read_U32()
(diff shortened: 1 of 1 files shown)


michelwi:
The same overwriting exists for the "feat" branch above (line 302ff). Maybe you can add the same fix there as well?

returnn/sprint/cache.py
```diff
@@ -313,11 +313,10 @@ def _raw_read(self, size, typ):
 
     elif typ in ["align", "align_raw"]:
       type_len = self.read_U32()
-      typ = self.read_str(type_len)
-      assert typ == "flow-alignment"
+      assert self.read_str(type_len) == "flow-alignment"
```
albertz, line 316: Can you keep that as before (and maybe just rename the var)? That makes the error reporting much better (it will print what you got here from `read_str`).
I.e.
```suggestion
      type_ = self.read_str(type_len)
      assert type_ == "flow-alignment"
```

albertz (approved):

PR #1073, opened 2022-07-08
extend check for ndarray in CachedDataset

michelwi:
I have a MetaDataset which combines multiple HDFDatasets. During construction, the case occurred that `self._seq_index` was a `numpy.ndarray` but `seq_index` was of type `list`.

### DIFF
--- a/returnn/datasets/cached.py
+++ b/returnn/datasets/cached.py
@@ -86,7 +86,7 @@ class CachedDataset(Dataset):
     old_index_map = self._index_map[:]
     self._index_map = range(len(seq_index))  # sorted seq idx -> seq_index idx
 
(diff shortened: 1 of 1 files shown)


albertz (approved):

albertz:
I'm merging this already, as this itself should be fine.

But there might be further things missing. @patrick-wilken you should check this.

patrick-wilken:
Ah, yes, MetaDataset may initialize first with 'default' and then with  e.g. 'sorted' sequence order and then the seq_index lists can have different types. The fix is correct.

(For context: I introduced `typing.Sequence` as return value of `get_seq_order_for_epoch()` mainly to support ranges in case of 'default' sequence order, which is much more efficient than lists.)

PR #1076, opened 2022-07-13
ConcatSeqsDataset, option to match alignment

albertz:
Example: you have downsampling factor 6, i.e. ceildiv(data_len, 6) == align_len.

Now it could happen that ceildiv(data_len1 + data_len2, 6) < align_len1 + align_len2.

This option would repeat intermediate ending frames such that data_len1 % 6 == 0.

Also part of this PR is some rewrite of `StaticDataset`, somewhat related to https://github.com/rwth-i6/returnn/pull/329.

### DIFF
--- a/returnn/datasets/generating.py
+++ b/returnn/datasets/generating.py
@@ -854,7 +854,7 @@ class DummyDatasetMultipleDataKeys(DummyDataset):
     return DatasetSeq(seq_idx=seq_idx, features=features, targets=None)
 
 
-class StaticDataset(GeneratingDataset):
+class StaticDataset(CachedDataset2):
   """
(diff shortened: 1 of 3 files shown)


patrick-wilken:
You you don't see a use case for sequence ordering of on-the-fly data? Or are you saying again that "real" datasets should never be created from a full in-memory list?

albertz:
> You you don't see a use case for sequence ordering of on-the-fly data? Or are you saying again that "real" datasets should never be created from a full in-memory list?

The latter. Only toy datasets fit into memory. Or even if your real dataset does currently fit into memory, I would still argue to not use it, because scaling it up later will then require more work because you need to change the dataset then.

PR #1085, opened 2022-07-20
Option for setting blank index in forced_align Layer

Atticus1806:
Similar to e.g. `ctc_loss`  in `native_op.py` this adds the option to set the blank_index for the forced_align layer. Default is still the same as before (last index in the vocab). 
Also adds documentation for the parameter to a few other functions which have the same parameter.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -9099,11 +9099,12 @@ class ForcedAlignmentLayer(_ConcatInputLayer):
   """
   layer_class = "forced_align"
 
-  def __init__(self, align_target, topology, input_type, **kwargs):
+  def __init__(self, align_target, topology, input_type, blank_index=-1, **kwargs):
     """
     :param LayerBase align_target:
     :param str topology: e.g. "ctc" or "rna" (RNA is CTC without label loop)
     :param str input_type: "log_prob" or "prob"
+    :param int blank_index: vocab index of the blank symbol
     """
     from returnn.tf.native_op import get_ctc_fsa_fast_bw, fast_viterbi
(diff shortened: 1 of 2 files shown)


Atticus1806:
I found that the naming is not consistent in all functions, `blank_index` vs `blank_idx`, maybe it makes sense to agree on one way to write it and change the other in this PR aswell. I am fine with both, not sure how the naming convention in RETURNN would suggest to name this.

albertz:
Regarding naming convention, I have no idea whether "idx" or "index" is nicer. How do other projects (PyTorch, TensorFlow) have this? Or also not consistent?

For the internal API (native functions), I think it's also less important (although some external code might also use that directly).

Atticus1806:
So for both Tensorflow and PyTorch found occurences of both idx and index in some part of the code so I guess they dont really care. 
Its fine for me to just leave it like it is, I just noticed it when doing the docs, then I guess this is all ready for review/merge.

albertz:
Specifically about blank_idx vs blank_index, we also have both in public APIs, right?

`FastBaumWelchLayer` `ctc_opts` are for `native_op.get_ctc_fsa_fast_bw` which uses `blank_idx`.

`CtcLoss` `ctc_opts` are for `native_op.ctc_loss` which uses `blank_index`.

Other examples? So, what should `ForcedAlignmentLayer` be more consistent to? I would maybe argue `FastBaumWelchLayer`, i.e. `blank_idx`?

Atticus1806:
Sounds good, parameter of the internal function is `blank_idx` aswell.

albertz (approved):

PR #1089, opened 2022-08-04
GatherLayer on batch axis

### LINKED ISSUE #1087: GatherLayer on batch axis
I have a case where I'd like to use the GatherLayer on the batch axis. This works, but the size placeholder is not gathered. Does anything speak against gathering on the batch axis or was this just not considered before and we should fix the size placeholder?

Not gathering the size placeholder leads to an error later in my model and it also doesn't make sense to keep the same size placeholder in general.

Here is a test case to demonstrate the issue:
```
def test_GatherLayer_batch_dim():
  with make_scope() as session:
    import numpy as np
    net = TFNetwork(extern_data=ExternData())
    batch_dim, time_dim, feature_dim = 3, 4, 2
    # [B, T, F]
    random = np.random.RandomState(42)
    values_seqs = random.rand(batch_dim, time_dim, feature_dim).astype('float32')
    values_size = np.array([4, 2, 3])
    values_placeholder = tf.constant(values_seqs, dtype=tf.float32)
    values_size_placeholder = {0: tf.constant(values_size, dtype=tf.int32)}
    values = InternalLayer(
      name="values", network=net,
      output=Data(
        name="values",
        batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=2,
        shape=[None, feature_dim],
        placeholder=values_placeholder,
        size_placeholder=values_size_placeholder,
      ))
    position_np = np.array([0, 2])
    position = InternalLayer(
      name="position", network=net,
      output=Data(
        name="position",
        placeholder=tf.constant(position_np, dtype=tf.int64),
        batch_dim_axis=0, shape=[], dtype="int64",
      ))
    values.output.sanity_check()
    position.output.sanity_check()

    # should become [B', T, F]
    layer = GatherLayer(
      name="gather", network=net,
      sources=[values], position=position, axis="B",
      output=GatherLayer.get_out_data_from_opts(
        name="gather", sources=[values], position=position, axis="B"))
    layer.output.sanity_check()
    out_seqs, out_size = session.run([layer.output.placeholder, layer.output.size_placeholder.as_dict()])
    assert isinstance(out_seqs, numpy.ndarray)

    np.testing.assert_equal(values_seqs[position_np, :], out_seqs)
    np.testing.assert_equal(values_size[position_np], out_size[0])
```

vieting:
This PR fixes #1087. As I face the issue in the context of supervised multilingual training, I added a more general test case also for that which does not necessarily need to go into the main branch. The fix is similar to how the `size_placeholder` is modified in the `ShiftAxisLayer`.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1341,6 +1341,17 @@ class GatherLayer(_ConcatInputLayer):
     # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for axis in self.output.size_placeholder:
+        new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+        from ..util.data import Dim
+        Dim(
+          kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+          dyn_size=new_size, batch=self.output.batch,
+          src_data=self.output, src_axis=axis, auto_generated=True)
+        self.output.size_placeholder[axis] = new_size
+
+
   @classmethod
   def _get_common_input_position_axes(cls, input_data, position_data, old_gather_axis):
     """
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1342,14 +1342,16 @@ class GatherLayer(_ConcatInputLayer):
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
     if input_data.dim_tags[old_gather_axis].is_batch_dim():
-      for axis in self.output.size_placeholder:
-        new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
-        from ..util.data import Dim
-        Dim(
-          kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
-          dyn_size=new_size, batch=self.output.batch,
-          src_data=self.output, src_axis=axis, auto_generated=True)
-        self.output.size_placeholder[axis] = new_size
+      for dim_tag in self.output.dim_tags:
+        if dim_tag.is_spatial_dim():
+          axis = self.output.get_batch_axis_excluding_batch(self.output.get_axis_by_tag_name(dim_tag.description))
+          new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+          from ..util.data import Dim
+          Dim(
+            kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+            dyn_size=new_size, batch=self.output.batch,
+            src_data=self.output, src_axis=axis, auto_generated=True)
+          self.output.size_placeholder[axis] = new_size
 
 
   @classmethod
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -5456,6 +5456,51 @@ def test_GatherLayer_broadcast_dim():
   })
 
 
+def test_GatherLayer_batch_dim():
+  with make_scope() as session:
+    import numpy as np
+    net = TFNetwork(extern_data=ExternData())
+    batch_dim, time_dim, feature_dim = 3, 4, 2
+    # [B, T, F]
+    random = np.random.RandomState(42)
+    values_seqs = random.rand(batch_dim, time_dim, feature_dim).astype('float32')
+    values_size = np.array([4, 2, 3])
+    values_placeholder = tf.constant(values_seqs, dtype=tf.float32)
+    values_size_placeholder = {0: tf.constant(values_size, dtype=tf.int32)}
+    values = InternalLayer(
+      name="values", network=net,
+      output=Data(
+        name="values",
+        batch_dim_axis=0, time_dim_axis=1, feature_dim_axis=2,
+        shape=[None, feature_dim],
+        placeholder=values_placeholder,
+        size_placeholder=values_size_placeholder,
+      ))
+    position_np = np.array([0, 2])
+    position = InternalLayer(
+      name="position", network=net,
+      output=Data(
+        name="position",
+        placeholder=tf.constant(position_np, dtype=tf.int64),
+        batch_dim_axis=0, shape=[], dtype="int64",
+      ))
+    values.output.sanity_check()
+    position.output.sanity_check()
+
+    # should become [B', T, F]
+    layer = GatherLayer(
+      name="gather", network=net,
+      sources=[values], position=position, axis="B",
+      output=GatherLayer.get_out_data_from_opts(
+        name="gather", sources=[values], position=position, axis="B"))
+    layer.output.sanity_check()
+    out_seqs, out_size = session.run([layer.output.placeholder, layer.output.size_placeholder.as_dict()])
+    assert isinstance(out_seqs, numpy.ndarray)
+
+    np.testing.assert_equal(values_seqs[position_np, :], out_seqs)
+    np.testing.assert_equal(values_size[position_np], out_size[0])
+
+
 def test_SliceNdLayer():
   n_batch = 5
   n_time = 7
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -9258,6 +9258,62 @@ def test_contrastive_loss():
     assert numpy.isfinite(loss_v)
 
 
+def test_supervised_multilingual_training():
+  n_batch = 3
+  out_dim = 5
+
+  net_dict = {
+    "encoder": {"class": "copy", "from": "data"},
+
+    "loss_language_0": {
+      "class": "subnetwork",
+      "from": "data:classes",
+      "is_output_layer": True,
+      "subnetwork": {
+        # create language id indices based on where we see non-silence in the alignment.
+        # this should be done in the dataset in a neat way.
+        "aux_0": {"class": "gather", "position": 0, "axis": "T", "from": "data"},
+        "aux_1": {"class": "compare", "kind": "greater", "value": 2, "from": "aux_0"},
+        "idx": {  # shape (B',) with B' < B (B' is the number of utterances in the batch for this language)
+          "class": "eval",
+          "eval": "tf.squeeze(tf.where(source(0)), axis=-1)",
+          "out_type": {"dtype": "int64"},
+          "from": "aux_1",
+        },
+
+        # gather targets and encoder outputs
+        "tgt": {"class": "gather", "from": "data", "axis": "B", "position": "idx"},  # B', T (sparse)
+        "enc_raw": {"class": "gather", "from": "base:encoder", "axis": "B", "position": "idx"},  # B', T, F
+        "enc": {"class": "reinterpret_data", "size_base": "tgt", "from": "enc_raw"},  # B', T, F
+        "logits": {"class": "linear", "n_out": out_dim, "from": "enc"},  # B', T, n_out
+        "output": {
+          "class": "sparse_softmax_cross_entropy_with_logits",
+          "logits": "logits",
+          "targets": "tgt",
+          "loss": "as_is",
+        },
+      },
+    },
+  }
+
+  with make_scope() as session:
+    config = Config({
+      "extern_data": {
+        "data": {"dim": 7, "batch_dim_axis": 0, "time_dim_axis": 1, "feature_dim_axis": 2},
+        "classes": {"dim": out_dim, "sparse": True}},
+      "debug_print_layer_output_shape": True,  # "debug_print_layer_output": True,
+    })
+    net = TFNetwork(config=config, train_flag=True)
+    net.construct_from_dict(net_dict)
+    loss = net.get_total_loss()
+
+    tf_compat.v1.set_random_seed(1)
+    net.initialize_params(session)
+    loss_v = session.run(loss, feed_dict=make_feed_dict(net.extern_data, same_time=True, n_batch=n_batch))
+    print("loss:", loss_v)
+    assert numpy.isfinite(loss_v)
+
+
 if __name__ == "__main__":
   try:
     better_exchook.install()
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -5457,13 +5457,14 @@ def test_GatherLayer_broadcast_dim():
 
 
 def test_GatherLayer_batch_dim():
+  from returnn.tf.util.data import batch_dim
   with make_scope() as session:
     import numpy as np
     net = TFNetwork(extern_data=ExternData())
-    batch_dim, time_dim, feature_dim = 3, 4, 2
+    b_dim, time_dim, feature_dim = 3, 4, 2
     # [B, T, F]
     random = np.random.RandomState(42)
-    values_seqs = random.rand(batch_dim, time_dim, feature_dim).astype('float32')
+    values_seqs = random.rand(b_dim, time_dim, feature_dim).astype('float32')
     values_size = np.array([4, 2, 3])
     values_placeholder = tf.constant(values_seqs, dtype=tf.float32)
     values_size_placeholder = {0: tf.constant(values_size, dtype=tf.int32)}
@@ -5475,6 +5476,7 @@ def test_GatherLayer_batch_dim():
         shape=[None, feature_dim],
         placeholder=values_placeholder,
         size_placeholder=values_size_placeholder,
+        dim_tags=(batch_dim, SpatialDim("time"), FeatureDim("feature", feature_dim)),
       ))
     position_np = np.array([0, 2])
     position = InternalLayer(


vieting:
Hi @albertz, what do you think about the way the size placeholder and dim tag are modified in general? Right now there is a failing test case, where we first do `flatten_batch` and then gather on the batch axis. I'm not very sure how the desired behavior in this case would look like. It'd be nice if you could comment on what you think here.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,17 @@ def __init__(self, position, axis, **kwargs):
...
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for axis in self.output.size_placeholder:
+        new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+        from ..util.data import Dim
+        Dim(
+          kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+          dyn_size=new_size, batch=self.output.batch,
+          src_data=self.output, src_axis=axis, auto_generated=True)
+        self.output.size_placeholder[axis] = new_size
```
albertz, line 1352: You don't use the `Dim` object you created?
Instead of assigning `size_placeholder`, I think it would be better to set the newly created dim tag.

tests/test_TFNetworkLayer.py
```diff
@@ -9213,6 +9258,62 @@ def _get_mask_eval_layer(source, **_kwargs):
...
+          "class": "eval",
+          "eval": "tf.squeeze(tf.where(source(0)), axis=-1)",
+          "out_type": {"dtype": "int64"},
+          "from": "aux_1",
+        },
+
+        # gather targets and encoder outputs
+        "tgt": {"class": "gather", "from": "data", "axis": "B", "position": "idx"},  # B', T (sparse)
+        "enc_raw": {"class": "gather", "from": "base:encoder", "axis": "B", "position": "idx"},  # B', T, F
+        "enc": {"class": "reinterpret_data", "size_base": "tgt", "from": "enc_raw"},  # B', T, F
```
albertz, line 9287: Why is this needed?

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,17 @@ def __init__(self, position, axis, **kwargs):
     # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for axis in self.output.size_placeholder:
```
albertz, line 1345: Instead of iterating over the `size_placeholder`, you should iterate over the dim tags.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,17 @@ def __init__(self, position, axis, **kwargs):
...
+        new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+        from ..util.data import Dim
+        Dim(
+          kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+          dyn_size=new_size, batch=self.output.batch,
```
albertz, line 1350: You miss `dyn_size_ext` here.

albertz:
When you modify the batch dim, you should create a new `BatchInfo` object as well, and assign that to `output`.

returnn/tf/layers/basic.py
albertz: You should not set `dyn_size` in case it is non-standard. Set `dyn_size_ext` instead.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,17 @@ def __init__(self, position, axis, **kwargs):
     # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for axis in self.output.size_placeholder:
+        new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
```
albertz, line 1346: You assume that `position_data` is of shape [new-batch]?

vieting: Right, in the case I have in mind yes. But for the failing test case, this is different and we need to take this into account.

vieting:
> When you modify the batch dim, you should create a new `BatchInfo` object as well, and assign that to `output`.

As I said, the fix is similar to what is done in the `ShiftAxisLayer`. Do you have another layer which modifies the batch axis and could serve as a good example?

albertz:
> > When you modify the batch dim, you should create a new `BatchInfo` object as well, and assign that to `output`.
> 
> As I said, the fix is similar to what is done in the `ShiftAxisLayer`. Do you have another layer which modifies the batch axis and could serve as a good example?

`ShiftAxisLayer` does not modify the batch dim. You probably mean the size adoption. That code is a bit ugly/outdated/deprecated/hacky in `ShiftAxisLayer`, and might not work correct in all cases (but anyway it's simpler because the batch dim is not changed).

albertz:
> Do you have another layer which modifies the batch axis and could serve as a good example?

Not many layers do that. I just recall `FlattenBatchLayer` right now.

returnn/tf/layers/basic.py
albertz: What is it in that case?

vieting: There it's of shape [B,T,F], however, in the input B and T are packed
```
>>> input_data                                                                                                                                                                                                                  
Data{'flat_output', [B&Packed{'time'},F|F'feature'(5)]}
>>> self.output                                                                                                                                                                                                                 
Data{'output_output', [B,T|'time'[B],'other-spatial'(7),F|F'feature'(5)]}
>>> position_data                                                                                                                                                                                                               
Data{'indices_flat_output', [B,T|'time'[B],F|'other-spatial'(7)], dtype='int32'}
```

albertz: Which test case is that? The one you added? `test_rand_indices`?
Why is `position_data` of this shape? As described, it should have some new-batch dim in it, right? Or basically just the shape [new-batch]? When you gather into the batch dim. It definitely should not have the old batch dim in its shape.

vieting: This is from `test_rand_indices` which is an existing test case. These are the layer templates
```
layer /'data': [B,T|'time'[B],F|F'feature'(5)] float32
layer /'flat': [B&Packed{'time'},F|F'feature'(5)] float32
layer /'length_flat': [] int32
layer /'indices_flat': [B,T|'time'[B],F|'other-spatial'(7)] int32
layer /'output': [B,T|'time'[B],'other-spatial'(7),F|F'feature'(5)] float32
```
It gathers the flattened input with random indices of shape B,T,F'. That's how the test was written. So it's gathering on the flattened B&T dim, not a regular batch dim

vieting: This is the net dict:
```
    sz = (batch_dim, time_dim, SpatialDim("other-spatial", 7))
    net.construct_from_dict({
      "flat": {"class": "flatten_batch", "from": "data"},
      "length_flat": {"class": "length", "from": "flat", "axis": batch_dim},
      "indices_flat": {"class": "rand_int", "shape": sz, "minval": 0, "maxval": "length_flat", "seed": 42},
      "output": {"class": "gather", "from": "flat", "axis": batch_dim, "position": "indices_flat"},
    })
```

albertz: Ah yes, sorry, yes, this is a valid case, the shape of `position_data` makes sense here.

What is the output shape of `tf.gather` here, i.e. of `new_size`?

Operating (assigning) `size_placeholder` is problematic here, as this always assumes shape [B]. But I think your `new_size` has actually the same shape as `position_data`.

What is `self.output.size_placeholder[axis]` actually before? Or rather the `dyn_size_ext` of the dim tags of `input_data`?

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,19 @@ def __init__(self, position, axis, **kwargs):
...
+      for dim_tag in self.output.dim_tags:
+        if dim_tag.is_spatial_dim():
+          axis = self.output.get_batch_axis_excluding_batch(self.output.get_axis_by_tag_name(dim_tag.description))
+          new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+          from ..util.data import Dim
+          Dim(
+            kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+            dyn_size=new_size, batch=self.output.batch,
+            src_data=self.output, src_axis=axis, auto_generated=True)
+          self.output.size_placeholder[axis] = new_size
```
albertz, line 1354: You should not assign `size_placeholder` but rather the dim tags.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,19 @@ def __init__(self, position, axis, **kwargs):
...
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for dim_tag in self.output.dim_tags:
+        if dim_tag.is_spatial_dim():
+          axis = self.output.get_batch_axis_excluding_batch(self.output.get_axis_by_tag_name(dim_tag.description))
+          new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
+          from ..util.data import Dim
+          Dim(
+            kind=Dim.Types.Spatial, description="%s_gather_axis" % self.name,
+            dyn_size=new_size, batch=self.output.batch,
```
albertz, line 1352: You should not assign `dyn_size` but rather `dyn_size_ext`.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,19 @@ def __init__(self, position, axis, **kwargs):
     # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for dim_tag in self.output.dim_tags:
+        if dim_tag.is_spatial_dim():
+          axis = self.output.get_batch_axis_excluding_batch(self.output.get_axis_by_tag_name(dim_tag.description))
+          new_size = tf.gather(params=self.output.size_placeholder[axis], indices=position_data.placeholder)
```
albertz, line 1348: You should not access `size_placeholder` but rather `dim_tag.dyn_size_ext`.

returnn/tf/layers/basic.py
```diff
@@ -1341,6 +1341,19 @@ def __init__(self, position, axis, **kwargs):
     # (BatchAxes.., InputAxesBeforeGatherAxis, PositionAxes.., InputAxesAfterGatherAxis..)
     self.output.placeholder = tf.gather(params=params, indices=indices, axis=gather_axis, batch_dims=batch_dims)
 
+    if input_data.dim_tags[old_gather_axis].is_batch_dim():
+      for dim_tag in self.output.dim_tags:
+        if dim_tag.is_spatial_dim():
+          axis = self.output.get_batch_axis_excluding_batch(self.output.get_axis_by_tag_name(dim_tag.description))
```
albertz, line 1347: This is:

* way too complicated: you can simply do `for axis, dim_tag in enumerate(self.output.dim_tags)`
* wrong: do not rely on `get_axis_by_tag_name` and `dim_tag.description`
* not necessary: just use `dim_tag.dyn_size_ext`

returnn/tf/layers/basic.py
vieting: What is the usual way to set dim tags? I can't just reassign `self.output.dim_tags`. `declare_same_as` is used elsewhere, but not sure if it applies here.

returnn/tf/layers/basic.py
vieting: Yes, `new_size` has the same shape as `position_data`.
```
>>> new_size                                                                                                                                                                                                                    
<tf.Tensor 'output/GatherV2_1:0' shape=(?, ?, 7) dtype=int32>
>>> position_data.placeholder                                                                                                                                                                                                   
<tf.Tensor 'indices_flat/random_uniform:0' shape=(?, ?, 7) dtype=int32>
>>> position_data                                                                                                                                                                                                               
Data{'indices_flat_output', [B,T|'time'[B],F|'other-spatial'(7)], dtype='int32'}
```

vieting: `self.output.size_placeholder[axis]` is of shape (B,) before.
```
>>> self.output.size_placeholder[axis]                                                                                                                                                                                          
<tf.Tensor 'extern_data/placeholders/data/data_dim0_size:0' shape=(?,) dtype=int32>
```
`dyn_size_ext` is not set on the dim tags in `input_data`.

returnn/tf/layers/basic.py
albertz: See most other layers. Usually you set `dim_tags` in `get_out_data_from_opts`. You should not assign a new dim tag in `__init__`. In `__init__`, you just might to assign the `dyn_size_ext` or `dyn_size_ext.placeholder` of a dim tag which was previously newly created in `get_out_data_from_opts`.

returnn/tf/layers/basic.py
albertz: I wonder a bit: All the output dim tags here are already defined ([B,T|'time'[B],'other-spatial'(7),F|F'feature'(5)]). It would be wrong to overwrite the `dyn_size_ext` of any of them. So either the output dim tags are wrong (you actually want to have another time dim here?) or nothing needs to be done for this case.

albertz: > `dyn_size_ext` is not set on the dim tags in `input_data`.

Yes because `input_data` does not have any dim tags which need that. But you are speaking about the dim tags of `self.output` here? But anyway, see my previous comment.

vieting: > I wonder a bit: All the output dim tags here are already defined ([B,T|'time'[B],'other-spatial'(7),F|F'feature'(5)]). It would be wrong to overwrite the dyn_size_ext of any of them. So either the output dim tags are wrong (you actually want to have another time dim here?) or nothing needs to be done for this case.

Yes, I also think `dyn_size_ext` should not be changed. Regarding the different time dim I'm not sure, that's what I meant initially when saying I don't know what the desired behavior is in this case. If you don't have a clear opinion on this, I'd vote for not changing anything.

albertz: Well, this should not be a question on opinion, but what makes sense. There should be a clear answer. Or rather, the ambiguity just comes from our lack of really understanding it, I think. But actually, if I understand it right, it's actually quite clear: The axis where the gather operates on is replaced by all the axes of `position_data`. The axes of `position_data` already exist, and all have already defined dim tags (incl dyn size), so nothing needs to be done about them, no new dyn size needs to be calculated. Maybe we still miss something in this understanding, but with this understanding, we should just leave it as it is.

albertz: Actually, ok, one case where it needs a change: there is really a new batch dim. So some batch dim gets replaced by a new batch dim.

tests/test_TFNetworkLayer.py
```diff
@@ -5456,6 +5456,53 @@ def test_GatherLayer_broadcast_dim():
...
+        dim_tags=(batch_dim, SpatialDim("time"), FeatureDim("feature", feature_dim)),
+      ))
+    position_np = np.array([0, 2])
+    position = InternalLayer(
+      name="position", network=net,
+      output=Data(
+        name="position",
+        placeholder=tf.constant(position_np, dtype=tf.int64),
+        batch_dim_axis=0, shape=[], dtype="int64",
+      ))
```
vieting, line 5488: @albertz do I need to change the creation of `position` in order to make it have a different batch axis dim tag here?

returnn/tf/layers/basic.py
albertz: So, basically, what needs to be done:

You need to check the case that you gather into the batch dim (as you do here):

Then you need to iterate over the output dim tags (similar as you do here).

For each, you need to check its dyn_size_ext, whether it contains the batch dim which you are going to replace/gather here. In your example here, it is `B&Packed{'time'}`, but this dim is not in any of the output dim tags, so nothing needs to be done. If it would be there, then you actually would need the same gather logic as you do for `output` itself.

tests/test_TFNetworkLayer.py
albertz: Yes. It's actually not so simple because of the special treatment of the batch dim tag. I'm not sure it's really possible currently.

In practice, in your real code, how would you end up with `position`?

returnn/tf/layers/basic.py
vieting: How do I check this? In the case with `B&Packed{'time'}`, I still get
```
>>> input_data.dim_tags[0] == self.output.dim_tags[0]                                                                                                                                                                           
True
```
Even though the input has the packed one and the output the global batch. See https://github.com/rwth-i6/returnn/blob/master/returnn/tf/util/data.py#L766 Do we need to extend that logic for this case?

albertz: Yea, that's the problem I mentioned elsewhere, that dim tag equality is handled differently for the batch dim...

Here, in this case, we need different logic for the equality check of the batch dim. For example, you can compare `batch_dim.batch` or `data.batch`.

tests/test_TFNetworkLayer.py
vieting: > In practice, in your real code, how would you end up with position?
Do you mean what dim tag I get there?
```
>>> position.output.dim_tags[0].description                                                                                                                                                                                     
'batch:position'
```
So it's not actually the global batch dim. I was just confused because I got
```
>>> position.output.dim_tags[0] == values.output.dim_tags[0]                                                                                                                                                                    
True
```
but this is because the check does not cover this case, see comment here: https://github.com/rwth-i6/returnn/pull/1089#discussion_r955964986

returnn/tf/layers/basic.py
albertz: In the output, you anyway already need to assign `output.batch`. So you can compare `input_data.batch != output.batch`. For each output dim tag, you should call `get_for_batch_ctx`. And then in `__init__`, the `dyn_size_ext` should be unknown for the case that there is a new batch dim, and if so, then you need the gather logic also on it. You get the old original dyn size again via `get_for_batch_ctx`.

vieting: `*.batch` is `None` here.

albertz: It should not be `None`. For `input_data`? Or `output`? Or `position_data`? For `output`, you need to assign it.

vieting: Right now, it's `None` for all.

vieting: I see that I need to assign it for `output`. But it should come from `position_data`, right?

albertz: Why is it `None` for `position_data`? I don't mean in the test case, I mean in the real case which motivated this test case. In the real case, you would not have such `InternalLayer`.

albertz: It should never be done if the data has a batch dim, unless sth is wrong. In case of the test case, then the test case is buggy.

vieting: Right, this is about the test case. However, in the case that I'm interested in, still `input_data.batch == position_data.batch` is `True`. This is probably because I'm using an `EvalLayer` to get the batch indices from a 0/1 vector with shape (B,) and that `EvalLayer` does not set the output correctly. Then we would need a layer which does that correctly, right?

albertz: An `EvalLayer` should never change the shape. If it does, and you are not very careful in setting the output data, then yes, this is a bug in your config.

vieting:
As discussed offline, it is possible to get the desired results in my use case using the `MaskedComputationLayer`. Instead of the indices to gather, we need a boolean mask over the batch axis. In my use case, I have this anyway and only computed the indices from the mask. We can use the mask like this:
```
network = {
    "encoder": {...},  # B, T, F
    "boolean_mask": {...},  # B
    "encoder_masked": {
        "class": "masked_computation",
        "mask": "boolean_mask",
        "unit": {"class": "copy", "from": "encoder"}
    },  # B', T, F
    ...
}
```
Since that does exactly what I need, I'll close this PR and the corresponding issue.

albertz:
Well, GatherLayer on batch axis is still maybe sometimes a valid thing someone wants to do. I would leave this PR open.

PR #1090, opened 2022-08-05
Pitch extraction

Atticus1806:
Adds librosa pitch extraction as option for feature extraction.

### DIFF
--- a/returnn/datasets/util/feature_extraction.py
+++ b/returnn/datasets/util/feature_extraction.py
@@ -201,6 +201,7 @@ class ExtractAudioFeatures:
         feature_data = _get_audio_linear_spectrogram(**kwargs)
       elif self.features == "f_0":
         kwargs.pop("num_feature_filters")
+        kwargs.pop("window_len")
         feature_data = _get_f0_values(**kwargs)
       else:
         raise Exception("non-supported feature type %r" % (self.features,))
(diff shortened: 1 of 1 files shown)


Atticus1806:
This is ready now and tested to produce plausible outputs. 
Swapped away from librosa to pyworld because it is a lot faster.
Since librosa is also not in the requirements I assume this is fine and if a user wants to use it they have to properly install it?

albertz:
What does the "WIP" in the title mean? If this is WIP, mark it as draft, and remove the WIP from the title. If this is not WIP, remove the WIP from the title.

returnn/datasets/util/feature_extraction.py
```diff
@@ -199,6 +199,10 @@ def get_audio_features(self, audio, sample_rate, seq_name=None):
         feature_data = _get_audio_db_mel_filterbank(**kwargs)
       elif self.features == "linear_spectrogram":
         feature_data = _get_audio_linear_spectrogram(**kwargs)
+      elif self.features == "f_0":
```
albertz, line 202: `f_0` is inconsistent to `f0` as used elsewhere. I would suggest here:
```suggestion
      elif self.features == "f0":
```

returnn/tf/layers/basic.py
```diff
@@ -9099,12 +9099,13 @@ class ForcedAlignmentLayer(_ConcatInputLayer):
   """
   layer_class = "forced_align"
 
-  def __init__(self, align_target, topology, input_type, blank_idx=-1, **kwargs):
+  def __init__(self, align_target, topology, input_type, blank_idx=-1, blank_included=False, **kwargs):
```
albertz, line 9102: These changes look like they don't belong here?

returnn/datasets/util/feature_extraction.py
```diff
@@ -467,3 +471,22 @@ def _get_random_permuted_audio(audio, sample_rate, opts, random_state):
...
+  Calls pyworld dio and stonemask to retrieve f_0 values and replaces all NaN positions(silence) with 0
+  :param numpy.ndarray audio: raw time signal
+  :param int sample_rate: e.g. 22050
+  :param float step_len: in seconds
+  :param int fmin: minimum frequency covered by mel filters
+  :param int|None fmax: maximum frequency covered by mel filters
+  :rtype: numpy.ndarray
+  :return: Pitch features for audio signal
+  """
+  import pyworld as pw  # noqa
```
albertz, line 487: Is this really not possible via librosa?

albertz: https://librosa.org/doc/main/generated/librosa.pyin.html ?

albertz: Ah I just read your comment:

> Swapped away from librosa to pyworld because it is a lot faster.

This should be here in the code as a comment.

Atticus1806: For 10000 sequences `librosa.pyin` (which is what some papers which dont use pyworld use) took 1-2 days while this takes about 2 hours and is more far used in the papers for pitch prediction I read. 
E.g. EspNet2 uses [dio](https://github.com/espnet/espnet/blob/b008ac7d58e9ced1a9f8c89cc85ee69d9e9461ab/espnet2/tts/feats_extract/dio.py) with pyworld.

PR #1092, opened 2022-08-09
init_seq_order: consistently treat empty seq_list/seq_order to mean 0 sequences

patrick-wilken:
A sampling size of 0 did not work in `CombinedDataset` with `TranslationDataset` as subdatasets because an empty `seq_order` list as parameter of `init_seq_order()` was interpreted as "no sequence order given" in `TranslationDataset`. The dataset then creates a sequence order itself with more sequences which makes the "all_sequences_used" check in `CombinedDataset._expand_dataset_sec_idxs()` fail.

I tried to fix it in all places now. Maybe a bit pedantic, but necessary at least for this special use case.

### DIFF
--- a/returnn/datasets/cached.py
+++ b/returnn/datasets/cached.py
@@ -444,9 +444,7 @@ class CachedDataset(Dataset):
 
   @property
   def num_seqs(self):
-    if self._index_map:
-      return len(self._index_map)
-    return self._num_seqs
+    return len(self._index_map)
 
   def is_cached(self, start, end, blocking = False):
     """
--- a/returnn/datasets/generating.py
+++ b/returnn/datasets/generating.py
@@ -55,7 +55,8 @@ class GeneratingDataset(Dataset):
     This is called when we start a new epoch, or at initialization.
     """
     super(GeneratingDataset, self).init_seq_order(epoch=epoch)
-    assert not seq_list and not seq_order, "predefined order doesn't make sense for %s" % self.__class__.__name__
+    assert seq_list is None and seq_order is None, (
+      "predefined order doesn't make sense for %s" % self.__class__.__name__)
     self.random.seed(self.fixed_random_seed or self._get_random_seed_for_epoch(epoch=epoch))
     self._num_timesteps = 0
     self.reached_final_seq = False
--- a/returnn/datasets/lm.py
+++ b/returnn/datasets/lm.py
@@ -1390,7 +1390,7 @@ class TranslationDataset(CachedDataset2):
 
     if seq_list is None and self.seq_list:
       seq_list = self.seq_list
-    if seq_order:
+    if seq_order is not None:
       self._seq_order = seq_order
     elif seq_list is not None:
       self._seq_order = [int(s[len(self._tag_prefix):]) for s in seq_list]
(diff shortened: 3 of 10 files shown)


returnn/datasets/cached.py
```diff
@@ -444,9 +444,7 @@ def delete(self, nframes):
 
   @property
   def num_seqs(self):
-    if self._index_map:
-      return len(self._index_map)
-    return self._num_seqs
+    return len(self._index_map)
```
patrick-wilken, line 447: Forgot to test with HDFDataset, after this `num_seqs=0` works also here.
`self. _index_map` is always set, so should be fine.

patrick-wilken: Hmm, ok depends on what we want `Dataset.num_seqs` to be before `init_seq_order` is called, see the changed test case. `CachedDataset2` would raise `NotImplementedError`. Not sure if we want that here.

albertz: You raise `NotImplementedError` when it is not implemented (or only in a derived class), and only then, not in any other case.

But I don't exactly understand anyway: If `self._index_map` is always set, this change here should not change anything?

albertz:
The change in all other datasets is maybe ok, but why not simply catching this directly in `CombinedDataset`? Then all the other changes are not needed.

returnn/datasets/cached.py
patrick-wilken: `__init__()` sets `self._index_map` to an empty range, which is falsy. So before an `init_seq_order()` call `CachedDataset.num_seq` is `self._num_seqs`. It would now be 0 after the change. Probably `None` would be the most logical value though?

> You raise `NotImplementedError` when it is not implemented

Yes, that would be nice. And that's why I don't want to use it. But what we have right now is `CachedDataset2` raising it depending on its state (whether `init_seq_order()` was called). And `ExternSprintDataset` even depending on the data loading speed...

albertz: > But what we have right now is `CachedDataset2` raising it depending on its state

Yes, so where is the problem? Is there any contradiction? It means it is implemented depending on the state. I.e. it is not implemented for one particular case.

patrick-wilken:
Well, the question is whether `seq_order=[]` in an `init_seq_order()` call is valid. If yes, there is nothing to fix in `CombinedDataset`. Of course we could catch it there nevertheless to not run into bugs, but fixing bugs is better, right? :)

albertz:
> Well, the question is whether `seq_order=[]` in an `init_seq_order()` call is valid.

Yea I don't know. Maybe, although it's quite unexpected. I assume many datasets would not really expect this. Even if you change the check to `is not None` in the `init_seq_order`, the remaining code might not expect this.

But if this works now for your use case, then we can just leave it as-is now.

returnn/datasets/cached.py
albertz: So when `init_seq_order` was not called yet, as this is somewhat undefined anyway, you could also just leave the current code.

patrick-wilken:
True, some Datasets might not handle it correctly. I only checked `HDFDataset` and `TranslationDataset`. But all that is required is for `init_seq_order()` to run through and then for `is_less_than_num_seqs()` to return False immediately, I think. At least that is what `CombinedDataset` uses.

albertz (approved):
Ok I think this is all fine then.

PR #1093, opened 2022-08-19
Support static time

albertz:


albertz:
Ok I remove the [better default time_dim_axis with dim tags](https://github.com/rwth-i6/returnn/pull/1093/commits/455c15e59961604f9b689e91572b2de8364e2250) again because it's not clear how much this potentially breaks. At least one test (`test_GenericAttentionLayer_weights_heads_auto_squeeze_time_end`) broke because now the output of `GenericAttentionLayer` had a time dim (`time_dim_axis` was set to some static spatial dim), while before it did not.

For the user, setting `time_dim_axis` explicitly in the `extern_data` dict also works (as can be seen in the updated tests).

So this PR does not introduce any changes to RETURNN but just these new test cases (which can be seen as examples).

albertz:
To add to the discussion:

One principal problem we have here is that the behavior of some layers, `Data` defaults etc is dependent on:

* Dimensions being static or dynamic. E.g. `Data` default `time_dim_axis` is dependent on that. But probably many layers as well. @JackTemaki reported that the `BatchNormLayer` is also dependent on this.
* `time_dim_axis` being a special attribute. E.g. `RecLayer` is dependent on this (by default when `axis` is not specified). Probably other layers as well.
* The order of axes, or spatial axes. E.g. the `Data` default `time_dim_axis` uses the first dynamic spatial dim.

Note that RETURNN-common resolves most (all?) of these problems because `time_dim_axis` should never play a role, all axes for every operation are always specified explicitly. So the default behavior of `time_dim_axis` should never matter.

albertz:
What was not really mentioned here: The support of `dyn_size_ext` in static dim tags is also not really supported. Or rather, there is some code spread around RETURNN which assumes that static dim tags do not have `dyn_size_ext`, and that needs to be fixed.

Issue #25, opened 2017-11-24
adapt model to single writer


xdsv:
Is there a way to adapt a model (that has been trained on IAM) to the handwriting of a single writer? 
Would that make a substantial difference to the recognition accuracy (for that writer)?

pvoigtlaender:
Hi,

you can first train on all data jointly and then start a second training, where you load the model from the first training, but this time you only train on the data of the specific writer.
This should probably boost the accuracy, but this of course depends on the amount of available adaptation data.
There are also many more elaborate methods for speaker/writer adaptation, but you would need to implement these yourself.

xdsv:
Can you elaborate on how to do this step by step?
Also, how much writer data is "enough" (approximately) in order to get good results?

albertz:
This is research. You have to try out and read other research papers. See option `import_model_train_epoch1` to start a new training where you import an existing model at the beginning.

As this is not really a bug in Returnn, I'm closing this issue now. If you want to discuss further, better ask by mail. But I am not sure if anyone from us can give much more advice than that. As said, this is research.

PR #1098, opened 2022-08-24
Allow subsampled targets for ExternSprintDataset

SimBe195:
This adds a parameter `reduce_target_factor` to the ExternSprintDataset class. This is used to modify the assertion that number of frames and number of targets are equal and checks the number of frames after subsampling by this factor instead.

An alternative solution would be to remove the assertion altogether.

### DIFF
--- a/returnn/datasets/sprint.py
+++ b/returnn/datasets/sprint.py
@@ -52,7 +52,7 @@ class SprintDatasetBase(Dataset):
 
   def __init__(self, target_maps=None, str_add_final_zero=False, input_stddev=1.,
                orth_post_process=None, bpe=None, orth_vocab=None,
-               suppress_load_seqs_print=False,
+               suppress_load_seqs_print=False, reduce_target_factor=1,
                **kwargs):
     """
     :param dict[str,str|dict] target_maps: e.g. {"speaker_name": "speaker_map.txt"}, with "speaker_map.txt" containing
@@ -64,9 +64,12 @@ class SprintDatasetBase(Dataset):
     :param None|dict[str] bpe: if given, will be opts for :class:`BytePairEncoding`
     :param None|dict[str] orth_vocab: if given, orth_vocab is applied to orth and orth_classes is an available target`
     :param bool suppress_load_seqs_print: less verbose
+    :param int|None reduce_target_factor: downsample factor to allow less targets than features
     """
     super(SprintDatasetBase, self).__init__(**kwargs)
     self.suppress_load_seqs_print = suppress_load_seqs_print
+    self.reduce_target_factor = reduce_target_factor
+    assert isinstance(self.reduce_target_factor, int) and self.reduce_target_factor >= 1
     if target_maps:
       assert isinstance(target_maps, dict)
       target_maps = target_maps.copy()
@@ -340,6 +343,8 @@ class SprintDatasetBase(Dataset):
     # is in format (feature,time)
     assert self.num_inputs == features.shape[0]
     num_frames = features.shape[1]
+    # features maybe downsampled via the network
+    reduce_num_frames = int((num_frames + self.reduce_target_factor - 1) / self.reduce_target_factor)
     # must be in format: (time,feature)
     features = features.transpose()
     assert features.shape == (num_frames, self.num_inputs)
@@ -355,8 +360,9 @@ class SprintDatasetBase(Dataset):
       targets = {"classes": targets}
     if "classes" in targets:
       # 'classes' is always the alignment
-      assert targets["classes"].shape == (num_frames,), (  # is in format (time,)
-        "Number of targets %s does not equal to number of features %s" % (targets["classes"].shape, (num_frames,)))
+      assert targets["classes"].shape == (reduce_num_frames,), (  # is in format (time,)
+          "Number of targets %s does not match number of features %s (reduce factor %d)"
+              % (targets["classes"].shape, (num_frames,), self.reduce_target_factor))
     if "speaker_name" in targets:
       targets["speaker_name"] = targets["speaker_name"].decode("utf8").strip()
     if "orth" in targets:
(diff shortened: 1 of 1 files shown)


returnn/datasets/sprint.py
```diff
@@ -340,6 +343,8 @@ def add_new_data(self, features, targets=None, segment_name=None):
     # is in format (feature,time)
     assert self.num_inputs == features.shape[0]
     num_frames = features.shape[1]
+    # features maybe downsampled via the network
+    reduce_num_frames = int((num_frames + self.reduce_target_factor - 1) / self.reduce_target_factor)
```
michelwi, line 347: would it make sense to make the rounding configurable?

albertz: This is a good default though, as you almost always have ceildiv, eg in pool, conv, etc. I'm not sure that you ever have sth else. So maybe we don't need to add it until we really need it.

michelwi: I had the case where the model (conformer) would do internal downsampling of factor 3, then upsample to FFR, then I downsample factor 2 to generate alignments.

Now depending on T%6 some weird edge cases arised. But I agree that this use case is contrived and that cieldiv is usually a good solution.

We can keep it like this, I only wanted to spring a discussion.

@DanEnergetics this is maybe also interesting for you.

returnn/datasets/sprint.py
```diff
@@ -64,9 +64,12 @@ def __init__(self, target_maps=None, str_add_final_zero=False, input_stddev=1.,
     :param None|dict[str] bpe: if given, will be opts for :class:`BytePairEncoding`
     :param None|dict[str] orth_vocab: if given, orth_vocab is applied to orth and orth_classes is an available target`
     :param bool suppress_load_seqs_print: less verbose
+    :param int|None reduce_target_factor: downsample factor to allow less targets than features
```
michelwi, line 67:
```suggestion
    :param int reduce_target_factor: downsample factor to allow less targets than features
```
I don't think None is allowed.

returnn/datasets/sprint.py
```diff
@@ -64,9 +64,12 @@ def __init__(self, target_maps=None, str_add_final_zero=False, input_stddev=1.,
     :param None|dict[str] bpe: if given, will be opts for :class:`BytePairEncoding`
     :param None|dict[str] orth_vocab: if given, orth_vocab is applied to orth and orth_classes is an available target`
     :param bool suppress_load_seqs_print: less verbose
+    :param int reduce_target_factor: downsample factor to allow less targets than features
```
michelwi, line 67: What about allowing arbitrary floats >0 (I.e. Also upsampling)?

SimBe195: In that case the ceildiv would not work properly. But maybe it would be good to allow an arbitrary callable. Then the user could manually specify how the sequence length is changed. This would also cover the downsample->upsample->downsample case you mentioned before.

albertz: Yes but I think I would only add that once we really need it.

returnn/datasets/sprint.py
```diff
@@ -355,8 +360,10 @@ def add_new_data(self, features, targets=None, segment_name=None):
       targets = {"classes": targets}
     if "classes" in targets:
       # 'classes' is always the alignment
-      assert targets["classes"].shape == (num_frames,), (  # is in format (time,)
-        "Number of targets %s does not equal to number of features %s" % (targets["classes"].shape, (num_frames,)))
+      assert targets["classes"].shape == (reduce_num_frames,), (  # is in format (time,)
+        "Number of targets %s does not match number of features %s (reduce factor %d)"
+        % (targets["classes"].shape, (num_frames,), self.reduce_target_factor)
+      )
```
albertz, line 366:
```suggestion
        % (targets["classes"].shape, (num_frames,), self.reduce_target_factor))
```

returnn/datasets/sprint.py
michelwi: It sure would still work. E.g. `ceildiv(100/1.8) = 55` or  `ceildiv(100/0.33) = 303`

albertz:
Btw, doesn't RASR need a similar parameter? I think RASR also had such check at some point. Or was that check just removed?

returnn/datasets/sprint.py
albertz: > It sure would still work. E.g. `ceildiv(100/1.8) = 55` or `ceildiv(100/0.33) = 303`

That does not make sense. Maybe you mean `math.ceil` here? But using `math.ceil` is problematic. There are much better ways to implement `ceildiv`, see [here](https://stackoverflow.com/questions/14822184/is-there-a-ceiling-equivalent-of-operator-in-python). The current implementation in the PR is also fine. But eg. this implementation would not work (as-is) for float, because it would return not an int.

Anyway, I think we can just leave it like that for now. I think I would only extend it once we really need it, and then maybe directly go with the callable for full flexibility.

returnn/datasets/sprint.py
DanEnergetics: The conv layer seems to give different output dimensions depending on the [selected padding](https://github.com/rwth-i6/returnn/blob/ce4366ff0caafc2e4b349fd2a189870f3c76f630/returnn/tf/layers/basic.py#L5228). This would cover the case that padding "SAME" is used but not "VALID" in all cases if I'm not mistaken. The pool layer probably behaves similar. Is "SAME" then more common or better in general?
Apart from that, this seems like a really useful change.

SimBe195:
> Btw, doesn't RASR need a similar parameter? I think RASR also had such check at some point. Or was that check just removed?

Yes, RASR also has this check. This is handled in Wei's label_sync_decoding branch but not in master at the moment. See [#25](https://github.com/rwth-i6/rasr/pull/25).

returnn/datasets/sprint.py
michelwi: Ok from my side

PR #1106, opened 2022-08-30
RecLastOutputLayer, resolve indirect layers

albertz:
This came up via https://github.com/rwth-i6/returnn_common/issues/199.


JackTemaki:
Ah okay so you want to make it more flexible on the RETURNN side, I guess that is the most simple solution.

albertz:
On RETURNN-common side, I think (without really looking at it) it might have been also a bit tricky, as automatic optimizations on what layer name to use could automatically do such replacements.

Issue #975, opened 2022-03-03
`Dim` internals and API should be refactored


albertz:
The logic involving `get_for_batch_ctx` and `get_same_base` is ugly, unintuitive, and a potential source of problems.

Direct assignments or reads from `dyn_size` or `dyn_size_ext`, but also all the other flags, even `description`, depends on `get_for_batch_ctx` or `get_same_base`.

Some older code ignores `get_for_batch_ctx` or `get_same_base` and directly accesses (reads or writes) any of the other flags like `dyn_size`. Which works when it is the correct instance. But otherwise it can lead to unexpected behavior.

Related is also `declare_same_as`, although this might not be much a problem, even less after such refactoring.

I'm not actually sure how much we really have a problem here. Maybe it works all fine already and nothing needs to be done?

I'm also not sure about a good way to solve this. Maybe `dyn_size` and `dyn_size_ext` should be hidden away, and only be accessible through functions `get_...` and `set_...`, which would also require the `batch` and `ctx`.

Such changes might break some older code. But on RETURNN side, this can all be fixed. And there should not be much (or any) external code yet using this.

albertz:
Note, in case we try to address this, i.e. clean up or refactor some of this: We should also check and run the test cases of returnn-common, as there are some problems which might only occur via those tests.

albertz:
One big problem is that we never really defined the logic around `get_for_batch_ctx` well, and it was added only at a later point. And similarly for some of the other concepts.

`CumConcatLayer` was one of the main reasons which introduced this, but then also the beam search logic.

Another big problem is the special role of the batch dim, and all the extra logic around `BatchInfo`, and when `Data.batch` or `Dim.batch` should be set, and what it actually should be set to. In principle, we should not need any special logic for the batch dim, and it should be treated just like other dims.

One feature which is available for the batch dim (`BatchInfo`) but not for other dims is the flattening logic, and also the special beam search logic.

I think the flattening should be done in a way that you could combine multiple dims (but any dims) and you would get a new flattened dim. Nothing would be specific about the batch dim. There would be meta info attached to the combined dim to be able to recover the individual dims.

The beam should just be another separate dim, not merged into the batch dim. And then there could also be meta information attached to it, what we basically have in `SearchBeam`.

albertz:
Related:

#1153

albertz:
[...]

PR #1107, opened 2022-08-30
Layer output, always set batch info

albertz:
This came up via https://github.com/rwth-i6/returnn_common/issues/199.

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -549,10 +549,11 @@ class LayerBase(object):
             break
         if not output.batch:
           output.batch = _set_global_batch_by_data(dyn_dim_tags_with_batch[0].dyn_size_ext)
-      else:
+      elif output.placeholder is not None:
         # No layers at all yet. This implies that the output must already have a placeholder.
         output.batch = _set_global_batch_by_data(output)
-    output.batch = output.batch.copy_set_beam(output.beam)
+    if output.batch:
+      output.batch = output.batch.copy_set_beam(output.beam)
     if output.control_flow_ctx != network.get_control_flow_ctx():
       x = output.placeholder
       output = output.copy_template_set_ctx(network.get_control_flow_ctx())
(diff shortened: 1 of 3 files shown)


albertz:
@JackTemaki This was important to fix the problem in https://github.com/rwth-i6/returnn_common/issues/199 because the output of the RecLayer had shape [k_dim], i.e. it did not had a batch dim, and thus `Data._adapt_batch_consistent_dim_tags` did not had an effect (due to its current implementation), and thus the dim tag of that output was not correctly adapted. This was easy to see in the debugger.

JackTemaki:
> @JackTemaki This was important to fix the problem in [rwth-i6/returnn_common#199](https://github.com/rwth-i6/returnn_common/issues/199) because the output of the RecLayer had shape [k_dim], i.e. it did not had a batch dim, and thus `Data._adapt_batch_consistent_dim_tags` did not had an effect (due to its current implementation), and thus the dim tag of that output was not correctly adapted. This was easy to see in the debugger.

What does `not correctly adapted` mean in this sense?

albertz:
> What does "not correctly adapted" mean in this sense?

What `Data._adapt_batch_consistent_dim_tags` does, i.e. switching to the right dim tag instance, via `Dim.get_for_batch_ctx`.

Note that #975 is somewhat related to this.

JackTemaki:
> > What does "not correctly adapted" mean in this sense?
> 
> What `Data._adapt_batch_consistent_dim_tags` does, i.e. switching to the right dim tag instance, via `Dim.get_for_batch_ctx`.
> 
> Note that #975 is somewhat related to this.

Hmm yes this function is hard to read. I also played around with the debugger, so I see `dyn_size` for the `k_dim` Dim is set correctly and then at some point suddenly not, but I would have no idea how to actually find from there the location where it is suddenly changed.

albertz:
> I also played around with the debugger, so I see `dyn_size` for the `k_dim` Dim is set correctly and then at some point suddenly not, but I would have no idea how to actually find from there the location where it is suddenly changed.

The point where it is not correctly set, this is a different layer (`RecLayer`, not `TopKLayer`), different layer output `Data`, and then a different instance of `Dim`, relating to the same `k_dim` (see `same_as` logic, `get_for_batch_ctx`).

PR #1114, opened 2022-09-01
Dim declare_same_as, prioritize explicit defined

albertz:
Via https://github.com/rwth-i6/returnn_common/issues/200.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -53,6 +53,8 @@ class Dim(object):
     Feature = Entity("feature")
     Types = (Batch, Spatial, Feature)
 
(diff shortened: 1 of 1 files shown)


returnn/tf/util/data.py
```diff
@@ -920,6 +925,12 @@ def declare_same_as(self, other):
       # We actually want it to be the other way around.
       other_same_base.declare_same_as(self_same_as)
       return
+    if self.is_dim_known and not other.is_dim_known():
```
albertz, line 928: Note this typo was fixed in a follow up commit (`self.is_dim_known()`).

PR #1117, opened 2022-09-09
Vocabulary, some refactor, more special labels 

albertz:
The refactor avoids loading labels/vocab to a dict when not needed (e.g. for SentencePieces).

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -875,6 +875,7 @@ class CombinedDataset(CachedDataset2):
     for dataset_key_tuple, data_key in data_map.items():
       dataset_key, dataset_data_key = dataset_key_tuple
       dataset = self.datasets[dataset_key]
+      assert isinstance(dataset, Dataset)
       if not data_dims:
         self.data_dims[data_key] = dataset.num_outputs[dataset_data_key]
       if dataset_data_key in dataset.labels:
--- a/returnn/datasets/util/vocabulary.py
+++ b/returnn/datasets/util/vocabulary.py
@@ -47,7 +47,8 @@ class Vocabulary(object):
     return clz(**opts)
 
   def __init__(self, vocab_file, seq_postfix=None,
-               unknown_label="UNK", bos_label=None, eos_label=None, pad_label=None, special_labels=None,
+               unknown_label="UNK", bos_label=None, eos_label=None, pad_label=None,
+               control_symbols=None, user_defined_symbols=None,
                num_labels=None, labels=None):
     """
     :param str|None vocab_file:
@@ -55,7 +56,10 @@ class Vocabulary(object):
     :param str|int|None bos_label: e.g. "<s>"
     :param str|int|None eos_label: e.g. "</s>"
     :param str|int|None pad_label: e.g. "<pad>"
-    :param dict[str,str|int]|None special_labels:
+    :param dict[str,str|int]|None control_symbols:
+      https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md
+    :param dict[str,str|int]|None user_defined_symbols:
+      https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md
     :param int num_labels: just for verification
     :param list[int]|None seq_postfix: labels will be added to the seq in self.get_seq
     :param list[str]|None labels:
@@ -75,7 +79,8 @@ class Vocabulary(object):
     self.bos_label_id = self.to_id(bos_label, allow_none=True)
     self.eos_label_id = self.to_id(eos_label, allow_none=True)
     self.pad_label_id = self.to_id(pad_label, allow_none=True)
-    self.special_label_ids = {name: self.to_id(label) for name, label in (special_labels or {}).items()}
+    self.control_symbol_ids = {name: self.to_id(label) for name, label in (control_symbols or {}).items()}
+    self.user_defined_symbol_ids = {name: self.to_id(label) for name, label in (user_defined_symbols or {}).items()}
     self.seq_postfix = seq_postfix or []
 
   def __repr__(self):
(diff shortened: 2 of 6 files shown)


patrick-wilken:
I really don't think it is a good idea to make `control_symbols` and `user_defined_symbols`, which are sentencepiece specific, parameters of the `Vocabulary` base class.

Also, `bos_label`, `eos_label`, `pad_label`: they are not used currently, except for the representation string. What is their purpose? Why is it the vocabularies job to know about their special nature? You are not planning to make e.g. the padding inside the network depend on the `pad_label`, or something along these lines, right?

And do we need to support setting those symbols via their id instead of the string? Makes it unnecessarily complicated in my view, maybe I'm missing something.

In general, I think the design of the `Vocabulary` class is quite bad already and with this it is not getting better. The vocabulary should be considered belonging to the network, actually we use it as an attribute of `Dim`/`Data`. Ideally, its only job should be to make sparse data interpretable by mapping ids to class labels.
Everything else (subword splitting, preventing splitting reserved words, interpretation of special symbols) is pre-/post-processing to my mind. Sure, you can do it on the fly, and sure, pre-processing must match the input vocabulary, but I don't think there is an advantage to mixing the implementation of those two things.

My background is that I never use any data processing in RETURNN, all I want is for RETURNN to handle the vocabulary lookup for me so I can feed/fetch interpretable (string) data. I don't like seeing this `Vocabulary` class, which could be just a wrapper around a dict with support for an unknown class, getting more complicated because of processing steps necessary for other people's tasks. 😉

albertz:
> I really don't think it is a good idea to make `control_symbols` and `user_defined_symbols`, which are sentencepiece specific, parameters of the `Vocabulary` base class.

They are not sentencepiece specific? You could have such control symbols or user defined symbols with any vocab. (I just linked the sentencepiece documentation because it has a good definition and explanation how you might use them.)

> Also, `bos_label`, `eos_label`, `pad_label`: they are not used currently, except for the representation string. What is their purpose? Why is it the vocabularies job to know about their special nature? You are not planning to make e.g. the padding inside the network depend on the `pad_label`, or something along these lines, right?

It's completely up to you.

E.g., I build models trainings now where the targets are automatically shifted to get the inputs for a decoder or LM, and for that I need to know the BOS. I already have access to the vocab (via the dim tag), and I thought the vocab is a reasonable place to store such meta information.

I did not want to always pass around a separate `bos_idx` parameter only because I could not directly store it inside the vocab. And the BOS is always defined together with the vocab, or not?

But it's completely optional.

I don't really understand your confusion here. You need to store this meta information **somewhere**, what the BOS/EOS/whatever symbol is, right? So why not store it directly in (or next to) the vocab?

But as said, it's completely optional, and you don't need to use it.

patrick-wilken:
> You could have such control symbols or user defined symbols with any vocab

`user_defined_symbols` are there to not split certain words, right? So no, does only apply to subword vocabularies. `control_symbols` maybe, they are tokens that should be removed from the output if present, right? But then you would need to add that functionality to the other vocabulary types, I don't think that's the plan.

But, ok, I could see that `bos_label_id`, `eos_label_id` and `pad_label_id` are special values that layers can insert into the data. Is that what you have in mind? In that case it accessing them via the vocabulary makes sense.

But if you use them only outside the network they should be handled wherever they are used. Because after the data is created there is no special logic about these symbols anywhere in the network, right? So they are merely "processing parameters" and are independent of the fact that there even is a vocabulary mapping. 😄
And also it becomes a bit arbitrary which special symbols to include into `Vocabulary` and which not. I can see that the three above are pretty universal though.

> But as said, it's completely optional, and you don't need to use it.

Well, it would be good if it's also optional to understand the related code. 😄 So to separate things if possible.

albertz:
> `user_defined_symbols` are there to not split certain words, right?

No, they are just custom user defined symbols. Whatever you want. For example, I probably will put "<blank>" in there. You could also put other special symbols in there.

> `control_symbols` maybe, they are tokens that should be removed from the output if present, right?

No. There is no logic at all for these. This is purely just meta information. E.g. you could put EOS/BOS/whatever also in here. But as they are so common, I thought it is useful to have them separate/explicit.

Before I called them `user_defined_symbols`/`control_symbols`, I called them just `special_symbols`. But then I read the SentencePiece documentation and I thought that this distinction between user-defined symbols and control symbols makes sense.

Only the unknown label has some special logic for the vocab (but this did not change in this PR).

> But, ok, I could see that `bos_label_id`, `eos_label_id` and `pad_label_id` are special values that layers can insert into the data. Is that what you have in mind? In that case it accessing them via the vocabulary makes sense.

So far there is no layer accessing them, and I don't plan to do that (although maybe there would be potential use cases as well). I want to access it purely from the returnn-common side, i.e. my config, i.e. when constructing the net dict, i.e. defining the net / computation. In returnn-common, you also have access to the `Data`, `Dim` and thus the vocab, including this meta information now on EOS/BOS/other special labels.

See [here](https://github.com/rwth-i6/i6_experiments/blob/a84b9e94e2353e7874e484ac5025073bad23eee0/users/zeyer/experiments/exp2022_07_21_transducer/pipeline_swb_2020.py#L263) as an example.

> But if you use them only outside the network they should be handled wherever they are used. Because after the data is created there is no special logic about these symbols anywhere in the network, right? So they are merely "processing parameters" and are independent of the fact that there even is a vocabulary mapping. 😄

I think they belong to the vocab. Whatever vocab you create, when you create the vocab, you also define those symbols. Where they are used, and how they are used, this is completely up to the user.

> And also it becomes a bit arbitrary which special symbols to include into `Vocabulary` and which not. I can see that the three above are pretty universal though.

Yes. I took those three because they were also explicit in SentencePiece, and I think they are really quite universal. Whenever you have others, you would put them now into `user_defined_symbols`/`control_symbols`.

I'm still thinking whether I maybe also should add some explicit `blank_label`, as it is so common as well, for CTC and transducer.


> Well, it would be good if it's also optional to understand the related code. 😄 So to separate things if possible.

It is? What do you mean?

PR #1135, opened 2022-09-29
global_step global var name scope fix

albertz:


albertz:
@Atticus1806 this should fix your problem.

Atticus1806:
I will test it, thanks.

PR #1136, opened 2022-09-29
Engine search, do not fall back to default_target

albertz:
We used `default_target` just for historically reasons here (earlier it was even hardcoded to `"classes"`).

I'm not really sure that anyone really depends on this, as you can just set the `target` on a layer explicitly.

It seems some tests are failing, which depends on this, i.e. they did not have `target` set on the layer (e.g. `decision` layer) and assumed that the `default_target` is set to the right target. There might be setups which have the same assumption, so they would break. I'm not really sure about the solution.

Note that the current behavior, this fallback to `default_target` breaks my search, as I have some output layer which does not correspond to a target. So with this change, it just behaves as I expect, it just does not try to serialize with any vocab.


albertz:
Well, as this potentially breaks setups, I think it requires a new behavior version.

PR #1137, opened 2022-09-29
PyTorch: save/load models

patrick-wilken:
".pt" file extension seems standard.
I think calling `model.load_state_dict()` in each epoch is the right way to do it if we really want `get_model_func` to be specific to the epoch. Although this probably has to be extended if the set of model parameters really changes between epochs.

The demo now carries over the model state between model epochs, saves model files, and can be interrupted and continued.

### DIFF
--- a/returnn/engine/base.py
+++ b/returnn/engine/base.py
@@ -56,6 +56,10 @@ class EngineBase(object):
           if os.path.exists(fn + ".index"):
             file_list[epoch] = fn
             break
+        elif util.BackendEngine.is_torch_selected():
+          if os.path.exists(fn + ".pt"):
+            file_list[epoch] = fn
+            break
     return file_list
 
   @classmethod
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -3,12 +3,14 @@ Main engine for PyTorch
 """
 
 import torch
+import os
 from torch.utils.data import DataLoader
 from random import random
 
 from returnn.log import log
 from returnn.engine.base import EngineBase
 from returnn.datasets.basic import init_dataset
+from returnn.util import basic as util
 from . import data_pipeline
 
 
@@ -23,9 +25,12 @@ class Engine(EngineBase):
     """
     super(Engine, self).__init__()
     self.config = config
+    self.model_filename = self.config.value('model', None)
     self.train_dataset = None
     self.eval_datasets = {}
     self.learning_rate = config.float("learning_rate", 1.)  # TODO LR control...
+
+    self._model_state = None
 
   def init_train_from_config(self, config=None, train_data=None, dev_data=None, eval_data=None):
     """
@@ -51,6 +56,11 @@ class Engine(EngineBase):
     """
     start_epoch, _ = self.get_train_start_epoch_batch(self.config)
     final_epoch = self.config_get_final_epoch(self.config)
+
+    if start_epoch == 1:
+      self._model_state = None
+    else:
+      self._model_state = self._load_model_state_from_file(epoch=start_epoch - 1)
 
     print("Starting training at epoch {}.".format(start_epoch), file=log.v3)
 
@@ -90,6 +100,8 @@ class Engine(EngineBase):
     sentinel_kw = {"__fwd_compatible_random_arg_%i" % int(random() * 100): None}
     model = get_model_func(epoch=self.epoch, **sentinel_kw)
     assert isinstance(model, torch.nn.Module)
+    if self._model_state:
+      model.load_state_dict(self._model_state)
     model.to(device)
     model.train()
     train_step_func = self.config.typed_value("train_step")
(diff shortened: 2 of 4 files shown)


returnn/engine/base.py
```diff
@@ -56,6 +56,10 @@ def get_existing_models(cls, config):
           if os.path.exists(fn + ".index"):
             file_list[epoch] = fn
             break
+        elif util.BackendEngine.is_torch_selected():
+          if os.path.exists(fn + ".pt"):
+            file_list[epoch] = fn
+            break
```
albertz, line 62: Why is this needed? I think this can be removed.

returnn/torch/engine.py
```diff
@@ -52,6 +57,11 @@ def train(self):
     start_epoch, _ = self.get_train_start_epoch_batch(self.config)
     final_epoch = self.config_get_final_epoch(self.config)
 
+    if start_epoch == 1:
+      self._model_state = None
+    else:
+      self._model_state = self._load_model_state_from_file(epoch=start_epoch - 1)
```
albertz, line 63: Why keeping this in an attrib? I would load the params directly into the model, and only when this is really needed.

returnn/torch/engine.py
```diff
@@ -115,6 +127,31 @@ def train_epoch(self):
...
+    """
+    :param int epoch:
+    :return: state saved via previous self._save_model_state_to_file() call for given epoch
+    :rtype: collections.OrderedDict
+    """
+    filename = self.get_epoch_model_filename(epoch=epoch) + util.get_model_filename_postfix()
+    print("Load model %s" % (filename,), file=log.v4)
+    return torch.load(filename)
+
+  def _save_model_state_to_file(self, model_state_dict):
```
albertz, line 143: I would keep this more consistent to the TF/Theano logic, and directly store it based on the current loaded model, not via model state dict.

returnn/util/basic.py
```diff
@@ -314,6 +314,8 @@ def get_model_filename_postfix():
   if BackendEngine.is_tensorflow_selected():
     # There will be multiple files but a *.meta file will always be present.
     return ".meta"
+  if BackendEngine.is_torch_selected():
+    return ".pt"
```
albertz, line 318: Why is this needed? I don't think you want that.

returnn/util/basic.py
```diff
@@ -329,6 +331,8 @@ def get_checkpoint_filepattern(filepath):
     return filepath[:-len(".meta")]
   elif filepath.endswith(".index"):
     return filepath[:-len(".index")]
+  elif filepath.endswith(".pt"):
+    return filepath[:-len(".pt")]
```
albertz, line 335: Remove this.

albertz:
> ".pt" file extension seems standard.

Why does that matter? There is no point in hardcoding this. Just leave it up to the user.

And note, the TF logic is very different there. It is about the **additional** postfix to the file, after the standard file extension (which is often sth like `.ckpt` but also totally up to the user), which is added by TF because a checkpoint consists  multiple files. This extra postfix logic is only needed when you have this case, i.e. multiple files each with some additional file extension. This extra postfix is predefined by TF and cannot be changed. That's why it needs this extra logic for TF.

albertz:
> Although this probably has to be extended if the set of model parameters really changes between epochs.

This is important. We should avoid the GPU -> CPU and CPU -> GPU transfer when it is not needed.

Maybe the interface for `get_model_func` is bad? If this is difficult to do in practice, due to the interface of `get_model_func`, we should change that interface.

patrick-wilken:
> > ".pt" file extension seems standard.
> 
> Why does that matter? There is no point in hardcoding this. Just leave it up to the user.

With the current config options (`model`) the user cannot specify a file extension because the epoch (and possibly "pretrain") is added to the path. (i.e. `model.pt.005` instead of `model.005.pt`). If we want to reuse the existing code to create and find the paths something like what I implemented is needed. Or we could say we don't need the file extension...

Regarding the model: I don't think it is practical to change an existing `Module` instance (add parameters, replace the `forward` function) between epochs. So if we want to support that we get a new network each epoch this means we get a new `Module` instance, as it is currently implemented. (Maybe `get_model_func` could return `None` to indicate same as previous epoch? Then we could keep the instance.)
Of course, if we don't need that functionality we would create the `Module` instance once and load the parameters directly into it.

albertz:
Ok then on the filename postfix.

Regarding the model, I was not saying at all that you should change an existing `Module` instance. Where did you read that?

Although, I think I would also leave that up to the user. E.g. there could be some `on_new_epoch(model)` callback, and the user could maybe just change some other hyper params, like dropout or do whatever.

> So if we want to support to have a new network each epoch this means we get a new `Module` instance, as it is currently implemented.

There are other ways to implement it, this is not strictly necessary. This is why I questioned whether our `get_model_func` interface should maybe be changed.

> (Maybe `get_model_func` could return `None` to indicate same as previous epoch? Then we could keep the instance.)

Yes, this is one example.

In any case, we should keep it efficient, and avoid introducing inefficiencies at all cost.

albertz:
Regarding the proof of concept, maybe completely ignore such pretraining aspect for now, to keep it simple. So `get_model_func` would not get an `epoch` parameter.

patrick-wilken:
This would be the version without the `epoch` parameter, so calling `get_model` once and then keep the `Module` as attribute.

returnn/torch/engine.py
```diff
@@ -52,6 +57,8 @@ def train(self):
     start_epoch, _ = self.get_train_start_epoch_batch(self.config)
     final_epoch = self.config_get_final_epoch(self.config)
 
+    self._load_model(epoch=start_epoch)
```
albertz, line 60: This should not be here but in `init_train_from_config` or `init_network_from_config`.

albertz: Just like `start_epoch` etc. Maybe I just merge this for now.

PR #1164, opened 2022-10-21
Remove Theano Backend

### LINKED ISSUE #593: Dropping Theano Support
I know that we do not want to do it right away, but I still wanted to make an issue about this, as I would be in favor of doing this rather sooner than later. I think the biggest argument for doing this are the Theano only parameters which often cause confusion, but there are also many other small things that sometimes cause extra work or at least some attention. Many could be fixed separately (e.g. in some conditions the docs to not compile without theano installed), but this would be wasted effort in my opinion.

I would also volunteer to do the actual PR about this (and e.g. keeping a single tagged version).

JackTemaki:
A draft PR to work on the removal of the Theano backend.

Fix #593.


JackTemaki:
I reactivated the old branch and rebased it, and fixed the most obvious issues. What I do not understand is:
 - Why did LSTMBlock disappear? This should not have happened
 - Why the log test fails
 
Unfortunately the SprintInterface and SprintDataset tests have a theano dependency although they are not really theano-specific tests. Here is some additional work necessary.

returnn/sprint/error_signals.py
```diff
@@ -442,7 +442,7 @@ def get_batch_loss_and_error_signal(self, log_posteriors, seq_lengths, tags=None
       jobs[j].append(i)
       joblen[j] += l
 
-    if not BackendEngine.is_theano_selected() and self.max_num_instances > 1:
+    if BackendEngine.is_tensorflow_selected() and self.max_num_instances > 1:
```
albertz, line 445:
```suggestion
    if self.max_num_instances > 1:
```

JackTemaki:
@albertz before you go over each detail, this is WIP and there are plenty of theano only functions remaining.

JackTemaki:
I just wanted to start first with the obvious imports to see how the tests behave. When this is dealt with I can start removing all theano only code I find.

returnn/util/basic.py
```diff
@@ -123,10 +123,9 @@ def is_64bit_platform():
 class BackendEngine:
   """
   Stores which backend engine we use in RETURNN.
-  E.g. Theano or TensorFlow.
+  Currently only TensorFlow.
```
albertz, line 126:
```suggestion
  E.g. TensorFlow or PyTorch.
```

returnn/util/basic.py
```diff
@@ -147,9 +146,6 @@ def select_engine(cls, engine=None, config=None):
       if config is None:
         from returnn.config import get_global_config
         config = get_global_config()
-      engine = None
```
albertz, line 150: This line should stay there.

returnn/util/basic.py
JackTemaki: As I said, this is WIP, do not look at those details yet. When I made this commit we did not have PyTorch yet.

returnn/util/debug_helpers.py
```diff
@@ -8,17 +8,8 @@
...
-  # noinspection PyUnresolvedReferences,PyPackageRequirements
-  import theano.tensor as tt
-  # noinspection PyUnresolvedReferences,PyPackageRequirements
-  import theano.sandbox.cuda as cuda
-  from returnn.theano.util import make_var_tuple
-  from returnn.theano.network import LayerNetwork
-else:
-  theano = None
+
+theano = None
```
albertz, line 12: Is this needed?

albertz:
You should remove the Theano tests from CI.

JackTemaki:
> You should remove the Theano tests from CI.

Are we are sure nothing is needed from there? E.g. the pretrain test was independent but still listed under theano. Not sure if this is needed at all anymore but this is what I want to have checked/discussed.

albertz:
> > You should remove the Theano tests from CI.
> 
> Are we sure nothing is needed from there? E.g. the pretrain test was independent but still listed under theano. Not sure if this is needed at all anymore but this is what I want to have checked/discussed.

Yes sure, you should check that.

albertz:
> Why did LSTMBlock disappear? This should not have happened

LSTMBlock already disappeared long ago.

The logic is this:
```python
    if not tf_compat.have_contrib and name.lower() in ["LSTMBlock".lower(), "LSTMBlockFused".lower()]:
      # E.g. TF 2 does not have the contrib module and also does not have LSTMBlock anymore.
      # (Actually, the raw op is still there, but not the wrapper class...)
      if cell_only:
        name = "StandardLSTM"
      else:
        name = "NativeLSTM2"
```
I'm not really sure why that fails now though.

albertz:
> Why the log test fails

They start RETURNN with the no-op task and then analyze stdout for certain properties. Now it has gotten much longer and that is unexpected. This should be fixed by filtering out some of the other TF, CUDA, CuDNN warnings, in `filter_out`., e.g.:
```python
    if "tensorflow/core/" in s or "tensorflow/stream_executor/" in s:  # some TF warnings
      i += 1
      continue
```
But even after the filtering, the output might just be longer, so we might need to adapt the numbers in the test.

returnn/util/basic.py
albertz: I don't exactly understand the complaint. Sure it can only help or not?

JackTemaki: Ok if it does not bother you then sure, you can list all that already.

returnn/native_op.py
```diff
@@ -340,7 +340,6 @@ def map_layer_inputs_to_op(cls, Z, V_h, i):
     Map.
     """
     import returnn.util.basic
-    assert returnn.util.basic.BackendEngine.is_theano_selected()
```
albertz, line 343: X

JackTemaki: For me the comment was shown

albertz: Only here but not in the file diff view for me. But anyway.

JackTemaki:
> > > You should remove the Theano tests from CI.
> > 
> > 
> > Are we sure nothing is needed from there? E.g. the pretrain test was independent but still listed under theano. Not sure if this is needed at all anymore but this is what I want to have checked/discussed.
> 
> Yes sure, you should check that. But you can already strip it down a lot. E.g. Device, EngineTask, EngineUtil, etc, are all obviously pure Theano.

So from what I see after looking at it they can all be removed.

JackTemaki:
We have some tools that only work with Theano, I assume they can all be removed? Those are for now:
 - import-sprint-nn.py (we are not using those NNs anymore anyway)
 - dump-network-json.py (was used to convert from the legacy .txt format to json I guess, not really needed in any current setup)

JackTemaki:
pretrain.py has quite the high interaction with Theano. I can remove everything that leads to code calling theano imports, but this might need extra care.

albertz:
> We have some tools that only work with Theano, I assume they can all be removed? Those are for now:
> 
> * import-sprint-nn.py (we are not using those NNs anymore anyway)

Yes ok.

> * dump-network-json.py (was used to convert from the legacy .txt format to json I guess, not really needed in any current setup)

I think this is still useful, e.g. when you have pretraining or so, or even with RETURNN-common, to actually see what your net dict looks like.

JackTemaki:
> > * dump-network-json.py (was used to convert from the legacy .txt format to json I guess, not really needed in any current setup)
> 
> I think this is still useful, e.g. when you have pretraining or so, or even with RETURNN-common, to actually see what your net dict looks like.

Ok. Just be aware that this tool is currently non-functioning for any config with pre-training. If we want to use it in the future this would need a re-write.

JackTemaki:
> > > * dump-network-json.py (was used to convert from the legacy .txt format to json I guess, not really needed in any current setup)
> > 
> > 
> > I think this is still useful, e.g. when you have pretraining or so, or even with RETURNN-common, to actually see what your net dict looks like.
> 
> Ok. Just be aware that this tool is currently non-functioning for any config with pre-training. If we want to use it in the future this would need a re-write.

How should I proceed here? I could remove all theano only code in pretrain.py and add a`NotImplementedError` in `dump-network-json.py` for the pretrain case...

albertz:
> How should I proceed here? I could remove all theano only code in pretrain.py and add a`NotImplementedError` in `dump-network-json.py` for the pretrain case...

I don't really understand why `dump-network-json.py` needs any special treatment here. I don't see any Theano specific in there. So just leave it as it is.

About `pretrain.py`, yea, that could be cleaned up, but as you said, be careful. If there is some left over dead code, it's also not so critical now and can be removed later.

albertz:
> I am just saying you can not use the tool right now because of the Theano dependency:
> https://gist.github.com/JackTemaki/56e713739df2feceb1cd5ecb85ae798e

Ah, I did not realize `get_network_for_epoch` is Theano specific.
Actually the code is slightly broken also for the non-pretrain case.
Just do this:
```python
  if pretrain:
    json_data = pretrain.get_network_json_for_epoch(args.epoch)
  else:
    json_data = network_json_from_config(config)
```

> > About `pretrain.py`, yea, that could be cleaned up, but as you said, be careful. If there is some left over dead code, it's also not so critical now and can be removed later.
> 
> I will not be able to pass the PyCharm test if I do not remove the code.

I'm not sure what you mean. Of course you should remove all Theano code. I was speaking about other left-over dead code. Why would PyCharm complain about that?

JackTemaki:
> > I am just saying you can not use the tool right now because of the Theano dependency:
> > https://gist.github.com/JackTemaki/56e713739df2feceb1cd5ecb85ae798e
> 
> Ah, I did not realize `get_network_for_epoch` is Theano specific. Actually the code is slightly broken also for the non-pretrain case. Just do this:
> 
> ```python
>   if pretrain:
>     json_data = pretrain.get_network_json_for_epoch(args.epoch)
>   else:
>     json_data = network_json_from_config(config)
> ```

Got it.

> > > About `pretrain.py`, yea, that could be cleaned up, but as you said, be careful. If there is some left over dead code, it's also not so critical now and can be removed later.
> > 
> > 
> > I will not be able to pass the PyCharm test if I do not remove the code.
> 
> I'm not sure what you mean. Of course you should remove all Theano code. I was speaking about other left-over dead code. Why would PyCharm complain about that?

Okay then I just misunderstood you.

JackTemaki:
Pretrain should also be clean now. The "greedy" flag is probably not usable anymore but this belongs to code that can be cleaned later on.

JackTemaki:
The LSTMBlock error does not happen when I test locally (also TF2.3), so I am not sure how to debug that. I am not sure if it has any implications if we remove the check for contrib completely and always map it.

.github/workflows/main.yml
```diff
@@ -90,7 +90,10 @@ jobs:
           - TEST=HDFDataset
           - TEST=TranslationDataset
           - TEST=LearningRateControl
+          - TEST=Log
+          - TEST=Pretrain
           - TEST=SprintDataset
+          - TEST=SprintInterface
```
albertz, line 96: This test requires TF, right?

.github/workflows/main.yml
```diff
@@ -90,7 +90,10 @@ jobs:
           - TEST=HDFDataset
           - TEST=TranslationDataset
           - TEST=LearningRateControl
+          - TEST=Log
```
albertz, line 93: This test requires TF.

tests/test_Pretrain.py
```diff
@@ -85,20 +37,3 @@ def test_config1():
   assert_in("hidden_0", net2_json)
   assert_in("hidden_1", net2_json)
   assert_equal(net2_json, net3_json)
-
-
-@unittest.skipIf(not theano, "not theano")
```
albertz, line 90: Does this test actually really depends on Theano? Why?

tests/test_Pretrain.py
```diff
@@ -85,20 +37,3 @@ def test_config1():
...
-def test_config2():
-  config = Config()
-  config.update(config2_dict)
-  pretrain = pretrain_from_config(config)
-  assert_equal(pretrain.get_train_num_epochs(), 3)
-
-
-@unittest.skipIf(not theano, "not theano")
```
albertz, line 98: Same here.

tests/test_Pretrain.py
JackTemaki: Because it does very explicitly. (`use_theano: True`), but sure, I can re-write it instead.

.github/workflows/main.yml
JackTemaki: Still, the log-test should not be removed just because the TF backend would be removed, but re-written. Thus I would prefer it under "base-tests"

.github/workflows/main.yml
JackTemaki: In this case we can move it, I already thought of doing that but then went with base-tests because I thought this should work either way (the test used Theano before, but could be re-written for TF, so it is a rather independent feature that is tested here).

.github/workflows/main.yml
albertz: What do you mean by removed? Why do you think it should be removed?

I just said it requires TF. So it means it must be under the TF tests.

It cannot be under the base tests when it requires TF.

I don't really see an option to rewrite it as well. It really tests TF specific things. Maybe rename the test to test_TF_Log.py if you like, and then you can have a separate test_Log.py to really only test base things, but I think we don't really have anything for that.

tests/test_Pretrain.py
albertz: Yea obviously just remove that line (`use_theano`). My question is, does it really depend on Theano? If not, it means nothing needs to be rewritten.

.github/workflows/main.yml
albertz: But you did exactly that, or not? I.e. you just rewrote it that it depends on TF?

.github/workflows/main.yml
JackTemaki: > Why do you think it should be removed?

I do not think that. 

> So it means it must be under the TF tests.

I will just move it, then we do not need to discuss.

.github/workflows/main.yml
JackTemaki: Yes sure. But we never defined if the tf-tests are "Tests that are testing TF specific features" or "Any test that depends somehow on TF". From your comments it now is clear you want it structured in the second way, so I will oblige to that.

albertz: No I did not say that. I did not say I want to have it structured this way.

This is just the only difference of the current structure in the CI tests, that one have a TF dependency and one have not. In no way does it define what kinds of tests are in what structure, except of this aspect what dependencies they need, because that is how the structures differ.

If you really want, you can create another section for tests which have a TF dependency but are not really TF specific. But I'm not really sure this adds anything?

tests/test_Pretrain.py
```diff
@@ -27,13 +17,13 @@
 }
 
 config2_dict = {
-  "use_theano": True,
+  "use_tensorflow": True,
```
albertz, line 20: Why is this needed at all? Does it actually init the engine?

tests/test_Pretrain.py
```diff
@@ -27,13 +17,13 @@
...
   "pretrain": "default",
   "num_inputs": 40,
   "num_outputs": 4498,
   "bidirectional": True,
   "hidden_size": (500, 500, 500),
-  "hidden_type": "lstm_opt",
+  "hidden_type": "rec",
```
albertz, line 26: Why is this change needed? Does it actually test whether the layer class exist? Or does it depend on the layer class in any way?

.github/workflows/main.yml
```diff
@@ -183,6 +181,8 @@ jobs:
           - TEST=Config
           - TEST=Fsa
           - TEST=Log
+          - TEST=Pretrain
```
albertz, line 184: Why do the Pretrain tests actually depend on TF?

tests/test_Pretrain.py
JackTemaki: lstm_opt is a layer that only exists for Theano, so it will fail for TF.

.github/workflows/main.yml
JackTemaki: I did not care about the "why" yet, but it does.

This is what I got before changing it:
```python3
  File "/u/rossenbach/src/returnn/returnn/tf/layers/basic.py", line 11668, in get_layer_class
    line: raise Exception("unknown layer class %r" % name)
    locals:
      Exception = <builtin> <class 'Exception'>
      name = <local> 'lstm_opt', len = 8
Exception: unknown layer class 'lstm_opt'
```

tests/test_Pretrain.py
JackTemaki: No, it does not init the engine. But it does call TF code, see other comment.

.github/workflows/main.yml
albertz: What is the full backtrace?

tests/test_Pretrain.py
albertz: Why does it fail? Why does it actually depend on the layer class?

.github/workflows/main.yml
JackTemaki: https://gist.github.com/JackTemaki/57a51b9e52451341876c711b2ee1947f

albertz: Ok this is in `LayerNetworkDescription`. I think I would just remove those tests which use `LayerNetworkDescription`.

tests/test_Pretrain.py
albertz: Can you put a normal net dict here instead?

tests/test_Pretrain.py
albertz: I think we should change that. I don't really see a good reason why it really needs TF.

.github/workflows/main.yml
albertz: So far I don't see a good reason why it really needs TF. If there is no good reason, we should maybe fix things and move this over to the base tests.

tests/test_Pretrain.py
JackTemaki: For this test it makes sense, for the other not really as it really tests this old legacy format. So I will change it here to a dict and remove the other test.

albertz: Which do you want to remove? I think all the tests are valid? They just test `get_train_num_epochs` in the end.

JackTemaki: This does not even make sense, this is literally test 3 then. I will remove both.

albertz: I think `test_config1` is also important.

albertz: Yea `test_config2` becomes obsolete then, as it is the same as `test_config3`. but `test_config1` is different and should stay.

JackTemaki: Did a full rework now.

tests/test_Pretrain.py
```diff
@@ -76,17 +75,9 @@ def test_config1():
...
-
-def test_config2():
-  config = Config()
-  config.update(config2_dict)
-  pretrain = pretrain_from_config(config)
-  assert_equal(pretrain.get_train_num_epochs(), 3)
-
-
-def test_config3():
+def test_config_net_dict2():
```
albertz, line 78: Missing newline.

tests/test_Pretrain.py
```diff
@@ -1,41 +1,17 @@
...
-config3_dict = {
-  "use_theano": True,
+config_dict = {
   "pretrain": "default",
   "num_inputs": 40,
   "num_outputs": 4498,
 }
 
-config3_json = """
+config_json = """
```
albertz, line 14: How is this still relevant with `net_dict2`?

tests/test_Pretrain.py
```diff
@@ -1,41 +1,17 @@
...
-  "bidirectional": True,
-  "hidden_size": (500, 500, 500),
-  "hidden_type": "rec",
-  "activation": "sigmoid",
-  "dropout": 0.1,
-}
-
-config3_dict = {
-  "use_theano": True,
+config_dict = {
```
albertz, line 8: You removed one empty line here.

tests/test_Pretrain.py
```diff
@@ -54,17 +30,40 @@
...
+  "lstm2_fw": {"class": "lstm_opt", "n_out": 500, "dropout": 0.1, "sampling": 1, "reverse": False,
+               "from": ["lstm1_fw", "lstm1_bw"]},
+  "lstm2_bw": {"class": "lstm_opt", "n_out": 500, "dropout": 0.1, "sampling": 1, "reverse": True,
+               "from": ["lstm1_fw", "lstm1_bw"]},
+  "output":   {"class": "softmax", "loss": "ce", "from": ["lstm2_fw", "lstm2_bw"]}
+}
+
 
-def test_init_config1():
+def test_config_topology_json():
```
albertz, line 55: I think an empty line is missing before this function.

tests/test_Pretrain.py
JackTemaki: It does trigger slightly different code. This can probably (did not check all details) be remove in the future as well (including the code related to this in config.py).

albertz: But isn't it exactly the same? I don't see the difference.

albertz: Also, the naming of these vars is confusing to me. Why call one `config_json`, and the other `net_dict`?

JackTemaki: Because one is a json string that gets parsed in the config.py via specifically setting the `network_topology_json`, the other is just our regular `net_dict`. I never knew this existed, and as I said we can probably remove it. Still, the tests are not doing exactly the same.

JackTemaki: Lets not discuss this, I will just remove it.

albertz: But it should not be removed if there is a difference.

tests/test_Pretrain.py
```diff
@@ -13,26 +13,6 @@
...
-  "from" : ["lstm0_fw", "lstm0_bw"] },
-
-"lstm2_fw" : { "class" : "lstm_opt", "n_out" : 500, "dropout": 0.1, "sampling" : 1, "reverse" : false,
-  "from" : ["lstm1_fw", "lstm1_bw"] },
-"lstm2_bw" : { "class" : "lstm_opt", "n_out" : 500, "dropout": 0.1, "sampling" : 1, "reverse" : true,
-  "from" : ["lstm1_fw", "lstm1_bw"] },
-
-"output" :   { "class" : "softmax", "loss" : "ce", "from" : ["lstm2_fw", "lstm2_bw"] }
-}
-"""
```
albertz, line 65: You said there was a difference. (Not JSON but sth else?)

Then this should not be removed.

tests/test_Pretrain.py
JackTemaki: Nothing that is directly relevant to the pre-training I think.

albertz: But what exactly is different? I still don't understand this.

JackTemaki: `config.py 581 ff`
```python3
  if config.has("network") and config.is_typed("network"):
    json_content = config.typed_value("network")
    assert isinstance(json_content, dict)
    assert json_content
  elif config.network_topology_json:
    start_var = config.network_topology_json.find('(config:', 0)  # e.g. ..., "n_out" : (config:var), ...
    while start_var > 0:
      end_var = config.network_topology_json.find(')', start_var)
      assert end_var > 0, "invalid variable syntax at " + str(start_var)
      var = config.network_topology_json[start_var+8:end_var]
      assert config.has(var), "could not find variable " + var
      config.network_topology_json = (
        config.network_topology_json[:start_var] + config.value(var, "") + config.network_topology_json[end_var+1:])
      print("substituting variable %s with %s" % (var, config.value(var, "")), file=log.v4)
      start_var = config.network_topology_json.find('(config:', start_var+1)
    try:
      import json
      json_content = json.loads(config.network_topology_json)
    except ValueError as e:
      print("----- BEGIN JSON CONTENT -----", file=log.v3)
      print(config.network_topology_json, file=log.v3)
      print("------ END JSON CONTENT ------", file=log.v3)
      assert False, "invalid json content, %r" % e
    assert isinstance(json_content, dict)
    if 'network' in json_content:
      json_content = json_content['network']
    if not json_content:
      raise Exception("network json is empty: %s" % config.network_topology_json)
```
As I said before (and now verified) it is really just the json parsing. So it is different code, but not relevant to the pre-training, so I think the test can be removed and PROBABLY the whole code here as well.

JackTemaki: (In a future PR where we check this properly, that is)

albertz: But as I said, JSON vs net dict, this is not a difference for this test.

I'm asking, if there is any real difference. Is there? If yes, then the test should stay (but remove the JSON aspect). And clarify what the difference is. If no, then remove it.

albertz: I don't see the answer? So the answer is, "no, there is no difference"?

JackTemaki: It was the post before:

> As I said before (and now verified) it is really just the json parsing. So it is different code, but not relevant to the pre-training, so I think the test can be removed and PROBABLY the whole code here as well

Which says:
 - There is a difference with respect to loading the JSON vs. not loading the JSON
 - There is no difference with respect to the pre-training

albertz: But as I said, the JSON aspect is not relevant. Let's not repeat that over and over.

I'm asking if there is any other difference. Not just w.r.t. pretraining, but potentially other things as well. I just want to know.

JackTemaki:
There are definitely still leftovers, but I think they can be removed over time. For everything not removed now I either missed it or I lack sufficient understanding of the code-based to correctly estimate if it can be removed.

albertz (approved):

PR #1180, opened 2022-10-25
remove CTC targets from Datasets

### LINKED ISSUE #1179: Remove CTC Targets from all Datasets
From my understanding this very specific feature was only used in combination with Theano setups. If yes, it can be removed.

JackTemaki:
resolves #1179

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -621,15 +621,6 @@ class Dataset(object):
     # For new-style subclasses, which just provide get_data.
     return self.get_data(sorted_seq_idx, target)
 
-  def get_ctc_targets(self, sorted_seq_idx):
-    """
-    Warning: This is deprecated/obsolete.
-
-    :param int sorted_seq_idx:
-    :rtype: numpy.ndarray|None
-    """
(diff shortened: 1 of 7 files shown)


returnn/datasets/basic.py
```diff
@@ -893,20 +877,6 @@ def serialize_data(self, key, data):
...
-    :rtype: numpy.ndarray
-    """
-    priori = numpy.zeros((self.num_outputs[target][0],), dtype=numpy.float32)
-    i = 0
-    while self.is_less_than_num_seqs(i):
-      self.load_seqs(i, i + 1)
-      for t in self.get_targets(target, i):
-        priori[t] += 1
-      i += 1
-    return numpy.array(priori / self.get_num_timesteps(), dtype=numpy.float32)
```
albertz, line 908: This is somewhat unrelated to this PR. But I guess also ok.

JackTemaki: Ah yes, `update_weights` is extra as well. I forgot to exclude it from the commit.

albertz (approved):

PR #1192, opened 2022-10-27
fix option stft-pad_last_frame in MultiChannelMultiResolutionStftLayer

michelwi:
The original code did not work without eager execution mode. And even then the if statement does not work element-wise over the batch dim.
This now makes use of  integer division round up.
```
int(ceil(x / float(y))) -> (x-1)/y + 1
```

### DIFF
--- a/returnn/tf/layers/signal_processing.py
+++ b/returnn/tf/layers/signal_processing.py
@@ -448,14 +448,10 @@ class MultiChannelMultiResolutionStftLayer(_ConcatInputLayer):
 
     def _compute_size_placeholder():
       size_placeholder_dict = {}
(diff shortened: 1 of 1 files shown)


albertz:
I think @vieting is the only one using this, so he should check. The code is extremely outdated and likely should be rewritten anyway.

vieting (approved):
I'm not really using it either. Agree with Albert that this code is outdated. I added the `StftLayer` which is clearer and more in line with recent RETURNN style.

Anyway, the fix looks good to me, so if you want to use it I don't see any problem with merging it.

albertz:
Ok, merging this. @michelwi but check whether your code can be rewritten using `StftLayer`. Or if not, what's missing? At some point we will probably clean up and remove those layers like `MultiChannelMultiResolutionStftLayer`.

PR #1193, opened 2022-10-27
add option to specify min_chunk_size as dict

michelwi:
If we have dataset keys with different length specify chunk_size and chunk_step as dicts we might also want to specify the min_chunk_size as a dict with different sizes per key.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -134,10 +134,7 @@ class Dataset(object):
     self._num_timesteps = 0
     self._num_seqs = 0
     self._estimated_num_seqs = estimated_num_seqs
-    if isinstance(min_chunk_size, int):
-      self.min_chunk_size = NumbersDict(broadcast_value=min_chunk_size)
-    else:
-      self.min_chunk_size = NumbersDict(min_chunk_size)
+    self.min_chunk_size = NumbersDict(min_chunk_size)
     self.chunking_variance = chunking_variance
     self.chunk_size, self.chunk_step, self.custom_chunking_func = self._parse_chunking(chunking)
     if isinstance(context_window, (tuple, list)):
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -54,7 +54,7 @@ class Dataset(object):
     set_or_remove("chunking", config.opt_typed_value("chunking", None))
     set_or_remove("seq_ordering", config.value("batching", None))
     set_or_remove("shuffle_frames_of_nseqs", config.int('shuffle_frames_of_nseqs', 0) or None)
-    set_or_remove("min_chunk_size", config.int('min_chunk_size', 0) or None)
+    set_or_remove("min_chunk_size", config.opt_typed_value('min_chunk_size', 0) or None)
     set_or_remove("chunking_variance", config.float("chunking_variance", 0))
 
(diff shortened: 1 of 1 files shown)


returnn/datasets/basic.py
```diff
@@ -134,7 +134,10 @@ def __init__(self, name=None,
     self._num_timesteps = 0
     self._num_seqs = 0
     self._estimated_num_seqs = estimated_num_seqs
-    self.min_chunk_size = min_chunk_size
+    if isinstance(min_chunk_size, int):
+      self.min_chunk_size = NumbersDict(broadcast_value=min_chunk_size)
+    else:
+      self.min_chunk_size = NumbersDict(min_chunk_size)
```
albertz, line 140:
```suggestion
    self.min_chunk_size = NumbersDict(min_chunk_size)
```

returnn/datasets/basic.py
```diff
@@ -951,8 +954,9 @@ def iterate_seqs(self, recurrent_net=True, used_data_keys=None):
           if chunk_step[key] == chunk_step[default_key]:
             raise Exception("Chunking with multiple data-keys of different length: %r" % length)
           else:
-            limit = limit_default = 1
-            if self.min_chunk_size == chunk_size[default_key]:
+            limit = self.min_chunk_size[key] or 1
+            limit_default = self.min_chunk_size[default_key] or 1
+            if self.min_chunk_size[default_key] == chunk_size[default_key]:
```
albertz, line 959: Hm, this code seems somehow wrong to me...
But not just due to your change, but already before.
Or does it make sense for you?
I also mean specifically the two lines below:
```python
limit = chunk_size[key]
limit_default = chunk_size[default_key]
```

I see also a somewhat strange edit history for this.
I very roughly remember that I did this for Tobias, but not sure anymore. Maybe also @vieting.

Who exactly uses `min_chunk_size`? I think in all my setups, it's 0 anyway.

albertz: If we remove this `if` including its code, is it correct then?

michelwi: This only makes sense to me, if indeed the data-keys had different length and then 
`min_chunk_size == chunk_step` for the default key implies "always full chunks for all keys"

I have kept this only for preserving the old behavior and existing setups. We can indeed get the same behavior with my proposed change if the `min_chunk_size` is set to `chunk_step` for each data-key.

I would be fine with removing this if and its code.

albertz: Before your change, `min_chunk_size` was anyway for all keys.

After your change, you also fixed the `limit` above. So in any case the logic here is different now. But it looks more correct.

I think we should remove this `if` and its code.

Anyway, this whole code here is only to perform the sanity check, for nothing else.

returnn/datasets/basic.py
```diff
@@ -951,8 +951,9 @@ def iterate_seqs(self, recurrent_net=True, used_data_keys=None):
...
               limit = chunk_size[key]
               limit_default = chunk_size[default_key]
```
michelwi, line 958:
```suggestion
```

returnn/datasets/basic.py
michelwi: ok, it is removed now.

albertz (approved):

PR #1203, opened 2022-11-08
SelfAttentionLayer: add global inf_value option

Gerstenberger:
Add global config option "inf_value" to the SelfAttentionLayer.
We need this to convert acoustic models to ONNX. The current value results in nan values.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -2104,7 +2104,8 @@ class SoftmaxOverSpatialLayer(_ConcatInputLayer):
         start=start.output if start else None,
         window_start=window_start.output if isinstance(window_start, LayerBase) else window_start,
         window_size=window_size.output if isinstance(window_size, LayerBase) else window_size)
-      energy = where_bc(energy_mask, energy, float("-inf"), name="energy_masked")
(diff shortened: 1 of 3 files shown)


Gerstenberger:
> Looks good. Can you search for other places of `inf` in RETURNN? Those other places probably should also use this. E.g. `SoftmaxOverSpatialLayer`.

Sure, i can do that.

albertz:
Also `SeqLenMaskLayer`, `DotAttentionLayer`, `ConcatAttentionLayer`.

albertz:
And `softmax_cross_entropy_over_size`.

albertz:
As far as I can see, those are the relevant uses.

albertz:
`SeqLenMaskLayer` missing?

Gerstenberger:
I adapted `SoftmaxOverSpatialLayer`, `DotAttentionLayer`, `ConcatAttentionLayer` and `softmax_cross_entropy_over_size`.

`SeqLenMaskLayer` looks okay to me as is. The only inf values i can find is the check for `self.allow_inf_in_output`.

albertz (approved):

PR #1206, opened 2022-11-08
Decoupled constraints better, grad accum and others deterministic

### LINKED ISSUE #1205: train step incr and read order non-deterministic, possible effect on gradient accumulation
While working on some changes on decoupled constraints, I noticed that the order of `incr_step_op`(increasing `global_train_step`) and any read of the `global_train_step` seems non-deterministic. This means, it might already be increased or not.

This has an effect wherever the train step is used. For example for the grad accumulation logic:
```python
                tf.cond(
                  tf.equal(
                    tf_compat.v1.mod(tf_util.py_print(self.global_train_step, ["accum grad train step", self.global_train_step]), accum_grad_multiple_num_steps),
                    accum_grad_multiple_num_steps - 1),
...
```

And this is the `incr_step_op` (here slightly extended):
```python
    # Do this at the very end.
    with tf.control_dependencies([self.optim_op]):
      incr_step_op = self.network.global_train_step.assign_add(1, name="global_train_step_increment", read_value=False)
      tf_util.add_control_input(incr_step_op, self.optim_op)
      tf_util.add_control_input(incr_step_op, tf_util.py_print(1, ["incr_step_op train step", self.network.global_train_step]))
    self.optim_op = tf.group(self.optim_op, incr_step_op, name="optim_and_step_incr")
```

This results in such a log:
```
Run step 0
[incr_step_op train step][0]
[accum grad train step][1]
Run step 1
[accum grad train step][1]
[incr_step_op train step][1]
Run step 2
[incr_step_op train step][2]
[accum grad train step][2]
Run step 3
[accum grad train step][3]
[incr_step_op train step][3]
```

albertz:
The decoupled constraints are calculated in a more direct way, not using the SGD optimizer. This should make it (much) more efficient (at compilation time), while being equivalent otherwise. But I actually had memory problems and this solved it.

I also fixed/changed the logic of decoupled constraints when used with gradient accumulation: With gradient accumulation N, the constraints are only applied every Nth step. This is because this is actually equivalent if you would increase the batch size by N instead of using gradient accumulation.

Along the way, I discovered and fixed non-deterministic behavior on global train step incr and read order, which potentially has an effect on gradient accumulation (#1205). Thus it introduces the new behavior version 15 for this. Note that this is necessary for the gradient accumulation logic of the decoupled constraints as well. At least on my computer (MacBook M1), the test `test_Updater_decouple_constraints_simple_graph_grad_accum` does not pass with behavior version 14.

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -22,6 +22,18 @@ and not listing legacy/deprecated parameters.
 Version History
 ---------------
 
+Behavior version 15 (2022-11-08)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``TFNetwork.global_train_step`` should deterministically
+always return the initial value of the current step.
+Before, it could non-deterministically return the current step or the next step.
+This has an influence on any code which makes use of it.
+The biggest effect is on gradient accumulation
+where the old non-deterministic behavior likely was wrong.
+
+See issue `#1205 <https://github.com/rwth-i6/returnn/issues/1205>`__.
+
 Behavior version 14 (2022-10-19)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -1438,9 +1438,16 @@ class LayerBase(object):
     :return: None or scalar
     :rtype: tf.Tensor|None
     """
+    decouple_constraints = self.network.get_config().bool("decouple_constraints", False)
     c = 0
     if self.L2:
-      c += self.L2 * self.get_params_l2_norm()
+      if decouple_constraints:
+        # Keep this logic in sync with TFUpdater.
+        for _, param in self.params.items():
+          assert isinstance(param, tf.Variable)
+          setattr(param, "RETURNN_constraint_L2", self.L2)
+      else:
+        c += self.L2 * self.get_params_l2_norm()
     if self.spatial_smoothing:
       c += self.spatial_smoothing * self.get_output_spatial_smoothing_energy()
     if self.darc1:
--- a/returnn/tf/network.py
+++ b/returnn/tf/network.py
@@ -2294,7 +2294,7 @@ class TFNetwork(object):
         continue
       if BehaviorVersion.get() <= 14:
         return net.global_train_step_var
-      with tf_util.reuse_name_scope("", absolute=True), tf.device("/cpu:0"):
+      with tf_util.default_control_flow_ctx(), tf_util.reuse_name_scope("", absolute=True), tf.device("/cpu:0"):
         net._global_train_step = net.global_train_step_var.read_value()
       return net._global_train_step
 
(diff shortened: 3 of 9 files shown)


albertz:
@JackTemaki @mmz33 Can you maybe test some training which uses grad accumulation with this PR and behavior version 15, and see if there is any difference which is more than just noise due to non-determinism. Or maybe also test if the non-determinism with behavior version 15 is lower than the non-determinism with behavior version 14. For this test, I think only a few epochs are enough.

JackTemaki:
> @JackTemaki @mmz33 Can you maybe test some training which uses grad accumulation with this PR and behavior version 15, and see if there is any difference which is more than just noise due to non-determinism. Or maybe also test if the non-determinism with behavior version 15 is lower than the non-determinism with behavior version 14. For this test, I think only a few epochs are enough.

I need to create a setup for that first but yes, I can try.

albertz:
Btw, independent of testing it, can you also review?

JackTemaki:
> Btw, independent of testing it, can you also review?

Not really, I went through the code but I do not understand what is really happening there and how the changes are working.

albertz:
> > Btw, independent of testing it, can you also review?
> 
> Not really, I went through the code but I do not understand what is really happening there and how the changes are working.

What do you not understand? I can try to put some more comments. It's important that everyone understands this.

albertz:
Note that in the log of #1205, actually only the first step was inconsistent, while all the other steps  However, the log does not tell the true story because what you see there in the `py_print` is not necessarily the same what is actually then used for the `mod`/`cond`. In my tests, it looked like the `true` branch was executed more often than it should have been, about 90% of the cases or so, although with `accum_grad_multiple_num_steps=2`, it should only execute 50% of the time.

albertz:
Note that I asked you for review not really for the decoupled constraints, but about potentially new behavior on the gradient accumulation and related things. You can ignore the decoupled constraints part if you never used it.

albertz:
Regarding the attrib: Yea, I know, but I did not really see another good option for doing this, which would not be much more complicated, and maybe even more error prone. This is also a `tf.Variable`, so it cannot easily be replaced by another object.

albertz:
@JackTemaki Do you have results?

albertz:
@JackTemaki @patrick-wilken @michelwi @mmz33 Can you please give some feedback on the gradient accumulation? You have noticed that this here potentially changes the behavior of it? So it's fine?

JackTemaki:
> @JackTemaki Do you have results?

No, not yet. I can see if it fits in today, but I can not promise anything.

> You have noticed that this here potentially changes the behavior of it? So it's fine?

Yes, this is fine.

patrick-wilken:
> Can you please give some feedback on the gradient accumulation?

Yes, looks good. And does not affect my setups anyway. 🙃

albertz:
You don't use gradient accumulation?

patrick-wilken:
Ah, sorry, thought you mean the gradient accumulation related things inside the `if self.decouple_constraints:`.
So you are basically asking about the `read_value()` call? I have no clue how exactly it would work to be honest.

albertz:
> Ah, sorry, thought you mean the gradient accumulation related things inside the `if self.decouple_constraints:`. So you are basically asking about the `read_value()` call? I have no clue how exactly it would work to be honest.

Yes exactly. First, issue #1205 is also about that, and this changes it now.

I don't really know whether gradient accumulation was buggy all the time or not. Or maybe it was buggy but still somehow worked. In any case, this here can potentially change how gradient accumulation works, so that is why it is very important that some of you check this, as gradient accumulation is an important feature.

albertz:
Ok, due to lack of activity: I'm merging this now. But please check whether the gradient accumulation behavior somehow has changed for you.

PR #1220, opened 2022-11-16
More restrictive is_equal_opts for dim tag comparison

### LINKED ISSUE #865: Dim tags matching should use `allow_same_spatial_dim=False`

### LINKED ISSUE #1219: GatherLayer resolved wrong common axes
The code in `_get_common_input_position_axes` used `dict(allow_same_spatial_dim=True)`. For some different static dim tags which happen to have the same static dim value, this would see them as equal, which is wrong in general.

We could introduce a new behavior version to be more restrict. But this might make it difficult for some people to update?

Or only match auto-generated same spatial/feature dims but never user-generated?

Very related is also #1154 and #865.

albertz:
Fix #865.
Fix #1219.

But while implementing this, I am now thinking whether I should instead implement the suggestion from #865 using `auto_generated`.

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -22,6 +22,11 @@ and not listing legacy/deprecated parameters.
 Version History
 ---------------
 
+Behavior version 16 (2022-11-16)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+XXX
+
 Behavior version 15 (2022-11-08)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -1395,9 +1395,12 @@ class GatherLayer(_ConcatInputLayer):
     batch dim.
     """
     from returnn.util import BehaviorVersion
-    is_equal_opts = dict(allow_same_spatial_dim=True)
-    if BehaviorVersion.get() < 11:
-      is_equal_opts["broadcast_matches"] = True
+    if BehaviorVersion.get() >= 16:
+      is_equal_opts = {}
+    elif BehaviorVersion.get() >= 11:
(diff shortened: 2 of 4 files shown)


albertz:
A better fix is in #1222, so closing this one.

PR #1227, opened 2022-11-21
fixes for depthwise 1d conv

vieting:
The `ConvLayer` implements depthwise convolution via `tf.nn.depthwise_conv2d`. This is used in a testcase in https://github.com/rwth-i6/pytorch-to-returnn/pull/128.

There are currently two issues:
1) It seems to me that the strides are confused for `"NCHW"` and `"NHWC"`.
2) When we actually have a 1D case, an additional dim is added and still `tf.nn.depthwise_conv2d` is used. An additional stride of 1 is added for this dim. However, I get
```
InvalidArgumentError: Current implementation only supports equal length strides in the row and column dimensions.
```
in this case. Since the added dim is anyway 1, we can also have the same stride as for the spatial dim which is then supported.


albertz (approved):

albertz:
Right, good catch.

PR #1234, opened 2022-11-29
verbose error for invalid datasets

JackTemaki:
Just a more verbose error message for some issue I stumbled upon when using an `LmDataset` within a `MetaDataset`.

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -286,7 +286,13 @@ class MetaDataset(CachedDataset2):
       assert isinstance(default_dataset, Dataset)
       print("Reading sequence list for MetaDataset %r from sub-dataset %r" % (self.name, default_dataset.name),
             file=log.v3)
-      seq_list = default_dataset.get_all_tags()
+      try:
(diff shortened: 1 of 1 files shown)


albertz:
You ignore the message from the original exception, which might also have some useful information. You should include that.

You can also do raise ... from ... but this is Python 3 only. I forgot if we want to start using that now, i.e. dropping Python 2 with this.

JackTemaki:
> You ignore the message from the original exception, which might also have some useful information.

The original message is still printed for me.

albertz:
In the code you throw away the original exception, i.e. don't use it.

JackTemaki:
Nevertheless it will be printed. I do not understand what you are getting at, is it about presenting the original information to the user? This is still done automatically. It will first print the original exception, and then something like "during handling of the exception another exception occurred", followed by the new exception. If it is to allow for further handling of the original exception in the code, I do not see why this is necessary.

albertz:
Ah right I forgot about that behavior.

PR #1239, opened 2022-12-06
Add doc for cleanup behavior

Atticus1806:
Adds documentation to the possible parameters of `cleanup_old_models`


Atticus1806:
Found this https://returnn.readthedocs.io/en/latest/configuration_reference/training.html  which already explains it, but for me that link does not show in the search of the docs.

albertz:
> but for me that link does not show in the search of the docs.

Maybe open an issue that the doc search does not work properly.

PR #1244, opened 2022-12-06
axis description for any tag

vieting:
It is already possible to use axis descriptions for spatial tags like `"axis": "stag:<...>"`. I'd like to add that functionality for static dims as well, e.g. via `"axis": "tag:<...>"` for any tag.

### DIFF
--- a/returnn/tf/util/data.py
+++ b/returnn/tf/util/data.py
@@ -4907,6 +4907,8 @@ class Data(object):
         _, idx_s, name = axes.split(":", 2)  # stag-single:<idx>:<name>
         idx = int(idx_s)
         return [self.get_axes_by_tag_name(name, spatial_only=True)[idx]]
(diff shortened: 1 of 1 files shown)


albertz:
Why do you need this?

I think it was already discussed somewhere else that even `stag:...` is deprecated and we do not want to encourage more use of that and extend that.

The problem is that the description of a dim tag is ambiguous and also we do not guarantee that it will not change, and thus any config using this might break at some future point.

The reliable way is to just use the dim tag object directly.

vieting:
It would simplify an issue that I have with the pytorch-to-returnn converter a lot. I admit it is not the cleanest way possible, though. Should I open an issue for the converter with that case?

vieting:
Sounds good to me. Should I add a comment that it is not recommended to use that option as it might change behavior in the future?

albertz (approved):
Ok for now.

PR #1246, opened 2022-12-08
require dim tags to be defined

albertz:
"Defined" here means that `dyn_size_ext` is set for dynamic sizes. This can be a template (no `placeholder`). But the shape is important. Usually [B] but not always.

Returnn-common requires this but this should be consistent anyway.

The first commit adds this to the `Data.sanity_check`.

A number of layers don't do this correctly. Namely:

- [x] `ShiftAxisLayer`, e.g. `test_shift_layer`
- [x] `ReinterpretDataLayer` with `set_dim_tags`, e.g. `test_ReinterpretDataLayer_change_batch_to_spatial`
- [x] `SliceNdLayer`, e.g. `test_SliceNdLayer_ReinterpretDataLayer`
- [x] `MergeDimsLayer`
- [x] `UnflattenNdLayer`, via `test_unflatten_2d`
- [x] `MaskedComputationLayer` with given but undefined `out_spatial_dim`, e.g. `test_MaskedComputationLayer_dyn_size_none`, etc
- [x] `RecLayer`, e.g. `test_compile_tf_graph_basic`, `test_reclayer_prev_template_with_out_shape`, etc
- [x] `CumConcatLayer`, e.g. `test_CumConcatLayer_self_attention_equal_to_SelfAttentionLayer`

A few other things broke because of this. This is again a reminder for https://github.com/rwth-i6/returnn/issues/975...

### DIFF
--- a/returnn/tf/layers/base.py
+++ b/returnn/tf/layers/base.py
@@ -517,6 +517,10 @@ class LayerBase(object):
       # In all cases set output.batch, even if the output has no batch dim,
       # as this is important in Data._adapt_batch_consistent_dim_tags().
 
+      global_batch = network.get_root_network().extern_data.get_batch_info(allow_none=True)
+      if global_batch:
+        assert global_batch.is_global_batch()
+
       def _set_global_batch_by_data(data):
         """
         :param Data data:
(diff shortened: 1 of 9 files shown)


albertz:
As expected, a number of layers don't do this correctly. Namely:

- [ ] `SliceNdLayer`, e.g. `test_SliceNdLayer_ReinterpretDataLayer`
- [ ] `MaskedComputationLayer` with given but undefined `out_spatial_dim`
- [ ] ...

PR #1248, opened 2022-12-27
rename_var for initial_state placeholder

Gerstenberger:
This PR does two things:
* rename state vars to have good names when `initial_state == "placeholder"` is used
* fix naming for `SelfAttentionLayer` k_left and v_left state variables

### DIFF
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -4915,6 +4915,9 @@ class RnnCellLayer(_ConcatInputLayer):
           tf.float32, shape=shape_invariant, name="initial_state_placeholder_%s" % key_name)
 
       def rename_var():
+        """
+        Renames final hidden states to have a readable name.
+        """
         if isinstance(rec_layer, RecLayer) and isinstance(rec_layer.cell, _SubnetworkRecCell):
           final_rec_vars = rec_layer.cell.get_final_rec_vars(name)
           last_state = cls.get_state_by_key(final_rec_vars[state_key], key=key, shape=initial_shape)
--- a/returnn/tf/layers/rec.py
+++ b/returnn/tf/layers/rec.py
@@ -4913,6 +4913,16 @@ class RnnCellLayer(_ConcatInputLayer):
       with rec_layer.var_creation_scope():
         ph = tf_compat.v1.placeholder(
           tf.float32, shape=shape_invariant, name="initial_state_placeholder_%s" % key_name)
+
+      def rename_var():
+        if isinstance(rec_layer, RecLayer) and isinstance(rec_layer.cell, _SubnetworkRecCell):
+          final_rec_vars = rec_layer.cell.get_final_rec_vars(name)
+          last_state = cls.get_state_by_key(final_rec_vars[state_key], key=key, shape=initial_shape)
+        else:
+          last_state = rec_layer.get_last_hidden_state(key=key)
+        tf.identity(last_state, name=key_name)
+
+      rec_layer.post_init_hooks.append(rename_var)
       return ph
     else:
       raise Exception("invalid initial state type %r for sub-layer %r, key %r" % (initial_state, name, key))
@@ -6960,11 +6970,13 @@ class SelfAttentionLayer(_ConcatInputLayer):
           prev_k_left = RnnCellLayer.get_rec_initial_state_inner(
             initial_state=initial_state, name=self.name, rec_layer=self,
             state_key="k_left",
+            key="k_left",
             initial_shape=(batch_dim, num_heads, 0, total_key_dim // num_heads),
             shape_invariant=(None, num_heads, None, total_key_dim // num_heads))
           prev_v_left = RnnCellLayer.get_rec_initial_state_inner(
             initial_state=initial_state, name=self.name, rec_layer=self,
             state_key="v_left",
+            key="v_left",
             initial_shape=(batch_dim, num_heads, 0, total_value_dim // num_heads),
             shape_invariant=(None, num_heads, None, total_value_dim // num_heads))
         else:


curufinwe (approved):

albertz:
You need to rebase.

albertz:
`key` has a different meaning in this function. See its usages, e.g.:
```python
            def update_var():
                """
                :return: nothing, calls :func:`TFNetwork.register_post_control_dependencies`.
                """
                if isinstance(rec_layer, RecLayer) and isinstance(rec_layer.cell, _SubnetworkRecCell):
                    final_rec_vars = rec_layer.cell.get_final_rec_vars(name)
                    last_state = cls.get_state_by_key(final_rec_vars[state_key], key=key, shape=initial_shape)
                else:
                    last_state = rec_layer.get_last_hidden_state(key=key)
```

But it seems you are only setting `key` to influence the tensor name? I think we can rewrite the `key_name` logic a bit:
```python
if key is not None:
    key_name = key
elif state_key != "state":
    key_name = state_key
else:
    key_name = "var"
```

Gerstenberger:
I forgot to mention:
Compiling a TF Transformer Graph with the current upstream results in the following error:
```
...
  File "/efs/agerstenberger/tools/returnn/returnn/tf/network.py", line 1284, in TFNetwork._create_layer
    line: layer.post_init(layer_desc)
    locals:
      layer = <local> <SelfAttentionLayer output/'dec_0_self_att_att' out_type=Tensor{[B,T|'time:var:extern_data:delayed'[B],F|F'dec_0_self_att_att_self_att_feat'(512)]}>
      layer.post_init = <local> <bound method LayerBase.post_init of <SelfAttentionLayer output/'dec_0_self_att_att' out_type=Tensor{[B,T|'time:var:extern_data:delayed'[B],F|F'dec_0_self_att_att_self_att_feat'(512)]}>>
      layer_desc = <local> {'attention_dropout': 0.0, 'attention_left_only': True, 'forward_weights_init': "variance_scaling_initializer(mode='fan_in', distribution='uniform', scale=1.0)", 'initial_state': 'keep_over_epoch_no_init', 'n_out': 512, 'num_heads': 8, 'total_key_dim': 512, '_network': <TFNetwork '/output(rec-sub..., len = 13
  File "/efs/agerstenberger/tools/returnn/returnn/tf/layers/base.py", line 328, in LayerBase.post_init
    line: func()
    locals:
      func = <local> <function RnnCellLayer.get_rec_initial_state_inner.<locals>.update_var at 0x7f53815a7550>
  File "/efs/agerstenberger/tools/returnn/returnn/tf/layers/rec.py", line 5508, in RnnCellLayer.get_rec_initial_state_inner.<locals>.update_var
    line: last_state = rec_layer.get_last_hidden_state(key=key)
    locals:
      last_state = <not found>
      rec_layer = <local> <SelfAttentionLayer output/'dec_0_self_att_att' out_type=Tensor{[B,T|'time:var:extern_data:delayed'[B],F|F'dec_0_self_att_att_self_att_feat'(512)]}>
      rec_layer.get_last_hidden_state = <local> <bound method LayerBase.get_last_hidden_state of <SelfAttentionLayer output/'dec_0_self_att_att' out_type=Tensor{[B,T|'time:var:extern_data:delayed'[B],F|F'dec_0_self_att_att_self_att_feat'(512)]}>>
      key = <local> None
  File "/efs/agerstenberger/tools/returnn/returnn/tf/layers/base.py", line 1946, in LayerBase.get_last_hidden_state
    line: assert not self.rec_vars_outputs
    locals:
      self = <local> <SelfAttentionLayer output/'dec_0_self_att_att' out_type=Tensor{[B,T|'time:var:extern_data:delayed'[B],F|F'dec_0_self_att_att_self_att_feat'(512)]}>
      self.rec_vars_outputs = <local> {'k_left': <tf.Tensor 'output/rec/dec_0_self_att_att/concat:0' shape=(?, 8, ?, 64) dtype=float32>, 'v_left': <tf.Tensor 'output/rec/dec_0_self_att_att/concat_1:0' shape=(?, 8, ?, 64) dtype=float32>}
AssertionError
```
Setting the keys fixes that error. Is this the right way to do it?

albertz:
> The commit [8fcd984](https://github.com/rwth-i6/returnn/commit/8fcd98409e11594a7e7cb7342c212fb56baa746f) (PR #348) seems to be the issue. The commit before that runs fine.

That is from Aug 29, 2020. So you say that no-one has used any more recent RETURNN version since then for this?

Gerstenberger:
Now i have time to look into this again to finalize exporting LMs to Onnx.

> > The commit [8fcd984](https://github.com/rwth-i6/returnn/commit/8fcd98409e11594a7e7cb7342c212fb56baa746f) (PR #348) seems to be the issue. The commit before that runs fine.
> 
> That is from Aug 29, 2020. So you say that no-one has used any more recent RETURNN version since then for this?

Maybe some old scripts floating around :( Until last year i personally haven't touched Transformer LMs for some time anymore but this came up when Javier was working on his Transformer LM setups. For us, it helped to set the key also, but this might be the wrong thing to do.

> > The commit [8fcd984](https://github.com/rwth-i6/returnn/commit/8fcd98409e11594a7e7cb7342c212fb56baa746f) (PR #348) seems to be the issue. The commit before that runs fine.
> 
> I wonder a bit what change specifically broke this. Also before that, it used `state_key`, with `state_key="kv_left"`.

I think it would also return the current rec_layer (SelfAttentionLayer); at least that is what is passed into `rec_layer` for `get_rec_initial_state_inner` in the `SelfAttention` layer. The size of `rec_vars_output` was 1 so it never got through to the assertion as there is a specific check for size 1 in `get_last_hidden_state`:
```python
def get_last_hidden_state(self, key):
        """
        If this is a recurrent layer, this would return the last hidden state.
        Otherwise, we return None.

        :param int|str|None key: also the special key "*"
        :rtype: tf.Tensor | None
        :return: optional tensor with shape (batch, dim)
        """
        if key in self.rec_vars_outputs:
            return self.rec_vars_outputs[key]
        if key is None and len(self.rec_vars_outputs) == 1:
            return list(self.rec_vars_outputs.values())[0]
        assert not self.rec_vars_outputs
        return None
```

I can also add a simple test case for compiling self-attention networks in the next few days.

curufinwe:
@Gerstenberger friendly reminder

albertz:
> What is `rec_layer` here? I would have expected it is the _outer_ rec layer.

I'm not sure anymore why I thought this, but when setting a breakpoint in `get_rec_initial_state_inner`, you will see, it is the inner rec layer. This is called via `RecLayer.__init__`:
```python
self._initial_state = RnnCellLayer.get_rec_initial_state(...)
```

But then the true branch in this `update_var` function doesn't make sense, specifically:
```python
final_rec_vars = rec_layer.cell.get_final_rec_vars(name)
```
`name` is the layer name.

albertz:
> rename state vars to have good names when `initial_state == "placeholder"` is used

This should be fixed via #1353.

albertz:
> > fix naming for SelfAttentionLayer k_left and v_left state variables
> 
> Basically, this part of the PR here should be fixed now via #1352.

@curufinwe mentioned [here](https://github.com/rwth-i6/returnn/pull/1352#issuecomment-1618266325) that there might be still another issue? I really need some example config to understand what the problem is.

albertz:
I noticed, when you actually run the exported graph, there is another bug in `SelfAttentionLayer`. This is fixed via #1354.

albertz:
Ok, now that we have #1352, #1353, #1354 and #1355 merged, I think everything from this PR here should be addressed and fixed. I'm closing this now.

@curufinwe (or also @Gerstenberger) can you check if your use cases now work correctly?

curufinwe:
I can confirm that compilation now works with master. I have not tested running the graph. I leave that to @Gerstenberger ^^.

Issue #1245, opened 2022-12-07
LMDataset output size


JackTemaki:
I found a small nuisance while working on the LM pipeline. In the `LmDataset` we define the number of outputs based on the length of label list (`num_labels = len(self.labels["data"])`) instead of the length of the indices. But our LM vocabularies usually start like:

```
<sb> 0
<s> 0
</s> 0
THE 1
AND 2
OF 3
[...]
```

So in this case the output size will be too large by 2 entries. Kazuki did just remove `<s>` and `</s>` from the vocab for RETURNN training and set start and end token to `<sb>` instead of `<s>` and `</s>`  as a workaround, but I consider this to be quite ugly making the pipeline more inconsistent.

Any idea how this can be fixed without breaking the behavior?

albertz:
We always have `labels` as a list, with `len(labels) == num_labels`. If this is not the case, there is a bug in `labels`. We must have this, as we often use `labels[label_idx]`, e.g. for serialization.

This implies that there can only be one label in `labels` per index. However, in the vocab you can have multiple.

So the bug rather seems to be this:
```python
self.labels["data"] = list(self.orth_symbols)
```
And related code.

JackTemaki:
I got the point. For the `Vocabulary` class we do:

```python
 labels = {idx: label for (label, idx) in sorted(d.items())}
 [...]
  self._labels = [label for (idx, label) in sorted(labels.items())]
```
which is making the mapping bijective...

Will do the same for the LM dataset.

PR #1249, opened 2023-01-06
LmDataset fix `labels`

JackTemaki:
Fix for #1245 

Should not break old behavior, as for correct bijective mappings this results in the same sizes, and surjective mappings gave wrong sizes, resulting in models which I assume nobody could use correctly.

### DIFF
--- a/returnn/datasets/lm.py
+++ b/returnn/datasets/lm.py
@@ -125,8 +125,7 @@ class LmDataset(CachedDataset2):
       orth_symbols = open(orth_symbols_file).read().splitlines()
       self.orth_symbols_map = {sym: i for (i, sym) in enumerate(orth_symbols)}
       self.orth_symbols = orth_symbols
-      reverse_map = {i: sym for (i, sym) in enumerate(orth_symbols)}
-      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map)]
+      self.labels["data"] = orth_symbols
       self.seq_gen = None
     if orth_symbols_map_file and orth_symbols_map_file.endswith('.pkl'):
       import pickle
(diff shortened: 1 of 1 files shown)


returnn/datasets/lm.py
```diff
@@ -178,26 +178,27 @@ def __init__(self,
     if word_end_symbol and not word_based:  # Character-based modeling and word_end_symbol is specified.
       self.orth_replace_map[" "] = [word_end_symbol]  # Replace all spaces by word_end_symbol.
 
-    num_labels = len(self.labels["data"])
```
albertz, line 181: I don't see why you changed this. This should always be correct. If not, then `labels` is wrong.

michelwi: there is a problem if two or more tokens get mapped to the same output e.g.
```
<s> 0
<\s> 0
<UNK> 1
hello 2
world 3
```
in this case `num_labels` would be 5 but we only want 4 outputs.

albertz: If `labels` is correct, then `num_labels = 4`. If `num_labels != 4`, then `labels` is incorrect.

albertz: What you listed is not what `labels` is. `labels` is just a list, with the condition that `labels[label_idx] == label`.

albertz: > then maybe the code that generates the `labels` from the `orth_symbols_map` is wrong?

Yes this looks wrong. At least it sounds as if `labels` are wrong here. Because we always must have `len(labels) == num_labels`.

JackTemaki: Yes, `labels` is wrong here, see: https://github.com/rwth-i6/returnn/issues/1245#issuecomment-1373451235

returnn/datasets/lm.py
```diff
@@ -125,14 +125,16 @@ def __init__(self,
       orth_symbols = open(orth_symbols_file).read().splitlines()
       self.orth_symbols_map = {sym: i for (i, sym) in enumerate(orth_symbols)}
       self.orth_symbols = orth_symbols
-      self.labels["data"] = orth_symbols
+      reverse_map = {i: sym for (i, sym) in enumerate(orth_symbols)}
+      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map)]
```
albertz, line 129: I don't understand this. Isn't this just the same as before?

returnn/datasets/lm.py
```diff
@@ -125,14 +125,16 @@ def __init__(self,
...
       self.seq_gen = None
     if orth_symbols_map_file and orth_symbols_map_file.endswith('.pkl'):
       import pickle
       with open(orth_symbols_map_file, 'rb') as f:
         self.orth_symbols_map = pickle.load(f)
       self.orth_symbols = self.orth_symbols_map.keys()
-      self.labels["data"] = list(self.orth_symbols)
+      reverse_map = {i: sym for (sym, i) in self.orth_symbols_map.items()}
+      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map)]
```
albertz, line 137:
```suggestion
      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map.items())]
```
?

returnn/datasets/lm.py
```diff
@@ -154,7 +156,8 @@ def __init__(self,
       assert orth_symbols_imap_list[0][0] == 0
       self.orth_symbols_map = {sym: i for (i, sym) in orth_symbols_imap_list}
       self.orth_symbols = [sym for (i, sym) in orth_symbols_imap_list]
-      self.labels["data"] = self.orth_symbols
+      reverse_map = {i: sym for (i, sym) in orth_symbols_imap_list}
+      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map)]
```
albertz, line 160:
```suggestion
      self.labels["data"] = [sym for (i, sym) in sorted(reverse_map.items())]
```

returnn/datasets/lm.py
```diff
@@ -178,26 +181,26 @@ def __init__(self,
     if word_end_symbol and not word_based:  # Character-based modeling and word_end_symbol is specified.
       self.orth_replace_map[" "] = [word_end_symbol]  # Replace all spaces by word_end_symbol.
 
-    num_labels = len(self.labels["data"])
```
albertz, line 181: I would keep the old variable name `num_labels`.

returnn/datasets/lm.py
JackTemaki: Ah yes, here we have no explicit indices anyway.

returnn/datasets/lm.py
JackTemaki: I do not like it but ok.

returnn/datasets/lm.py
```diff
@@ -132,7 +132,8 @@ def __init__(self,
       with open(orth_symbols_map_file, 'rb') as f:
         self.orth_symbols_map = pickle.load(f)
       self.orth_symbols = self.orth_symbols_map.keys()
-      self.labels["data"] = list(self.orth_symbols)
+      reverse_map = {i: sym for (sym, i) in self.orth_symbols_map.items()}
```
albertz, line 135: I think we need a `sorted` here to have it deterministic on older Python versions.

albertz (approved):

PR #1250, opened 2023-01-09
get_seq_order_for_epoch: fix seq_tags_filter

patrick-wilken:
@sarahberanek Can you try? Should be all that is needed to get seq_tags_filter working again.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -481,7 +481,7 @@ class Dataset(object):
       seq_index = list(seq_index) * repeat_epoch
     if self.seq_tags_filter is not None:
       # Note: This is as generic as possible, but requires that get_all_tags is implemented.
-      assert seq_index
+      assert len(seq_index)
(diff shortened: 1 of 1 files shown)


JackTemaki (changes requested):
Doesn't line 490 need a change as well?

patrick-wilken:
Not necessarily, there it is definitively a list. The problem was that `assert seq_index` does not work for numpy arrays.

(I changed `seq_index` to a generic `typing.Sequence` for efficiency: numpy sorting and random permutations are much faster, and for "default" sequence order we can just use a range, which uses up no memory.)

albertz (approved):

albertz:
So we wait for @sarahberanek ? Otherwise looks fine.

JackTemaki (approved):

sarahberanek:
Hi all, I just tried it and this seems to fix the issue: Training starts smoothly. Thanks everyone!

PR #1251, opened 2023-01-26
add fake quantization layer

JackTemaki:
I am not sure if the `get_out_data_from_opts` is correctly defined, but this is still not clear at all for me how to figure out what should be put there. I found multiple solutions that work, and multiple that did not work without me understanding why, so I just picked the one that calls the least code and works...

### DIFF
--- a/tests/test_TFNetworkLayer.py
+++ b/tests/test_TFNetworkLayer.py
@@ -8132,6 +8132,23 @@ def test_PostfixInTimeLayer():
         assert out[0, src_seq_lens[0] + repeat - 1, 0] == -7
 
 
+def test_FakeQuantizeStaticLayer():
+  with make_scope() as session:
+    num_inputs = 3
+    config = Config()
+    config.update({
+      "extern_data": {"data": {"dim": num_inputs}},
+      "network": {
+        "output": {"class": "fake_quantize_static", "min": -6, "max":6, "num_bits":2, "from": "data"}
+      }})
+    network = TFNetwork(config=config, train_flag=True)
+    network.construct_from_dict(config.typed_value("network"))
+    feed = {network.extern_data.get_data("data").placeholder: numpy.asarray([[[-5,0,3]]])}
+    out = network.get_default_output_layer().output.placeholder
+    result = session.run(out, feed_dict=feed)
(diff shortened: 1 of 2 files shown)


albertz:
> I am not sure if the `get_out_data_from_opts` is correctly defined

I'm not really sure what the problem is/was? Or whether sth is unclear to you? What exactly is unclear?

I think you don't necessarily need to derive from `_ConcatInputLayer` here. You can just derive from `LayerBase`, and assert that you have exactly one input. That makes it a bit simpler.

albertz:
Your current `get_out_data_from_opts` just returns basically the same as the input (concatenated). I was wondering esp about the dtype. So if the input dtype is float32, the output dtype will also be float32.

From the doc on `fake_quant_with_min_max_args`, it seems this is correct for float32, and would anyway fail for any other type. So I guess this is ok.

tests/test_TFNetworkLayer.py
```diff
@@ -8132,6 +8132,23 @@ def test_PostfixInTimeLayer():
...
+    num_inputs = 3
+    config = Config()
+    config.update({
+      "extern_data": {"data": {"dim": num_inputs}},
+      "network": {
+        "output": {"class": "fake_quantize_static", "range_min": -6, "range_max":6, "num_bits":2, "from": "data"}
+      }})
+    network = TFNetwork(config=config, train_flag=True)
+    network.construct_from_dict(config.typed_value("network"))
+    feed = {network.extern_data.get_data("data").placeholder: numpy.asarray([[[-5,0,3]]])}
```
albertz, line 8146: I would rather use the `make_feed_dict` function, which is more correct. (You miss many things here, like the sizes, etc.)

JackTemaki: I did not want to use `make_feed_dict` (I had this in the beginning) because it does not allow me to put manual values.

albertz:
Btw, this layer basically does nothing on the shape/dtype, and simply wraps `tf.quantization.fake_quant_with_min_max_args`. In such cases, I usually would not introduce a new layer and just use eval layer. Or wrap it on returnn-common side, via eval layer. You will find a number of similar examples in returnn-common.

So, I'm not sure if we really need this? I would only add layers which will be often used by many people.

JackTemaki:
ok...

PR #1254, opened 2023-02-01
RelativePositionalEncodingLayer, fix dyn dim tag dyn_size_ext template

### LINKED ISSUE #1253: Attention-ASR setup not working anymore
Our current ASR setup does not work beginning from commit 7eee625cef492601cc2f639674860ce320cd5356.

The full log (including network and stacktrace) can be found here: https://gist.github.com/JackTemaki/a338415827154833662615cbd9a8d8c3

Unfortunately I do not have time to debug this.

albertz:
Fix #1253

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -4568,6 +4568,7 @@ class RepeatLayer(_ConcatInputLayer):
     :param Dim|None out_dim:
     """
     super(RepeatLayer, self).__init__(out_dim=out_dim, **kwargs)
+    from returnn.tf.util.data import BatchInfo
     self.repetitions = repetitions
     if isinstance(self.repetitions, int):
       repetitions_data = Data.from_tensor(tf.constant(self.repetitions))
@@ -4594,8 +4595,8 @@ class RepeatLayer(_ConcatInputLayer):
 
       :param Data data:
       :param BatchInfo|None other_batch: use this batch info if we should add some
-      :rtype: tf.Tensor
-      :return: the placeholder, with shape [B, T, ...], maybe with added batch dim.
+      :rtype: Data
+      :return: data, with shape [B, T, ...], maybe with added batch dim.
       """
       if self.output.have_batch_axis():
         if data.have_batch_axis():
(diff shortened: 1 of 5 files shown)


albertz:
There are some somewhat independent commits in here, so this should be merged, not squashed.

albertz:
This is still somewhat incomplete. `RelativePositionalEncodingLayer` is actually not really recommended anymore, as it is right now, as it does not properly handle dim tags. Specifically, this code:
```
    if data.have_time_axis():
      time_dim_tag = data.get_time_dim_tag()
      # TODO using same dim tag twice will not be supported at some future point...
      data = data.copy_template_new_dim_tags((time_dim_tag, time_dim_tag, feature_dim_tag))
```
This TODO is a problem. It means the order of dim tags is relevant. So with new behavior version, no other layer can properly distinguish those dims without an error. This must be fixed. That's also why the test case does not work yet.

albertz:
> This is still somewhat incomplete. `RelativePositionalEncodingLayer` is actually not really recommended anymore, as it is right now, as it does not properly handle dim tags. Specifically, this code:
> 
> ```python
>     if data.have_time_axis():
>       time_dim_tag = data.get_time_dim_tag()
>       # TODO using same dim tag twice will not be supported at some future point...
>       data = data.copy_template_new_dim_tags((time_dim_tag, time_dim_tag, feature_dim_tag))
> ```
> 
> This TODO is a problem. It means the order of dim tags is relevant. So with new behavior version, no other layer can properly distinguish those dims without an error. This must be fixed. That's also why the test case does not work yet.

Ok, I worked around that problem now by using `SelfAttentionLayer` with the `key_shift` option in the test, which expects the output of `RelativePositionalEncodingLayer` in this specific way.

Note, with returnn_common, there is a more direct implementation of rel pos encoding, and also for self-attention, so both those layers are not needed, and can maybe be treated as deprecated.

patrick-wilken:
> [...] and can maybe be treated as deprecated.

Please don't. Just because there is an implementation in another repo does not mean the layer is not useful anymore in (core) RETURNN.

albertz:
This is not about returnn_common. returnn_common is just a repo to make it simpler to write RETURNN net dicts. You can also write the net dict manual, using just core RETURNN. So even in core RETURNN, I don't think the `SelfAttentionLayer` is useful anymore. It has become way too bloated, and you can do the same with other more low level layers (#391). I would not recommend anyone to use it anymore. It does not make sense to further extend it more and more by whatever future feature, and make it even more complex and bloated. And it already does not support many important features, like more modern relative positional encoding variants.

Note that by deprecated, I don't mean to remove it. We are quite strict in keeping compatibility to older configs. I just mean that the documentation might have a passage discouraging people from using it, and suggesting better alternatives (#391). And I mean that it does not make sense to extend it more and more by new features. But if you want, you can take ownership of that layer, and be responsible for it, and then add whatever you want. Maybe we move it to a separate file.

patrick-wilken:
Oh, sorry, I only meant `RelativePositionalEncodingLayer`, that was not clear. I didn't want to reopen the discussion about `SelfAttentionLayer`. 😅 But `RelativePositionalEncodingLayer` is a pretty simple layer and does not have an inherent problem, or does it?

albertz:
Ah ok. Actually I meant both layers.

> But `RelativePositionalEncodingLayer` is a pretty simple layer and does not have an inherent problem, or does it?

Mostly the same arguments apply for `RelativePositionalEncodingLayer`, although it is much less bloated so far. But only because it also only supports a single case (which is not really up-to-date anymore).

But then, more importantly, as already explained above, this layer is also somewhat broken in its current state. Specifically, the order of the dimensions matters, which is a violation of the [RETURNN principles](https://github.com/rwth-i6/returnn/wiki/RETURNN-principles), and which means that potentially things can break when you use it together with other layers, as many layers can reorder dimensions. Further, it uses the same dimension tag twice, which is also against the RETURNN principles, as it's not possible to distinguish dimensions then. It currently only really works together with `SelfAttentionLayer` and otherwise it would be broken.

PR #1259, opened 2023-02-09
Fix `-no_ogg` option in bliss-to-ogg-zip.py tool

JackTemaki:
The current code had undefined behavior with respect to the written duration or even crashed for some bliss datasets with missing audio.

Now the duration is always given by the text length if the `-no_ogg` option is set. This does change the behavior, but I am not sure how we want to deal with it in tools. If this behavior change is not accepted, I can keep my local version for experiments that require this. I do not want to add an extra flag because this would make the way the tool works even more complicated and non-understandable.

### DIFF
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -180,7 +180,7 @@ class OggZipDataset(CachedDataset2):
         first_entry = data[0]
         assert isinstance(first_entry, dict)
         assert isinstance(first_entry["text"], str)
-        assert isinstance(first_entry["duration"], float)
+        assert isinstance(first_entry["duration"], (float, int))
         # when 'audio' is None and sequence names are given, this dataset can be used in text-only mode
         if "file" in first_entry:
             assert isinstance(first_entry["file"], str)
--- a/tools/bliss-to-ogg-zip.py
+++ b/tools/bliss-to-ogg-zip.py
@@ -345,8 +345,11 @@ def main():
         rec_filename = seq.recording_filename
         assert os.path.isfile(rec_filename) or args.no_ogg
         assert (seq.start_time < seq.end_time and seq.delta_time > 0) or args.no_ogg
-        duration = seq.delta_time
-        assert duration > 0 or args.no_ogg
+        if args.no_ogg:
+            duration = len(seq.orth.split(" "))
+        else:
+            duration = seq.delta_time
+        assert duration > 0
         total_duration += duration
         assert rec_filename.startswith(rec_filename_common_prefix) and rec_filename.endswith(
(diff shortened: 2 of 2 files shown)


albertz (changes requested):
In principle, such an option is ok. But I would rename it to make it more clear. Specifically, I would add an option like `--duration-type` or so, and this can be either "audio" or "words".

But on the problem itself: I think we should fix the problem itself. I think using the audio seq length as duration makes more sense. In the end, this is used for sorting in the dataset, and in most cases, the audio seq length is much more relevant than the number of words, e.g. w.r.t. padding and memory consumption, the audio len will always dominate.

Do we actually have an issue about the problem itself, where we can discuss this further, and potential fixes? I see this here just as a bad workaround, with the described disadvantage.

JackTemaki:
> In principle, such an option is ok. But I would rename it to make it more clear. Specifically, I would add an option like `--duration-type` or so, and this can be either "audio" or "words".

This PR is not about a new option, but an missing use case in an existing one which crashes the script.

> 
> **Edit** I just realize, your option does more than just that. But I would not mix this. If you also want to be able to not process the audio at all, add an option like `--audio-format`, which can be "ogg" or "none". But this was actually incomplete here? You still do the ffmpeg call, even with your option.

No, ffmpeg was never called with the option.

> 
> But on the problem itself: I think we should fix the problem itself. I think using the audio seq length as duration makes more sense. In the end, this is used for sorting in the dataset, and in most cases, the audio seq length is much more relevant than the number of words, e.g. w.r.t. padding and memory consumption, the audio len will always dominate.
> 

I have a case where this duration does not exist, thus the change.

> Do we actually have an issue about the problem itself, where we can discuss this further, and potential fixes? I see this here just as a bad workaround, with the described disadvantage.

Using the OggZipDataset without Audio is a workaround for itself, and I just do this in lack of another Dataset offering the same functionality. I will keep the branch around for my experiments, but lets not merge this then. For the next experiment setup I will think of something completely different.

albertz:
> This PR is not about a new option, but an missing use case [for the existing `--no_ogg`] which crashes the script.

Ah sorry, I did not realize this option was already there.

> I have a case where this duration does not exist, thus the change.

But this still sounds like you are mixing things. `--no_ogg` does not necessarily imply that there is no duration. I still would introduce a separate option for that, like my suggested `--duration-type`. But otherwise, you can keep your code.


> Using the OggZipDataset without Audio is a workaround for itself, and I just do this in lack of another Dataset offering the same functionality.

What exactly do you miss? But this should better be a separate issue, not discussed here.


> but lets not merge this then

Why? What's the problem?

PR #1261, opened 2023-02-14
Tensor abstraction

### LINKED ISSUE #1165: Abstract and move ExternData/Data/Dim
Basically the whole file `returnn/tf/util/data.py` and then also the class `ExternData` should be moved to a framework common place, maybe like `returnn/data`, maybe separare submodules there like `extern_data.py`, `data.py`, `dim.py`, `batch_info.py`.

Then all TF specific code should be removed. This should not be much. This is `auto_create_placeholders` and a couple of the `copy_...` functions.

`Data` and co was always designed in a way to be generic and not TF specific. Now that we work on PyTorch (#1120), it becomes relevant for that as well (https://github.com/rwth-i6/returnn/issues/1120#issuecomment-1286887444).

`Data` and `Dim` should ideally also be simplified, or most importantly the core object and the `__init__`. The `__init__` should execute very fast, as this would be required for eager mode. Also see #975.

### LINKED ISSUE #1260: Remove `Dim` `generic` flag
I think this is basically not used yet, and we want to simplify the class (#1165, #975).

So, any reasons to keep it?

albertz:
* This moves `Data` and `Dim` to a more central place `returnn.tensor`
* Rename `Data` to `Tensor`.
* Abstracts away from TF specific code, and allows for any raw tensor type. This is via the new frontends.
* Separates the core of `Tensor` to additional attribs in `_TensorExtra` which is mostly only needed for legacy code, and most methods are in `_TensorMixin`. Similarly for `Dim`. To the user, these are implementation details and should not matter. For a new user interested in `Tensor`/`Dim`, it should be enough to just look the main classes.
* Fix #1165.
* Fix #1260.
* Related: #975.

### DIFF
--- a/returnn/tensor/dim.py
+++ b/returnn/tensor/dim.py
@@ -92,3 +92,9 @@ batch_dim = Dim(kind=Dim.Types.Batch, description="global batch", dimension=None
 
 # This indicates to perform a single step execution of some layer which can potentially have recurrent state.
 single_step_dim = Dim(description="single-step", kind=Dim.Types.Spatial, special=True, dimension=1)
+
+
+class VerifyOutShapeException(Exception):
+    """
+    Exception via :func:`Data.verify_out_shape`.
+    """
--- a/returnn/tensor/dim.py
+++ b/returnn/tensor/dim.py
@@ -82,6 +82,10 @@ class Dim(_DimMixin):
         if kwargs:
             # TODO filter kwargs, avoid creation of _DimExtra if not needed
             self._extra = _DimExtra(dim=self, **kwargs)
+            if kwargs.get("dyn_size") is not None:
+                self.dyn_size = kwargs["dyn_size"]
+            if self.derived_from_op and self.is_dynamic():
+                self.complete_dyn_size()
 
     def __repr__(self):
         return "Dim{%s}" % self.short_repr()
--- a/returnn/tensor/dim.py
+++ b/returnn/tensor/dim.py
@@ -80,12 +80,7 @@ class Dim(_DimMixin):
         self._extra = None
 
         if kwargs:
-            # TODO filter kwargs, avoid creation of _DimExtra if not needed
-            self._extra = _DimExtra(dim=self, **kwargs)
-            if kwargs.get("dyn_size") is not None:
-                self.dyn_size = kwargs["dyn_size"]
-            if self.derived_from_op and self.is_dynamic():
-                self.complete_dyn_size()
+            self._handle_extra_kwargs(**kwargs)
 
     def __repr__(self):
         return "Dim{%s}" % self.short_repr()
--- a/returnn/frontend_api.py
+++ b/returnn/frontend_api.py
@@ -143,6 +143,17 @@ class Frontend(Generic[T]):
         raise NotImplementedError
 
     @staticmethod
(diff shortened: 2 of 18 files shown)


albertz:
Ok, this fails:
```python
def test_Data_copy_move_axis_time_to_end():
    d1 = Data(name="att_weights", shape=(None, None, 4))
    d2 = d1.copy_move_axis(d1.time_dim_axis, -1)
    assert d2.shape == (None, 4, None) and d2.feature_dim_axis == 2 and d2.time_dim_axis == 3
```
Because by default we leave time_dim_axis as not-specified, and the `copy_move_axis` (`copy_transpose`) now does not modify this. I wonder what we should do about this. Whether there are any setups depending on this old behavior. Maybe new behavior version?

albertz:
I don't like the idea of using the behavior version too much, as it adds a dependency, and also makes the behavior maybe less intuitive.

The current behavior in the PR is:
`feature_dim_axis`/`time_dim_axis` are NotSpecified by default and stored like this, and when code accesses them, it infers a reasonable default according to some somewhat outdated rules.

The behavior in current master is mostly the same but actually a small difference: `time_dim_axis` is not stored as NotSpecified but it directly infers it in `__init__`. So later on we don't know anymore whether it was not specified or specified.

A more reasonable new default would be:
`feature_dim_axis`/`time_dim_axis` are None by default. We don't really need them at all anymore.

The problem with the new default is, this obviously would break lots of code, so it's not really an option.

The problem with the old defaults is, this adds overhead, in runtime and also code logic, which further adds potential bugs. In the current PR implementation, it would create that `_extra` object, i.e. also overhead in memory.

Potential solutions:

Maybe there could be a separate version tag just for `Tensor` which switches between these defaults? However, how to determine the version automatically? What's the default version? We could maybe say, when `dims` or `dim_tags` was specified, we select the new version, otherwise (with `shape` etc), we select the old version. But this is still wrong in a number of cases. Or only when `dims` is selected (as this is a new arg introduced here in the PR). This should be safe. But is this a good idea?

albertz:
Milestone achieved: test_TFUtil passes 100%.

albertz:
Finally, all tests are passing.

albertz:
I will try to separate a few unrelated commits from here. But otherwise, I don't see how this can be structured well, so I guess I just squash everything together.

returnn/tensor/dim.py
```diff
@@ -0,0 +1,99 @@
...
+        self,
+        dimension: Optional[Union[int, _t.Tensor]],
+        *,
+        name: Optional[str] = None,
+        capacity: Optional[int] = None,
+        dyn_size_ext: Optional[_t.Tensor] = None,
+        description: Optional[str] = None,
+        **kwargs,
+    ):
+        self.name = name or description
```
Icemole, line 63: Right now if `name` is passed, `description` is ignored. I think `description` could be a useful addition for debugging purposes as an attribute separate to `name`. @albertz what do you think?

albertz: Currently `description` in all old code (also external code) is treated like a `name`.

albertz: And in this PR, `description` and `name` are just aliases of each other.

albertz:
If there are no obvious important issues, I will merge it soon. We can further improve it iteratively then.

albertz:
Ok merged now.

PR #1266, opened 2023-03-06
TorchFrontend: dot product

patrick-wilken:
Making this a pull request because I added a new test module, want to make sure it is executed correctly.
Also, there are probably things to discuss. 🙃 

So, I added `dot()` to `Frontend` and implemented it for the `TorchFrontend`. It seems to work for pretty much all cases, see the tests.
Masking is still missing as I didn't understand the implementation in DotLayer right away, will look into it... Also, as discussed, handling sparse inputs is not yet implemented.

### DIFF
--- a/returnn/torch/frontend/_internal_frontend.py
+++ b/returnn/torch/frontend/_internal_frontend.py
@@ -1,0 +1,42 @@
+"""
+Internal frontend for PyTorch frontend.
+"""
+
+from __future__ import annotations
+from typing import Optional, Tuple
+
+import torch
+
+from returnn._internal_frontend_api import InternalFrontend
+from returnn.tensor import Tensor
+
+_TT = Tensor[torch.Tensor]
+
+
+# Ignore this warning until we implemented everything.
+# noinspection PyAbstractClass
+class TorchInternalFrontend(InternalFrontend[torch.Tensor]):
+    """
+    Internal frontend for PyTorch frontend.
+    """
+
+    @staticmethod
+    def get_dtype_name_raw(raw_tensor: torch.Tensor) -> str:
+        """
+        :return: dtype of raw tensor, as string
+        """
+        return str(raw_tensor.dtype).replace("torch.", "")
+
+    @staticmethod
+    def get_ndim_raw(raw_tensor: torch.Tensor) -> int:
+        """
+        :return: ndim of raw tensor
+        """
+        return raw_tensor.dim()
+
+    @staticmethod
+    def get_known_shape_raw(raw_tensor: torch.Tensor) -> Tuple[Optional[int]]:
+        """
+        :return: shape of raw tensor; here for PyTorch the full shape is always known
+        """
+        return tuple(raw_tensor.size())
--- a/returnn/torch/frontend/frontend.py
+++ b/returnn/torch/frontend/frontend.py
@@ -3,10 +3,17 @@ Frontend for exposing PyTorch-specific functionality.
 """
 
 from __future__ import annotations
+from typing import Union, Sequence
 
 import torch
 
 from returnn.frontend_api import Frontend
+from ._internal_frontend import TorchInternalFrontend
+from returnn.tensor import Tensor, Dim
+
+from returnn.util.basic import prod
+
+_TT = Tensor[torch.Tensor]
 
 
 # Ignore this warning until we really expect that we implemented everything.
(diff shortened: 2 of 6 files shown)


returnn/torch/frontend/_internal_frontend.py
```diff
@@ -0,0 +1,42 @@
...
+
+from returnn._internal_frontend_api import InternalFrontend
+from returnn.tensor import Tensor
+
+_TT = Tensor[torch.Tensor]
+
+
+# Ignore this warning until we implemented everything.
+# noinspection PyAbstractClass
+class TorchInternalFrontend(InternalFrontend[torch.Tensor]):
```
patrick-wilken, line 18: Only implemented what I needed for the dot tests so far...

returnn/torch/frontend/frontend.py
```diff
@@ -17,3 +24,78 @@ class TorchFrontend(Frontend[torch.Tensor]):
...
+        assert all(
+            dim in b_dims for dim in reduce
+        ), f"'b' does not have the specified reduce dim(s) {reduce} (b dims: {b_dims})"
+
+        reduce_dims_sorted_by_a = [dim for dim in a_dims if dim in reduce]
+
+        a_reduce_axes = [a_dims.index(reduce_dim) for reduce_dim in reduce_dims_sorted_by_a]
+        b_reduce_axes = [b_dims.index(reduce_dim) for reduce_dim in reduce_dims_sorted_by_a]
+
+        a_unique_axes = [i for (i, dim) in enumerate(a_dims) if dim not in b_dims]
```
patrick-wilken, line 52: Naming ok? I found "unique" clearer than "var". But I can also make it consistent with what is used in `DotLayer`.

tests/test_torch_frontend.py
```diff
@@ -0,0 +1,273 @@
...
+tests for returnn.torch.frontend
+"""
+
+import _setup_test_env  # noqa
+
+import torch
+import pytest
+
+from returnn.tensor import Tensor, Dim, batch_dim
+from returnn.torch.frontend import TorchFrontend as fe
```
albertz, line 11: Our convention was to use `rf` (RETURNN frontend).

returnn/torch/frontend/frontend.py
```diff
@@ -17,3 +24,74 @@ class TorchFrontend(Frontend[torch.Tensor]):
...
+
+        reduce_axes_total_dimension = prod(reduce_axes_shape)
+
+        a_raw = torch.permute(a_raw, a_common_axes + a_unique_axes + a_reduce_axes)
+        a_raw = torch.reshape(a_raw, common_axes_shape + (a_unique_axes_total_dimension, reduce_axes_total_dimension))
+
+        b_raw = torch.permute(b_raw, b_common_axes + b_reduce_axes + b_unique_axes)
+        b_raw = torch.reshape(b_raw, common_axes_shape + (reduce_axes_total_dimension, b_unique_axes_total_dimension))
+
+        raw_result = torch.matmul(a_raw, b_raw)
```
albertz, line 88: I'm not sure if we really want to call `matmul` here. I think calling either `torch.bmm` or `torch.mm` directly makes the code more clear, maybe faster, also more bug-proof.

For `torch.bmm`, I think you need to merge all the `common_axes_shape` together. If that product is 1, you can just call `torch.mm`.

tests/test_torch_frontend.py
```diff
@@ -0,0 +1,273 @@
+"""
+tests for returnn.torch.frontend
+"""
+
+import _setup_test_env  # noqa
+
+import torch
+import pytest
+
+from returnn.tensor import Tensor, Dim, batch_dim
```
albertz, line 10: `batch_dim` should not be used here. Rather, just create your own `Dim(...)` for it.

albertz: Note: It's not really so clear yet whether we need the `batch_dim` at all for the new frontend API. I think not. Maybe it should be moved elsewhere, or only specific for the old RETURNN net dict. Not sure.

Related is also `single_step_dim`. I didn't really thought about it yet.

But just avoid using it, unless there is a good reason, but then let's discuss first.

returnn/torch/frontend/frontend.py
patrick-wilken: Hmm, if there's a difference at all I would assume it is slower to do the shape check and the extra reshape from Python. It's true that matmul has many different behaviours, but the special cases are about 1-D inputs and broadcasting. Here we have neither. I think this is the implementation: https://github.com/pytorch/pytorch/blob/758ad516f32708de243f194144ee0f7b9e0f5117/aten/src/ATen/native/LinearAlgebra.cpp#L340
There is a direct fallback to `mm` and reshaping before calling `bmm`, so pretty much exactly what we would do here. But it's also obvious how to do it here, so I can change it if you want.

albertz: There is quite some logic involved in matmul, and we don't need any of that, we know directly how to call `bmm` or `mm`. Also, at least for me, when looking at the matmul code or documentation, it is not immediately obvious that it really does what we need.

patrick-wilken: Ok, changed it.

albertz (approved):

albertz:
As a follow-up comment, one this missing here is the masking logic. This will be handled in a further update.

Issue #1265, opened 2023-03-06
`Dim` no longer has `dimension` attribute


JackTemaki:
Code in in the convolution layer in `returnn_common` is trying to set the `dimension` attribute of a `Dim` tag, which no longer exists after the last re-work. I assume we are missing a setter here?

```python
  File "/u/rossenbach/experiments/tts_asr_2021/recipe/returnn_common/nn/conv.py", line 479, in make_conv_out_spatial_dims
    line: out_spatial_dims[i].dimension = ConvLayer.calc_out_dim(
            in_dim=in_spatial_dims[i].dimension,
            filter_size=filter_size[i],
            stride=strides[i],
            dilation_rate=dilation_rate[i],
            padding=padding)
    locals:
      out_spatial_dims = <local> [Dim{'conformer_aed_model/encoder/input_layer/pool1d/_pool_nd:out-spatial-dim0'[?]}]
      i = <local> 0
      dimension = <not found>
      ConvLayer = <local> <class 'returnn.tf.layers.basic.ConvLayer'>
      ConvLayer.calc_out_dim = <local> <bound method ConvLayer.calc_out_dim of <class 'returnn.tf.layers.basic.ConvLayer'>>
      in_dim = <not found>
      in_spatial_dims = <local> [Dim{F'extractor_mel_feature'(80)}]
      filter_size = <local> [2]
      stride = <not found>
      strides = <local> [2]
      dilation_rate = <local> [1]
      padding = <local> 'same'
AttributeError: can't set attribute

```

JackTemaki:
I would have tried to fix it myself, but it is not clear how this should be resolved. I understand that `dimension` is something that should no longer be used explicitly, and this information is now distributed among `capacity`, `size` and `dyn_size_ext` (this variable should definitely be renamed).

I now see two options:
 1. There should be a setter which has the logic from `dimension` to the three variables just as the current constructor has.
 2. `dimension` should not used anymore and the code in the convolution layer should be changed accordingly.
 
Unfortunately it is not clear from the code which of these two options (or even a third one I do not see) is the preferable one to solve this issue.

albertz:
Well, I don't know.

[...]

PR #1272, opened 2023-03-10
Dim: add set_sizes function

JackTemaki:
Related to #1265 . I am not sure about the best name of the function...


JackTemaki:
I just saw that this is no longer needed....

albertz:
> I just saw that this is no longer needed....

Why?

JackTemaki:
> > I just saw that this is no longer needed....
> 
> Why?

See #1265

If you think it is still useful and would simplify code in returnn-common we can keep it of course.

PR #1274, opened 2023-03-13
Frontend combine, tensor ops

Icemole:
Creates `combine()`, handles tensor operations (e.g. Tensor + Tensor), and some more improvements.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -261,6 +261,7 @@ jobs:
           - TEST=demos RETURNN_DISABLE_TF=1
           - TEST=PTDataset
           - TEST=torch_frontend
+          - TEST=torch_internal_frontend
 
     steps:
     - uses: actions/checkout@v3
--- a/returnn/frontend_api.py
+++ b/returnn/frontend_api.py
@@ -42,6 +42,21 @@ class Frontend(Generic[T]):
         """
         raise NotImplementedError
 
+    @staticmethod
+    def _tensor_op_sanity_check(t1: Tensor, t2: Tensor):
+        """
+        Performs basic checks common to all _TensorMixin binary operations,
+        like _TensorMixin.__eq__ or _TensorMixin.__add__.
+
+        :param t1:
+        :param t2:
+        """
+        assert isinstance(t1, _t.Tensor) and isinstance(
+            t2, _t.Tensor
+        ), "The two operands of a binary operation must be of class Tensor."
+        assert t1.raw_frontend == t2.raw_frontend, "Cannot combine tensors from two different frontends, e.g. TF and PT"
+        assert t1.dtype == t2.dtype, "For now only operations with Tensors of the same dtypes are supported."
+
     @classmethod
     def compare(
         cls,
@@ -64,6 +79,7 @@ class Frontend(Generic[T]):
         """
         a = cls.convert_to_tensor(a)
         b = cls.convert_to_tensor(b)
+        cls._tensor_op_sanity_check(a, b)
         all_dims = []
         for dim in a.dims + b.dims:
             if dim not in all_dims:
@@ -83,6 +99,61 @@ class Frontend(Generic[T]):
         a = a.copy_compatible_to(out, check_sparse=False, check_dtype=False)
         b = b.copy_compatible_to(out, check_sparse=False, check_dtype=False)
         out.raw_tensor = cls._internal_frontend.compare_raw(a.raw_tensor, kind, b.raw_tensor)
+        return out
+
+    @classmethod
+    def combine(
+        cls,
+        a: Union[Tensor, RawTensorTypes],
+        kind: str,
+        b: Union[Tensor, RawTensorTypes],
+        *,
+        allow_broadcast_all_sources: Optional[bool] = None,
+        dim_order: Optional[Sequence[Dim]] = None,
+    ) -> Tensor:
+        """
+        :param a:
+        :param kind: "add"|"+", "sub"|"-", "mul"|"*", "truediv"|"/", "floordiv"|"//", "mod"|"%", "pow"|"**",
+            "max"|"maximum", "min"|"minimum", "logical_and", "logical_or", "squared_difference"
+        :param b:
+        :param allow_broadcast_all_sources: if True, it is allowed that neither a nor b has all dims of the result.
+            Not needed when out_dims is specified explicitly.
+        :param dim_order: defines the order of the resulting dims. if None, it is automatically inferred from a and b.
+            Not all the dims of a and b need to be specified here, and there could also be other dims in the dim_order.
+        :return: element-wise combination of a and b
+        """
+        # Truediv checks for int/int division
+        if kind in {"truediv", "/"}:
+            if isinstance(b, (int, float, numpy.number)) and b == 1:
+                return a
+            elif isinstance(b, (int, float, numpy.number)):
+                assert not a.dtype.startswith(
+                    "int"
+                ), "Dividing a Tensor of type int by an integer is disallowed. Please convert the Tensor to float."
+        a = cls.convert_to_tensor(a)
+        b = cls.convert_to_tensor(b)
+        cls._tensor_op_sanity_check(a, b)
+        all_dims = []
+        for dim in a.dims + b.dims:
+            if dim not in all_dims:
+                all_dims.append(dim)
+        if all(set(x.dims) != set(all_dims) for x in (a, b)):
+            if allow_broadcast_all_sources is False:
+                raise ValueError(f"combine: sources {a!r} {b!r} not allowed with allow_broadcast_all_sources=False")
+            elif allow_broadcast_all_sources is None:
+                raise ValueError(f"combine: sources {a!r} {b!r} require explicit allow_broadcast_all_sources=True")
+            elif allow_broadcast_all_sources is True:
+                pass
+            else:
+                raise TypeError(f"invalid type for allow_broadcast_all_sources: {type(allow_broadcast_all_sources)}")
+        if dim_order:
+            all_dims.sort(key=lambda d: dim_order.index(d) if d in dim_order else len(dim_order))
+        assert a.dtype == b.dtype, "For now only operations with Tensors of the same dtypes are supported."
+        out = _t.Tensor("combine", dims=all_dims, dtype=a.dtype)
+        a = a.copy_compatible_to(out, check_sparse=False, check_dtype=False)
+        b = b.copy_compatible_to(out, check_sparse=False, check_dtype=False)
+        out.raw_tensor = cls._internal_frontend.combine_raw(a.raw_tensor, kind, b.raw_tensor)
+
         return out
 
     @staticmethod
--- a/returnn/tensor/_tensor_extra.py
+++ b/returnn/tensor/_tensor_extra.py
@@ -609,6 +609,83 @@ class _TensorMixin:
     def __hash__(self):
         return id(self)
 
+    # _TensorMixin.__eq__ is disabled as per the following error in some TF tests:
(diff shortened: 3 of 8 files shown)


returnn/tensor/tensor.py
```diff
@@ -21,11 +21,13 @@
...
 
 from __future__ import annotations
-from typing import Optional, Sequence, Tuple, Generic, TypeVar
+from typing import Optional, Sequence, Tuple, Generic, TypeVar, Union
+from numpy import number as numpy_number
 
 from .dim import Dim
 from ._tensor_extra import _TensorExtra, _TensorMixin
 from . import _tensor_extra
+from .. import frontend_api as _frontend_api
```
albertz, line 30: As I told you, remove this here, or make it conditional with TYPE_CHECKING.

returnn/tensor/tensor.py
```diff
@@ -123,5 +125,141 @@ def raw_tensor(self, value: Optional[RawTensorType]):
         self._raw_tensor = value
         self.sanity_check(assume_complete=False)
 
+    def _tensor_op_sanity_check(self, other: Union[_frontend_api.RawTensorTypes, Tensor]):
```
albertz, line 128: Remove this function here. This should be directly called in `combine` and `compare`.

returnn/tensor/tensor.py
```diff
@@ -123,5 +125,141 @@ def raw_tensor(self, value: Optional[RawTensorType]):
...
+
+    def __rtruediv__(self, other: Union[_frontend_api.RawTensorTypes, Tensor]) -> Tensor:
+        self._tensor_op_sanity_check(other)
+        if isinstance(other, (int, float, numpy_number)):
+            if other == 1:
+                return self
+            else:
+                assert not self.dtype.startswith(
+                    "int"
+                ), "Dividing a Tensor of type int by an integer is disallowed. Please convert the Tensor to float."
```
albertz, line 215: Remove such logic here. This logic should be in `combine`.

returnn/tensor/tensor.py
```diff
@@ -123,5 +125,141 @@ def raw_tensor(self, value: Optional[RawTensorType]):
...
+
+    def __truediv__(self, other: Union[_frontend_api.RawTensorTypes, Tensor]) -> Tensor:
+        self._tensor_op_sanity_check(other)
+        if isinstance(other, (int, float, numpy_number)):
+            if other == 1:
+                return self
+            else:
+                assert not self.dtype.startswith(
+                    "int"
+                ), "Dividing a Tensor of type int by an integer is disallowed. Please convert the Tensor to float."
```
albertz, line 203: Remove this here. This should be in `combine`.

.github/workflows/main.yml
```diff
@@ -261,6 +261,7 @@ jobs:
           - TEST=demos RETURNN_DISABLE_TF=1
           - TEST=PTDataset
           - TEST=torch_frontend
+          - TEST=torch_internal_frontend
```
albertz, line 264: As discussed, the tests should be torch independent. We should not have torch specific tests here.

albertz:
All the merge commits should be removed. You rather should rebase.

albertz:
Please check tests. You should configure your environment correctly. It seems you do not have Black configured. Please refer to the wiki.

tests/test_torch_internal_frontend.py
```diff
@@ -4,15 +4,16 @@
...
 
 from returnn.tensor import Tensor, Dim
 from returnn.torch.frontend import TorchFrontend as rf
 
 
 def test_compare_eq():
-    a_raw = rf.Tensor([2.0, 2.0, 2.0])
-    b_raw = rf.Tensor([1.0, 2.0, 3.0])
+    a_raw = torch.tensor([2.0, 2.0, 2.0])
+    b_raw = torch.tensor([1.0, 2.0, 3.0])
```
albertz, line 16: Don't do this. Use only `rf` functions.

albertz:
Why did you add those functions to `Tensor` and not to `_TensorMixin`?

Icemole:
> Why did you add those functions to `Tensor` and not to `_TensorMixin`?

Because `compare` and `combine` required a `Tensor` as parameter, which is not always the case if I do `_TensorMixin.__add__` as a `_TensorMixin` is not always a `Tensor`.
But if you want I can change those to `_TensorMixin` and place the `__add__` functions etc in `_TensorMixin`.

.github/workflows/main.yml
Icemole: Since we need this PR merged I won't do this here, we can implement the unified tests later on.

albertz:
> > Why did you add those functions to `Tensor` and not to `_TensorMixin`?
> 
> Because `compare` and `combine` required a `Tensor` as parameter, which is not always the case if I do `_TensorMixin.__add__` as a `_TensorMixin` is not always a `Tensor`. But if you want I can change those to `_TensorMixin` and place the `__add__` functions etc in `_TensorMixin`.

`_TensorMixin` is supposed to be a mixin, i.e. `self` is always a `Tensor`. You can also specify this explicitly via type hints. See some of the other methods.

tests/test_torch_internal_frontend.py
Icemole: I will use `convert_to_tensor`.

.github/workflows/main.yml
albertz: Ok.

tests/test_torch_internal_frontend.py
Icemole: For these I believe we should have `rf.constant()` as we discussed in the past, which is not available yet. If this is correct, I suggest we temporarily leave it like this.

albertz: Just add `rf.constant` then. Or use `rf.range` or sth else. This really should not be complicated.

returnn/frontend_api.py
```diff
@@ -42,6 +42,20 @@ def convert_to_tensor(value: Union[Tensor, T, RawTensorTypes]) -> Tensor[T]:
         """
         raise NotImplementedError
 
+    @staticmethod
+    def _tensor_op_sanity_check(t1: Union[RawTensorTypes, Tensor], t2: Union[RawTensorTypes, Tensor]):
+        """
+        Performs basic checks common to all _TensorMixin binary operations,
+        like _TensorMixin.__eq__ or _TensorMixin.__add__.
```
albertz, line 49: There should be an empty line after the descriptions, before the params. (Keep consistency.)

returnn/frontend_api.py
```diff
@@ -42,6 +42,20 @@ def convert_to_tensor(value: Union[Tensor, T, RawTensorTypes]) -> Tensor[T]:
...
+        :param t1:
+        :param t2:
+        """
+        if isinstance(t1, _t.Tensor) and isinstance(t2, _t.Tensor):
```
albertz, line 53: The way you call `_tensor_op_sanity_check`, you always expect that you have `isinstance(t1, _t.Tensor) and isinstance(t2, _t.Tensor)`, right? Then don't make this an `if` check, but rather an `assert`.

returnn/tensor/_tensor_extra.py
```diff
@@ -609,6 +610,116 @@ def __repr__(self):
...
+            return self
+        return self.raw_frontend.combine(self, "*", other)
+
+    def __rmul__(self: Tensor, other: Union[_frontend_api.RawTensorTypes, Tensor]) -> Tensor:
+        if isinstance(other, (int, float, numpy_number)) and other == 1:
+            return self
+        return self.raw_frontend.combine(other, "*", self)
+
+    def __truediv__(self: Tensor, other: Union[_frontend_api.RawTensorTypes, Tensor]) -> Tensor:
+        if isinstance(other, (int, float, numpy_number)):
```
albertz, line 667: This check is duplicated here, right? You already do the same check in `combine`. Then just leave it in `combine`, and remove it here?

returnn/tensor/_tensor_extra.py
```diff
@@ -609,6 +610,116 @@ def __repr__(self):
...
+                return self
+            else:
+                assert not self.dtype.startswith(
+                    "int"
+                ), "Dividing a Tensor of type int by an integer is disallowed. Please convert the Tensor to float."
+
+        return self.raw_frontend.combine(self, "/", other)
+
+    def __rtruediv__(self: Tensor, other: Union[_frontend_api.RawTensorTypes, Tensor]) -> Tensor:
+        if isinstance(other, (int, float, numpy_number)):
```
albertz, line 678: Same.

returnn/frontend_api.py
```diff
@@ -85,6 +100,67 @@ def compare(
...
+            Not all the dims of a and b need to be specified here, and there could also be other dims in the dim_order.
+        :return: element-wise combination of a and b
+        """
+        a = cls.convert_to_tensor(a)
+        b = cls.convert_to_tensor(b)
+        cls._tensor_op_sanity_check(a, b)
+        # Truediv checks for now
+        if kind in {"truediv", "/"}:
+            if isinstance(a, (int, float, numpy.number)) and a == 1:
+                return b
```
albertz, line 130: This is wrong. 1/b != b.

returnn/frontend_api.py
```diff
@@ -85,6 +100,67 @@ def compare(
...
+        :param allow_broadcast_all_sources: if True, it is allowed that neither a nor b has all dims of the result.
+            Not needed when out_dims is specified explicitly.
+        :param dim_order: defines the order of the resulting dims. if None, it is automatically inferred from a and b.
+            Not all the dims of a and b need to be specified here, and there could also be other dims in the dim_order.
+        :return: element-wise combination of a and b
+        """
+        a = cls.convert_to_tensor(a)
+        b = cls.convert_to_tensor(b)
+        cls._tensor_op_sanity_check(a, b)
+        # Truediv checks for now
```
albertz, line 127: It should be consistent. You should move all such checks here.

returnn/tensor/_tensor_extra.py
```diff
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 from typing import TYPE_CHECKING, Optional, Union, Type, Sequence, Tuple, List, Dict
+from numpy import number as numpy_number
```
albertz, line 8: This is very uncommon. Just do `import numpy` instead, and then refer to it as `numpy.number`.

albertz:
> All the merge commits should be removed. You rather should rebase.

Can you still do that? And also clean up some unnecessary commits (reorder, squash).

returnn/frontend_api.py
```diff
@@ -85,6 +100,67 @@ def compare(
...
+        if kind in {"truediv", "/"}:
+            if isinstance(a, (int, float, numpy.number)) and a == 1:
```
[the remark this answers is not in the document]
Icemole: In fact when we convert the scalar value (int/float/np.number) to a `Tensor`, we don't have to do all these checks, do we?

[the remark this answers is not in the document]
Icemole: The `truediv` checks are correct, I mean the rest of the checks that return `self`: multiplication * 1, addition + 0, etc.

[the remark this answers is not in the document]
albertz: I don't understand. It's just the same?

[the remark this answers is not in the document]
Icemole: I hadn't seen your latest message. Locally I've moved this check before creating tensors `a` and `b`, which should be correct. What I meant was that I found redundant to return `self` with the identity operation.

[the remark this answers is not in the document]
albertz: I don't understand. What exactly is redundant?

Icemole:
I've squashed my commits into 5 relevant commits. In these I've also applied the feedback you suggested.

PR #1279, opened 2023-03-20
Set forward batch size explicitly for task Search

mmz33:
Using default value 0 is not a good idea. Also i had a job where i did not set the `forward_batch_size` and the job got stuck with default value 0.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1092,7 +1092,7 @@ class Dataset(object):
         :param set(str)|None used_data_keys:
         """
         if not batch_size:
-            raise Exception("batch_size must be set or greater than 0")
+            raise Exception("batch_size must be set and be greater than 0")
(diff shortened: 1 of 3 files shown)


albertz (approved):

mmz33:
> Is probably not a good idea? Maybe we should even make the exception there?

Yes I agree.

returnn/datasets/basic.py
```diff
@@ -1092,7 +1092,7 @@ def _generate_batches(
         :param set(str)|None used_data_keys:
         """
         if not batch_size:
-            batch_size = sys.maxsize
+            raise Exception("batch_size must be set or greater than 0")
```
albertz, line 1095:
```suggestion
            raise Exception("batch_size must be set and be greater than 0")
```

albertz:
In `test_engine_end_layer`, in `forward_to_hdf`, set some `batch_size`.

PR #1292, opened 2023-03-28
preload_from_files for PT engine

vieting:
As discussed in https://github.com/rwth-i6/returnn/issues/1120, `preload_from_files` or something equivalent should be added for the PT engine as well. This is certainly not complete, but could be helpful as a starting point. It works as a proof-of-concept to load a wav2vec 2.0 checkpoint.

What do you think in general?

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -289,19 +289,19 @@ class Engine(EngineBase):
         self._model = get_model_func(**sentinel_kw)
         assert isinstance(self._model, torch.nn.Module)
 
-        preload_from_files = self.config.typed_value("preload_from_files", {})
         if epoch > 1:
             filename = self.get_epoch_model_filename(epoch=epoch - 1) + util.get_model_filename_postfix()
             print("Load model %s" % (filename,), file=log.v4)
             model_state = torch.load(filename)
             self._model.load_state_dict(model_state)
-        elif preload_from_files:
+        preload_from_files = self.config.typed_value("preload_from_files", {})
+        if preload_from_files:
             # see `preload_from_files` in tf engine and `returnn.tf.network.CustomCheckpointLoader`
             is_training = self.config.value("task", "train") == "train"
             is_first_train_epoch = epoch == 1 and (
                 is_training or self.config.value("task", "train") == "initialize_model"
             )
-            for key, opts in preload_from_files.items():
+            for preload_key, opts in reversed(sorted(preload_from_files.items())):
                 assert isinstance(opts, dict) and "filename" in opts
                 if opts.get("init_for_train", False):
                     if not is_first_train_epoch:
@@ -309,7 +309,7 @@ class Engine(EngineBase):
                 else:  # default: init for recog
                     if is_training:
                         continue
-                print(f"Pre-load weights for key '{key}' from {opts['filename']}", file=log.v3)
+                print(f"Pre-load weights for key '{preload_key}' from {opts['filename']}", file=log.v3)
                 preload_model_state = torch.load(opts["filename"])
                 if opts.get("checkpoint_key", None) is not None:
                     preload_model_state = preload_model_state[opts["checkpoint_key"]]
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -289,11 +289,26 @@ class Engine(EngineBase):
         self._model = get_model_func(**sentinel_kw)
         assert isinstance(self._model, torch.nn.Module)
 
+        preload_from_files = self.config.typed_value("preload_from_files", {})
         if epoch > 1:
             filename = self.get_epoch_model_filename(epoch=epoch - 1) + util.get_model_filename_postfix()
             print("Load model %s" % (filename,), file=log.v4)
             model_state = torch.load(filename)
             self._model.load_state_dict(model_state)
+        elif preload_from_files:
+            for key, opts in preload_from_files.items():
+                assert isinstance(opts, dict) and "filename" in opts
+                print(f"Pre-load weights for key '{key}' from {opts['filename']}")
+                preload_model_state = torch.load(opts["filename"])
+                if opts.get("checkpoint_key", None) is not None:
+                  preload_model_state = preload_model_state[opts["checkpoint_key"]]
+                if opts.get("prefix", ""):
+                    preload_model_state = {
+                        opts["prefix"] + key: value for key, value in preload_model_state.items()}
+                strict = not opts.get("ignore_missing", False)
+                missing_keys, unexpected_keys = self._model.load_state_dict(preload_model_state, strict=strict)
+                print(f"Missing keys: {missing_keys}")
+                print(f"Unexpected keys: {unexpected_keys}")
 
         self._model.to(self._device)
 
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -325,9 +325,11 @@ class Engine(EngineBase):
                     # Currently, this only supports single level dicts, but it could be extended if needed.
                     preload_model_state = preload_model_state[opts["checkpoint_key"]]
                 if opts.get("prefix", ""):
-                    # Only params with this prefix should be loaded. They are expected to be in the checkpoint without
-                    # this prefix. By adding the prefix to all params, we make sure that only params matching this
-                    # condition are loaded. This is in line with the behavior of the tf engine.
+                    # Only params with this prefix should be loaded.
+                    # They are expected to be in the checkpoint without this prefix.
+                    # By adding the prefix to all params,
+                    # we make sure that only params matching this condition are loaded.
+                    # This is in line with the behavior of the TF engine.
                     preload_model_state = {opts["prefix"] + key: value for key, value in preload_model_state.items()}
                 ignore_params = opts.get("ignore_params", [])
                 ignore_params_prefixes = opts.get("ignore_params_prefixes", [])
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -301,6 +301,10 @@ class Engine(EngineBase):
             is_first_train_epoch = epoch == 1 and (
                 is_training or self.config.value("task", "train") == "initialize_model"
             )
+            # we use the reversed sorted order here to achieve consistent behavior with the tf engine. There, the keys
+            # are used in sorted order but if a variable is loaded, it will not be considered anymore afterwards. So
+            # the first occurrence is used. Here, we overwrite variables even if they have been loaded before. In order
+            # to get consistent behavior, we use the reversed order.
             for preload_key, opts in reversed(sorted(preload_from_files.items())):
                 assert isinstance(opts, dict) and "filename" in opts
                 if opts.get("init_for_train", False):
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -301,10 +301,12 @@ class Engine(EngineBase):
             is_first_train_epoch = epoch == 1 and (
                 is_training or self.config.value("task", "train") == "initialize_model"
             )
-            # we use the reversed sorted order here to achieve consistent behavior with the tf engine. There, the keys
-            # are used in sorted order but if a variable is loaded, it will not be considered anymore afterwards. So
-            # the first occurrence is used. Here, we overwrite variables even if they have been loaded before. In order
-            # to get consistent behavior, we use the reversed order.
+            # We use the reversed sorted order here to achieve consistent behavior with the TF engine.
+            # There, the keys are used in sorted order but if a variable is loaded,
+            # it will not be considered anymore afterwards.
+            # So the first occurrence is used.
+            # Here, we overwrite variables even if they have been loaded before.
+            # In order to get consistent behavior, we use the reversed order.
             for preload_key, opts in reversed(sorted(preload_from_files.items())):
                 assert isinstance(opts, dict) and "filename" in opts
                 if opts.get("init_for_train", False):
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -301,10 +301,9 @@ class Engine(EngineBase):
                 print(f"Pre-load weights for key '{key}' from {opts['filename']}")
                 preload_model_state = torch.load(opts["filename"])
                 if opts.get("checkpoint_key", None) is not None:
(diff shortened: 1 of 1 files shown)


albertz:
The behavior should be consistent to TF. Most importantly, `init_for_train`. Currently you seem to ignore this? You actually seem to assume a different default (`init_for_train=True`) than we have in TF (`init_for_train=False`), which we should definitely not do.

albertz:
> Regarding `init_for_train` specifically, how would I know in `_load_model()` whether this is training?

The logic in TF is this:
```python
        is_training = config.value("task", "train") == "train"
        is_first_train_epoch = not epoch and (is_training or config.value("task", "train") == "initialize_model")

...

            for model_name, opts in sorted(self.preload_from_files.items()):
                assert isinstance(opts, dict)
                if opts.get("init_for_train", False):
                    if not is_first_train_epoch:
                        continue
                else:  # default: init for recog
                    if is_training:
                        continue
```
`epoch` is None in this code if there was no existing model found. But I think you can just test `epoch == 1` instead.

albertz:
Your `prefix` logic is also not consistent. It would only consider the params in the model with this prefix and ignore the others. `ignore_missing` refers to only those params with the prefix, and all other params are always ignored.

returnn/torch/engine.py
```diff
@@ -254,11 +254,49 @@ def _load_model(self, epoch):
...
+                        continue
+                print(f"Pre-load weights for key '{key}' from {opts['filename']}", file=log.v3)
+                preload_model_state = torch.load(opts["filename"])
+                if opts.get("checkpoint_key", None) is not None:
+                    preload_model_state = preload_model_state[opts["checkpoint_key"]]
+                if opts.get("prefix", ""):
+                    preload_model_state = {opts["prefix"] + key: value for key, value in preload_model_state.items()}
+                ignore_params = opts.get("ignore_params", [])
+                ignore_params_prefixes = opts.get("ignore_params_prefixes", [])
+                for key in list(preload_model_state.keys()):
```
albertz, line 285: You are overwriting the `key` from the parent loop. Use a different variable name.

returnn/torch/engine.py
```diff
@@ -254,11 +254,49 @@ def _load_model(self, epoch):
...
+                    if key in ignore_params or any(
+                        [key.startswith(ignore_key) for ignore_key in ignore_params_prefixes]
+                    ):
+                        print(f"Ignoring variable {key}", file=log.v3)
+                        preload_model_state.pop(key)
+                for new_name, name_in_checkpoint in opts.get("var_name_mapping", {}).items():
+                    preload_model_state[new_name] = preload_model_state.pop(name_in_checkpoint)
+                missing_keys, unexpected_keys = self._model.load_state_dict(preload_model_state, strict=False)
+                if not opts.get("ignore_missing", False):
+                    prefix_keys = [key for key in self._model.state_dict() if key.startswith(opts.get("prefix", ""))]
```
albertz, line 295: You are overwriting the `key` from the parent loop. Use a different variable name.

returnn/torch/engine.py
```diff
@@ -254,11 +254,49 @@ def _load_model(self, epoch):
         self._model = get_model_func(**sentinel_kw)
         assert isinstance(self._model, torch.nn.Module)
 
+        preload_from_files = self.config.typed_value("preload_from_files", {})
         if epoch > 1:
             filename = self.get_epoch_model_filename(epoch=epoch - 1) + util.get_model_filename_postfix()
             print("Load model %s" % (filename,), file=log.v4)
             model_state = torch.load(filename)
             self._model.load_state_dict(model_state)
+        elif preload_from_files:
```
albertz, line 263: This is wrong here. This would only be executed with `epoch == 1`. But this is wrong. It must always be executed.

albertz:
We also should make the order, or actually the preference for var loading consistent to TF. In TF, we first go through `preload_from_files`, and those entries sorted by key name, and then afterwards to the normal loading. However, every variable we will load, we will mark (see `set_as_custom_init`), and this makes sure that any subsequent loading will not load it anymore. So it means the values from where it is loaded first, they stay.

Now, you don't have any such logic in PT. This effectively means, some will get loaded multiple times, and also, the values from where it is loaded *last*, they stay. So the opposite order. I think then you need to do the normal loading first, and then you need to iterate over `reversed(sorted(preload_from_files.items()))`.

I'm not sure if I'm maybe missing sth else here.

albertz (approved):
Looks ok to me now, despite my last comment. I hope I did not miss anything. As said, I think consistency to the TF logic is important.

@patrick-wilken or @JackTemaki or someone should also review.

returnn/torch/engine.py
```diff
@@ -266,6 +266,10 @@ def _load_model(self, epoch):
             is_first_train_epoch = epoch == 1 and (
                 is_training or self.config.value("task", "train") == "initialize_model"
             )
+            # we use the reversed sorted order here to achieve consistent behavior with the tf engine. There, the keys
+            # are used in sorted order but if a variable is loaded, it will not be considered anymore afterwards. So
+            # the first occurrence is used. Here, we overwrite variables even if they have been loaded before. In order
+            # to get consistent behavior, we use the reversed order.
```
albertz, line 272: Use newlines properly.
```suggestion
            # We use the reversed sorted order here to achieve consistent behavior with the TF engine.
            # There, the keys are used in sorted order but if a variable is loaded,
            # it will not be considered anymore afterwards.
            # So the first occurrence is used.
            # Here, we overwrite variables even if they have been loaded before.
            # In order to get consistent behavior, we use the reversed order.
```

returnn/torch/engine.py
```diff
@@ -259,6 +259,50 @@ def _load_model(self, epoch):
...
+                    if not is_first_train_epoch:
+                        continue
+                else:  # default: init for recog
+                    if is_training:
+                        continue
+                print(f"Pre-load weights for key '{preload_key}' from {opts['filename']}", file=log.v3)
+                preload_model_state = torch.load(opts["filename"])
+                if opts.get("checkpoint_key", None) is not None:
+                    preload_model_state = preload_model_state[opts["checkpoint_key"]]
+                if opts.get("prefix", ""):
```
patrick-wilken, line 287: In the TF backend the `"prefix"` option gets passed to `load_if_prefix` of `CustomCheckpointLoader`, so the inverse option to `"ignore_prefix"`. What you implement here is something different, right? You just add the prefix to all variables. I don't think this is useful? Anyways should not have a different meaning than for the TF backend.

returnn/torch/engine.py
```diff
@@ -259,6 +259,50 @@ def _load_model(self, epoch):
...
+                assert isinstance(opts, dict) and "filename" in opts
+                if opts.get("init_for_train", False):
+                    if not is_first_train_epoch:
+                        continue
+                else:  # default: init for recog
+                    if is_training:
+                        continue
+                print(f"Pre-load weights for key '{preload_key}' from {opts['filename']}", file=log.v3)
+                preload_model_state = torch.load(opts["filename"])
+                if opts.get("checkpoint_key", None) is not None:
```
patrick-wilken, line 285: Why this new option? Do some external models come with checkpoints that do not only store the state dict but also other things? I'm not sure we have to try to handle that. Because it could be in a tuple, or in a nested dict. I would rather require the user to convert the checkpoint if necessary such that it looks like it has been saved via `torch.save(model.state_dict())`. Or is there some other convention I'm missing? This is not for `torch.save(model)`, right?

vieting: Good point. I tried importing a wav2vec 2.0 checkpoint from [here](https://github.com/facebookresearch/fairseq/blob/main/examples/wav2vec/README.md). When I `torch.load()` it, it is a dict with `dict_keys(['args', 'model', 'optimizer_history', 'extra_state', 'last_optimizer_state'])`. Since my goal was to load this model and the option requires minimal implementation, I added it. I'm not sure if there are conventions, though.

returnn/torch/engine.py
albertz: I also had to look how it is done `CustomCheckpointLoader` because I always keep forgetting what `prefix` means. But as I understand the TF code, it expects that you have some params in the model named like `<prefix>paramN` or so, and then in the checkpoint file it expects it to be as `paramN` (without the prefix).

And if I understand the PT code correctly here, it should have the same effect, right?

albertz: The model was usually prepared in a way that it has a subnetwork in `output/lm` or so for an external LM, or maybe it does model combination where there was a complete separate model, all with prefix `model2_`, and then you would load an existing checkpoint, but in the checkpoint it would be without this predix (`output/lm/` or `model2_`).

returnn/torch/engine.py
albertz: I think also here it would help maybe to add this example with the w2v checkpoint as a comment.

returnn/torch/engine.py
patrick-wilken: Oof, I see. I would find it easier to understand if you get the list of model parameters first (`model.named_parameters()`?) and then select the parameters from `preload_model_state` that appear in the model parameters with the prefix.
And then also `ignore_params` and `ignore_params_prefixes` could be checked against the model parameters directly rather than against the altered checkpoint parameter names.

As it is now aren't valid parameters that are both in the checkpoint and the model, but just don't have the "prefix" printed as `unexpected_keys`? Even with the prefix which they never had.

albertz: You mean you suggest to reimplement this, in a way you outlined?

Currently this code uses `load_state_dict`, but what you are suggesting cannot really use `load_state_dict` then, but have to replicate that logic. As I remember when looking at `load_state_dict`, it does a bit more than that, though. See code [here](https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.load_state_dict) and then also `_load_from_state_dict`. It is also about persistent buffers, `set_extra_state`, pre-hooks, and maybe more.

patrick-wilken:
By the way, what defines the parameter names in the frontend? That's what `rf.Module.named_parameters()` / rf.Module.named_parameters()` seem to do, right? And it uses the attribute names of the modules. So to add the prefix we are talking about here you would rename the Module attribute? 😕

returnn/torch/engine.py
patrick-wilken: > it would be the same, or not?

Yes.
Except for the `unexpected_keys` message. Or do we expect that if "prefix" is used all parameters from the checkpoint will be present in the "prefix subnetwork" in all cases?

patrick-wilken: Sorry, I didn't get the thing about `load_state_dict`. Of course the dict you pass to it would use the model parameter names, so with prefix. 🤔

returnn/torch/engine.py
patrick-wilken: > Shouldn't we maybe also follow more this standard convention?

It's farfetched to call anything a convention here. But true, many frameworks seem to add additional information into the checkpoint files and it's also in the docs https://pytorch.org/tutorials/beginner/saving_loading_models.html#save

```
torch.save({
            'epoch': epoch,
            'model_state_dict': model.state_dict(),
            'optimizer_state_dict': optimizer.state_dict(),
            'loss': loss,
            ...
            }, PATH)
```
(Note the dots...)

@Icemole implemented saving the optimizer state to separate files. I think it can be convenient this way because you directly get the checkpoint you need for inference. But of course, if we needed to save more things then there is a limit to this approach.

> We can also extend this option, to accept a list of keys, so it can go deeply through dict keys, list items, etc.

Yeah, and allow it to be a callable so really everything is possible... :ok_woman:

I would rather have a defined interface and just support loading what we save. And yes, maybe this `checkpoint_key` for 1-level dicts because it's common.

albertz:
> So to add the prefix we are talking about here you would rename the Module attribute?

No, you would simply put it into a submodule. Where you probably have the model anyway. E.g. if you have trained a LM, using the module `TransformerLm`, now you would create `TransformerLm` as a submodule into your main model, like:
```python
class Model(nn.Module):
  def __init__(self):
    super().__init__()
    ...
    self.ext_lm = TransformerLm(...)
```
In that example, the prefix is simply `ext_lm.`.

returnn/torch/engine.py
vieting: You're right about the unexpected keys. I added the prints because it seemed handy to me, but we can think about whether this should be modified. Maybe it could be combined for all contents of `preload_from_files` because if a key is missing in the first model but is loaded in the second, it could be misleading if the log shows it's missing in the first place. I often find the log output of the tf `preload_from_files` confusing, so not sure how much it can serve as a guidance.

returnn/torch/engine.py
vieting: I added a comment. Is this thread resolved then?

returnn/torch/engine.py
```diff
@@ -294,6 +294,57 @@ def _load_model(self, *, epoch: int, step: int):
...
+                    # This can be used if an external checkpoint saves a checkpoint a different structure that just the
+                    # model state dict. E.g., if a checkpoint is created using
+                    # `torch.save({"model": model.state_dict(), "optimizer": optimizer.state)_dict(), ...})`
+                    # we can set checkpoint_key = "model" to load the model.
+                    # Currently, this only supports single level dicts, but it could be extended if needed.
+                    preload_model_state = preload_model_state[opts["checkpoint_key"]]
+                if opts.get("prefix", ""):
+                    # Only params with this prefix should be loaded. They are expected to be in the checkpoint without
+                    # this prefix. By adding the prefix to all params, we make sure that only params matching this
+                    # condition are loaded. This is in line with the behavior of the tf engine.
```
albertz, line 330: Please do proper newlines in comments.
```suggestion
                    # Only params with this prefix should be loaded.
                    # They are expected to be in the checkpoint without this prefix.
                    # By adding the prefix to all params,
                    # we make sure that only params matching this condition are loaded.
                    # This is in line with the behavior of the TF engine.
```

returnn/torch/engine.py
patrick-wilken: You mean a TODO comment that we want to improve the output?
Anyways, as I see now the behaviour, also in terms of logging, should be like the TF backend. So good to merge.

albertz: No, I meant a comment describing what `prefix` actually does, and why we implement it this way, why we use `load_state_dict`.

PR #1293, opened 2023-03-28
unify learning rates in EngineBase

JackTemaki:
learning rate handling is the same for TF and PT.

### DIFF
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -23,7 +23,7 @@ from tensorflow.python.client import timeline
 from returnn.engine.base import EngineBase
 from returnn.datasets.basic import Dataset, Batch, BatchSetGenerator, init_dataset
 from returnn.datasets.util.vocabulary import Vocabulary
-from returnn.learning_rate_control import load_learning_rate_control_from_config, LearningRateControl
+from returnn.learning_rate_control import load_learning_rate_control_from_config
 from returnn.log import log
 from returnn.pretrain import pretrain_from_config
 import returnn.tf.compat as tf_compat
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -877,7 +877,6 @@ class Engine(EngineBase):
         self.tf_session = None  # type: typing.Optional[tf.compat.v1.Session]
         self.network = None  # type: typing.Optional[TFNetwork]
         self.updater = None  # type: typing.Optional[Updater]
-        self.learning_rate_control = None  # type: typing.Optional[LearningRateControl]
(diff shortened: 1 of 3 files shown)


returnn/tf/engine.py
```diff
@@ -877,20 +877,17 @@ def __init__(self, config=None):
         self.tf_session = None  # type: typing.Optional[tf.compat.v1.Session]
         self.network = None  # type: typing.Optional[TFNetwork]
         self.updater = None  # type: typing.Optional[Updater]
-        self.learning_rate_control = None  # type: typing.Optional[LearningRateControl]
         self._checked_uninitialized_vars = False
         self._merge_all_summaries = None
         self.dataset_batches = {}  # type: typing.Dict[str,BatchSetGenerator]
         self.dataset_provider = None  # type: typing.Optional[DatasetDataProvider]
         self.train_data = None  # type: typing.Optional[Dataset]
-        self.eval_datasets = {}  # type: typing.Dict[str,Dataset]
```
albertz, line 886: Why did you remove this?

returnn/tf/engine.py
```diff
@@ -1080,6 +1078,7 @@ def init_train_from_config(self, config=None, train_data=None, dev_data=None, ev
         :param returnn.datasets.basic.Dataset|None dev_data:
         :param returnn.datasets.basic.Dataset|None eval_data:
         """
+        super().init_train_from_config(config=config)
```
albertz, line 1081: It should come after `if not config: config = self.config`.

returnn/tf/engine.py
```diff
@@ -1091,6 +1089,7 @@ def init_train_from_config(self, config=None, train_data=None, dev_data=None, ev
             from returnn.datasets.basic import set_config_extern_data_from_dataset
 
             set_config_extern_data_from_dataset(config=config, dataset=train_data or dev_data or eval_data)
+        super().init_train_from_config(config=config)
```
albertz, line 1092: This should be as early as possible, i.e. right after `if not config: config = self.config`.

returnn/tf/engine.py
JackTemaki: I went a little ahead and also checked for unification of datasets, and this was a leftover I missed when doing the PR.

JackTemaki:
@albertz please do not send each comment one by one, but group them all and send them once, this makes it easier to fix it all at once and saves time.

albertz:
Btw, this would have been better squashed instead of rebased and merged, as those commits all belong together. But this needs to be decided individually per PR.

Issue #1146, opened 2022-10-14
learning_rate_control_error_measure should match exactly


albertz:
Currently when `learning_rate_control_error_measure` (the error key) does not match exactly, it will use some heuristics. See `LearningRateControl.get_error_key`.

This is also because the error key can change depending on whether there is a single loss or multiple, see `LearningRateControl.set_epoch_error`.

This behavior can potentially hide user mistakes like providing an invalid error key.

This was already mentioned in https://github.com/rwth-i6/returnn_common/issues/211 where an option like `learning_rate_control_error_measure_must_match` was suggested to enforce an exact match. However, after some discussion with @mmz33, we think that having a new behavior version might be a another, maybe a better option instead.

Such new behavior (either via new behavior version, or via such option) would need to change both `get_error_key` and `set_epoch_error`.

albertz:
No feedback here?

We can also first introduce such an option and later change the default of this option via a new behavior version. Or maybe also right away, and then people can either disable it to get the old behavior, or adapt it.

albertz:
Despite this, there is also the inconsistency of a single vs multiple losses. In the single case, it would just store `dev_score` etc, and only with multiple losses it appends the loss name. I think this option should also enable to always store the full name.

patrick-wilken:
I see that storing the full name in all cases would require a new option / behaviour version. But with the current loss names, can't we already add a check that if `learning_rate_control_error_measure` is set it should match one of the losses? (Maybe optionally including an "_output" postfix, that's currently in the code.) That is a bug fix rather than a new behaviour, right? Nobody sets this option and expects it to be ignored.

PR #1295, opened 2023-03-29
add training losses to learning rates file

JackTemaki:
Also @Icemole @patrick-wilken, I would prefer to write `train_loss_<name>` and `dev_loss_<name>` instead of `score`, what is your opinion on that?

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -128,19 +128,28 @@ class Engine(EngineBase):
 
         self._model.train()
 
+        accumulated_losses_dict = NumbersDict()
         step_idx = 0
         for data in self._train_dataloader:
-            loss, _ = self._run_step(data)
+            total_loss, losses_dict = self._run_step(data)
 
             self._updater.get_optimizer().zero_grad()
-            loss.backward()
+            total_loss.backward()
             self._updater.get_optimizer().step()
 
-            print("step %i, loss: %f" % (step_idx, loss.detach().cpu().numpy()), file=log.v4)
+            losses_dict = {
+                "train_loss_" + name: float(loss.detach().cpu().numpy()) for name, loss in losses_dict.items()
+            }
+            accumulated_losses_dict += NumbersDict(losses_dict)
+            print("step %i, loss: %f" % (step_idx, total_loss.detach().cpu().numpy()), file=log.v4)
 
             step_idx += 1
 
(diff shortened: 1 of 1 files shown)


albertz:
Note the failing tests. Not sure if you want to fix that in here or separately.

albertz:
Note that `LearningRateControl.get_error_key` (and maybe other places) assume "score", not "loss", and thus they break. But break silently, i.e. the heuristics will not work and will return some other things instead, in certain cases. This might be very unexpected, esp. when people switch from TF to PT. So I would extend at least `get_error_key` to also behave the same for "loss". And check if there are other places where we assume "score".

JackTemaki:
> > Note the failing tests. Not sure if you want to fix that in here or separately.
> 
> I do not understand at all why it fails now and not before. Why should it have been able to write models to the same path but not the learning_rates file now?

My first solution would be now to just create the dir within the config, but this does not feel right...

albertz:
Can you maybe rebase?

JackTemaki:
> > but rather enforce that this is set explicitly for the torch backend if necessary.
> 
> Note that this currently is not really possible. Even when setting `learning_rate_control_error_measure` in the config, this will still be used as part of the heuristics, and might not need to match exactly, and if not found, it will fall back to the other heuristics. This is actually #1146.

Yes but it can be made possible. But lets continue with this on #1146, I forgot this existed.

albertz (approved):
For me, this is fine.

I would like if we do this for TF as well, but this can be a separate commit/PR.

I think @patrick-wilken or @Icemole should maybe also comment.

Icemole:
I haven't been following the discussion that much as I haven't had the time to look into anything, but I also like these changes and I actually prefer the term `loss` rather than the more ambiguous term `score`.

I assume the assumptions about `score...` that you discussed have already been taken care of?

Icemole (approved):

albertz:
> I assume the assumptions about `score...` that you discussed have already been taken care of?

Not sure. I did not check this. But we should check for other usages of "score" somewhere. I just pointed to `get_error_key`.

patrick-wilken:
"loss" instead of "score" because that's what we minimize, right? Agreed.

I don't find usages of `train_score` / `dev_score` strings outside of the LR control code.

patrick-wilken (approved):

PR #1296, opened 2023-03-29
use os.makedirs for learning rates file

JackTemaki:
Returnn should try to create the folder for the learning-rates file location if not existing yet.


albertz:
I pushed sth. Can you test?

michelwi:
> I pushed sth. Can you test?

Yes, works now. (tested only with TF backend). Thanks!

PR #1297, opened 2023-03-30
share Config between engines

JackTemaki:
Preparation for shared model cleanup code.

### DIFF
--- a/returnn/engine/base.py
+++ b/returnn/engine/base.py
@@ -8,7 +8,7 @@ import os
 import sys
 from typing import Optional
 
-from returnn.config import Config
+from returnn.config import Config, get_global_config
 from returnn.learning_rate_control import load_learning_rate_control_from_config, LearningRateControl
 from returnn.log import log
 from returnn.pretrain import Pretrain
@@ -24,8 +24,10 @@ class EngineBase(object):
         """
         :param config:
         """
-        self.epoch = 0
+        if config is None:
+            config = get_global_config(auto_create=True)
         self.config = config
+        self.epoch = 0
         self.pretrain = None  # type: Optional[Pretrain]
         self.model_filename = None  # type: Optional[str]
         self.learning_rate = 0.0  # set in init_train_epoch
--- a/returnn/tf/engine.py
+++ b/returnn/tf/engine.py
@@ -858,10 +858,6 @@ class Engine(EngineBase):
         """
         :param returnn.config.Config|None config:
         """
-        if config is None:
-            from returnn.config import get_global_config
-
-            config = get_global_config(auto_create=True)
         super(Engine, self).__init__(config=config)
         if not log.initialized:
             log.init_by_config(config)


returnn/engine/base.py
```diff
@@ -20,8 +20,12 @@ class EngineBase(object):
     Base class for a backend engine, such as :class:`TFEngine.Engine`.
     """
 
-    def __init__(self):
+    def __init__(self, config: Optional[Config] = None):
+        """
+        :param config:
+        """
         self.epoch = 0
+        self.config = config
```
albertz, line 28: Why don't we move the logic with `get_global_config` from TF engine also here?

JackTemaki: The code is Tensorflow specific or not? Also this can be done later if needed.

albertz: No, it's not TF specific. Why do you think so?

I don't like that the TF engine would behave different here. I don't see why. Just make the same behavior for all engines regarding the config, and use of `get_global_config()`. That's much cleaner.

JackTemaki: > Why do you think so?

```python
    from returnn.tf.compat import v1 as tf_v1

    graph = tf_v1.get_default_graph()
    # We could use collection refs, but this could cause other problems,
    # and is more complicated than what we need.
    # We just use a custom own attrib.
    attrib_name = "_RETURNN_config_in_graph"
```

But I will not start a discussion, I will just put this behind a backend check and unify the rest.

albertz: It is already safe. You don't need to touch that code.

JackTemaki: > No, it's not TF specific. Why do you think so?

btw. it would help me if you would answer "It might have TF specific parts, but they will not be active". This would be much less confusing for me instead of answering with a counter question and the whole explanation for the reasoning.

albertz: I did not even remember that there are TF specific parts. By default, whatever is not in the `returnn.tf` namespace, you can think of being TF independent (or at least it will definitely work without TF). I don't really know how you got the idea that anything in `returnn.config`, esp a function like `get_global_config`, could be TF specific. I still wonder a lot how you get the idea that a function like `get_global_config` could be TF specific. That's why I asked. It's totally unclear to me.

JackTemaki: Okay let me explain:

 - it was before reading it carefully unclear to me why this function is exactly needed, and the first thing I saw is a `_get_or_set_config_via_tf_default_graph`
 - Just because it is generic, it does not mean that it is actually useful for the PyTorch Engine and that we want to keep this in the future (except then for the TFEngine for not destroying anything existing)
 - config.py also contains `tf_should_use_gpu`, which is definitely TF specific code.

So, of course I would not automatically assume that the code is definitely not TF specific and/or relevant for the PyTorch Engine .
 
While we are at it: We have a "global" config both as `__main__.config` and as `config._global_config` , this does not seem to be a good idea. But this has nothing to do with the PR.

albertz: So any suggestion to improve the code such that such confusion would be avoided? Maybe rename `_get_or_set_config_via_tf_default_graph` to `_opt_get_or_set_config_via_tf_default_graph`?

albertz: The reason this code is there is really independent from the TF engine. It was added to make embedding of RETURNN code into some external script easier. Such that the user can simply write:
```python
engine = TFEngine()
engine.network = TFNetwork(...)
engine.train()
```
Without needing to create some config.

I don't see why the same reasoning would not apply for the PT engine as well. Actually, maybe it does even more so. We often said that we want to have anything in our PT code modular and usable also outside of RETURNN, so that a user could also write an independent training script, using some parts (any parts, in any combination) of the RETURNN PT code (that is what is meant by "modular code"). The user could use the `PTEngine` together with some own custom models and not using the RETURNN config otherwise.

At least this is the idea. And this is just one little snippet to make this easier. There might be other issues with the current implementation not really allowing this, or making it maybe harder than necessary, but if you encounter them, then please open separate issues on them.

albertz (approved):

PR #1301, opened 2023-04-03
RF: cross_entropy, implemented for torch

patrick-wilken:
@albertz asked for this, so already sharing.

The most important case `estimated_type="logits"` seems to work, see tests. But it's not yet complete, other `estimated_type`s don't work yet (sparse would need implementation of `gather`). Also, the case of additional dims other than batch and class label dim is not tested. Will continue tomorrow...

### DIFF
--- a/returnn/frontend/_backend.py
+++ b/returnn/frontend/_backend.py
@@ -276,6 +276,18 @@ class Backend(Generic[T]):
         raise NotImplementedError
 
     @staticmethod
+    def softmax_cross_entropy_with_logits(logits: Tensor, targets: Tensor, axis: Dim):
+        """
+        Efficient cross entropy.
+
+        :param logits: target estimates given as inputs to softmax (i.e. unnormalized)
+        :param targets: probabilities, i.e. normalized, can also be sparse
+        :param axis: class labels dim over which softmax is computed
+        :return: cross entropy (same Dims as 'logits' but without 'axis')
+        """
+        raise NotImplementedError
+
+    @staticmethod
     def sequence_mask_raw(lengths: T, *, batch_major: bool = True) -> T:
         """
         Like tf.sequence_mask().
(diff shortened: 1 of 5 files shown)


returnn/frontend/loss.py
```diff
@@ -0,0 +1,49 @@
+"""
+Loss functions
+"""
+
```
albertz, line 4:
```suggestion

from __future__ import annotations
```

tests/test_torch_frontend.py
albertz, line 1: Please stop using this file in the future, and instead make backend independent test. All the tests from here should be rewritten and moved. To `test_rf_base.py`, or maybe we should make separate files for different groups of functionality.

returnn/torch/frontend/_backend.py
```diff
@@ -153,6 +153,59 @@ def log_softmax(tensor: Tensor, *, axis: Dim) -> Tensor:
         out.raw_tensor = torch.log_softmax(tensor.raw_tensor, dim=tensor.dims.index(axis))
         return out
 
+    @staticmethod
+    def softmax_cross_entropy_with_logits(logits: Tensor, targets: Tensor, axis: Dim):
```
albertz, line 157:
```suggestion
    def softmax_cross_entropy_with_logits(*, logits: Tensor, targets: Tensor, axis: Dim):
```

returnn/frontend/_backend.py
```diff
@@ -275,6 +275,18 @@ def log_softmax(tensor: Tensor, *, axis: Dim) -> Tensor:
         """
         raise NotImplementedError
 
+    @staticmethod
+    def softmax_cross_entropy_with_logits(logits: Tensor, targets: Tensor, axis: Dim):
```
albertz, line 279:
```suggestion
    def softmax_cross_entropy_with_logits(*, logits: Tensor, targets: Tensor, axis: Dim):
```

Issue #1303, opened 2023-04-04
Batch dim not defined for PyTorch backend


JackTemaki:
I tried to solve some error related to an unset Batch Dimension, but it is unclear what should actually be set. Currently neither `.dimension` or `.dyn_size_ext` is set.

So for a single update step, should this be static (an integer) or marked as dynamic (and then set e.g. from `TorchSize`)? It is unclear to me if this was intentionally left out (I do not see a reason why though), or if this is an error and should be set to one of those two variants. I would say `dyn_size_ext` should be a torch tensor on the target device, but I wanted to confirm beforehand.

JackTemaki:
I will first go with setting a static `.dimension` and we can check better options later.

albertz:
I'm not really sure I understand the problem. Can you give some example, some code which leads to an error?

Is this about the RF? If not, if this is about pure PT code, why do you even care?

The batch dim was unfortunately often treated in a special way, with different behavior to other dims. We also have this `batch` attrib, and the `BatchInfo` object, to give more meta information about the batch-dim (and potential extended variants, including beam or whatever). That's why the batch dim originally did not had a `dyn_size_ext` but accessing its value (e.g. `Dim.get_dim_value()`) had a special case for the batch dim.

In case of the RF, I don't want to use this really. We probably would set `dyn_size_ext` in all cases to have some consistent logic. I already do that in some cases now.

JackTemaki:
> I'm not really sure I understand the problem. Can you give some example, some code which leads to an error?
> 
I think the actual problem is not at all relevant/helpful here. In `_run_step`  in `extern_data` for any `Tensor` and `batch_dim` both `.dimension` and `dyn_size_ext` are not set, this is what is relevant.

> Is this about the RF? If not, if this is about pure PT code, why do you even care?
> 

Lets for simplicity assume this is for RF. But as RF is now a core part of the engine, also for pure PT code this now became unfortunately relevant.

> The batch dim was unfortunately often treated in a special way, with different behavior to other dims. We also have this `batch` attrib, and the `BatchInfo` object, to give more meta information about the batch-dim (and potential extended variants, including beam or whatever). That's why the batch dim originally did not had a `dyn_size_ext` but accessing its value (e.g. `Dim.get_dim_value()`) had a special case for the batch dim.
> 

Seems to me this special treatement is not really relevant for the PyTorch Backend or RF in general then.

[...]

PR #1305, opened 2023-04-04
Add `fill_tensor_from_raw_tensor`

JackTemaki:
Introduces a new `fill_tensor_from_raw_tensor` function.

Also solves #1303


albertz:
I think this is not really needed anymore.

PR #1312, opened 2023-04-14
Fix Error Messages

vieting:
When I got the exception for mismatching shapes of raw tensor vs. tensor, the printed output was very large and I guess just the shape is needed, right? The other change is just a typo.

### DIFF
--- a/returnn/tensor/_tensor_extra.py
+++ b/returnn/tensor/_tensor_extra.py
@@ -323,7 +323,7 @@ class _TensorMixin(_TensorMixinBase):
                     continue  # we allow anything in the placeholder
                 if raw_shape[i] != self.batch_shape[i]:
                     raise Exception(
-                        f"Mismatching shape: Raw tensor {self._raw_tensor.shape} vs Tensor {self};\n"
+                        f"Mismatching shape: Raw tensor {raw_shape} vs Tensor {self};\n"
                         + backend.format_graph_output(self._raw_tensor, max_depth=3)
                     )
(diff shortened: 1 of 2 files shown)


returnn/tensor/_tensor_extra.py
```diff
@@ -323,7 +323,7 @@ def sanity_check(self, ignore_placeholder=False, assume_complete=True):
                     continue  # we allow anything in the placeholder
                 if raw_shape[i] != self.batch_shape[i]:
                     raise Exception(
-                        f"Mismatching shape: Raw tensor {self._raw_tensor} vs Tensor {self};\n"
+                        f"Mismatching shape: Raw tensor {self._raw_tensor.shape} vs Tensor {self};\n"
```
albertz, line 326: `.shape` is not correct for all potential raw tensor types. But you have the shape via `raw_shape`. So you could do:
```suggestion
                        f"Mismatching shape: Raw tensor {raw_shape} vs Tensor {self};\n"
```

Note that this is however maybe not the best way depending on the framework. For TF in graph mode, printing the tensor repr is actually nicer, because it gives you also the name (useful to see where it comes from) and otherwise the shape but not the content. So now this change would remove this useful information.

I think we should maybe introduce a function `short_tensor_repr_raw` in the backend to cover for different variants.

albertz: But I will just merge this for now anyway. This can be done as a next step. I'm waiting for tests.

albertz (approved):

PR #1315, opened 2023-04-24
fix: Correct assertion on HDF data dtype

NeoLegends:
We convert the dtype to `str` but then in the assertion try comparing for equality with the original, uncoverted type. This assertion triggered for me even though it does not seem right.


JackTemaki:
There is no "original unconverted type", is it?

I just see it is set by: `self.dtype = str(seq_data.dtype)`, so the assert should not fail if everything is alright.

JackTemaki (approved):

PR #1320, opened 2023-05-03
PT weight_decay fix for shared params

### LINKED ISSUE #1319: PyTorch `weight_decay` ambiguity with shared params
The code logic of `_get_optimizer_param_groups` branch with activated weight decay can not run in its current form. Iterating over all named modules while at the same time iterating recursively over parameters will yield the same parameter multiple times with a different module reference. 

In my case this was:
```python
<class 'i6_experiments.users.rossenbach.experiments.librispeech.tts_architecture_improvement_23.pytorch_networks.ctc_aligner_v1.Model'>                                                                            
<class 'torch.nn.modules.sparse.Embedding'> 
```
for
`speaker_embedding.weight`

This means we need a completely new logic if we want to exclude some modules.

Icemole:
Should only add one parameter to the weight decay set if two or more modules share common parameters.

Fix #1319.

### DIFF
--- a/returnn/torch/updater.py
+++ b/returnn/torch/updater.py
@@ -9,6 +9,7 @@ import torch
 import typing
 
 from returnn.log import log
+from returnn.util.basic import RefIdEq
 
 _OptimizerClassesDictInitialized = False
 _OptimizerClassesDict = {}
@@ -242,12 +243,13 @@ class Updater(object):
         no_wd_params = set()
         blacklist_wd_modules = (torch.nn.LayerNorm, torch.nn.Embedding)
         # Tracker of visited parameters to only add each parameter once, in case two modules share common parameters.
-        visited_params = set()
+        # We need the wrapper class RefIdEq because Parameters are compared by value and not by reference.
+        visited_params: typing.Set[RefIdEq[torch.nn.Parameter]] = set()
         for mn, m in self.network.named_modules():
             for pn, p in m.named_parameters():
-                if p in visited_params:
+                if RefIdEq(p) in visited_params:
                     continue
-                visited_params.add(p)
+                visited_params.add(RefIdEq(p))
                 fpn = "%s.%s" % (mn, pn) if mn else pn  # Full param name
                 if pn.endswith("bias"):
                     no_wd_params.add(fpn)
--- a/returnn/torch/updater.py
+++ b/returnn/torch/updater.py
@@ -241,8 +241,13 @@ class Updater(object):
         wd_params = set()
         no_wd_params = set()
         blacklist_wd_modules = (torch.nn.LayerNorm, torch.nn.Embedding)
+        # Tracker of visited parameters to only add each parameter once, in case two modules share common parameters.
+        visited_params = set()
         for mn, m in self.network.named_modules():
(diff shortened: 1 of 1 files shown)


returnn/torch/updater.py
```diff
@@ -9,6 +9,7 @@
 import typing
```
albertz, line 9:
```suggestion
import typing
from typing import Set
```

albertz: Note on the (somewhat loose) import rules:

Import often used things directly, e.g. things from `typing`.

Do not import rarely used things, or also maybe ambiguous things. Instead, just import the parent module under a common name. Keep consistency. We often use `util` for `returnn.util.basic`.

returnn/torch/updater.py
```diff
@@ -241,8 +242,14 @@ def _get_optimizer_param_groups(self, optim_class, optimizer_opts):
         wd_params = set()
         no_wd_params = set()
         blacklist_wd_modules = (torch.nn.LayerNorm, torch.nn.Embedding)
+        # Tracker of visited parameters to only add each parameter once, in case two modules share common parameters.
+        # We need the wrapper class RefIdEq because Parameters are compared by value and not by reference.
+        visited_params: typing.Set[RefIdEq[torch.nn.Parameter]] = set()
```
albertz, line 247:
```suggestion
        visited_params: Set[util.RefIdEq[torch.nn.Parameter]] = set()
```

returnn/torch/updater.py
```diff
@@ -241,8 +242,14 @@ def _get_optimizer_param_groups(self, optim_class, optimizer_opts):
...
         for mn, m in self.network.named_modules():
             for pn, p in m.named_parameters():
+                if RefIdEq(p) in visited_params:
```
albertz, line 250:
```suggestion
                if util.RefIdEq(p) in visited_params:
```

returnn/torch/updater.py
```diff
@@ -9,6 +9,7 @@
 import typing
 
 from returnn.log import log
+from returnn.util.basic import RefIdEq
```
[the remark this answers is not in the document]
albertz: Well, this is maybe a matter of preference. If you want, keep it as it is, and ignore my suggestion here.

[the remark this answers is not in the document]
Icemole: I originally based on `rf.Module.named_parameters`, which was imported like:
```python
from returnn.util.basic import OptionalNotImplementedError, RefIdEq
```
and then used as `RefIdEq` directly. But really I don't mind, if you prefer this way of setting it I'll change it.

returnn/torch/updater.py
Icemole: I can do that. For consistency I could also change:
```
self.optimizer = None  # type: typing.Optional[torch.optim.Optimizer]
```
to
```
self.optimizer = None  # type: Optional[torch.optim.Optimizer]
```
and
```suggestion
from typing import Set, Optional
```

albertz: Yea ok this is a bit unrelated to this PR, but if you want, you can do that.

PR #1330, opened 2023-05-17
Add seq_idx, seq_tag to torch datapipe

JackTemaki:
- Changes the torch data pipeline to allow for non-tensor data to be passed.
- Extended typing

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -19,31 +19,40 @@ other PyTorch datasets more directly, including also HuggingFace datasets.
 """
 
 from __future__ import annotations
-from typing import List, Dict
+from typing import List, Dict, Iterable, Union
 import sys
 from copy import deepcopy
 
-import numpy
+import numpy as np
 import torch
 import torch.utils.data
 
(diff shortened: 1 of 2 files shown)


albertz:
The recent commits added seq_tag to extern_data. It just uses a Numpy ndarray for that, which supports strings. It would also support any custom Python object, so it is generic to support any potential other case you might want to have later. So I add seq_tag as `numpy.array`. The only real remaining changes w.r.t. handling Numpy arrays are:

- `create_tensor`: Return Numpy array as-is for certain types. This is currently only the string type, but could be extended to Python object type, or other types which do not have meaningful PyTorch types.
- `collate_batch`: Also handle Numpy arrays.

Further, seq_tag is handled a bit as a special case, and is automatically added to extern_data if it is not specified.

I did not add seq_idx because the idx might not be meaningful. It depends on the current ordering of the dataset. Using `get_corpus_seq_idx` might be more meaningful. But I noticed that we anyway also have this wrong in the TF backend and it was probably never really used (as it was incorrect), so I left it away for now.

I guess this PR can be closed.

PR #1335, opened 2023-05-19
PyTorch DistributedDataParallel Multi-GPU training

### LINKED ISSUE #1332: PyTorch multi-GPU training, multi-node or single node, async or synced
For the PyTorch engine (#1120).

Multi-GPU training. Multi-node or single node. Async vs sync. It should also work well with consumer GPUs, also in the multi-node case. Currently distributed data parallel synchronized training, using `DistributedDataParallel`, and using our own dataset sampling logic, extending `Dataset._get_default_random_seed_offset`.

This issue is to track the process, discuss anything relevant.

Judyxujj:
For PyTorch (#1120).

1. Multi GPU training with `DistributedDataParallel`
2. Gradient accumulation
3. Add elapsed training time for each sub-epoch

Fix #1332.

Question: how do we set the multi-GPU flag in the returnn config?  I saw for the automated mix precision, the opts dict should be defined in the returnn config. Should I also use a dict opts for the DistributedDataParallel in the returnn config? So that we can define parameters like `bucket_cap_mb`, `find_unused_parameters` in the opts dict?

### DIFF
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -397,6 +397,11 @@ def finalize(error_occurred=False):
                 import horovod.tensorflow as hvd  # noqa
 
                 hvd.shutdown()
+        elif BackendEngine.is_torch_selected():
+            if config.typed_value("torch_distributed") is not None:
+                from torch.distributed import destroy_process_group
+
+                destroy_process_group()
 
 
 def need_data():
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -335,6 +335,7 @@ def init_backend_engine():
             returnn.tf.distributed.init_distributed_tf(config)
     elif BackendEngine.is_torch_selected():
         if config.is_true("use_DDP"):
+            # initializes the distributed backend which will take care of sychronizing nodes/GPUs
             from torch.distributed import init_process_group
             init_process_group("nccl")
         print("PyTorch:", util.describe_torch_version(), file=log.v3)
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -18,7 +18,6 @@ __license__ = "RWTHOCR"
 __maintainer__ = "Patrick Doetsch"
 __email__ = "doetsch@i6.informatik.rwth-aachen.de"
 
-
 import os
 import sys
 import time
@@ -334,11 +333,16 @@ def init_backend_engine():
 
             returnn.tf.distributed.init_distributed_tf(config)
     elif BackendEngine.is_torch_selected():
-        if config.is_true("use_DDP"):
-            # initializes the distributed backend which will take care of sychronizing nodes/GPUs
-            from torch.distributed import init_process_group
+        if config.typed_value("torch_distributed") is not None:
+            import socket
+            import returnn.torch.distributed
+            torch_distributed = returnn.torch.distributed.get_ctx(config=config)
+            print(
+                "Torch: Hostname %s, pid %i, using GPU %s."
+                % (socket.gethostname(), os.getpid(), str(torch_distributed.local_rank())),
+                file=log.v3,
+            )
 
-            init_process_group("nccl")
         print("PyTorch:", util.describe_torch_version(), file=log.v3)
     else:
         raise NotImplementedError
@@ -504,12 +508,12 @@ def execute_main_task():
         engine.init_network_from_config(config)
         engine.web_server(port=config.int("web_server_port", 12380))
     elif task.startswith("config:"):
-        action = config.typed_dict[task[len("config:") :]]
+        action = config.typed_dict[task[len("config:"):]]
         print("Task: %r" % action, file=log.v1)
         assert callable(action)
         action()
     elif task.startswith("optional-config:"):
-        action = config.typed_dict.get(task[len("optional-config:") :], None)
+        action = config.typed_dict.get(task[len("optional-config:"):], None)
         if action is None:
             print("No task found for %r, so just quitting." % task, file=log.v1)
         else:
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -337,6 +337,7 @@ def init_backend_engine():
         if config.is_true("use_DDP"):
             # initializes the distributed backend which will take care of sychronizing nodes/GPUs
             from torch.distributed import init_process_group
+
             init_process_group("nccl")
         print("PyTorch:", util.describe_torch_version(), file=log.v3)
     else:
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -334,6 +334,9 @@ def init_backend_engine():
 
             returnn.tf.distributed.init_distributed_tf(config)
     elif BackendEngine.is_torch_selected():
+        if config.is_true("use_DDP"):
+            from torch.distributed import init_process_group
+            init_process_group("nccl")
         print("PyTorch:", util.describe_torch_version(), file=log.v3)
     else:
         raise NotImplementedError
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -336,6 +336,7 @@ def init_backend_engine():
         if config.typed_value("torch_distributed") is not None:
             import socket
             import returnn.torch.distributed
+
             torch_distributed = returnn.torch.distributed.get_ctx(config=config)
             print(
                 "Torch: Hostname %s, pid %i, using GPU %s."
@@ -508,12 +509,12 @@ def execute_main_task():
         engine.init_network_from_config(config)
         engine.web_server(port=config.int("web_server_port", 12380))
     elif task.startswith("config:"):
-        action = config.typed_dict[task[len("config:"):]]
+        action = config.typed_dict[task[len("config:") :]]
         print("Task: %r" % action, file=log.v1)
         assert callable(action)
         action()
     elif task.startswith("optional-config:"):
-        action = config.typed_dict.get(task[len("optional-config:"):], None)
+        action = config.typed_dict.get(task[len("optional-config:") :], None)
         if action is None:
             print("No task found for %r, so just quitting." % task, file=log.v1)
         else:
--- a/returnn/__main__.py
+++ b/returnn/__main__.py
@@ -18,6 +18,7 @@ __license__ = "RWTHOCR"
 __maintainer__ = "Patrick Doetsch"
 __email__ = "doetsch@i6.informatik.rwth-aachen.de"
 
+
 import os
 import sys
 import time
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -245,9 +245,10 @@ class Dataset(object):
 
             if returnn.tf.horovod.get_ctx().is_dataset_distribution_random_seed_offset():
                 return returnn.tf.horovod.get_ctx().rank() * 16127
-        
+
         if config.is_true("use_DDP"):
             import torch.distributed as dist
+
             return dist.get_rank() * 16127
         return 0
 
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -246,10 +246,10 @@ class Dataset(object):
             if returnn.tf.horovod.get_ctx().is_dataset_distribution_random_seed_offset():
                 return returnn.tf.horovod.get_ctx().rank() * 16127
 
-        if config.is_true("use_DDP"):
-            import torch.distributed as dist
+        if config.typed_value("torch_distributed") is not None:
+            import returnn.torch.distributed
 
-            return dist.get_rank() * 16127
+            return returnn.torch.distributed.get_ctx().rank() * 16127
         return 0
 
     @staticmethod
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -245,6 +245,10 @@ class Dataset(object):
 
             if returnn.tf.horovod.get_ctx().is_dataset_distribution_random_seed_offset():
                 return returnn.tf.horovod.get_ctx().rank() * 16127
+        
+        if config.is_true("use_DDP"):
+            import torch.distributed as dist
+            return dist.get_rank() * 16127
         return 0
 
     @staticmethod
--- a/returnn/log.py
+++ b/returnn/log.py
@@ -192,6 +192,15 @@ class Log:
                 fn_ext = ".horovod-%i-%i%s" % (hvd.rank(), hvd.size(), fn_ext)
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
+        
+        if config.is_true("use_DDP"):
+            new_logs = []
+            for fn in logs:
+                fn_prefix, fn_ext = os.path.splitext(fn)
+                fn_ext = ".nccl-%i-%i%s" % (int(os.environ['LOCAL_RANK']), int(os.environ['WORLD_SIZE']), fn_ext)
+                new_logs.append(fn_prefix + fn_ext)
+            logs = new_logs
+
         self.initialize(logs=logs, verbosity=log_verbosity, formatter=log_format)
 
     def print_warning(self, text, prefix_text="WARNING:", extra_text=None):
--- a/returnn/log.py
+++ b/returnn/log.py
@@ -193,11 +193,14 @@ class Log:
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
 
-        if config.is_true("use_DDP"):
+        if config.typed_value("torch_distributed") is not None:
+            import returnn.torch.distributed
+
+            torch_distributed = returnn.torch.distributed.get_ctx(config=config)
             new_logs = []
             for fn in logs:
                 fn_prefix, fn_ext = os.path.splitext(fn)
-                fn_ext = ".nccl-%i-%i%s" % (int(os.environ["LOCAL_RANK"]), int(os.environ["WORLD_SIZE"]), fn_ext)
+                fn_ext = ".torch-distrib-%i-%i%s" % (torch_distributed.rank(), torch_distributed.size(), fn_ext)
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
 
--- a/returnn/log.py
+++ b/returnn/log.py
@@ -192,12 +192,12 @@ class Log:
                 fn_ext = ".horovod-%i-%i%s" % (hvd.rank(), hvd.size(), fn_ext)
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
-        
+
         if config.is_true("use_DDP"):
             new_logs = []
             for fn in logs:
                 fn_prefix, fn_ext = os.path.splitext(fn)
-                fn_ext = ".nccl-%i-%i%s" % (int(os.environ['LOCAL_RANK']), int(os.environ['WORLD_SIZE']), fn_ext)
+                fn_ext = ".nccl-%i-%i%s" % (int(os.environ["LOCAL_RANK"]), int(os.environ["WORLD_SIZE"]), fn_ext)
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
 
--- a/returnn/torch/distributed.py
+++ b/returnn/torch/distributed.py
@@ -59,7 +59,7 @@ class DistributedContext:
 
 
 _is_set_up = False
-_ctx = None  # type: typing.Optional[HorovodContext]
+_ctx = None  # type: Optional[DistributedContext]
 
 
 def get_ctx(config=None):
--- a/returnn/torch/distributed.py
+++ b/returnn/torch/distributed.py
@@ -3,6 +3,7 @@ torch.distributed utils
 """
 
 from __future__ import annotations
+import itertools
 from typing import Optional
 import os
 import socket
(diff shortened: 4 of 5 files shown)


returnn/torch/engine.py
```diff
@@ -64,6 +68,10 @@ def __init__(self, config: Config):
         self._device = _get_device_from_config(config)
         print("Using device:", self._device, file=log.v2)
 
+        self._use_DDP = config.is_true("use_DDP")  # type: Optional[bool]
```
albertz, line 71: As discussed, we want to have this a dict, and rename it, like `torch_distributed`.

albertz: There would be one `class` or `type` inside that dict, which would be `"DistributedDataParallel"`.

returnn/torch/engine.py
albertz, line 1: I think we should move out the DDP code into a separate file as much as possible, similar to how we have `returnn/tf/horovod.py` as a separate file, and then similar as `HorovodContext`, have some `DistributedContext` or so.

returnn/torch/engine.py
```diff
@@ -105,6 +113,11 @@ def init_train_from_config(
         """
         assert config is self.config
         super().init_train_from_config(config=config)
+        if self._use_DDP:
+            ddp_local_rank = int(os.environ["LOCAL_RANK"])
```
albertz, line 117: You should not use the `os.environ` but rather the `torch.distributed` functions.

returnn/torch/engine.py
```diff
@@ -130,6 +143,13 @@ def init_train_from_config(
         self._load_model(epoch=self._start_epoch)
         self._save_model_epoch_interval = config.int("save_interval", 1)
 
+        if self._use_DDP:
+            # wrap the model use DDP
+            self._pt_model = DDP(
+                self._pt_model,
+                device_ids=[int(os.environ["LOCAL_RANK"])],
+                find_unused_parameters=self._DDP_find_unused_parameters,
```
albertz, line 151: Instead of passing this, you would pass all the options from the user.

returnn/torch/engine.py
```diff
@@ -138,6 +158,11 @@ def init_train_from_config(
         self._train_step_func = self.config.typed_value("train_step")
         assert self._train_step_func, "train_step not defined"
 
+        self._gradient_accumulation_steps = config.int("gradient_accumulation_steps", 1)
```
albertz, line 161: In the TF engine, the option is called `accum_grad_multiple_step`. Why do you call it different here? I think we should just use the same name.

returnn/torch/engine.py
```diff
@@ -138,6 +158,11 @@ def init_train_from_config(
...
+        if self._gradient_accumulation_steps > 1:
+            assert self._gradient_accumulation_steps % torch.cuda.device_count() == 0
+            self._gradient_accumulation_steps //= torch.cuda.device_count()
```
albertz, line 164: As discussed, I think this is unexpected, so we should not do this.

returnn/torch/engine.py
```diff
@@ -359,9 +424,70 @@ def _run_step(self, extern_data_raw: Dict[str, torch.Tensor], *, train_flag: boo
...
+                    # this forward pass, to ensure we short circuit reduction for any
+                    # unused parameters. Only if `find_unused_parameters` is set.
+                    if self._pt_model.find_unused_parameters and not self._pt_model.static_graph:
+                        # Do not need to populate this for static graph.
+                        self._pt_model.reducer.prepare_for_backward(
+                            list(torch.nn.parallel.distributed._find_tensors(output))
+                        )
+                    else:
+                        self._pt_model.reducer.prepare_for_backward([])
+                else:
```
albertz, line 484: I assume you just copied all this code? Did you somehow adapt this?

You are calling a few private functions here. Not sure if this is a good idea.

In any case, we should move this code out here, into some context manager helper.

returnn/torch/engine.py
```diff
@@ -359,9 +424,70 @@ def _run_step(self, extern_data_raw: Dict[str, torch.Tensor], *, train_flag: boo
...
+
+                if self._pt_model._join_config.enable:
+                    # Notify joined ranks whether they should sync in backwards pass or not.
+                    self._pt_model._check_global_requires_backward_grad_sync(is_joined_rank=False)
+                with self._pt_model._inside_ddp_forward():
+                    with autocast(
+                        device_type=self._device, dtype=self._autocast_dtype
+                    ) if self._use_autocast else nullcontext():
+                        # If the flag use_DDP is enabled, we need to return the forward output in the train step function
+                        output = self._train_step_func(model=self._orig_model, extern_data=extern_data, **sentinel_kw)
```
albertz, line 465: There is no output.

returnn/torch/engine.py
```diff
@@ -359,9 +424,70 @@ def _run_step(self, extern_data_raw: Dict[str, torch.Tensor], *, train_flag: boo
...
+                    self._pt_model.require_forward_param_sync = True
+                    # We'll return the output object verbatim since it is a freeform
+                    # object. We need to find any tensors in this object, though,
+                    # because we need to figure out which parameters were used during
+                    # this forward pass, to ensure we short circuit reduction for any
+                    # unused parameters. Only if `find_unused_parameters` is set.
+                    if self._pt_model.find_unused_parameters and not self._pt_model.static_graph:
+                        # Do not need to populate this for static graph.
+                        self._pt_model.reducer.prepare_for_backward(
+                            list(torch.nn.parallel.distributed._find_tensors(output))
```
albertz, line 480: Use the loss instead. Access it via `rf.get_run_ctx()...`.

returnn/datasets/basic.py
```diff
@@ -245,6 +245,11 @@ def _get_default_random_seed_offset():
 
             if returnn.tf.horovod.get_ctx().is_dataset_distribution_random_seed_offset():
                 return returnn.tf.horovod.get_ctx().rank() * 16127
+
+        if config.is_true("use_DDP"):
```
albertz, line 249: `torch_distributed`

returnn/log.py
```diff
@@ -192,6 +192,15 @@ def init_by_config(self, config):
                 fn_ext = ".horovod-%i-%i%s" % (hvd.rank(), hvd.size(), fn_ext)
                 new_logs.append(fn_prefix + fn_ext)
             logs = new_logs
+
+        if config.is_true("use_DDP"):
+            new_logs = []
+            for fn in logs:
+                fn_prefix, fn_ext = os.path.splitext(fn)
+                fn_ext = ".nccl-%i-%i%s" % (int(os.environ["LOCAL_RANK"]), int(os.environ["WORLD_SIZE"]), fn_ext)
```
albertz, line 200: Why "nccl"? Call it "torch-distrib" or so. Also, do use `os.environ`.

returnn/torch/engine.py
Judyxujj: Yes, I have copied the codes. Where should be put the context manager helper, I now put it under returnn.torch.distributed

returnn/log.py
Judyxujj: I have to use `os.environ` to get the local rank, the `get_rank()` function from torch.distributed return rank and I couldn't find other function from torch.distributed to get the local rank

returnn/torch/engine.py
```diff
@@ -400,6 +526,8 @@ def _load_model(self, *, epoch: int):
         print("Model:", self._pt_model, file=log.v4)
 
         if checkpoint_state is not None:
+            # remove the prefix of "module." from the DDP state dict
+            torch.nn.modules.utils.consume_prefix_in_state_dict_if_present(checkpoint_state["model"], "module.")
```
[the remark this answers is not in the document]
Judyxujj: I am not sure in which cases we would need to keep the prefix module, if there are such cases, then I can add a condition here

returnn/torch/engine.py
albertz: I don't really understand what you mean. What do you mean by "the variable would be marked twice"? What does it mean to mark a variable?

To me it sounds like there is still some issue in the way we wrap the `forward` call.

returnn/__main__.py
```diff
@@ -499,12 +508,12 @@ def execute_main_task():
         engine.init_network_from_config(config)
         engine.web_server(port=config.int("web_server_port", 12380))
     elif task.startswith("config:"):
-        action = config.typed_dict[task[len("config:") :]]
+        action = config.typed_dict[task[len("config:"):]]
```
albertz, line 511: Why this change?

albertz:
Where is the `returnn/torch/distributed.py` file?

returnn/torch/engine.py
```diff
@@ -155,6 +185,8 @@ def train(self):
             self.epoch += 1
             self._epoch_mp_shared.value = self.epoch
 
+        if self._use_torch_distributed:
+            destroy_process_group()
```
albertz, line 189: This should be moved to `torch/distributed.py` as well.

returnn/torch/engine.py
```diff
@@ -359,9 +438,24 @@ def _run_step(self, extern_data_raw: Dict[str, torch.Tensor], *, train_flag: boo
...
 
-        with autocast(device_type=self._device, dtype=self._autocast_dtype) if self._use_autocast else nullcontext():
-            sentinel_kw = {"__fwd_compatible_random_arg_%i" % int(random() * 100): None}
-            self._train_step_func(model=self._orig_model, extern_data=extern_data, **sentinel_kw)
+        sentinel_kw = {"__fwd_compatible_random_arg_%i" % int(random() * 100): None}
+
+        with autocast(
+            device_type=self._device, dtype=self._autocast_dtype
+        ) if self._use_autocast else nullcontext():
+            if not self._use_torch_distributed:
```
albertz, line 446: There should be no if-statement here. You always should call the same `_train_step_func` func, in the same way.

The if-statement should be for the context manager, so inside the `with` statement above.

returnn/torch/engine.py
```diff
@@ -359,9 +438,24 @@ def _run_step(self, extern_data_raw: Dict[str, torch.Tensor], *, train_flag: boo
...
         rf.init_train_step_run_ctx(train_flag=train_flag)
 
-        with autocast(device_type=self._device, dtype=self._autocast_dtype) if self._use_autocast else nullcontext():
-            sentinel_kw = {"__fwd_compatible_random_arg_%i" % int(random() * 100): None}
-            self._train_step_func(model=self._orig_model, extern_data=extern_data, **sentinel_kw)
+        sentinel_kw = {"__fwd_compatible_random_arg_%i" % int(random() * 100): None}
+
+        with autocast(
+            device_type=self._device, dtype=self._autocast_dtype
+        ) if self._use_autocast else nullcontext():
```
albertz, line 445:
```suggestion
        ) if self._use_autocast else nullcontext(), ... if self._use_torch_distributed else nullcontext():
```
Like this. (Properly formatted.)

returnn/torch/engine.py
```diff
@@ -8,6 +8,10 @@
 
 import os
 import torch
+import time
+from torch.distributed import init_process_group, destroy_process_group
+from torch.nn.parallel import DistributedDataParallel as DDP
+from torch.distributed.algorithms.join import Join
```
albertz, line 14: Do not import such generic names (`Join`) into the main namespace.

albertz: Do you even need this here? I thought this is now in `returnn.torch.distributed`?

returnn/torch/engine.py
```diff
@@ -193,13 +250,29 @@ def train_epoch(self):
                 {name: float(_to_raw(loss.get_inv_norm_factor())) for name, loss in train_ctx.losses.items()}
             )
 
-            if self._grad_scaler is not None:
-                self._grad_scaler.scale(total_loss).backward()
-                self._grad_scaler.step(self._updater.get_optimizer())
-                self._grad_scaler.update()
+            if self._use_torch_distributed and (step_idx % self._accum_grad_multiple_step) != (
```
albertz, line 253: Do not put the `if` statement here. Do not repeat the same code below twice. Rather, Just have the code once, but have the context manager optionally, and use `nullcontext` otherwise.

returnn/torch/distributed.py
```diff
@@ -0,0 +1,135 @@
...
+            return None
+    _is_set_up = True
+    if config.typed_value("torch_distributed") is None:
+        return None
+    _ctx = DistributedContext(config=config)
+    return _ctx
+
+
+@contextmanager
+def _ddp_ctx(pt_model):
```
albertz, line 83: As you use this outside of this module, make it public. Also choose a more describing name, like `ddp_train_ctx` or so.

returnn/torch/engine.py
```diff
@@ -105,6 +119,14 @@ def init_train_from_config(
         """
         assert config is self.config
         super().init_train_from_config(config=config)
+        if self._use_torch_distributed:
+            import returnn.torch.distributed
+
+            torch_distributed = returnn.torch.distributed.get_ctx(config=config)
+            local_rank = torch_distributed.local_rank()
+            print(f"Start running basic DDP example on local rank {local_rank}.", file=log.v2)
```
albertz, line 127: Why "DDP example"?

returnn/torch/engine.py
```diff
@@ -130,14 +152,23 @@ def init_train_from_config(
         self._load_model(epoch=self._start_epoch)
         self._save_model_epoch_interval = config.int("save_interval", 1)
 
-        self._updater = Updater(self.config, self._pt_model, self.learning_rate)
+        if self._use_torch_distributed:
```
albertz, line 155: `self._ddp_pt_model = None` should be set outside/before this.

returnn/torch/engine.py
```diff
@@ -155,6 +186,8 @@ def train(self):
             self.epoch += 1
             self._epoch_mp_shared.value = self.epoch
 
+        if self._use_torch_distributed:
+            destroy_process_group()
```
albertz, line 190: Why are you calling this here? As far as I understood, this function should be called at the end of the process? E.g. in our `finalize` function.

returnn/torch/engine.py
```diff
@@ -130,14 +152,23 @@ def init_train_from_config(
         self._load_model(epoch=self._start_epoch)
         self._save_model_epoch_interval = config.int("save_interval", 1)
 
-        self._updater = Updater(self.config, self._pt_model, self.learning_rate)
+        if self._use_torch_distributed:
+            # wrap the model use torch distributed class
+            self._ddp_pt_model = self._torch_distributed_class(
+                self._pt_model, device_ids=[int(os.environ["LOCAL_RANK"])], **self._torch_distributed_options
```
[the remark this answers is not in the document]
albertz: After discussion:

The reason `os.environ` was used is because this was the only known way to get the local rank (as opposed to the global rank).

It's a bit ambiguous what we should use this, whether it is the local rank, global rank, the device id 0, or what else. It also matter how `CUDA_VISIBLE_DEVICES` is set up, single node vs multi node training, etc. Also how it is run later on, via SGE, mpirun, whatever. Using the local rank is probably reasonable for our current use cases though.

The next question is, how to get the local rank. `torch.distributed` does not provide such function?

So, we agreed to introduce some helper functions (in our `returnn.torch.distributed` module):
```python
def get_device_ids():
  # It depends on the specific setup what to return here,
  # how CUDA_VISIBLE_DEVICES is set up, etc.
  # This is currently a reasonable assumption
  # but we might extend the logic later,
  # or make it configurable.
  return [get_local_rank()]

def get_local_rank():
  # torch.distributed does not seem to provide a function for this.
  # Via mpirun (OpenMPI), this env variable would be set.
  # It should fail with an error otherwise.
  return int(os.environ["LOCAL_RANK"])
```

returnn/torch/engine.py
```diff
@@ -177,11 +210,35 @@ def train_epoch(self):
...
-        for data in self._train_dataloader:
-            self._updater.get_optimizer().zero_grad()
+        epoch_start_time = time.time()
+
+        data_iter = iter(self._train_dataloader)
+        elapsed_computation_time = 0
+
+        while True:
+            data = next(data_iter, None)
+            _has_data = torch.tensor([data is not None], dtype=torch.int8).to(self._device)
```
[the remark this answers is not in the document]
Judyxujj: if I do not use `to(self._device)`, I get the error RuntimeError: Tensors must be CUDA and dense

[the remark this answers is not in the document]
albertz: Where do you get that error?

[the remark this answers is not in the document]
Judyxujj: from here `torch.distributed.all_reduce(_has_data, op=torch.distributed.ReduceOp.MIN)`

returnn/torch/engine.py
```diff
@@ -130,14 +152,23 @@ def init_train_from_config(
...
 
-        self._updater = Updater(self.config, self._pt_model, self.learning_rate)
+        if self._use_torch_distributed:
+            # wrap the model use torch distributed class
+            self._ddp_pt_model = self._torch_distributed_class(
+                self._pt_model, device_ids=[int(os.environ["LOCAL_RANK"])], **self._torch_distributed_options
+            )
+            self._updater = Updater(self.config, self._ddp_pt_model, self.learning_rate)
+        else:
+            self._updater = Updater(self.config, self._pt_model, self.learning_rate)
```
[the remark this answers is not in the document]
Judyxujj: I have tested, to pass the DDP wrapped module or not has indeed no impact, so now I just pass `self._pt_mode ` here

returnn/torch/engine.py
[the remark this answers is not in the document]
Judyxujj: as suggested, I added the helper functions in `returnn.torch.distributed` module

returnn/torch/engine.py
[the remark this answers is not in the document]
albertz: Oh ok. I did not expect that it can only use one single mechanism for communication, namely NCCL here. This might be a bit suboptimal here (roundtrip CPU -> GPU -> NCCL -> GPU -> CPU; I was rather expecting CPU -> ethernet -> CPU). But some big warning comment then, and maybe we later need to deal with this.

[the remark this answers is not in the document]
albertz:
```suggestion
            # WARNING: torch.distributed works only for the registered device,
            # as it uses only one mechanism for communication, like NCCL.
            # This is suboptimal here as we have the roundtrip CPU -> GPU -> NCCL -> GPU -> CPU.
            # TODO: Use more direct CPU -> Ethernet -> CPU communication.
            _has_data = torch.tensor([data is not None], dtype=torch.int8).to(self._device)
```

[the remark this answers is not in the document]
Judyxujj: okay, I have added the warning comment here

returnn/torch/distributed.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def size(self):
+        """
+        :rtype: int
+        """
+        return self._size
+
+
+_is_set_up = False
+_ctx = None  # type: typing.Optional[HorovodContext]
```
albertz, line 56:
```suggestion
_ctx = None  # type: Optional[DistributedContext]
```

Judyxujj:
> There are conflicts. Can you rebase?
> 
> There are failing tests. Can you make sure they are passing?

I have merged the master branch to this branch and passed all tests now. In the commit [738b091](https://github.com/rwth-i6/returnn/commit/738b0910ee35d3d8ec3328a23a21eb9f3a249956) , `seq_tag` is added to the dataset key, it will cause error when doing the chunking. So I added an if statement to do chunking only on `data_dict['data']` as in commit [33cc459](https://github.com/rwth-i6/returnn/pull/1335/commits/33cc459dc4abfaba93bf2141fcb338e0737dd37e#). I am not sure if this is the right solution for it.
```
data = data_dict[data_key]
chunks = [
     data[start_index : start_index + chunk_size] for start_index in range(0, len(data), chunk_step)
 ]
```

returnn/torch/data/pipeline.py
```diff
@@ -114,22 +114,23 @@ def __iter__(self):
...
-
-                if num_chunks is None:
-                    num_chunks = len(chunks)
-                else:
-                    assert num_chunks == len(
-                        chunks
-                    ), "Chunking resulted in different number of chunks for different data keys."
-
-                data_chunks[data_key] = chunks
+                if data_key == "data":
```
albertz, line 117: You never should hardcode any data key name here (except maybe of the special keys like `seq_tag`, but `data` is really not special). Also, there should be no logic specific for some data key (except maybe the special keys like `seq_tag`).

albertz:
> I have merged the master branch to this branch and passed all tests now. In the commit [738b091](https://github.com/rwth-i6/returnn/commit/738b0910ee35d3d8ec3328a23a21eb9f3a249956) , `seq_tag` is added to the dataset key, it will cause error when doing the chunking. So I added an if statement to do chunking only on `data_dict['data']` as in commit [33cc459](https://github.com/rwth-i6/returnn/pull/1335/commits/33cc459dc4abfaba93bf2141fcb338e0737dd37e#). I am not sure if this is the right solution for it.

No it's not. I commented already.

But it's anyway not really related to this PR. This is a completely separate issue. Can you open a separate issue on it?

returnn/torch/data/pipeline.py
albertz: I see you added this as a workaround for chunking, which breaks because of `seq_tag`. This does not belong here into the PR. Can you undo this change?

For testing now without having this fixed, just don't use chunking. Let us fix chunking in a separate commit or PR.

albertz (changes requested):
Remove the data pipeline change. Otherwise I think we can merge now and then improve the code further when needed.

Judyxujj:
I removed the data pipeline changes and opened a separate issue for that. All the tests are passed and this branch is ready to be merged.

albertz (approved):

albertz:
Why do you actually have `_torch_distributed_class` (`class` in `torch_distributed` config dict) configurable? I recently changed the default of this to be `DistributedDataParallel`, but I wonder, are there really any potential cases where this can be sth different?

Judyxujj:
What about the fully sharded data parallel (FSDP)?(https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html) 
But I think to support this, we need to modify more things

Issue #1333, opened 2023-05-19
PyTorch ONNX export


albertz:
This issue is to track any aspects and issues on PyTorch (#1120) ONNX export.

* [x] Working script for conversion (`export_to_onnx.py`)
* [x] Fix issue with convolution
* [x] Rename script to `torch_export_to_onnx.py`
* [x] Check `model_outputs`.
* [x] Use `model_outputs` dims when not specified in `mark_as_output` (esp in case of PT).
* [x] The input names for sizes should be better (currently they are like "classes:size1" but should be like "data:size1"). (Fixed via #1362.)
* [x] The batch dim (`data:size0`), or any scalar dyn_size_ext, should not be an input, as it is redundant. (Fixed via #1362.)
* [x] `model_outputs`: Ignore dim tag equality. Check static dims, dyn dims information
* [x] Out seq lens are actually relevant, and should not be filled with dummy values? Clarify. (https://github.com/rwth-i6/returnn/issues/1333#issuecomment-1625179457) (Fixed via #1362.)
* [x] Working demo-rf + test case
* [x] Test case: Also perform ONNX inference (Fixed via #1362.)
* [ ] RF Conformer works
* [ ] Some real-world pure PT model works
* [ ] Running the resulting ONNX model in RASR
* [x] Writing a sisyphus job to run ONNX export in i6_core. (Fixed via https://github.com/rwth-i6/i6_core/pull/429.)

Other things which are less important for now:

* Avoiding `TracerWarning`s for some sanity checking code: I think it's not really possible in an easy way.

albertz:
The initial version is done already (`export_to_onnx.py`) and works for pure PT code, so I think the main work is done. We can already close the issue. But let's anyway add further aspects on this here.

albertz:
One aspect is that convolution with `padding="same"` is not supported by ONNX. However, we can simply work around that by doing the padding manually. The code is already there for the RF, and just needs to be used. We can use `torch.onnx.is_in_export_mode()` to test whether we should use it.

Icemole:
As a form of summary: the tool is mainly done, but there are some [operators that aren't yet supported](https://pytorch.org/docs/stable/onnx_supported_aten_ops.html), one of which is the convolutional layer padding when specified as a string (`"same"` or `"valid"`). This prevents the demos from being run with the tool, which is one of my top priorities.

[...]

PR #1343, opened 2023-06-14
Reuse global batch dim

albertz:
https://github.com/rwth-i6/returnn/issues/1333#issuecomment-1563052008
https://github.com/rwth-i6/returnn/issues/1333#issuecomment-1590965034

It was mentioned that the recreation of the batch dim was unexpected and caused problems because in PyTorch engine, we would set the `dyn_size_ext` of the batch dim, and otherwise not use the batch info mechanism, so a separate instance of such dim tag would not correctly have the `dyn_size_ext`. The recreation happens here: https://github.com/rwth-i6/returnn/blob/04c14d6a3745f90f0bd10b28ce4d51b1d61afb82/returnn/tensor/_tensor_extra.py#L3577

This is only for `model_outputs`.

It is anyway a bit bad that we have all that special handling for batch dim tags, so that the behavior depends so much on the dim tag type.

### DIFF
--- a/returnn/tensor/_dim_extra.py
+++ b/returnn/tensor/_dim_extra.py
@@ -436,6 +436,8 @@ class _DimMixin:
             # We ignore the ctx for the batch dim currently.
             if self.batch == batch:
                 return self
+            if batch.is_global_batch():
+                return _d.batch_dim  # reuse global batch dim if possible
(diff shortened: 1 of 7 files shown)


albertz:
Without having looked at it in detail, I assume the failing tests are because a new global batch info is created and then it gets confused when the global batch dim tag still references the old global batch info.

PR #1344, opened 2023-06-14
Min and max seq len for Pytorch

Atticus1806:
This PR adds `min_seq_length` and `max_seq_length` as parameters to the torch configs, similar to tensorflow.

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -240,3 +240,40 @@ class BatchingIterDataPipe(torch.utils.data.IterDataPipe):
 
         if current_batch:
             yield current_batch
+
+
+class LenFilterDataPipe(torch.utils.data.IterDataPipe):
+    """
+    Removes sequences which are either too long or too short from a dataset
+    Returns dataset yielding list of data lengths within the defined range
+    """
+
+    def __init__(self, dataset: torch.utils.data.IterableDataset,
+        min_seq_len: Union[int, NumbersDict] = None,
+        max_seq_len: Union[int, NumbersDict] = None):
+        """
+        :param dataset: dataset to apply the filter to
+        :param min_seq_len: minimum sequence length either in general or per data_key via dict
+        :param max_seq_len: maximum sequence length either in general or per data_key via dict
+        """
+        super().__init__()
+        self._dataset = dataset
+        self._min_seq_len = NumbersDict(0 if min_seq_len is None else min_seq_len)
+        self._max_seq_len = NumbersDict(sys.maxsize if max_seq_len is None else max_seq_len)
+
+    def __iter__(self):
+        """
+        :return: generator providing filtered data where each sequence is a dict
+          data_key -> data_array.
+        :rtype: Iterable[dict[str, numpy.ndarray]]
+        """
+        for data_dict in self._dataset:
+            # TODO: This assumes all data has time as first dimension. Currently we can't know better..
+            sequence_lengths = NumbersDict(
+                {data_key: data.shape[0] for data_key, data in data_dict.items() if data.shape}
+            )
+            if sequence_lengths.any_compare(self._min_seq_len, lambda a, b: a < b):
+                continue
+            if sequence_lengths.any_compare(self._max_seq_len, lambda a,b: a > b):
+                continue
+            yield data_dict
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -248,8 +248,6 @@ class LenFilterDataPipe(torch.utils.data.IterDataPipe):
     Returns dataset yielding list of data lengths within the defined range
     """
 
-    def __getitem__(self, index):
-        raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")
 
     def __init__(
         self,
@@ -283,3 +281,6 @@ class LenFilterDataPipe(torch.utils.data.IterDataPipe):
             if sequence_lengths.any_compare(self._max_seq_length, lambda a, b: a > b):
                 continue
             yield data_dict
+
+    def __getitem__(self, index):
+        raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")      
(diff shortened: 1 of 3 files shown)


Atticus1806:
Which abstract class is he missing? `getitem`? I dont think this makes sense here tbh.
Would you raise an error or just ignore the warning? Both are done above so not sure which one is better

albertz:
> Which abstract class is he missing? `getitem`? I dont think this makes sense here tbh. Would you raise an error or just ignore the warning? Both are done above so not sure which one is better

I assume you refer to the PyCharm warning?

> `returnn/torch/data/pipeline.py:245: WEAK WARNING PyAbstractClassInspection: Class LenFilterDataPipe must implement all abstract methods`

I don't know without looking at the base class carefully. PyCharm can tell you that automatically. Just rightclick -> implement abstract methods (or sth like that).

tests/test_torch_engine.py
```diff
@@ -4,6 +4,8 @@
 
 import _setup_test_env  # noqa
 import torch
+
+import returnn.datasets.generating
```
albertz, line 8: This import seems unnecessary.

returnn/torch/engine.py
```diff
@@ -416,7 +423,9 @@ def _create_data_loader(self, dataset: Dataset) -> DataLoader2:
         )
 
         wrapped_dataset = returnn_dataset_wrapper.ReturnnDatasetIterDataPipe(dataset, reset_callback=dataset_reset)
-
+        wrapped_dataset = data_pipeline.LenFilterDataPipe(
```
albertz, line 426: For simplicity of the data pipeline, I would only apply this if `_min_seq_len` or `_max_seq_len` is set.

returnn/torch/engine.py
```diff
@@ -57,6 +58,12 @@ def __init__(self, config: Config):
 
         self._start_epoch = None  # type: Optional[int]
         self._final_epoch = None  # type: Optional[int]
+        self._min_seq_len = config.typed_value("min_seq_len", None) or config.int(
+            "min_seq_len", 0
+        )  # type: Union[int,float,Dict[str,int],NumbersDict]
+        self._max_seq_len = config.typed_value("max_seq_len", None) or config.int(
+            "max_seq_len", sys.maxsize
+        )  # type: Union[int,float,Dict[str,int],NumbersDict]
```
albertz, line 66: The options are called `min_seq_length` and `max_seq_length`. Why this inconsistency? I would just use the same name.

returnn/torch/engine.py
```diff
@@ -57,6 +58,12 @@ def __init__(self, config: Config):
 
         self._start_epoch = None  # type: Optional[int]
         self._final_epoch = None  # type: Optional[int]
+        self._min_seq_len = config.typed_value("min_seq_len", None) or config.int(
+            "min_seq_len", 0
+        )  # type: Union[int,float,Dict[str,int],NumbersDict]
+        self._max_seq_len = config.typed_value("max_seq_len", None) or config.int(
+            "max_seq_len", sys.maxsize
```
albertz, line 65: I think I would use 0 or None as the default, not `sys.maxsize`. That makes it easier to check whether it is enabled, easier in debugging.

Atticus1806:
I think the only error left is not caused by this PR.

returnn/torch/data/pipeline.py
```diff
@@ -26,6 +26,7 @@
 import numpy
 import torch
 import torch.utils.data
+from torch.utils.data.dataset import T_co
```
albertz, line 29: You don't really need that. That also causes a warning. It is not supposed to be imported.

returnn/torch/data/pipeline.py
```diff
@@ -240,3 +241,46 @@ def __iter__(self):
...
             yield current_batch
+
+
+class LenFilterDataPipe(torch.utils.data.IterDataPipe):
+    """
+    Removes sequences which are either too long or too short from a dataset
+    Returns dataset yielding list of data lengths within the defined range
+    """
+
+    def __getitem__(self, index) -> T_co:
```
albertz, line 252:
```suggestion
    def __getitem__(self, index):
```

returnn/torch/data/pipeline.py
albertz, line 29:
```suggestion
```

returnn/torch/data/pipeline.py
```diff
@@ -240,3 +240,46 @@ def __iter__(self):
...
+
+
+class LenFilterDataPipe(torch.utils.data.IterDataPipe):
+    """
+    Removes sequences which are either too long or too short from a dataset
+    Returns dataset yielding list of data lengths within the defined range
+    """
+
+    def __getitem__(self, index):
+        raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")
```
albertz, line 252: The `__init__` function should come first.

```suggestion
```

returnn/torch/data/pipeline.py
```diff
@@ -240,3 +240,46 @@ def __iter__(self):
...
+        for data_dict in self._dataset:
+            # TODO: This assumes all data has time as first dimension. Currently we can't know better..
+            sequence_lengths = NumbersDict(
+                {data_key: data.shape[0] for data_key, data in data_dict.items() if data.shape}
+            )
+            if sequence_lengths.any_compare(self._min_seq_length, lambda a, b: a < b):
+                continue
+            if sequence_lengths.any_compare(self._max_seq_length, lambda a, b: a > b):
+                continue
+            yield data_dict
```
albertz, line 285:
```suggestion
            yield data_dict

    def __getitem__(self, index):
        raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")      
```

returnn/torch/data/pipeline.py
```diff
@@ -240,3 +240,47 @@ def __iter__(self):
...
         if current_batch:
             yield current_batch
+
+
+class LenFilterDataPipe(torch.utils.data.IterDataPipe):
+    """
+    Removes sequences which are either too long or too short from a dataset
+    Returns dataset yielding list of data lengths within the defined range
+    """
+
```
albertz, line 250:
```suggestion
```

returnn/torch/data/pipeline.py
Atticus1806: Oh. I think this was automatically added by the PyCharm auto generation. But agreed, this can be dropped.

albertz:
Can you apply Black?

albertz (approved):

PR #1346, opened 2023-06-19
PT load checkpoint to device

### LINKED ISSUE #1345: CUDA out of memory issue when resuming Torch model training
When running with multi-gpu, once the training job is interrupted, when restarting, when it loads the existing checkpoint, I always get` OutOfMemoryError: CUDA out of memory` at the first training step (the same GPUs are used). But such error does not happen when I run training with single GPU.

The possible reason could be that with multi-GPU training, we save the checkpoint when `rank==0`. Thus when loading the checkpoint, instead of loading the checkpoints to each device respectively, they are all loaded to device 0.

Judyxujj:
load the checkpoint_state to self._device or cpu first

Fix #1345.


returnn/torch/updater.py
```diff
@@ -122,8 +122,9 @@ def load_optimizer(self, filename):
         :param str filename: File from which to load the optimizer state.
         """
         print("Load optimizer %s" % filename, file=log.v4)
-        optimizer_state = torch.load(filename)
+        optimizer_state = torch.load(filename, map_location=torch.device("cpu"))
```
albertz, line 125: Why? Why not `self._device`?

Judyxujj: is the optimiser also send to self._device? I couldn't find code that sends the optimiser to self._device

albertz: It should. It would be very inefficient/slow if it is not. It would probably also throw an error, as you calculate gradients on GPU, the parameters are on GPU, you calculate the gradient updates on GPU, and that involves the optimizer state (momentum etc), so if that state is on CPU, you should get an error. So I'm pretty sure it is normally on GPU. I'm not sure what code exactly moves it to GPU. Maybe it is automatically on the same device as the parameters.

returnn/torch/engine.py
```diff
@@ -577,6 +577,8 @@ def _load_model(self):
                     assert not missing_prefix_keys, f"Missing keys and ignore_missing=False: {missing_prefix_keys}"
                 print(f"Missing keys: {missing_keys}", file=log.v4)
 
+        del checkpoint_state
+        torch.cuda.empty_cache()
```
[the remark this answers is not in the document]
Judyxujj: I use` gc.collect() `instead now

returnn/torch/updater.py
```diff
@@ -80,7 +81,7 @@ class Updater(object):
     Wraps a torch.optim.Optimizer, and extends it by some further functionality.
     """
 
-    def __init__(self, config, network, initial_learning_rate=1.0):
+    def __init__(self, config, network, device, initial_learning_rate=1.0):
```
albertz, line 84: I think we should enforce kwargs-only usage here.
```suggestion
    def __init__(self, *, config, network, device, initial_learning_rate=1.0):
```

returnn/torch/engine.py
```diff
@@ -577,6 +578,8 @@ def _load_model(self):
                     assert not missing_prefix_keys, f"Missing keys and ignore_missing=False: {missing_prefix_keys}"
                 print(f"Missing keys: {missing_keys}", file=log.v4)
 
+        del checkpoint_state
```
albertz, line 581: There should be some comment pointing to the issue, or this PR here, or so.

returnn/torch/updater.py
```diff
@@ -122,8 +124,10 @@ def load_optimizer(self, filename):
         :param str filename: File from which to load the optimizer state.
         """
         print("Load optimizer %s" % filename, file=log.v4)
-        optimizer_state = torch.load(filename)
+        optimizer_state = torch.load(filename, map_location=self._device)
         self.optimizer.load_state_dict(optimizer_state)
+        del optimizer_state
```
albertz, line 129: Same here.

returnn/torch/engine.py
albertz:
```suggestion
        # https://github.com/rwth-i6/returnn/issues/1345
        del checkpoint_state
```

returnn/torch/updater.py
albertz:
```suggestion
        https://github.com/rwth-i6/returnn/issues/1345
        del optimizer_state
```

returnn/torch/engine.py
```diff
@@ -481,7 +482,7 @@ def _load_model(self):
         if model_epoch_filename:
             filename = model_epoch_filename + util.get_model_filename_postfix()
             print("Load model %s" % (filename,), file=log.v4)
-            checkpoint_state = torch.load(filename)
+            checkpoint_state = torch.load(filename, map_location=self._device)
```
albertz, line 485:
```suggestion
            checkpoint_state = torch.load(filename, map_location=self._device)
```

returnn/torch/engine.py
```diff
@@ -187,7 +188,7 @@ def init_train_from_config(
             self._ddp_pt_model = self._torch_distributed_class(
                 self._pt_model, device_ids=get_device_ids(), **self._torch_distributed_options
             )
-        self._updater = Updater(self.config, self._pt_model, self.learning_rate)
+        self._updater = Updater(self.config, self._pt_model, self._device, self.learning_rate)
```
albertz, line 191: Use kwargs.

albertz:
So this PR is considered ready now? And you verified that this PR here would fix the issue described in https://github.com/rwth-i6/returnn/issues/1345?

Judyxujj:
For my multi-GPU training, this can fix the CUDA out of the memory issue. I have tried to print out the device of the tensor in the checkpoint_state, without specifying `map_location`, the device is `device cuda:0` for all the process. So all the checkpoints would be loaded to device 0 and it will cause CUDA OOM.  Maybe do you have other suggestions what else should I check.

albertz:
I don't exactly understand. If you get it all on the same device, it means the problem is not fixed? Or does this change now via this PR here? Does this PR here actually fix anything? Does it fix #1345 or not? So it is not on the same device anymore?

Judyxujj:
> I don't exactly understand. If you get it all on the same device, it means the problem is not fixed? Or does this change now via this PR here? Does this PR here actually fix anything? Does it fix #1345 or not? So it is not on the same device anymore?

Sorry for the confusion. Without this PR, all checkpoints will be loaded to` device cuda:0`. With this PR, the checkpoint will be loaded to the corresponding device respectively. So this PR can fix #1345

albertz (approved):

PR #1347, opened 2023-06-23
Add get_current_seq_order to MetaDataset

michelwi:
This is required by `MultiProcDataset._seq_order_proc_loop` https://github.com/rwth-i6/returnn/blob/1cee4aa1dadb1e216b9311dd65d4753753b7beb6/returnn/datasets/multi_proc.py#L131

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -427,6 +427,13 @@ class MetaDataset(CachedDataset2):
             dataset.init_seq_order(epoch=epoch, seq_list=self.seq_list_ordered[dataset_key])
         return True
 
+    def get_current_seq_order(self):
+        """
+        :return: returnsthe current seq order for the current epoch, after self.init_seq_order was called.
+        :rtype: list[int]
+        """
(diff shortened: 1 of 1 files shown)


returnn/datasets/meta.py
```diff
@@ -427,6 +427,13 @@ def get_seq_len(s):
             dataset.init_seq_order(epoch=epoch, seq_list=self.seq_list_ordered[dataset_key])
         return True
 
+    def get_current_seq_order(self):
+        """
+        :return: returnsthe current seq order for the current epoch, after self.init_seq_order was called.
```
albertz, line 432:
```suggestion
        :return: current seq order for the current epoch, after self.init_seq_order was called.
```

returnn/datasets/meta.py
```diff
@@ -427,6 +427,13 @@ def get_seq_len(s):
...
+        :rtype: list[int]
+        """
+        return [self.tag_idx[t] for t in self.seq_list_ordered[self.default_dataset_key]]
```
albertz, line 435:
```suggestion
        return [self.tag_idx[tag] for tag in self.seq_list_ordered[self.default_dataset_key]]
```

PR #1348, opened 2023-06-23
PyTorch ONNX export test

albertz:
Address task on #1333


albertz:
Currently it is failing not because the Torch ONNX export is not working correctly but actually because the Torch ONNX export script fails to find RETURNN.

```
  File "tools/torch_export_to_onnx.py", line 14, in <module>
    from returnn.config import Config
ModuleNotFoundError: No module named 'returnn'
```

PR #1349, opened 2023-06-23
rename _get_data to get_data in MetaDataset

michelwi:
Original problem was that `MultiProcDataset` calls the `get_data` method of the included dataset. Since `MetaDataset` does not implement it, the method of `CachedDataset2` is used, which does not work with `MetaDataset`.

But then I noticed that in almost all other datasets the equivalent method is called `get_data` and only the datasets from `meta.py` and `lm.py` have `_get_data` instead. Is there any reason for this? if not, I propose to rename it.

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -471,7 +471,7 @@ class MetaDataset(CachedDataset2):
         self_seq_tag = self.seq_list_ordered[dataset_key][seq_idx]
         assert dataset_seq_tag == self_seq_tag
 
-    def _get_data(self, seq_idx, data_key):
+    def get_data(self, seq_idx, data_key):
         """
         :type seq_idx: int
         :type data_key: str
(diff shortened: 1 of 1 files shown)


albertz:
> the method [`get_data`] of `CachedDataset2` is used, which does not work with `MetaDataset`.

I don't understand. Why does it not work? This function is used all the time, e.g. when loading the data for mini-batch preparation, here:
https://github.com/rwth-i6/returnn/blob/dbcf75cc0f689b3e74f859c73e2423594376009f/returnn/tf/data_pipeline.py#L353
Or here:
https://github.com/rwth-i6/returnn/blob/dbcf75cc0f689b3e74f859c73e2423594376009f/returnn/torch/data/returnn_dataset_wrapper.py#L80

It definitely should work. Maybe it is used incorrectly by `MultiProcDataset`?

`_get_data` and `get_data` are also not exactly the same here. The public `get_data` internally uses `_collect_single_seq` and uses the caching mechanism. The internal `_get_data` is used by `_collect_single_seq` once, and then it is cached.

michelwi:
sorry, upon further investigation I do not think this PR would solve the problem any more.
I habe created a test case to illustrate the problem in https://github.com/rwth-i6/returnn/pull/1374 and I think we can go from there.

PR #1351, opened 2023-06-29
enable both gloo and nccl backend for torch distributed

Judyxujj:
This PR enables both nccl and gloo backend for torch.distributed. The gloo backend will be used for collectives with CPU tensors and the nccl backend will be used for collectives with CUDA tensors. Before this PR, we only used nccl backend which is only  available with CUDA tensors. Thus for the has_data all reduce operation at the beginning of each training step iteration, we have to make a roundtrip to send has_data Tensor through CPU -> GPU -> NCCL -> GPU -> CPU, which is very inefficient.  With this PR, gloo backed is available for the CPU tensors. Therefore, has_data all reduce can be operated on cpu directly and save the detour.

### DIFF
--- a/returnn/torch/distributed.py
+++ b/returnn/torch/distributed.py
@@ -27,7 +27,7 @@ class DistributedContext:
         """
         import torch.distributed as dist
 
-        dist.init_process_group("nccl")
+        dist.init_process_group("gloo")
 
         self._config = config
         self._local_rank = os.environ["LOCAL_RANK"]
--- a/returnn/torch/distributed.py
+++ b/returnn/torch/distributed.py
@@ -27,7 +27,10 @@ class DistributedContext:
         """
         import torch.distributed as dist
 
-        dist.init_process_group()
+        # when no backend is specified, both gloo and nccl backends will be created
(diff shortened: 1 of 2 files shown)


returnn/torch/distributed.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, config):
         """
         import torch.distributed as dist
 
-        dist.init_process_group("nccl")
+        dist.init_process_group()
```
albertz, line 30: Your PR description says that you create a Gloo backend, but I thought this code here would create both a Gloo and a NCCL backend? See the documentation. This is also important. We really want to have the NCCL backend, so the PR title is misleading? We are not changing the NCCL backend to the Gloo backend. We want to *add* the Gloo backend as an *additional* backend.

You also should add a comment which explains that, and fix the PR title, and the PR description to explain that.

Judyxujj: Thanks for suggestion. I have added more description and changed the title. Would it be informative enough now?

albertz (approved):

PR #1352, opened 2023-07-03
TF initial state key name fix, keep_over_epoch, var, placeholder

albertz:
The `key_name` is used for `initial_state` cases `keep_over_epoch`, `var`, `placeholder`


albertz:
What is your config?

curufinwe:
https://gist.github.com/curufinwe/1b913ecad24a75b721118271e54f0600

albertz:
Thanks, I see the problem. I will make another PR to fix this. It's yet another problem. I will post an update to #1248.

PR #1357, opened 2023-07-05
PT dynamic_learning_rate support

Judyxujj:
This PR enables the user to define a custom dynamic learning rate function in the returnn config. At each train step, the current step learning rate will be then calculated based on the dynamic learning rate function and updated at each step. 
One example of the dynamic learning rate function would look like this

```
def dynamic_learning_rate(*, global_train_step, learning_rate, **kwargs):
    max_lr = 5e-4
    end_lr = 1e-8
    total_num_updates = 400000 * 3
    num_warm_up = total_num_updates * 0.2
    if global_train_step <= num_warm_up:
        warmup_factor = global_train_step / num_warm_up
        lr = warmup_factor * max_lr
    elif global_train_step > total_num_updates:
        lr = end_lr
    else:
        lr_range = max_lr - end_lr
        pct_remaining = 1 - (global_train_step - num_warm_up) / (
            total_num_updates - num_warm_up
        )
        lr = lr_range * pct_remaining + end_lr
    return lr
```

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -305,8 +305,7 @@ class Engine(EngineBase):
 
             # update the learning rate per step
             if self._use_dynamic_lr:
-                lr = self._updater.get_current_step_learning_rate(self.global_train_step)
-                self._updater.set_learning_rate(lr)
+                self._updater.set_current_step_learning_rate(lr)
 
             # only update the weights when every gradient accumulation loop ends
             if (step_idx % self._accum_grad_multiple_step) == (self._accum_grad_multiple_step - 1):
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -305,7 +305,7 @@ class Engine(EngineBase):
 
             # update the learning rate per step
             if self._use_dynamic_lr:
-                self._updater.set_current_step_learning_rate(lr)
+                self._updater.set_current_step_learning_rate(self.global_train_step)
 
             # only update the weights when every gradient accumulation loop ends
             if (step_idx % self._accum_grad_multiple_step) == (self._accum_grad_multiple_step - 1):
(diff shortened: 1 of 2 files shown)


returnn/torch/engine.py
```diff
@@ -82,6 +82,9 @@ def __init__(self, config: Config):
         self._device = _get_device_from_config(config)
         print("Using device:", self._device, file=log.v2)
 
+        self._use_dynamic_lr = False if self.config.typed_value("dynamic_learning_rate", None) is None else True
+        print("Using dynamic learning rate scheduler that updates based on global train steps", file=log.v2)
```
albertz, line 86: As we discussed, this should not be in the engine, but just in the updater.

returnn/torch/engine.py
```diff
@@ -216,7 +219,9 @@ def train(self):
             self.epoch += 1
             self._epoch_mp_shared.value = self.epoch
 
-            self.init_train_epoch()
+            # update learning rate sub-epoch wise
+            if not self._use_dynamic_lr:
+                self.init_train_epoch()
```
albertz, line 224: As discussed, you should not change this.

returnn/torch/engine.py
```diff
@@ -298,6 +303,10 @@ def train_epoch(self):
                 else:
                     total_loss.raw_tensor.backward()
 
+            # update the learning rate per step
+            if self._use_dynamic_lr:
+                self._updater.set_current_step_learning_rate(self.global_train_step)
```
albertz, line 308: As discussed, you would just unconditionally call some `self._updater.set_current_train_step(self.global_train_step)` or so here.

Judyxujj: I have refactored the code and also renamed the function to set_current_train_step now

returnn/torch/updater.py
```diff
@@ -102,11 +116,14 @@ def set_learning_rate(self, value):
         for param_group in self.optimizer.param_groups:
             param_group["lr"] = value
```
albertz, line 117: Here you also need to assign `set_learning_rate`, such that you can use the correct value in `set_current_train_step`.

albertz (approved):

PR #1359, opened 2023-07-06
fix the is_first_train_epoch check for preload_from_files

Judyxujj:
Now the is_first_train_epoch is assigned to False even at the first training epoch. This PR fix the the check is_first_train_epoch for preload_from_files

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -76,7 +76,7 @@ class ChunkingIterDataPipe(torch.utils.data.IterDataPipe):
     So it transforms one sequences into multiple sequences.
     """
 
-    def __init__(self, dataset: torch.utils.data.IterableDataset, chunking):
+    def __init__(self, dataset: torch.utils.data.IterableDataset, chunking, *, min_chunk_size=0):
         """
         :param dataset: dataset to apply chunking to
         :param None|int|(int,int)|dict|(dict,dict) chunking: tuple (chunk_size, chunk_step).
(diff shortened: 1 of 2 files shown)


albertz:
Unfortunately there seem to be some inconsistency between TF and PT. Currently this line of code is exactly the same as in TF:
```python
is_first_train_epoch = not epoch and (is_training or config.value("task", "train") == "initialize_model")
```
And there, `not epoch` is the correct check, because `epoch` is the epoch of an existing model, or None if there is no existing model and it would start with the first training epoch. If there is an existing model, that could be from the first epoch, then we have `epoch == 1` here. But this means we would load it and then continue with epoch 2. So your code change would not make sense.

In PyTorch, the `epoch` is just the same in case there is an existing model. Then it is the epoch of the existing model. So e.g. if you trained one epoch, and now it finds that existing model, you will get `epoch == 1`, and then you would want to continue with epoch 2 actually, so `is_first_train_epoch` should be False. So your change is wrong here.

In PyTorch, the actual bug is for the case there is no existing model. In that case, we should not yet set `epoch`. So basically:
```python
# Check existing model. This takes `load` and `load_epoch` into account,
# and also whether we are in train or eval mode.
epoch, model_epoch_filename = self.get_epoch_model(self.config)

checkpoint_state = None
if model_epoch_filename:
    filename = model_epoch_filename + util.get_model_filename_postfix()
    checkpoint_state = torch.load(filename, map_location=self._device)
    if epoch is None:
        epoch = checkpoint_state["epoch"]
    step = checkpoint_state["step"]
    print(f"  epoch {epoch}, global train step {step}", file=log.v4)
    # The checkpoint was saved when the step was already increased (but not the epoch yet).
    # Restore the last step.
    # Below, we will increase the step again in case we are training.
    step -= 1
else:
    step = 0
    epoch = self._start_epoch or 1  # do not do this here yet !

is_training = self.config.value("task", "train") == "train"
is_first_train_epoch = not epoch and (is_training or self.config.value("task", "train") == "initialize_model")
```
The `else:` branch should be moved. Similarly to how we have it in case of TF.

So basically like this:
```python
if model_epoch_filename:
    ...
# (no else branch)

is_training = self.config.value("task", "train") == "train"
is_first_train_epoch = not epoch and (is_training or self.config.value("task", "train") == "initialize_model")

if is_first_train_epoch:
    step = 0
    epoch = self._start_epoch or 1
```

albertz:
Why did you close this?

PR #1360, opened 2023-07-10
PT cleanup_old_models support

Atticus1806:
Adapts the cleanup logic from tf and adds it to torch.


returnn/torch/engine.py
```diff
@@ -669,6 +672,139 @@ def forward_with_callback(self, *, dataset: Dataset, callback: ForwardCallbackIf
 
             callback.finish()
 
+    @staticmethod
+    def delete_model(filename):
```
albertz, line 676: This looks like basically the same logic, even almost the same code from the TF engine, copy and pasted here. Why? Why not move it to the engine base class instead, to share the same code for both?

returnn/torch/engine.py
```diff
@@ -706,7 +706,7 @@ def cleanup_old_models(self, ask_for_confirmation=False):
         if hasattr(self, "learning_rate_control"):
             lr_control = self.learning_rate_control
         else:
-            raise NotImplementedError
+            raise load_learning_rate_control_from_config(self.config)
```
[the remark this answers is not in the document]
Atticus1806: whoops that was an overlook yes. Currently testing both tf and torch with the base enginge class, will push update once test succeeded.

returnn/torch/engine.py
```diff
@@ -669,6 +671,28 @@ def forward_with_callback(self, *, dataset: Dataset, callback: ForwardCallbackIf
...
+        :return: accumulated file-size in bytes of deleted files
+        :rtype: int
+        """
+        # This assumes PyTorch models here.
+        # They consist of a file with the extension ".pt".
+        from glob import glob
+
+        count_bytes = 0
+        assert os.path.exists(filename + ".pt")
+        for fn in glob(filename + "*"):
```
albertz, line 687: Why this? This looks potentially dangerous to me. I think we should iterate over exactly the potential available files which we expect.

Also, from the code below, it looks like you simply want anyway this here?
```suggestion
        for fn in glob(filename + "*.pt"):
```

Atticus1806: This was from the tf variant of the function. But I guess there we need to check for multiple different file endings. So if I am not mistaken 
```
            if fn_ext not in [".pt"]:
                continue
```

can just be dropped right?

albertz: Yes that can be dropped. But actually, as I said, I would drop the whole `glob` here. Or why is it necessary?

albertz (approved):

PR #1364, opened 2023-07-17
Check for learning_rate_control is not None during cleanup

Atticus1806:
When using the cleanup old models tool it happens that `self.learning_rate_control` is not set yet. This check then causes proper loading of the lr control for cleanup.

### DIFF
--- a/returnn/engine/base.py
+++ b/returnn/engine/base.py
@@ -281,7 +281,7 @@ class EngineBase:
 
         opts = CollectionReadCheckCovered(self.config.get_of_type("cleanup_old_models", dict, {}))
         existing_models = self.get_existing_models(config=self.config)
-        if hasattr(self, "learning_rate_control"):
+        if hasattr(self, "learning_rate_control") and self.learning_rate_control is not None:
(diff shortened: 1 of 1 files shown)


albertz (approved):

albertz:
I wonder if that code is even up-to-date. The `hasattr` looks bad. Is this really needed? Why?

albertz:
Yes I think this changed at some point. So then remove the `hasattr` as well.

returnn/engine/base.py
```diff
@@ -281,7 +281,7 @@ def cleanup_old_models(self, ask_for_confirmation=False):
 
         opts = CollectionReadCheckCovered(self.config.get_of_type("cleanup_old_models", dict, {}))
         existing_models = self.get_existing_models(config=self.config)
-        if hasattr(self, "learning_rate_control"):
+        if hasattr(self, "learning_rate_control") and self.learning_rate_control is not None:
```
albertz, line 284:
```suggestion
        if self.learning_rate_control is not None:
```

PR #1366, opened 2023-07-17
PyTorch engine: print total number of model params

vieting:
In TF, the total number of model parameters was printed at startup. I think it would be nice to add this for the torch engine, too.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -533,6 +533,8 @@ class Engine(EngineBase):
             raise TypeError(f"get_model returned {model} of type {type(model)}, expected rf.Module or torch.nn.Module")
         assert isinstance(self._pt_model, torch.nn.Module)
         print("Model:", self._pt_model, file=log.v4)
+        params = sum([parameter.data.size().numel() for parameter in self._pt_model.parameters()])
+        print(f"Total number of parameters: {params}", file=log.v4)
 
         if checkpoint_state is not None:
             self._pt_model.load_state_dict(checkpoint_state["model"])
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -533,8 +533,8 @@ class Engine(EngineBase):
             raise TypeError(f"get_model returned {model} of type {type(model)}, expected rf.Module or torch.nn.Module")
         assert isinstance(self._pt_model, torch.nn.Module)
         print("Model:", self._pt_model, file=log.v4)
-        params = sum([parameter.data.size().numel() for parameter in self._pt_model.parameters()])
-        print(f"Total number of parameters: {params}", file=log.v4)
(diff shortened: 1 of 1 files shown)


returnn/torch/engine.py
```diff
@@ -533,6 +533,8 @@ def _load_model(self):
             raise TypeError(f"get_model returned {model} of type {type(model)}, expected rf.Module or torch.nn.Module")
         assert isinstance(self._pt_model, torch.nn.Module)
         print("Model:", self._pt_model, file=log.v4)
+        params = sum([parameter.data.size().numel() for parameter in self._pt_model.parameters()])
+        print(f"Total number of parameters: {params}", file=log.v4)
```
albertz, line 537: The TF engine does this:
```python
print("net params #:", self.get_num_params(), file=log.v2)
print("net trainable params:", self.get_trainable_params(), file=log.v2)
```
So maybe we should make this consistent? E.g. I have a couple of scripts which parse `"net params #:"` from a log.

Maybe the list of trainable params is not so important for now.

Then, there are two issues in your code?

- `.data` is not really needed, right?
- `.size().numel()` is wrong. This gives you the num elements of a shape. E.g. for tensor of shape [2,3,4], the `size` will be the shape, i.e. the vector [2,3,4], i.e. it has 3 elements. You want just `.numel()`.

```suggestion
        num_params = sum([parameter.numel() for parameter in self._pt_model.parameters()])
        print(f"net params #: {params}", file=log.v2)
```

SimBe195:
There are also some external libraries that neatly print information about shape and parameter count of each layer and the whole model such as [this one](https://github.com/TylerYep/torchinfo). What's your opinion on using something like that?

albertz:
> There are also some external libraries that neatly print information about shape and parameter count of each layer and the whole model such as [this one](https://github.com/TylerYep/torchinfo). What's your opinion on using something like that?

First of all, I think this discussion does not belong here but into a separate issue. This discussion here should be solely about this PR.

But then, in general, I don't like adding additional dependencies without very good reason. This does not seem to give too much benefit. Actually I personally would not want this, as it looks way too verbose for my taste. Maybe for debugging, for a one-time call, but not every time.

returnn/torch/engine.py
JackTemaki: While `.numel()` is correct (and desired), `size().numel()` is actually not wrong:

```python
In [1]: import torch

In [2]: lin = torch.nn.Linear(20,40)

In [3]: for p in lin.parameters():
   ...:     print(p.size().numel())
      ...: 
800
40
```

vieting: Doesn't really matter imho, so I just applied the suggestion.

albertz (approved):

PR #1369, opened 2023-07-19
torch onnx export: load model on device

vieting:
When running this script in a CPU-only image, I got
```
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu
') to map your storages to the CPU.
```
even though I specified `--device cpu`. This PR should fix this.


albertz:
Can you resolve the conflict?

albertz:
I wonder, why does this happen? Is there anything CUDA specific in the checkpoint? Why? What is it exactly? Can you post the full error?

albertz (approved):

PR #1393, opened 2023-09-19
Dim eq, batch dim eq only for old behavior version

albertz:
This changes the behavior of equality of batch dim for behavior version >=16 to be more strict, and treat batch dim just like any other dim type, i.e. equality only if it is the same instance or via same_as.

With behavior version <=15, nothing changes.

We could have introduced another behavior version for this change. However, this should not affect much (or only internal things; let's wait for tests) and this keeps it much simpler. **Edit** Unfortunately, too much internal code depends on this and would require more complicated logic in other places, so we don't do this.

### DIFF
--- a/returnn/tensor/_dim_extra.py
+++ b/returnn/tensor/_dim_extra.py
@@ -454,9 +454,7 @@ class _DimMixin:
             # We ignore the ctx for the batch dim currently.
             if self.batch == batch:
                 return self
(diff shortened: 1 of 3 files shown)


albertz:
Unfortunately, too much internal code depends on this and would require more complicated logic in other places, so we don't do this for now.

Some of the other commits are probably still useful, so I moved them to #1395.

PR #1409, opened 2023-10-05
Tensor copy_compatible_to_dims native

albertz:


albertz:
> Now it fails because it triggers issue https://github.com/rwth-i6/returnn/issues/1410.

Fixed via #1411.

PR #1434, opened 2023-10-16
ConvLayer and co: use_time_mask and same_static padding

robin-p-schmitt:
As for other layers, this adds the 'use_time_mask' parameter to the ConvLayer, PoolLayer, TransposedConvLayer, StftLayer and IstftLayer.

Additionally, we add the 'same_static' option for the padding parameter of the ConvLayer and PoolLayer. This padding leads to the same output dims as 'same' padding but changes the left/right distribution of the padding for dynamic axes (i.e. see documentation of ConvLayer). We add this to guarantee that the convolution and pooling operations lead to the same output for single seqs and batched seqs, which previously was not the case.

### DIFF
--- a/returnn/tf/layers/basic.py
+++ b/returnn/tf/layers/basic.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 
 from typing import Optional, Union, Sequence, List, Tuple, Dict
 import typing
+
+import numpy
 import tensorflow as tf
 import contextlib
 import returnn.tf.compat as tf_compat
@@ -6114,6 +6116,7 @@ class ConvLayer(_ConcatInputLayer):
         filter=None,
         filter_perm=None,
         bias=None,
+        use_time_mask=None,
         **kwargs,
     ):
         """
@@ -6121,7 +6124,13 @@ class ConvLayer(_ConcatInputLayer):
             for 1D/2D/3D conv.
             The input data ndim must match, or you can add dimensions via input_expand_dims or input_add_feature_dim.
             It will automatically swap the batch-dim to the first axis of the input data.
-        :param str padding: "same" or "valid"
+        :param str padding: "same", "valid" or "same_static".
+            "same_static" is calculated differently depending on whether an axis is static or dynamic.
+            For static axes, "same_static" padding is the same as "same" padding,
+            i.e. filter_size - 1 - (T + strides - 1) % strides.
+            For dynamic axes, "same_static" calculates the total padding size as
+            filter_size - 1, i.e. it is independent of the length T of the axis and the striding. To avoid skipping
+            any frames on the right, we set left_padding = (filter_size - strides) // 2.
         :param int|Sequence[int] strides: strides for the spatial dims,
             i.e. length of this tuple should be the same as filter_size, or a single int.
         :param int|Sequence[int] dilation_rate: dilation for the spatial dims
@@ -6144,11 +6153,12 @@ class ConvLayer(_ConcatInputLayer):
         :param LayerBase|None filter: if given, will not create an own parameter, but use this as the filter
         :param dict[str,str]|None filter_perm: transposes the filter (input filter as layer)
         :param LayerBase|None bias: if given, will not create an own parameter, but use this as the bias
+        :param bool use_time_mask:
         """
         from returnn.util import BehaviorVersion
 
         padding = padding.upper()
-        assert padding in ["SAME", "VALID"], "no other padding supported at the moment"
+        assert padding in ["SAME", "VALID", "SAME_STATIC"], "no other padding supported at the moment"
         assert "out_type" not in kwargs, "don't set out_type explicitly for this layer"
         assert len(filter_size) in (1, 2, 3), "only 1D conv, 2D conv or 3D conv supported"
         super(ConvLayer, self).__init__(in_dim=in_dim, out_dim=out_dim, **kwargs)
@@ -6175,6 +6185,7 @@ class ConvLayer(_ConcatInputLayer):
             input_expand_dims=input_expand_dims,
             input_split_feature_dim=input_split_feature_dim,
             input_add_feature_dim=input_add_feature_dim,
+            use_time_mask=use_time_mask,
         )
         if self.output.feature_dim_axis == num_batch_dims:
             out_batch_feature_major = True
@@ -6273,7 +6284,19 @@ class ConvLayer(_ConcatInputLayer):
         data_format = None
         if out_batch_feature_major:
             data_format = {1: "NCW", 2: "NCHW", 3: "NCDHW"}[len(filter_size)]
-        x = input_data.placeholder
+
+        if padding == "SAME_STATIC":
+            x = self.get_input_placeholder_with_same_static_padding(
+                input_data=input_data,
+                num_batch_dims=num_batch_dims,
+                filter_size=[size.dimension for size in filter_size],
+                strides=strides,
+                out_batch_feature_major=out_batch_feature_major,
+            )
+            padding = "VALID"  # input is now already "same" padded, therefore use "valid" padding from here
+        else:
+            x = input_data.placeholder
+
         extended_batch_shape = None
         if num_batch_dims > 1:
             x_shape = tf.shape(x)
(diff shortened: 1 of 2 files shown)


returnn/tf/layers/basic.py
```diff
@@ -6,6 +6,8 @@
 
 from typing import Optional, Union, Sequence, List, Tuple, Dict
 import typing
+
```
albertz, line 9: Why the extra added empty line?

returnn/tf/layers/basic.py
```diff
@@ -6114,14 +6116,21 @@ def __init__(
         filter=None,
         filter_perm=None,
         bias=None,
+        use_time_mask=None,
```
albertz, line 6119: Why `None` as the default value? I thought this is supposed to be a bool?

albertz:
```suggestion
        use_time_mask=False,
```

returnn/tf/layers/basic.py
```diff
@@ -6114,14 +6116,21 @@ def __init__(
...
             The input data ndim must match, or you can add dimensions via input_expand_dims or input_add_feature_dim.
             It will automatically swap the batch-dim to the first axis of the input data.
-        :param str padding: "same" or "valid"
+        :param str padding: "same", "valid" or "same_static".
+            "same_static" is calculated differently depending on whether an axis is static or dynamic.
+            For static axes, "same_static" padding is the same as "same" padding,
+            i.e. filter_size - 1 - (T + strides - 1) % strides.
+            For dynamic axes, "same_static" calculates the total padding size as
+            filter_size - 1, i.e. it is independent of the length T of the axis and the striding. To avoid skipping
+            any frames on the right, we set left_padding = (filter_size - strides) // 2.
```
albertz, line 6133: Make proper newlines.
```suggestion
            filter_size - 1, i.e. it is independent of the length T of the axis and the striding.
            For dynamic axes, to avoid skipping any frames on the right,
            we set left_padding = (filter_size - strides) // 2.
```

returnn/tf/layers/basic.py
```diff
@@ -6433,6 +6456,7 @@ def transform_input(
         input_expand_dims=0,
         input_split_feature_dim=None,
         input_add_feature_dim=False,
+        use_time_mask=None,
```
albertz, line 6459: Why None?
```suggestion
        use_time_mask=False,
```

returnn/tf/layers/basic.py
```diff
@@ -6533,8 +6558,73 @@ def transform_input(
...
+    def get_input_placeholder_with_same_static_padding(
+        cls,
+        input_data,
+        num_batch_dims,
+        filter_size,
+        strides,
+        out_batch_feature_major,
+    ):
+        """
+        Returns the placeholder of input_data with same_static padding applied to it.
```
albertz, line 6586: There should always be an empty line after the docstring, before the params.
```suggestion
        Returns the placeholder of input_data with same_static padding applied to it.

```

returnn/tf/layers/basic.py
```diff
@@ -6533,8 +6558,73 @@ def transform_input(
...
         return input_data, num_batch_dims
 
+    @classmethod
+    def get_input_placeholder_with_same_static_padding(
+        cls,
+        input_data,
+        num_batch_dims,
+        filter_size,
+        strides,
+        out_batch_feature_major,
```
albertz, line 6583: Types not specified. I assume like this?
```suggestion
        input_data: Data,
        num_batch_dims: int,
        filter_size: Sequence[int],
        strides: Sequence[int],
        out_batch_feature_major: bool,
```
?

returnn/tf/layers/basic.py
```diff
@@ -6533,8 +6558,73 @@ def transform_input(
...
+    ):
```
albertz, line 6584: Return type:
```suggestion
    )- > tf.Tensor:
```

returnn/tf/layers/basic.py
```diff
@@ -6780,12 +6872,19 @@ def __init__(
         out_dim=None,
         out_spatial_dims=None,
         use_channel_first=NotSpecified,
+        use_time_mask=None,
```
albertz, line 6875:
```suggestion
        use_time_mask=False,
```
?

returnn/tf/layers/basic.py
```diff
@@ -6780,12 +6872,19 @@ def __init__(
...
         :param str mode: "max" or "avg"
         :param tuple[int] pool_size: shape of the window of each reduce
-        :param str padding: "valid" or "same"
+        :param str padding: "same", "valid" or "same_static".
+            "same_static" is calculated differently depending on whether an axis is static or dynamic.
+            For static axes, "same_static" padding is the same as "same" padding,
+            i.e. filter_size - 1 - (T + strides - 1) % strides.
+            For dynamic axes, "same_static" calculates the total padding size as
+            filter_size - 1, i.e. it is independent of the length T of the axis and the striding. To avoid skipping
+            any frames on the right, we set left_padding = (filter_size - strides) // 2.
```
albertz, line 6887: Same as before.

returnn/tf/layers/signal_processing.py
```diff
@@ -765,6 +765,7 @@ def __init__(
         in_spatial_dims=None,
         out_spatial_dims=None,
         out_dim=None,
+        use_time_mask=None,
```
albertz, line 768:
```suggestion
        use_time_mask=False,
```

returnn/tf/layers/signal_processing.py
```diff
@@ -870,6 +876,7 @@ def __init__(
...
```
albertz, line 879:
```suggestion
        use_time_mask=False,
```

returnn/tf/layers/basic.py
albertz:
```suggestion
```

robin-p-schmitt: same here, this shouldn't be there

returnn/tf/layers/basic.py
robin-p-schmitt: i looked at other layers which already had this option (e.g. ReduceLayer) and there the default was also `None`

returnn/tf/layers/basic.py
robin-p-schmitt: ah yes, this is correct

returnn/tf/layers/basic.py
albertz: But there the situation is different. First, they accept the type `Optional[bool]`, which allows None. Sometimes in other cases you might also see `Union[bool, NotSpecified]`. Basically, `None` is treated like `NotSpecified`, meaning that it is not specified. Then, in the code, the default value is not simply a True or False (otherwise this would not be needed), but it depends on the behavior version, and maybe other things.

However, here in your case, we simply have one default value (at least for now), which is simply `False`. None of the logic is needed then. You should only allow the type `bool`.

robin-p-schmitt: Okay, yes, that makes sense.

returnn/tf/layers/basic.py
```diff
@@ -6,6 +6,8 @@
 
 from typing import Optional, Union, Sequence, List, Tuple, Dict
 import typing
+
+import numpy
```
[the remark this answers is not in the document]
albertz: It's still there?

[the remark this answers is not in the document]
albertz: Did you forget to push? Or not done yet?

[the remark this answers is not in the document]
albertz:
```suggestion
```

returnn/tf/layers/basic.py
albertz: But you did not fix this yet?

returnn/tf/layers/basic.py
albertz: Not done yet?

returnn/tf/layers/basic.py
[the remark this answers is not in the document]
robin-p-schmitt: i removed it now

albertz (approved):

PR #1453, opened 2023-11-07
TF: add onnx_export option and onnx_comp_floor_div

### LINKED ISSUE #1328: Onnx: ceildiv yields wrong result during conversion
We had one particular problem when converting a Conformer acoustic model from TensorFlow to ONNX:
Calculating the sequence lengths after convolution resulted in wrong calculations on ONNX side.
The issue is the `ceildiv` operation which is currently defined as:

```
ceildiv(a, b) := -floordiv(-a, b)
```
In order for conversion to work properly, i suggest to change it to:

```
ceildiv'(a, b) := (a + b - 1) / b
```
which gave the correct results. Thoughts?

Gerstenberger:
Add the suggestion by Albert, which resolves #1328 
I'm actually not sure if this is how its supposed to be/correct (at least it works), please let me know.

### DIFF
--- a/returnn/tensor/_dim_extra.py
+++ b/returnn/tensor/_dim_extra.py
@@ -1182,6 +1182,8 @@ class _DimMixin:
             elif kind == "mul":
                 return a * b
             elif kind in ("floordiv", "truediv"):  # truediv assumes there is no remainder
+                if util.is_onnx_export_global():
+                    return tf_util.onnx_compat_floor_div(a, b)
                 return a // b
             elif kind == "ceildiv":
                 if util.is_onnx_export_global():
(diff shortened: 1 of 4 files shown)


returnn/tensor/_dim_extra.py
```diff
@@ -1184,6 +1184,8 @@ def _bin_op_tf(a, b):
             elif kind in ("floordiv", "truediv"):  # truediv assumes there is no remainder
                 return a // b
```
Gerstenberger, line 1185: For completeness, should I also add it here?

albertz: Yes.

albertz (approved):

PR #1460, opened 2023-11-09
use tf.cond instead of tf.control_flow_ops.cond

### LINKED ISSUE #1459: TF 2.14 compatibility issues for tf util `cond()`
In tf util [`cond()`](https://github.com/rwth-i6/returnn/blob/master/returnn/tf/util/basic.py#L3698), we have

```
from tensorflow.python.ops import control_flow_ops

return control_flow_ops.cond(pred, true_fn, false_fn, name=name)
```

With TF 2.14, this seems to face compatibility issues.

```
AttributeError: module 'tensorflow.python.ops.control_flow_ops' has no attribute 'cond'
```

vieting:
Fixes #1459.

### DIFF
--- a/returnn/tf/util/basic.py
+++ b/returnn/tf/util/basic.py
@@ -3697,7 +3697,7 @@ def debug_register_better_repr():
 
 def cond(pred, true_fn, false_fn, name=None):
     """
-    This is a wrapper around tf.control_flow_ops.cond().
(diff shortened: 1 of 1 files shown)


vieting:
The failing pycharm inspection is not caused by the changes in this PR.

albertz (approved):

Issue #1450, opened 2023-10-31
Linear memory growth leads to Freeze of training


Max-Ryujin:
# Discription
This issue appeared while testing `feature_args = {"class": "ScfNetwork", "size_tf": 256 // 2, "stride_tf": 10 // 2}` in a CTC setup. 
An identical setup with `feature_args = {"class": "LogMelNetwork", "wave_norm": True, "frame_size": 200, "frame_shift": 80, "fft_size": 256}`
works fine. The training runs for a few epochs while memory usage climbs linearly until the limit is reached. At this point, the step times grow fast until there is nearly no progress at all. In rare cases, the training crashes with an out-of-memory error. I am using the newest version of returnn. This runs in an apptainer.
The setup is `run_scf_baseline_big` in `/u/maximilian.kannen/setups/20230406_feat/recipe/i6_experiments/users/vieting/experiments/switchboard/ctc/feat/experiments.py`
https://gist.github.com/Max-Ryujin/690647f79773d6cd8338c524be039040

# Relavant Logs

## Example Freeze `/u/maximilian.kannen/setups/20230406_feat/alias/experiments/switchboard/ctc/feat/train_nn/conformer_bs5k_scf_baseline_big/log.run.1`
https://gist.github.com/Max-Ryujin/b2da56c72afaf28850dd7384c97a5b2a
## Example OOM Error `/u/maximilian.kannen/setups/20230406_feat/alias/experiments/switchboard/ctc/feat/train_nn/conformer_bs5k_audio_perturbation_scf_conf-wei-oldspecaug-audio_perturbation_speed0.4_0.8_1.2/log.run.1`
https://gist.github.com/Max-Ryujin/c559f169bc27f8f77e1e39af0146282b

Max-Ryujin:
The returnn config:
https://gist.github.com/Max-Ryujin/f7323769568c9803bd19e3eba623c2d1

albertz:
As far as I see, the only thing a bit uncommon is the `fast_bw` loss here, using Sprint (RASR) in a subprocess to compute the FSA. I wonder if that causes the problems.

albertz:
Can you run this in a memory profiler and report the observations?

albertz:
Also, can you add some more details:

- Since when does this happen?
- What did you change after this happens? TensorFlow version? RETURNN version? Something else?

albertz:
[...]

PR #1468, opened 2023-11-28
ConvLayer: add pad_seq_len_to_power

vieting:
We encountered a problem with linearly growing memory here https://github.com/rwth-i6/returnn/issues/1450. It seems to be cause by a problem in tensorflow https://github.com/tensorflow/tensorflow/issues/62441. Many different input shapes for the convolution lead to a large CPU memory consumption. As a workaround, we can introduce padding in the `ConvLayer` to reduce the number of different sequence lengths/input shapes. This reduces the CPU memory consumption. After the convolution, the padding can be removed again.


returnn/tf/layers/basic.py
```diff
@@ -6361,6 +6378,11 @@ def __init__(
             )
         if num_batch_dims > 1:
             y = tf.reshape(y, tf.concat([extended_batch_shape, tf.shape(y)[1:]], axis=0))
+
+        if pad_seq_len_to_power is not None:
+            for ax in self.output.get_dynamic_axes():
+                y = tf.gather(y, tf.range(tf.reduce_max(self.output.get_dynamic_size(ax))), axis=ax)
```
albertz, line 6384: Use `tf.slice`.

returnn/tf/layers/basic.py
```diff
@@ -6301,6 +6304,20 @@ def __init__(
         else:
             x = input_data.placeholder
 
+        if pad_seq_len_to_power is not None:
+            padding_for_power = []
+            for ax in range(input_data.batch_ndim):
+                if input_data.is_axis_dynamic(ax):
```
albertz, line 6310: In your previous test, I think you only padded the spatial dims, not the batch dim. Is this intended that you also pad the batch dim here?

albertz: Oh you are right. `is_axis_dynamic` is a bit confusing.

albertz: In your test, it was already constant mem consumption (after a while of running) when only doing this?

vieting: I started a training for my real config with this option in the first two layers. However, there are more `ConvLayers`. I could add it to all or I could also create a config for testing e.g. with an adapted `Task12AXDataset`.

returnn/tf/layers/basic.py
```diff
@@ -6158,6 +6159,8 @@ def __init__(
         :param dict[str,str]|None filter_perm: transposes the filter (input filter as layer)
         :param LayerBase|None bias: if given, will not create an own parameter, but use this as the bias
         :param bool use_time_mask:
+        :param Optional[float] pad_seq_len_to_power: pad seq len to power of given number to reduce number of different
+            seq lens. See RETURNN #1450 and https://github.com/tensorflow/tensorflow/issues/62441.
```
albertz, line 6163:
```suggestion
        :param Optional[float] pad_seq_len_to_power: pad seq len to power of given number
            to reduce number of different seq lens.
            See https://github.com/rwth-i6/returnn/issues/1450 and https://github.com/tensorflow/tensorflow/issues/62441.
```

vieting:
Btw, if I see it correctly, the pycharm-inspect test only fails for torch related files which are not affected by this PR.

returnn/tf/layers/basic.py
albertz: Is it needed for all? This was not clear yet, or was it? Or actually I thought in your other network where you only have the other layers, this problem does not occur?

albertz:
> Btw, if I see it correctly, the pycharm-inspect test only fails for torch related files which are not affected by this PR.

It's already fixed, maybe rebase or merge.

returnn/tf/layers/basic.py
vieting: I think the other layers also contribute to the memory increase. However, since they operate on subsampled sequences, the number of different input shape is not so large, so the increase is not as significant as on the waveform.

returnn/tf/layers/basic.py
```diff
@@ -6158,6 +6159,9 @@ def __init__(
         :param dict[str,str]|None filter_perm: transposes the filter (input filter as layer)
         :param LayerBase|None bias: if given, will not create an own parameter, but use this as the bias
         :param bool use_time_mask:
+        :param Optional[float] pad_seq_len_to_power: pad seq len to power of given number to reduce number of different
+            seq lens. See https://github.com/rwth-i6/returnn/issues/1450 and
+            https://github.com/tensorflow/tensorflow/issues/62441.
```
albertz, line 6164: "seq lens" should not be together in the same line with the next sentence, if the next sentence is partially split over again more lines.
```suggestion
        :param Optional[float] pad_seq_len_to_power: pad seq len to power of given number
            to reduce number of different seq lens.
            See https://github.com/rwth-i6/returnn/issues/1450 and
            https://github.com/tensorflow/tensorflow/issues/62441.
```

returnn/tf/layers/basic.py
albertz: Ah ok. So if this is a problem for all ConvLayers, instead of having this a per-layer option, this could also be a global option in the config.

vieting: How would you do that? Can I access the global config from within a layer?

vieting: `self.network._config` might work but is there a nice way to access it?

vieting: Something like `pad_seq_len_to_power = pad_seq_len_to_power or self.network.get_config().float("conv_pad_seq_len_to_power", None)`?

albertz: Yes `self.network.get_config()`. But I think I would also remove the layer option completely then (unless you think it makes sense to also do this per layer).

vieting: It depends a bit on how much this affects the performance I guess. Also, you might want to specify lower values here for initial layers (where sequence lengths are tens of thousands) and higher values for layers after subsampling.

albertz: It's logarithmic, so shouldn't it be the same?
But yes ok, if you think it's useful for you, then add it.

vieting: It only adds minimally to the code complexity, so I think I'd keep it.

vieting:
I compared padding the time dim for the first two layers on the waveform or for all `ConvLayer`s and additionally also padding the batch dim.

image

There is clearly an advantage when padding all `ConvLayer`s, however, padding the batch dim hardly makes any difference. In terms of training speed, padding all layers is ~6.3% slower than just the first two. Padding time + batch is ~10.5% slower than just padding time. The small reduction of memory does not seem to justify the slowdown in case of batch padding imho. If you still want to add it, I'd at least make it configurable.

We can also see that the linear memory growth is not yet fixed. There might be some other layer which leaks memory. However, the padding introduced in this PR largely helps reducing the leak. I'd therefore vote to merge it so it's already available and we might discuss other measures in a separate issue/PR.

albertz (approved):

PR #1481, opened 2023-12-22
torch export to onnx extern data size configurable

Judyxujj:
This PR adds `dyn_dim_min_sizes` and `dyn_dim_max_sizes` to the command line options of `tools/torch_export_to_onnx.py `. In default, extern data with time dimension size in range [2,25] would be generated for export, which might be too short for the cases where raw wave input is used. Thus these two options can be used to specify the min/max sizes for dim tags with dynamic sizes.

### DIFF
--- a/tools/torch_export_to_onnx.py
+++ b/tools/torch_export_to_onnx.py
@@ -182,6 +182,13 @@ def main():
     parser.add_argument("out_onnx_filename", type=str, help="Filename of the final ONNX model.")
     parser.add_argument("--verbosity", default=4, type=int, help="5 for all seqs (default: 4)")
     parser.add_argument("--device", type=str, default="cpu", help="'cpu' (default) or 'gpu'.")
+    parser.add_argument(
+        "--dyn_dim_min_sizes", type=dict, default=None, help="Specify min sizes for dim tags with dynamic sizes"
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


tools/torch_export_to_onnx.py
```diff
@@ -182,6 +182,13 @@ def main():
     parser.add_argument("out_onnx_filename", type=str, help="Filename of the final ONNX model.")
     parser.add_argument("--verbosity", default=4, type=int, help="5 for all seqs (default: 4)")
     parser.add_argument("--device", type=str, default="cpu", help="'cpu' (default) or 'gpu'.")
+    parser.add_argument(
+        "--dyn_dim_min_sizes", type=dict, default=None, help="Specify min sizes for dim tags with dynamic sizes"
```
albertz, line 186: Did you test this? What exactly did you pass as argument? I wonder how it would correctly convert it into a `dict`, and specifically with the right `Dim` objects as keys?

Judyxujj: I tested it by setting the `Dim` object like this
```
time_data_dim = SpatialDim('time:data')
extern_data_dict = {'data': {
        'dim_tags': [
            batch_dim,
            time_data_dim,
            FeatureDim('feat', 1),
        ],
        'dtype': 'float32',
    }}
dyn_dim_max_sizes={time_data_dim:100}
```

albertz: I mean, how did you start this script? How did you pass `--dyn_dim_min_sizes` when starting the script?

PR #1501, opened 2024-01-22
Check for torch version before average

Atticus1806:
Currently the tool uses the flag `mmap` which was only introduced in torch 2.1.0. 
This change makes the tool also useable with versions below 2.1.0.


tools/torch_avg_checkpoints.py
```diff
@@ -67,7 +67,12 @@ def merge_checkpoints(in_ckpts: Sequence[str], out_ckpt: str, extra_state: Optio
     out_state: Dict[str, Any] = {"model": out_model_state, "merged_epochs": [], "merged_steps": []}
     for in_ckpt in in_ckpts:
         print("read ckpt:", in_ckpt)
-        in_state = torch.load(in_ckpt, map_location=torch.device("cpu"), mmap=True)
+        torch_version = torch.__version__.split(".")
```
albertz, line 70:
```suggestion
        torch_version = tuple(int(s) for s in str(torch.__version__).split(".")[:2])
```
Or use some of the `packaging.version` or so.

albertz (approved):

PR #1510, opened 2024-03-27
Remove additional empty string before special tokens when parsing orthography

AtanasGruev:
In `returnn/tools/collect-words.py`, orthography is parsed via the `parse_orthography_into_symbols()` function. Specifically, when setting `word_based=True`, an empty string `''` is inserted prior to each special token, leading to incorrect parsing and biased counts. This PR fixes the mismatch.


albertz (approved):

albertz:
(Btw, for a PR, better make an own branch. Either in your fork or directly here in the repo.)

PR #1525, opened 2024-06-05
Make `ConcatFilesDataset` compatible with multimodal data

### LINKED ISSUE #1524: `ConcatFilesDataset` combines poorly with `MetaDataset`
I'm working on running a test config using `ConcatFilesDataset` (#1521, #1519).

The original setup I have uses `MetaDataset` to load the features and the alignment targets from two distinct sets of HDFs. This makes integrating the `ConcatFilesDataset` pretty difficult because it cannot deal with heterogenous data. Take the following example:

```python
train = {
    "class": "MetaDataset",
    "data_map": {"classes": ("alignments", "data"), "data": ("features", "data")},
    "datasets": {
        "alignments": {
            "class": "HDFDataset",
            "files": [
                "/alignment/files.hdf",
                "/..."
            ],
            "partition_epoch": 250,
            "seq_ordering": "random",
        },
        "features": {
            "class": "HDFDataset",
            "files": [
                "/feature/files.hdf",
                "/..."
            ],
        },
    },
    "seq_order_control_dataset": "alignments",
}
```

Where do we integrate the `ConcatFilesDataset`?

Thoughts:
1. We can place `ConcatFilesDataset` around the `MetaDataset`. This is problematic because we a) cannot give both the alignment caches and the feature caches to the `ConcatFilesDataset`'s `files` list, as they are fundamentally different in size and heterogenous in contents. We b) could give the `ConcatFilesDataset` only the features and always include all alignment caches in the `MetaDataset` emitted by `get_sub_epoch_dataset`. This would work but would be slow as we'd have to reload all alignment caches for every subepoch. We could also c) try and compute the relevant alignment caches given a list of feature caches in `get_sub_epoch_dataset`, but I'm not sure we can assume this always works. It might be worth precomputing this info though.
2. We can place `ConcatFilesDataset` inside the `MetaDataset` where the `HDFDataset`s currently are. This works as long as only one of the datasets is a `ConcatFilesDataset`, because it is the one that needs to be the `seq_order_control_dataset` (it does not accept a `seq_order` from outside https://github.com/rwth-i6/returnn/commit/02fa44b986042c7dbb6baa6a42a06205c2acc932#diff-6f16d6edae7b1113b8c292acded8b6ab78875f54dca08cd58edd11110709372fR189-R190). This means we can only use `ConcatFilesDataset` for one of the data streams, but the advantage over option 1b) is that we don't need to reload the HDFs for the other data stream after every subepoch. This might be enough if only one stream is data-heavy (features) and the other ones are small enough they can be loaded the normal way (targets). If we want both datasets to be `ConcatFilesDataset` to leverage the advanced prefetching and caching behavior we can precompute a mapping between sequence ID <-> containing file path. Then one of the datasets can be the `seq_order_control_dataset` and the others use that mapping to load up the data files and set up the sub dataset for precisely the relevant segments on-demand.
3. There might be issues with `partition_epoch` being specified on both the parent `MetaDataset` and on the `ConcatFilesDataset` but I haven't put much thought into that yet.

WDYT? Is there anything I'm missing?

NeoLegends:
Closes #1524

### DIFF
--- a/returnn/datasets/concat_files.py
+++ b/returnn/datasets/concat_files.py
@@ -25,7 +25,8 @@ _mp = NonDaemonicSpawnContext(process_pre_init_func=SubProcCopyGlobalConfigPreIn
 
 __all__ = ["ConcatFilesDataset"]
 
-FileTree = Union[str, Tuple["FileTree", ...], Dict[str, "FileTree"], List["FileTree"]]
+Filename = str
+FileTree = Union[Filename, Tuple["FileTree", ...], Dict[Any, "FileTree"], List["FileTree"]]
 
 
 class ConcatFilesDataset(CachedDataset2):
--- a/returnn/datasets/concat_files.py
+++ b/returnn/datasets/concat_files.py
@@ -395,7 +395,6 @@ class ConcatFilesDataset(CachedDataset2):
         return self._data_keys
 
 
-@staticmethod
 def _get_key_for_file_tree(t: FileTree) -> str:
     """generates a deterministic key given a file tree"""
     import tree
--- a/returnn/datasets/concat_files.py
+++ b/returnn/datasets/concat_files.py
@@ -13,7 +13,6 @@ import numpy
 from returnn.util.basic import try_run
 from returnn.util.multi_proc_non_daemonic_spawn import NonDaemonicSpawnContext
 from returnn.config import SubProcCopyGlobalConfigPreInitFunc
-import tree
 from .basic import init_dataset, DatasetSeq
 from .cached2 import CachedDataset2
 
@@ -199,6 +198,8 @@ class ConcatFilesDataset(CachedDataset2):
             exit_hook()
 
     def _lazy_init_file_sizes(self):
+        import tree
+
         if self._file_sizes:
             return
         self._file_sizes = {
@@ -359,6 +360,7 @@ class ConcatFilesDataset(CachedDataset2):
     @staticmethod
     def _get_key_for_file_tree(t: FileTree) -> str:
         """generates a deterministic key given a file tree"""
+        import tree
         return ":".join(tree.flatten(t))
 
     def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
--- a/returnn/datasets/concat_files.py
+++ b/returnn/datasets/concat_files.py
@@ -204,8 +204,7 @@ class ConcatFilesDataset(CachedDataset2):
         if self._file_sizes:
             return
         self._file_sizes = {
-            ConcatFilesDataset._get_key_for_file_tree(t): sum((os.path.getsize(fn) for fn in tree.flatten(t)), 0)
-            for t in self.files
+            _get_key_for_file_tree(t): sum((os.path.getsize(fn) for fn in tree.flatten(t)), 0) for t in self.files
         }
 
     def _lazy_init_file_cache_proc(self):
@@ -323,7 +322,7 @@ class ConcatFilesDataset(CachedDataset2):
         sub_epoch_idx = 0
         size_taken = 0
         for i, f_tree in enumerate(files_order):
-            size = file_sizes[ConcatFilesDataset._get_key_for_file_tree(f_tree)]
+            size = file_sizes[_get_key_for_file_tree(f_tree)]
             num_remaining = len(files_order) - i
             if num_remaining <= partition_epoch - sub_epoch_idx - 1:
                 # All remaining sub epochs must be filled.
@@ -357,13 +356,6 @@ class ConcatFilesDataset(CachedDataset2):
             sub_epoch_idx += 1
         assert all(files_per_sub_epochs)
         return files_per_sub_epochs
-
-    @staticmethod
-    def _get_key_for_file_tree(t: FileTree) -> str:
-        """generates a deterministic key given a file tree"""
-        import tree
-
-        return ":".join(tree.flatten(t))
 
     def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
         if seq_idx >= self._num_seqs:
@@ -400,6 +392,14 @@ class ConcatFilesDataset(CachedDataset2):
         if self._data_keys is None:
             self._lazy_init_num_outputs()
         return self._data_keys
+
+
+@staticmethod
+def _get_key_for_file_tree(t: FileTree) -> str:
+    """generates a deterministic key given a file tree"""
+    import tree
+
+    return ":".join(tree.flatten(t))
 
 
 class _WorkerProcParent:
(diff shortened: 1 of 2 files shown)


returnn/datasets/concat_files.py
```diff
@@ -24,6 +24,8 @@
 
 __all__ = ["ConcatFilesDataset"]
 
+FileTree = Union[str, Tuple["FileTree", ...], Dict[str, "FileTree"], List["FileTree"]]
```
albertz, line 27:
```suggestion
Filename = str
FileTree = Union[Filename, Tuple["FileTree", ...], Dict[Any, "FileTree"], List["FileTree"]]
```

returnn/datasets/concat_files.py
```diff
@@ -103,13 +143,13 @@ def __init__(
         self.preload_next_n_sub_epochs = preload_next_n_sub_epochs
         self.buffer_size = buffer_size
         self.file_cache_opts = file_cache_opts or {}
-        self._file_sizes: Optional[Dict[str, int]] = None  # file -> size. for equal distribution across sub epochs
+        self._file_sizes: Optional[Dict[str, int]] = None  # key -> size. for equal distribution across sub epochs
```
albertz, line 146: Maybe use the index (`int`) in the `files` list instead?
```suggestion
        self._file_sizes: Optional[Dict[int, int]] = None  # file index -> size. for equal distribution across sub epochs
```

returnn/datasets/concat_files.py
```diff
@@ -159,7 +199,9 @@ def _lazy_init_num_outputs(self):
     def _lazy_init_file_sizes(self):
         if self._file_sizes:
             return
-        self._file_sizes = {fn: os.path.getsize(fn) for fn in self.files}
+
+        import tree
+        self._file_sizes = {":".join(tree.flatten(t)): sum((os.path.getsize(fn) for fn in tree.flatten(t)), 0) for t in self.files}
```
albertz, line 204: I think we should not use such ad hoc key format.
```suggestion
        self._file_sizes = [sum((os.path.getsize(fn) for fn in tree.flatten(t)), 0) for t in self.files]
```

returnn/datasets/concat_files.py
NeoLegends: I was doing this but then we have to be really careful with shuffling the files list and making sure we don‘t mess up any indices. The ad hoc key circumvents this.

albertz: Hm right. Ok, as you prefer. But maybe make a separate function for it (`_key_for_files` or so).

NeoLegends:
I'll make this ready for reviews when I'm ready for further reviews, want to make tests green etc first

returnn/datasets/concat_files.py
```diff
@@ -13,6 +13,7 @@
 from returnn.util.basic import try_run
 from returnn.util.multi_proc_non_daemonic_spawn import NonDaemonicSpawnContext
 from returnn.config import SubProcCopyGlobalConfigPreInitFunc
+import tree
```
albertz, line 16: I would not put the import as a global import, as I'm not sure all people have installed this. Actually I guess many people have not installed it. It was only recently introduced to RETURNN. So it would break quite a lot of setups. I think it should only be imported when people make use of this dataset.

NeoLegends: Oh, hm, I see.

returnn/datasets/concat_files.py
```diff
@@ -354,6 +355,11 @@ def _get_files_per_sub_epochs(
         assert all(files_per_sub_epochs)
         return files_per_sub_epochs
 
+    @staticmethod
+    def _get_key_for_file_tree(t: FileTree) -> str:
```
albertz, line 361: I think you can move that to the global scope, instead of making a static method here. The whole file here (`concat_files.py`) is only about the dataset. That makes the code a bit shorter. Also, other things here are also moved out of the class.

NeoLegends: Hm, I'm not sure. I think the key format should remain specific to the dataset and be un-generic. I don't want other parts of the code to start relying on the concrete representation this function produces. If this comes up again we can always move it somewhere global.

albertz: I mean also as a hidden/private function (`_get_key_for_file_tree`), i.e. that no other parts of the code will use this.

returnn/datasets/concat_files.py
```diff
@@ -75,22 +78,61 @@ def get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+            ...
+          ],  # M entries
+          "get_sub_epoch_dataset": get_sub_epoch_dataset,
+          "partition_epoch": P,  # P << M
+        }
+
+    In this case the file sizes for sub epoch distribution are summed up per list entry
+    by iterating over the structure leaves.
+
     For some discussion, see https://github.com/rwth-i6/returnn/issues/1519.
```
albertz, line 119: Maybe also extend this by the new issue.

tests/test_Dataset.py
```diff
@@ -623,25 +623,31 @@ def __getitem__(self, item):
     assert res.features["data"].shape == (5, 3)
 
 
+NestedSize = Union[int, List["NestedSize"]]
```
albertz, line 626: I don't understand why you need this. You will never operate on nested sizes. We always take the sum over all files in case there is some nested structure, i.e. always a single `int`.

NeoLegends: The `_test` function gets this as input though?

NeoLegends: Sure we balance based on a single int but we should make sure the nested data is passed correctly into `get_subepoch_dataset`.

albertz: But in `_get_files_per_sub_epochs`, the `file_sizes` is not nested, it is just a single `int`. And this test here is specifically about `_get_files_per_sub_epochs`, nothing else. I think we should not make this test more complicated.

If you want to test the nested structure in some way, I would do that in a separate test. But I'm not sure if this is really needed, to have an own test for this. I don't see much potential that this can break. I usually introduce tests in case I feel that I can easily introduce a bug at some point to it, or just to easier develop it and test it directly within the test.

tests/test_Dataset.py
```diff
@@ -643,6 +643,27 @@ def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]]):
     _test([5, 5] + [10] * 7, 5, [[5, 5, 10], [10, 10], [10, 10], [10], [10]])
 
 
+NestedSize = Union[int, List["NestedSize"]]
+
+
+def test_ConcatFilesDataset_get_files_per_sub_epochs_nested():
```
albertz, line 649: I don't understand this test. `_get_files_per_sub_epochs` is not really about the nested files at all.

albertz: I don't think we need to have this test case. If you want to have a test case for the nested aspect, this would not be about `_get_files_per_sub_epochs`.

PR #1527, opened 2024-06-06
Print out list of files selected per subepoch

NeoLegends:


NeoLegends:
ah wrong commits in here

albertz:
> ah wrong commits in here

Why is this a reason to close the PR? Just remove the commits?

albertz:
But now that you opened #1528, I guess we can leave this closed. But next time, just modify the PR instead of opening a new one.

PR #1530, opened 2024-06-07
ConcatFilesDataset (DistributeFilesDataset): restart dataset worker proc once on error

NeoLegends:
just trying this out for debugging, not intending to merge this if this does not make a difference, no reviews needed for now


NeoLegends:
Does not work

Issue #1519, opened 2024-05-24
Support for larger scale datasets


albertz:
Most of our existing datasets have issues loading larger scale datasets. E.g. `HDFDataset`, `OggZipDataset` or others just take too long to load for really large data. The next problem is also the shuffling/sorting logic in those datasets. Even random shuffling is done offline for the whole dataset at once, and not on-the-fly. But when partially sorting (e.g. laplace order), it's even much slower, because that also needs to go through the whole dataset to check seq lens.

Related:

* I think this can all be solved on the dataset level, independent from the backend engine (TF, PyTorch, or whatever). However, conceptually, many relevant concepts and ideas has been discussed in this issue, but specific to TF, but we can adopt then and implement them framework-independently: https://github.com/rwth-i6/returnn/issues/292
* Another related idea is the ESPnet-style dataset shuffling/sorting: https://github.com/rwth-i6/returnn/issues/1506
* Some discussion on dataset alternatives, although those discussions were a bit off-topic from the actual reported issue on supporting arbitrary data in the `HDFDataset`: https://github.com/rwth-i6/returnn/issues/331
* Similarly, in this issue on `OggZipDataset` being too slow, there have been some discussion on dataset alternatives: https://github.com/rwth-i6/returnn/issues/1256
* The logic/code in `MultiProcDataset` could be relevant for some of the ideas here, to load things in the background.
* Dataset on-the-fly post processing (e.g. speed perturbation): https://github.com/rwth-i6/returnn/issues/1505

Some ideas:

* On-the-fly laplace: If data loading and random shuffling is still feasible, but just doing laplace sorting then makes it too slow:
  * We can make another `OnTheFlyBucketingBySeqLenDataset` or so, which acts like `MetaDataset` or `MultiProcDataset`, i.e. it gets some base dataset as an option. In that base dataset, it would use random shuffling. Then there would be some pre-defined seq length buckets, and they get filled from the data from the base dataset, and once some bucket is full, it can flush it. That's exactly like `bucket_by_sequence_length` in TF, but I think I have seen the same logic also elsewhere. This is usually to directly emit the mini batches. Maybe instead of implementing this in a separate dataset, it could also be implemented as a new way for our mini-batch creation logic. (Although the minibatch creation logic is currently framework specific in RETURNN. But it doesn't need to be that way.)
[...]

PR #1532, opened 2024-06-10
HDFDataset, use bisect in _get_file_index()

Icemole:
Helps in managing the training of big datasets (#1519).

Tests in a big dataset (1000 - 2000 HDF files, ~28 M segments) show that the time taken went from 50 minutes down to 15 minutes (approximately).

This code has been previously tested with an assertion that all file indices returned by the current implementation should be equal to those returned by the previous implementation. Either way, I think the implementation makes enough sense by itself.

As per `numpy.searchsorted` vs `bisect.bisect`, I looked into this, and it seems that `searchsorted` fails to give the best performance in single searches nested in a loop (as shown [here](https://stackoverflow.com/questions/62753467/is-python-bisect-faster-than-numpy-searchsorted)), which is pretty much the current use case. Either way, I assume both approaches would be fine since they would yield the same temporal cost.

Note that the current temporal cost of `get_file_index` is `O(log(num_hdf_files))`. This could in theory be reduced to `O(1)` by creating a dense dictionary with all real sequence indices pointing to their corresponding file indices. However, this would imply a spatial cost of `O(num_seqs)` dumped into a single variable, which in our case would definitely be non-negligible (28+ MB for a single variable).

PS: I tried implementing the `O(1)` temporal cost approach but so far I haven't succeeded. I only invested a short amount of time on this issue so I can look into it for longer if you really want, but I find that the speedup that this implementation provides with minimal changes is really valuable.

### DIFF
--- a/returnn/datasets/hdf.py
+++ b/returnn/datasets/hdf.py
@@ -457,10 +457,10 @@ class HDFDataset(CachedDataset):
         return ", ".join(["HDF dataset", "sequences: %i" % self.num_seqs, "frames: %i" % self.get_num_timesteps()])
 
     def _get_file_index(self, real_seq_idx):
-        file_index = 0
-        while file_index < len(self.file_start) - 1 and real_seq_idx >= self.file_start[file_index + 1]:
-            file_index += 1
-        return file_index
+        # bisect() returns the position for which all elements to the left of the returned index are <= real_seq_idx,
+        # so it actually returns the next file index in which the sequence can be found.
+        # Therefore, we subtract 1 to the index provided.
+        return bisect.bisect(self.file_start, real_seq_idx) - 1
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


returnn/datasets/hdf.py
```diff
@@ -457,10 +457,10 @@ def len_info(self):
...
 
     def _get_file_index(self, real_seq_idx):
-        file_index = 0
-        while file_index < len(self.file_start) - 1 and real_seq_idx >= self.file_start[file_index + 1]:
-            file_index += 1
-        return file_index
+        # bisect() returns the position for which all elements to the left of the returned index are <= real_seq_idx,
+        # so it actually returns the next file index in which the sequence can be found.
+        # Therefore, we subtract 1 to the index provided.
+        return bisect.bisect(self.file_start, real_seq_idx) - 1
```
albertz, line 463: `import bisect` missing?

albertz (approved):

albertz:
Note that the real solution is to just use `ConcatFilesDataset`.

PR #1533, opened 2024-06-10
Pass random_seed_offset via env var

NeoLegends:
Also correctly forward random_seed_offset in MetaDataset. Note this PR currently does not implement setting the environment variable _within_ one RETURNN process, the environment variable is currently only ever set for child processes spawned for dataset handling.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -28,6 +28,9 @@ from returnn.util.basic import try_run, NumbersDict, OptionalNotImplementedError
 from returnn.tensor import TensorDict
 
 
+RANDOM_SEED_OFFSET_ENV_VAR = "RETURNN_RANDOM_SEED_OFFSET"
+
+
 class Dataset(object):
     """
     Base class for any dataset. This defines the dataset API.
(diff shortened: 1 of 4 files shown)


returnn/datasets/basic.py
```diff
@@ -245,6 +248,13 @@ def _get_default_random_seed_offset():
         config = get_global_config(raise_exception=False)
         if not config:
             return 0
+
+        env_val = os.environ.get(RANDOM_SEED_OFFSET_ENV_VAR)
+        if env_val is not None:
+            try:
+                return int(env_val)
+            except ValueError:
+                pass
```
albertz, line 257: If it is set (non empty string), I think it should not silently ignore invalid values but raise an exception instead. I.e. simply like this:
```suggestion
        if env_val:
            return int(env_val)
```

NeoLegends: ah, fair point. will do

returnn/util/basic.py
```diff
@@ -4519,3 +4519,22 @@ def find_libcudart_from_runtime():
...
     _find_libcudart_from_runtime_cached = [None]
     return None
+
+
+@contextlib.contextmanager
+def override_env_var(var_name: str, value: str):
+    """
+    context manager for temporarily overriding the value of an env var
+    :param str var_name: the name of the environment variable to override
+    :param str value: the value to set while the context mgr is active
```
albertz, line 4529:
```suggestion
    context manager for temporarily overriding the value of an env var
    
    :param var_name: the name of the environment variable to override
    :param value: the value to set while the context mgr is active
```

NeoLegends:
Using this PR I can get a Horovod training to run.

Issue #1518, opened 2024-05-24
RF weight dropout and variational noise


albertz:
Currently we don't have weight dropout in the RF. We should add it.

(I thought there was an issue already about it but I don't find it.)

Related:

* Weight dropout in RETURNN common: `returnn_common.nn.utils.weight_dropout.weight_dropout`. Related issue: https://github.com/rwth-i6/returnn_common/issues/59, https://github.com/rwth-i6/returnn_common/issues/250
* Weight dropout for TF net-dict. See: https://github.com/rwth-i6/returnn/issues/735
* Related discussion (on weight norm): https://github.com/rwth-i6/returnn/issues/1264#issuecomment-1446276707

In general, there is a whole class of similar features on parameter reparameterizations, which would require a similar mechanism, like weight norm.

Regarding implementation:

I think we could follow the PyTorch implementation of similar logic (e.g. weight norm) by using a forward-pre-hook. We already have support for hooks via `rf.Module.register_forward_hook`/`rf.hooks.setup_post_hook_on_method`.

Regarding PyTorch:

* [`torch.nn.utils.weight_norm`](https://pytorch.org/docs/stable/generated/torch.nn.utils.weight_norm.html) uses `register_forward_pre_hook`. However, the doc says:
  > This function is deprecated. Use [torch.nn.utils.parametrizations.weight_norm()](https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.weight_norm.html#torch.nn.utils.parametrizations.weight_norm) which uses the modern parametrization API.
* [`torch.nn.utils.parametrizations.weight_norm`](https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.weight_norm.html#torch.nn.utils.parametrizations.weight_norm) uses the modern parametrization API, i.e. [`torch.nn.utils.parametrize.register_parametrization`](https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrize.register_parametrization.html#torch.nn.utils.parametrize.register_parametrization). On `register_parametrization` (it's somewhat ugly (although our own hooks mechanism is also a bit complicated), but this here looks even worse):
  * It uses `_inject_new_class` to replace the original module class by a new dummy one, which is going to be extended by adding a property for the parameter.
    (We cannot add the property to the object. A property can only be added to a class. See [the descriptor guide doc](https://docs.python.org/3/howto/descriptor.html).)
[...]

PR #1536, opened 2024-06-10
Gradient checkpointing experiments

NeoLegends:
I'm using this PR to run the CI. Albert don't bother reviewing yet.


returnn/frontend/_backend.py
```diff
@@ -1293,6 +1293,22 @@ def tensor_array_get_item(cls, tensor_array: TensorArrayType, index: Union[int,
             return tensor_array[index]
         raise NotImplementedError
 
+    @classmethod
+    def gradient_checkpoint(cls, fn: Callable[[Tensor], Tensor], *args: List[Tensor]) -> Tensor:
```
albertz, line 1297: Is this exactly like `recompute_grad`? I think i would rather then call it `recompute_grad`.

returnn/frontend/linear.py
```diff
@@ -40,6 +41,43 @@ def __call__(self, source: Tensor) -> Tensor:
         return out
 
 
+class LinearWithWeightDropout(rf.Module):
```
albertz, line 44: Why do you add this? We don't want to have specialized variants for every single existing module for every single variant of reparameterization (e.g. weight dropout, variational noise, weight normalization, ...). See #1518.

returnn/frontend/_backend.py
albertz, line 1297: I think the typing is wrong in multiple ways. If you accept multiple args for `fn`, shouldn't it be `Callable[[Tensor, ...], Tensor]` or so? And also, I think the type for `args` is just `Tensor` (I think you specify the type of individual elements).

tests/test_rf_base.py
```diff
@@ -231,6 +231,32 @@ def _forward_step(*, model: _Net, extern_data: TensorDict):
...
+
+        def __call__(self, x: Tensor) -> Tensor:
+            return self.linear(x)
+
+    # noinspection PyShadowingNames
+    def _forward_step(*, model: _Net, extern_data: TensorDict):
+        out = model(extern_data["data"])
+        out.mark_as_default_output()
+
+    run_model(extern_data, lambda *, epoch, step: _Net(), _forward_step)
```
albertz, line 257: Note that currently, by default, we test the TF-layers backend and the PyTorch backend. But you only have implemented the direct TF backend. I'm not saying that you also must implement the TF-layers backend, just that you maybe want to disable the TF-layers backend here, and maybe instead enable the direct TF backend.

albertz, line 257: Also note, this test will actually not test at all the weight dropout. It will operate in forward mode, i.e. disable any dropout. But even if you enable training, what you also would want to check is whether it did the grad checkpointing actually correctly. Although I have no idea how to test this. In TF, I have similar tests by really investigating the computation graph. Not sure how you can investigate the gradient tape in PT. Or how you can verify in any way that the gradient checkpointing worked.

tests/test_rf_base.py
```diff
@@ -231,6 +231,32 @@ def _forward_step(*, model: _Net, extern_data: TensorDict):
     run_model(extern_data, lambda *, epoch, step: _Net(), _forward_step)
 
 
+def test_linear():
```
albertz, line 234: The name of the test is misleading. This is not about testing `linear`. I guess you want to test weight dropout and/or gradient checkpointing here.

returnn/frontend/ckpt.py
```diff
@@ -0,0 +1,31 @@
+"""
```
albertz, line 1: The file name is misleading. `ckpt` is usually for (model) checkpoints.

In TF, there is a file `custom_gradient.py`, where you find functions like `custom_gradient` and `recompute_grad`. We actually already have `returnn/frontend/gradient.py` where you also find some similar related functions. I think the best fit would be there.

returnn/frontend/ckpt.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import annotations
+from typing import Callable, TypeVar
+
+__all__ = ["gradient_checkpoint"]
+
+
+T = TypeVar("T")
+
+
+def gradient_checkpoint(fn: Callable[..., T], *args: ...) -> T:
```
albertz, line 14: I think `recompute_grad` is a better name.

tests/test_rf_base.py
albertz, line 234: If you only want to test gradient checkpointing, `test_rf_gradient.py` is a better place for the test.

albertz:
> I'm using this PR to run the CI.

It's fine that you put intermediate work into a draft PR. But you don't need to use the CI to test things. You can simply run the relevant tests locally. I usually run the relevant tests inside a debugger (e.g. PyCharm) so I can directly see potential issues. It's actually a nice way to debug and develop, to write the test case and then run it in the debugger.

albertz:
Hey, what's the state here? Did you test this? Does it work? I.e. it doesn't store the intermediate activations in memory?

NeoLegends:
Hey, no new results yet.

albertz:
> Hey, no new results yet.

But what was the state? What were the old results? Did it work? Does it not store the intermediate activations in memory anymore? Or you don't know?

albertz:
Also, I don't exactly understand how this API here could be used for variational noise and/or weight dropout. Can you give an example?

NeoLegends:
Superseeded by #1559

PR #1538, opened 2024-06-11
`DistributeFilesDataset`: allow sharding files across GPU workers

### LINKED ISSUE #1531: `DistributeFilesDataset` with sharding on file level
This is for multi GPU training.

I think the random_seed_offset mechanism should be fine once you train multiple full epochs per each GPU worker.

However, when the data is so large that a single GPU worker will never see the whole data once, i.e. does never finish one full epoch, it might make sense to implement sharding on file-level.

It could look like this: At the beginning of a full epoch: when we shuffle the files (logic we currently have in `init_seq_order`: `random_generator.shuffle(files_order_flat)`), we would still use the same random seed for every GPU worker. Then we distribute the files for the GPU workers. This could actually use the same `_get_files_per_sub_epochs` which we use for the sub epochs, but now distributing over the workers, not the sub epochs. Then you pick the right part for the current GPU worker, and apply the remaining logic as before, i.e. now `_get_files_per_sub_epochs` for sub epochs. Whether the sub datasets now use the same random seed for all workers or a different random seed does not really matter so much, as they anyway all use different data.

NeoLegends:
Closes #1531.

See PR #1521 for the initial implementation, and issue #1519 for the initial discussion on `DistributeFilesDataset`.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -144,9 +144,7 @@ class Dataset(object):
         self.window = window
         self.seq_ordering = seq_ordering  # "default", "sorted" or "random". See self.get_seq_order_for_epoch().
         self.fixed_random_seed = fixed_random_seed
-        if random_seed_offset is None:
-            random_seed_offset = self._get_default_random_seed_offset()
-        self.random_seed_offset = random_seed_offset
+        self._random_seed_offset = random_seed_offset
         self.partition_epoch = partition_epoch or 1
         self.repeat_epoch = repeat_epoch or 1
         self._seq_list_filter_file = seq_list_filter_file
@@ -241,6 +239,12 @@ class Dataset(object):
 
         state = {attr: getattr(self, attr) for attr in ["epoch", "zpad"]}
         return Dataset._create_from_reduce, (self.__class__, kwargs, state)
+
+    @property
+    def random_seed_offset(self) -> int:
+        if self._random_seed_offset is None:
+            self._random_seed_offset = self._get_default_random_seed_offset()
+        return self._random_seed_offset
 
     def _uses_custom_distributed_sharding(self):
         """override if the dataset has its own sharding logic independent of TF/PT"""
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -246,7 +246,7 @@ class Dataset(object):
             self._random_seed_offset = self._get_default_random_seed_offset()
         return self._random_seed_offset
 
-    def _uses_custom_distributed_sharding(self):
+    def _uses_custom_distributed_sharding(self) -> bool:
         """override if the dataset has its own sharding logic independent of TF/PT"""
         return False
 
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -242,6 +242,7 @@ class Dataset(object):
 
     @property
     def random_seed_offset(self) -> int:
+        """:return: random seed offset for shuffling"""
         if self._random_seed_offset is None:
             self._random_seed_offset = self._get_default_random_seed_offset()
         return self._random_seed_offset
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -621,10 +621,12 @@ class Dataset(object):
 
         return seq_index
 
-    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1):
+    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1, include_random_seed_offset=True):
         """
         :param int|None epoch:
         :param int num_epochs_fixed: keep random seed fixed for n subsequent epochs
+        :param bool include_random_seed_offset: whether to add in the local random_seed_offset.
+            Can be disabled to get deterministic seeds in multi-process settings.
         :rtype: int
         """
         if self.fixed_random_seed is not None:
@@ -635,7 +637,7 @@ class Dataset(object):
             seed = (seed - 1) // partition_epoch + 1  # taking partitions requires constant seed during full epoch
         if num_epochs_fixed > 1:
             seed = (seed - 1) // num_epochs_fixed + 1
-        return seed + self.random_seed_offset
+        return seed + (self.random_seed_offset * include_random_seed_offset)
 
     def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
         """
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -248,7 +248,10 @@ class Dataset(object):
         return self._random_seed_offset
 
     def _uses_custom_distributed_sharding(self) -> bool:
-        """override if the dataset has its own sharding logic independent of TF/PT"""
+        """
+        :return: if dataset has its own data sharding logic independent of TF/PT.
+            Leads to a fixed random_seed_offset independent of the workers local rank.
+        """
         return False
 
     def _get_default_random_seed_offset(self):
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -621,12 +621,10 @@ class Dataset(object):
 
         return seq_index
 
-    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1, include_random_seed_offset=True):
+    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1):
         """
         :param int|None epoch:
         :param int num_epochs_fixed: keep random seed fixed for n subsequent epochs
-        :param bool include_random_seed_offset: whether to add in the local random_seed_offset.
-            Can be disabled to get deterministic seeds in multi-process settings.
         :rtype: int
         """
         if self.fixed_random_seed is not None:
@@ -637,7 +635,7 @@ class Dataset(object):
             seed = (seed - 1) // partition_epoch + 1  # taking partitions requires constant seed during full epoch
         if num_epochs_fixed > 1:
             seed = (seed - 1) // num_epochs_fixed + 1
-        return seed + (self.random_seed_offset * include_random_seed_offset)
+        return seed + self.random_seed_offset
 
     def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
         """
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -242,18 +242,22 @@ class Dataset(object):
         state = {attr: getattr(self, attr) for attr in ["epoch", "zpad"]}
         return Dataset._create_from_reduce, (self.__class__, kwargs, state)
 
-    @staticmethod
-    def _get_default_random_seed_offset():
+    def _uses_custom_distributed_sharding(self):
+        """override if the dataset has its own sharding logic independent of TF/PT"""
+        return False
+
+    def _get_default_random_seed_offset(self):
         """
         :return: 0 usually
         :rtype: int
         """
         from returnn.config import get_global_config
 
+        if self._uses_custom_distributed_sharding():
+            return 0
         config = get_global_config(raise_exception=False)
         if not config:
             return 0
-
         env_val = os.environ.get(RANDOM_SEED_OFFSET_ENV_VAR)
         if env_val is not None:
             return int(env_val)
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -70,6 +70,7 @@ class DistributeFilesDataset(CachedDataset2):
     or to an integer between 1..=NUM_GPU_WORKERS to manually select how many shards the
     data is split into.
     The dataset attempts to split the files as evenly as possible based on the file size.
+    When ``seq_ordering: "random"``, the files are shuffled before every full epoch before sharding.
 
     Example usage::
 
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -150,9 +150,7 @@ class DistributeFilesDataset(CachedDataset2):
         :param shard: set to true to shard the data across worker processes
         :param _meta_info_cache: for internal use
         """
-
         super().__init__(**kwargs)
-
         self.files = files
         self.get_sub_epoch_dataset = get_sub_epoch_dataset
         assert preload_next_n_sub_epochs >= 0
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -172,9 +172,9 @@ class DistributeFilesDataset(CachedDataset2):
         self._file_cache: Optional[_FileCacheProc] = None
         self._workers: Dict[int, _WorkerProcParent] = {}  # epoch -> worker
         self._files_order_cache: Dict[int, List[List[FileTree]]] = {}  # full epoch (0-indexed) -> files order
-        self._files_for_this_worker: Dict[int, List[FileTree]] = (
-            {}
-        )  # full epoch (0-indexed) -> file shard for this worker
+        self._files_for_this_worker: Dict[
+            int, List[FileTree]
+        ] = {}  # full epoch (0-indexed) -> file shard for this worker
 
         if _meta_info_cache:
             # This allows to skip the lazy init in self.initialize().
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -65,12 +65,8 @@ class DistributeFilesDataset(CachedDataset2):
     In case the dataset grows so large it is unreasonable to expect one worker to
     ever see all the data, this dataset can also shard the file list on a per-worker
     basis before distributing across subepochs.
-    This behavior can be configured with the property ``num_shards``, which can either
-    be set to the string ``per-worker``, which automatically selects the number of shards
-    or to an integer between 1..=NUM_GPU_WORKERS to manually select how many shards the
-    data is split into.
+    This behavior can be configured by setting the property ``"shard": "per-worker"``.
     The dataset attempts to split the files as evenly as possible based on the file size.
-    When ``seq_ordering: "random"``, the files are shuffled before every full epoch before sharding.
 
     Example usage::
 
@@ -141,7 +137,7 @@ class DistributeFilesDataset(CachedDataset2):
         preload_next_n_sub_epochs: int = 1,
         buffer_size: int = 1,
         file_cache_opts: Optional[Dict[str, Any]] = None,
-        num_shards: Union[None, int, str] = 1,
+        shard: Union[None, str] = None,
         _meta_info_cache: Optional[Dict[str, Any]] = None,
         **kwargs,
     ):
@@ -151,16 +147,13 @@ class DistributeFilesDataset(CachedDataset2):
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
-        :param num_shards: shard the data into this many pieces and distribute between worker processes
-            before distributing the data over subepochs.
-
-            Valid values here are a string "per-worker" to automatically select the number of shards
-            based on the number of GPU workers or integers from 1..=NUM_GPU_WORKERS. When an integer
-            is set, it must cleanly divide the number of workers to avoid introducing a bias in the
-            seen data.
+        :param shard: set to "per-worker" to shard the data across worker processes
         :param _meta_info_cache: for internal use
         """
         super().__init__(**kwargs)
+
+        assert shard is None or shard == "per-worker"
+
         self.files = files
         self.get_sub_epoch_dataset = get_sub_epoch_dataset
         assert preload_next_n_sub_epochs >= 0
@@ -170,7 +163,7 @@ class DistributeFilesDataset(CachedDataset2):
         self._file_sizes: Optional[Dict[str, int]] = None  # key -> size. for equal distribution across sub epochs
         self._data_keys: Optional[List[str]] = None
         self._num_seqs: Optional[int] = None
-        self._worker_index, self._num_shards = _parse_num_shards_input(num_shards)
+        self._worker_index, self._num_shards = _get_rank_and_size() if shard == "per-worker" else 0, 1
 
         self._file_cache: Optional[_FileCacheProc] = None
         self._workers: Dict[int, _WorkerProcParent] = {}  # epoch -> worker
@@ -463,6 +456,7 @@ def _get_key_for_file_tree(t: FileTree) -> str:
 def _get_rank_and_size() -> Tuple[int, int]:
     """
     gets the global rank and size for multiprocess trainings
+    :return: tuple of rank and size
     """
 
     from returnn.config import get_global_config
@@ -484,33 +478,6 @@ def _get_rank_and_size() -> Tuple[int, int]:
         return ctx.rank(), ctx.size()
     else:
         return 0, 1
-
-
-def _parse_num_shards_input(num_shards_input: Union[None, int, str]) -> Tuple[int, int]:
-    """
-    parses the given input for the number of shards into an integer
-    and returns the local shard index for this worker
-    :param num_shards_input: the user-configured value for the number of shards
-    """
-
-    if num_shards_input is None:
-        return 0, 1
-    rank, size = _get_rank_and_size()
-    if isinstance(num_shards_input, int):
-        if num_shards_input <= 0:
-            raise ValueError(f"num_shards must be positive")
-        if num_shards_input > size:
-            raise ValueError(f"cannot shard into more splits ({num_shards_input}) than number of workers ({size})")
-        if size % num_shards_input != 0:
-            raise ValueError(
-                f"number of shards ({num_shards_input}) must cleanly divide number "
-                f"of workers ({size}) to avoid introducing a bias in the seen data"
-            )
-        return rank, num_shards_input
-    elif num_shards_input == "per-worker":
-        return rank, size
-    else:
-        raise ValueError(f"invalid input for num_shards: {num_shards_input}, can be None, int or 'per-worker'")
 
 
 class _WorkerProcParent:
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -151,8 +151,10 @@ class DistributeFilesDataset(CachedDataset2):
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
         :param num_shards: shard the data into this many pieces before distributing over subepochs.
-            Valid values here are integers from 1..=NUM_GPU_WORKERS or a string "per-worker" to automatically
-            select the number of shards based on the number of GPU workers.
+            Valid values here are a string "per-worker" to automatically select the number of shards
+            based on the number of GPU workers or integers from 1..=NUM_GPU_WORKERS. When an integer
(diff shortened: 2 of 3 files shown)


returnn/datasets/distrib_files.py
```diff
@@ -140,6 +150,11 @@ def __init__(
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
+        :param num_shards: shard the data into this many pieces before distributing over subepochs.
+            Valid values here are a string "per-worker" to automatically select the number of shards
```
Icemole, line 157: We should probably add below something like

```
assert num_shards is None or type(num_shards) == int or num_shards == "per-worker", (
    f"Invalid num_shards parameter: {num_shards}."
)
```

returnn/datasets/distrib_files.py
```diff
@@ -245,8 +264,32 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
+                del self._files_for_this_worker[k]
+        for ep_ in range(epoch, epoch + self.preload_next_n_sub_epochs + 1):
+            full_epoch_0idx_ = (ep_ - 1) // self.partition_epoch
+            if full_epoch_0idx_ in self._files_for_this_worker:
+                continue
+            if self.seq_ordering == "default":
+                files = self.files
+            elif self.seq_ordering == "random":
+                # do not use `self._get_random_seed_for_epoch` to exclude `random_seed_offset`
+                # to ensure an equal distribution in every worker
```
Icemole, line 284:
```suggestion
                # to ensure an equal distribution in every worker
                # since the sequences might collide between workers if a different seed is chosen.
```

returnn/datasets/distrib_files.py
```diff
@@ -245,8 +266,32 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
+                random_generator = numpy.random.RandomState(full_epoch_0idx_)
```
NeoLegends, line 284: Is this the right way to ensure equal distribution across all workers, does this need to take any more info into account or is the full epoch number enough here?

returnn/datasets/distrib_files.py
```diff
@@ -245,8 +264,32 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
```
Icemole, line 285: One question, why is a new random state with a different seed per epoch being created here? I would expect a fixed seed like 0 so that we always have the same files ordering, or otherwise we could have repeated sequences in the training, even after having completed them previously and before finishing a full training epoch. Maybe I'm misunderstanding something though.

returnn/datasets/distrib_files.py
```diff
@@ -308,62 +351,62 @@ def _get_sub_dataset_dict(self, files: List[FileTree]) -> Tuple[Dict[str, Any],
         return dataset_dict, exit_hook
 
     @staticmethod
-    def _get_files_per_sub_epochs(
-        *, partition_epoch: int, file_sizes: Dict[str, int], files_order: Sequence[FileTree]
+    def _distribute_evenly_by_size(
+        *, num_bins: int, file_sizes: Dict[str, int], files_order: Sequence[FileTree]
     ) -> List[List[FileTree]]:
```
Icemole, line 359: If possible try to add a docstring that at least specifies that the `num_bins` should be something similar to either the `partition_epoch` parameter, or the number of workers, or whatever the user specifies.

returnn/datasets/distrib_files.py
```diff
@@ -413,6 +456,59 @@ def _get_key_for_file_tree(t: FileTree) -> str:
...
+        return ctx.rank(), ctx.size()
+    else:
+        return 0, 1
+
+
+def _parse_num_shards_input(num_shards_input: Union[None, int, str]) -> Tuple[int, int]:
+    """
+    parses the given input for the number of shards into an integer
+    and returns the local shard index for this worker
+    :param num_shards_input: the user-configured value for the number of shards
```
Icemole, line 492:
```suggestion
    :param num_shards_input: the user-configured value for the number of shards
    :return: Local shard index for this worker and global number of shards.
```
Maybe also delete the line above, `and returns...`?

Icemole (changes requested):
Have you tested this? It would be interesting to use this in our codebase so that we don't have to wait for a future release, but it's not that important (mostly quality of life).

returnn/datasets/distrib_files.py
```diff
@@ -245,8 +267,32 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
+                del self._files_for_this_worker[k]
+        for ep_ in range(epoch, epoch + self.preload_next_n_sub_epochs + 1):
+            full_epoch_0idx_ = (ep_ - 1) // self.partition_epoch
+            if full_epoch_0idx_ in self._files_for_this_worker:
+                continue
+            if self.seq_ordering == "default":
+                files = self.files
+            elif self.seq_ordering == "random":
+                # do not use `self._get_random_seed_for_epoch` to exclude `random_seed_offset`
+                # to ensure an equal distribution in every worker
```
albertz, line 284: This is wrong. We should use `_get_random_seed_for_epoch` in any case. We just should fix `_get_default_random_seed_offset` (just the same as for TF).

returnn/datasets/distrib_files.py
albertz: Full epoch is enough. But I don't like to just take it as-is as a random seed. Multiply it with some prime number and add some prime number. But also, just use `_get_random_seed_for_epoch` here.

returnn/datasets/distrib_files.py
```diff
@@ -245,8 +267,32 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
+        # Shard files across GPU workers
+        for k in list(self._files_for_this_worker.keys()):
+            if k < full_epoch_0idx:
+                del self._files_for_this_worker[k]
+        for ep_ in range(epoch, epoch + self.preload_next_n_sub_epochs + 1):
+            full_epoch_0idx_ = (ep_ - 1) // self.partition_epoch
+            if full_epoch_0idx_ in self._files_for_this_worker:
+                continue
+            if self.seq_ordering == "default":
+                files = self.files
```
albertz, line 281: This is wrong here. Sharding should always be done, independent of the ordering.

returnn/datasets/distrib_files.py
albertz: No, there should always be a `:return:` section in the docstring.

returnn/datasets/distrib_files.py
NeoLegends: Is there a good reason for that?

albertz:
Your current logic:
* Shuffle `files`.
* Take subset based on sharding.
* Shuffle subset again.
* Take subset based on subepoch.

This seems suboptimal for me. E.g. that you do the shuffling twice here.

I would suggest the following logic instead:
* Optionally shuffle `files` if `seq_order == "random"`, or otherwise leave order.
* Take subset based on (sharding, subepoch) tuple (i.e. `num_bins = rank * partition_epoch`).

returnn/datasets/distrib_files.py
```diff
@@ -140,6 +151,13 @@ def __init__(
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
+        :param num_shards: shard the data into this many pieces and distribute between worker processes
```
albertz, line 154: I don't think it makes sense to specify the num shards? You either enable sharding or not, and index is always based on rank of the distributed worker, and num workers is always the size.

returnn/datasets/distrib_files.py
NeoLegends: This piece of code only runs for every full epoch, not for every subepoch. Therefore, there is repetition anyway.

We could, of course, only seed the RNG once at the start with seed `0` and keep it around for the entire training, but we also do it this way below and I guess there isn't a strong reason to prefer one over the other.

returnn/datasets/distrib_files.py
NeoLegends: But it is always done, see L290-293

returnn/datasets/distrib_files.py
NeoLegends: Yeah I was debating this. I think it might be a valid use case to have e.g. 16 workers, and only shard the data into 4 or 8 parts if this makes it realistic that every one worker can complete a full epoch. Not sure though, and removing the `int` option would simplify some of the validation logic.

returnn/datasets/distrib_files.py
albertz: I just want to have a different random state at different places where we use the epoch for seeding. If we just use `RandomState(epoch)` in multiple places throughout RETURNN (e.g. for shuffling seqs, for data augmentation, for random param init, whatever), it means all those RNGs start from the same state. This could have some weird effects. I like that all RNGs start from different seeds, by doing those adjustments on the seed. It's a bit similar reasoning like [salting for cryptography](https://en.wikipedia.org/wiki/Salt_(cryptography)).

returnn/datasets/distrib_files.py
albertz: But this is effectively then just running multiple epochs? I don't really understand the use case.

returnn/datasets/distrib_files.py
albertz: Ah right, sorry. But see my other comment on this logic.

NeoLegends:
> This seems suboptimal for me. E.g. that you do the shuffling twice here.

I think we should absolutely shuffle twice, because the first shuffle must use the same seed across different workers, while the second one should rather not but use the `random_seed_offset` to ensure some randomness even when we don't shard.

returnn/datasets/distrib_files.py
NeoLegends: I don't understand what you mean. If I used `_get_random_seed_for_epoch` here every horovod worker process would "mix in" its random_seed_offset and there would be different seeds per process for the initial sharding per worker process. Specifically, I don't understand what you mean with "fixing" `_get_default_random_seed_offset `. It seems like it does do exactly what it's supposed to do?

albertz: > here every horovod worker process would "mix in" its random_seed_offset

No, this is only when the user configures `horovod_dataset_distribution = "random_seed_offset"`. But when the user configures `horovod_dataset_distribution = "shard"`, it will not do that. Despite, when the user configures an explicit `random_seed_offset`, it will also ignore the worker rank but just use the explicit `random_seed_offset`.

Currently for PyTorch, the `_get_default_random_seed_offset` does not do this. There is no such setting like `torch_distributed_dataset_distribution`, it always assumes "random_seed_offset". That's what I mean by fixing. `_get_default_random_seed_offset`/`_get_random_seed_for_epoch` should not depend on the worker rank when sharding is done. If they do, they don't work correct, i.e. should be fixed.

albertz:
> > This seems suboptimal for me. E.g. that you do the shuffling twice here.
> 
> I think we should absolutely shuffle twice, because the first shuffle must use the same seed across different workers, while the second one should rather not but use the `random_seed_offset` to ensure some randomness even when we don't shard. Otherwise all workers process the data in the exact same order, which seems off because then all the gradients would be the same?

I don't understand what you mean. When you shuffle `files` for one full epoch, they are already all shuffled. Why does shuffling again make this more random?

returnn/datasets/distrib_files.py
```diff
@@ -140,9 +147,13 @@ def __init__(
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
+        :param shard: set to "per-worker" to shard the data across worker processes
         :param _meta_info_cache: for internal use
         """
         super().__init__(**kwargs)
+
+        assert shard is None or shard == "per-worker"
```
albertz, line 155: I would move the check below to where it is used/assigned.

returnn/datasets/distrib_files.py
Icemole: Yes, that's why I asked this. If I remember correctly there wasn't a `:return:` section in the string, as I proposed in the suggestion above. My proposal here was removing

> and returns the local shard index for this worker

and adding it more explicitly as a `:return:` tag.

returnn/datasets/distrib_files.py
michelwi: > Multiply it with some prime number and add some prime number.

now the same prime number is chosen for both multiplier and summand which makes 
```
full_epoch_0idx_ * 127 + 127
```
equivalent to
```
(full_epoch_0idx_ + 1) * 127
```

which still works, I guess, but was this how it was intended?

albertz: No, I did not intend to use the same prime for both multiplier and summand. I think this should be a different prime. Also, I would use a much larger prime. Also, I like primes of the form 2^n - 1, but probably that doesn't matter.

As an example, see how we seed the PyTorch random number generator:
```python
random_seed = self.config.int("random_seed", 42)
random_seed = (epoch * 193939 + step * 19937 + random_seed * 27644437 + 479001599) % (2**31)
rf.set_random_seed(random_seed)
```

michelwi: > Also, I like primes of the form 2^n - 1

to be fair, 127 is `2^7 - 1` :P

NeoLegends:
> I don't understand what you mean. When you shuffle files for one full epoch, they are already all shuffled. Why does shuffling again make this more random?

It does not make it more random, but my understanding is that the first shuffle must be done the exact same way on all workers to be able to use its results for sharding. This is why the first shuffle has a fixed seed that is identical across all workers, while the second one, where we shuffle the data within one shard, takes into account the random seed offset (if any).

If we did this with only one shuffle we wouldn't be sharding properly, but we might have duplicate/left out files. Or am I missing something here?

michelwi:
> It does not make it more random, but my understanding is that the first shuffle must be done the exact same way on all workers to be able to use its results for sharding. This is why the first shuffle has a fixed seed that is identical across all workers,

yes.

>  while the second one, where we shuffle the data within one shard, takes into account the random seed offset (if any).
> 
> If we did this with only one shuffle we wouldn't be sharding properly, but we might have duplicate/left out files. Or am I missing something here?

I think the argument is not against the first shuffle but the second one (per worker), because by doing the first one (across all workers) also all files per worker are already sufficiently shuffled. Doing the second shuffling changes the order, but does not add any benefit.

> If we did this with only one shuffle we wouldn't be sharding properly, but we might have duplicate/left out files. Or am I missing something here?

If we only do the first shuffle, then there would not be problems, right?

NeoLegends:
Oh, right! I had always considered the first shuffle optional, not the second one. But this isn't actually true, since we now always do the "first" shuffle even with only one worker!

returnn/datasets/distrib_files.py
```diff
@@ -245,26 +258,45 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
             elif self.seq_ordering == "random":
-                rnd_seed = self._get_random_seed_for_epoch(full_epoch_0idx_ * self.partition_epoch + 1)
-                random_generator = numpy.random.RandomState(rnd_seed)
-                files_order_flat = list(self.files)
-                random_generator.shuffle(files_order_flat)
+                # do not use `self._get_random_seed_for_epoch` to exclude
+                # `random_seed_offset` to ensure an equal distribution in every
+                # worker since the sequences might collide between workers if a
+                # different seed is chosen.
+                random_generator = numpy.random.RandomState(full_epoch_0idx_ * 9941 + 4423)
```
albertz, line 278: Btw, it is still wrong that you do not use `_get_random_seed_for_epoch` here. This is actually the correct function to use. If `random_seed_offset` is set, it should be used. (What you actually want is to fix `_get_default_random_seed_offset` to also allow sharding, i.e. to not use a different random seed offset per worker.)

NeoLegends: > Btw, it is still wrong that you do not use _get_random_seed_for_epoch here. This is actually the correct function to use. If random_seed_offset is set, it should be used. (What you actually want is to fix _get_default_random_seed_offset to also allow sharding, i.e. to not use a different random seed offset per worker.)

I'm not sure I understand this completely here. What you say seems contradictionary to me.

It seems right to use the random seed offset for all cases except for this one, so I'm not sure why we should adapt this function. Like all the other dataset shuffling logic in the sub datasets should use the offset, and if we're in a worker process in a multi-GPU setting here loading data through `DistributeFilesDataset` we most probably will already have a random seed offset set (which you said _should_ then be used) based on the local rank.

In that sense the current logic seems right to me? Or do you just want to pass a flag to the function like `use_random_seed_offset` that runs the existing logic but specifically excludes the offset?

albertz: `_get_random_seed_for_epoch` is supposed to return the random seed for an epoch, in all cases, including the case here. If you want a different random seed offset per worker, then `_get_random_seed_for_epoch` should return that. If you want the same random seed offset per worker, then `_get_random_seed_for_epoch` should return that.

`_get_random_seed_for_epoch` should not just always return a different random seed per worker. That would be incorrect behavior of `_get_random_seed_for_epoch`.

`_get_random_seed_for_epoch` can also do other logic. E.g. there is `fixed_random_seed`. In the future, there might be other things. We should always use this function we need the random seed for the epoch.

If `_get_random_seed_for_epoch` does not return the correct random seed for the epoch, then you should consider this a bug of `_get_random_seed_for_epoch`. And you are not using it here because apparently it does not return the correct random seed for the epoch.

So, why do you not use `_get_random_seed_for_epoch`? Why is it not the correct seed? This is because of wrong/buggy/incomplete code in `_get_default_random_seed_offset`. As said before, `_get_default_random_seed_offset` handles both cases correctly in case for TF already, just not for PT.

returnn/datasets/basic.py
```diff
@@ -621,10 +621,12 @@ def _apply_partition_epoch(cls, seq_index, partition_epoch, epoch):
 
         return seq_index
 
-    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1):
+    def _get_random_seed_for_epoch(self, epoch, num_epochs_fixed=1, include_random_seed_offset=True):
```
albertz, line 624: That's wrong. The `random_seed_offset` should always be used.

I think what you actually want to change is `_get_default_random_seed_offset`, not this function here.

returnn/datasets/distrib_files.py
NeoLegends: My point is that even in the correctly configured TF case where `random_seed_offset` is used, we _still do not want_ the output of `_get_default_random_seed_offset` because we do not want to use it, but we _do_ want to use it in the sub datasets. It's an entirely separate case, no?

albertz: Of course we always want `_get_default_random_seed_offset`? Why do you think we would not want it? `_get_default_random_seed_offset` should just return the correct random seed offset depending on what behavior you want.

albertz: Very much the same argument for `_get_default_random_seed_offset`: `_get_default_random_seed_offset` should not always return a different random seed offset per worker. It should always return the correct random seed offset instead.

NeoLegends: But we only want this very specific behavior in this one dataset. And `_get_default_random_seed_offset ` is static, so it cannot be overridden. I find it strange to special case that generic implementation there just for this dataset?

albertz: No it's not only for this dataset. It's always when you want to have the same random seed offset per worker.

But you are right that we should discuss on how we should handle this in general. For TF/Horovod, it checks `is_dataset_distribution_random_seed_offset`. This is somewhat a global setting. As I said before, maybe we should have the same for PT. Or maybe instead of having such a global setting, it should be part of the base `Dataset` class? I'm not sure on this.

In any case, if `random_seed_offset` is explicitly set by the user, or by a parent dataset, or when `RANDOM_SEED_OFFSET_ENV_VAR` is set, it would be used over any such logic. This should basically always be the case when there is a parent dataset. So we are really only discussing the case for the top datasets here.

NeoLegends: Do you agree to merge this PR and decide on a better solution for the seed in a follow-up?

albertz: I think this should be done within this PR, as this is actually one of the core aspects of the PR.

albertz: In any case, the fix would also be very simple. We just should decide on what solution we prefer. Another idea: A dataset can specify whether it supports and uses its own custom sharding in distributed training (like in this case), so some function `Dataset._uses_custom_distributed_sharding`, which would be overwritten by the dataset here, and the `Dataset._get_default_random_seed_offset` would check for it.

NeoLegends:
> Oh, right! I had always considered the first shuffle optional, not the second one. But this isn't actually true, since we now always do the "first" shuffle even with only one worker!

Ah, no, this is all wrong. I let myself get mislead by y'all!

The issue is not with the shuffle itself, but w/ the replication across different workers. Imagine you have a files list like `abcdefghijklmnopqrstuvxyz`. You then train w/ >1 workers but no sharding. In the current implementation this means you first shuffle the list on each worker using the same seed. This shuffled list is then further divided up into sub epochs, i.e. you end up with e.g. `xao ceb hfv kij uys ptq mdn zrgl`.

My issue w/ the code as it is if you only shuffle a single time w/ a seed that is replicated across all workers is that now, this "final" shuffle result is identical on every worker. This means there is no difference in the data order between workers, because we don't include any random seed offset. I think the result will be that now we're equivalent to training on just one GPU at a time, because all the gradients/weights will always be the same and the parallelism will be averaged out.

Meanwhile if we shuffled twice, the following would happen: we start off with `abcdefghijklmnopqrstuvxyz`, then shuffle (just like before) w/ fixed seed into `xaocebhfvkijuysptqmdnzrgl`. However, now every worker shuffles again with its own offset seed, leading to unique results per worker:
```
Worker 1: reaxtvcozgnhpsqjbdulyfikm -> rea xtv coz gnh psq jbd uly fikm
Worker 2: fhtirubjkzqlysoxcmagvnpde -> fht iru bjk zql yso xcm agv npde
...
```
This means in the first subepoch the training now sees the data from `rea` and `fht` (etc.), instead of the data from `xao` `N` times.

This is only an issue if you don't shard since otherwise the data is already split up and not replicated. Am I missing something?

michelwi:
> This shuffled list is then further divided up into sub epochs, i.e. you end up with e.g. `xao ceb hfv kij uys ptq mdn zrgl.`

But would then not the random seed offset kick in and in the first sub-epoch the workers would use
```
worker 0: xao
worker 1: ceb
worker 2: hfv
...
```
?

albertz:
> You then train w/ >1 workers but _no sharding_. In the current implementation this means you first shuffle the list on each worker using the same seed.

Oh this is a bug I think. I wonder why it was done this way. I think we should have used `_get_random_seed_for_epoch` there (as I said, we always should use `_get_random_seed_for_epoch`).

But even right now, `_get_random_seed_for_epoch` is used correctly in the sub dataset, so they still operate on different seq orders then (but on the same file if I understand correctly).

returnn/datasets/basic.py
```diff
@@ -242,18 +240,28 @@ def __reduce__(self):
...
 
-    @staticmethod
-    def _get_default_random_seed_offset():
+    @property
+    def random_seed_offset(self) -> int:
+        if self._random_seed_offset is None:
+            self._random_seed_offset = self._get_default_random_seed_offset()
+        return self._random_seed_offset
+
+    def _uses_custom_distributed_sharding(self):
```
albertz, line 249:
```suggestion
    def _uses_custom_distributed_sharding(self) -> bool:
```

returnn/datasets/distrib_files.py
```diff
@@ -140,9 +147,12 @@ def __init__(
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
+        :param shard: set to true to shard the data across worker processes
```
albertz, line 150: I wonder if we maybe should choose a bit more detailed name, like `distributed_sharding` or so? Or is sharding really only relevant for distributed training?

returnn/datasets/distrib_files.py
```diff
@@ -140,9 +147,12 @@ def __init__(
...
         :param _meta_info_cache: for internal use
         """
+
         super().__init__(**kwargs)
+
```
albertz, line 155:
```suggestion
        super().__init__(**kwargs)
```

returnn/datasets/distrib_files.py
```diff
@@ -187,6 +198,9 @@ def _meta_info_cache(self):
             "file_sizes": self._file_sizes,
         }
 
+    def _uses_custom_distributed_sharding(self):
```
albertz, line 201:
```suggestion
    def _uses_custom_distributed_sharding(self) -> bool:
```

returnn/datasets/distrib_files.py
```diff
@@ -413,6 +436,33 @@ def _get_key_for_file_tree(t: FileTree) -> str:
     return ":".join(tree.flatten(t))
 
 
+def _get_rank_and_size() -> Tuple[int, int]:
+    """
+    gets the global rank and size for multiprocess trainings
+    :return: tuple of rank and size
```
albertz, line 442:
```suggestion
    :return: tuple (rank, size): the global rank and size for distributed trainings
```

returnn/datasets/distrib_files.py
```diff
@@ -255,17 +269,21 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
             elif self.seq_ordering == "random":
-                rnd_seed = self._get_random_seed_for_epoch(full_epoch_0idx_ * self.partition_epoch + 1)
-                random_generator = numpy.random.RandomState(rnd_seed)
-                files_order_flat = list(self.files)
-                random_generator.shuffle(files_order_flat)
+                random_generator = numpy.random.RandomState(
+                    self._get_random_seed_for_epoch(full_epoch_0idx_ * self.partition_epoch + 1)
+                )
+                files = list(self.files)
+                random_generator.shuffle(files)
```
albertz, line 278: It's actually here exactly the same code as before, only that you changed the var name. Maybe just keep the original code, so the diff looks cleaner?

returnn/datasets/distrib_files.py
NeoLegends: Hmm, I'm not sure either. I like that it's concise and short, but I also feel like I lack background knowledge to answer whether sharding is relevant in other cases. We could rename it to `shard_files` though to be clearer on the behavior.

albertz: But it's also specific to distributed training. I'm not sure if there are maybe other applications of sharding.

albertz: So `distrib_shard_files`?

albertz (approved):

Icemole (approved):

albertz:
Btw, when merging to master, I usually clean up the commit message. Esp I don't like having all those intermediate dummy commit messages in it ("black", "improve docs", ...). If at all, maybe there could be the PR description in it, or a summarized version of it. And I keep the Co-authored-by, but that's all.

NeoLegends:
I'm now running a test training for this.

PR #1543, opened 2024-06-18
`FileCache`: retry on ENOSPC, prealloc space, lock on whole cache dir

### LINKED ISSUE #1542: Possible race condition in `FileCache`?
I had a training using `DistributeFilesDataset` fail due to missing disk space while copying a file onto the local disk with `FileCache`. I don't see how this can happen given that we check for enough free disk space before copying every file other than a race condition with another `FileCache` implementation or the disk suddenly shrinking in size. Given that we (by default) pre-fetch the files for the next epoch there are indeed two `FileCache` implementations at work at the same time which might get into conflict w/ each other.

I think we should _not_ put work into fixing the race condition itself, but rather be more tolerant to failure by allowing a couple retries (3-5) for cleanup + copying onto the local disk before actually giving up.

NeoLegends:
testing if this makes a difference...

Closes #1542

### DIFF
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -339,6 +339,10 @@ def _copy_with_prealloc(src: str, dst: str):
     Note the function preallocates `size + 1` to allow detecting incompletely copied
     files by a mismatch in the file size, should the copy process be interrupted. The
     additional byte is then truncated away after copying.
+
+    In practice this function is used to copy to a temporary file first, so the
+    +1-size trick is technically not necessary -- but it also does not hurt leaving
+    it in.
     """
     file_size = os.stat(src).st_size
     with open(dst, "wb") as dst_file:
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -19,7 +19,6 @@ from collections import defaultdict
 from contextlib import contextmanager
 from threading import Thread, Event
 from .basic import expand_env_vars, LockFile, human_bytes_size
-from returnn.log import log
 
 
 __all__ = ["FileCache", "CachedFile"]
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -18,6 +18,7 @@ from collections import defaultdict
 from contextlib import contextmanager
 from threading import Thread, Event
 from .basic import expand_env_vars, LockFile, human_bytes_size
+from returnn.log import log
 
 
 __all__ = ["FileCache", "CachedFile"]
@@ -56,6 +57,7 @@ class FileCache:
         cleanup_files_always_older_than_days: float = 31.0,
         cleanup_files_wanted_older_than_days: float = 7.0,
         cleanup_disk_usage_wanted_free_ratio: float = 0.2,  # try to free at least 20% disk space
+        num_tries: int = 3,  # retry twice by default
     ):
         """
         :param cache_directory: directory where to cache files.
@@ -64,7 +66,11 @@ class FileCache:
         :param cleanup_files_wanted_older_than_days: if cleanup_disk_usage_wanted_free_ratio not reached,
             cleanup files older than this.
         :param cleanup_disk_usage_wanted_free_ratio: try to free at least this ratio of disk space.
+        :param num_retries: how many times to try caching a file before giving up
         """
+
+        assert num_tries > 0
+
         self.cache_directory = expand_env_vars(cache_directory)
         self._cleanup_files_always_older_than_days = cleanup_files_always_older_than_days
         self._cleanup_files_wanted_older_than_days = cleanup_files_wanted_older_than_days
@@ -72,6 +78,7 @@ class FileCache:
         self._touch_files_thread = _TouchFilesThread(cache_base_dir=self.cache_directory)
         self._touch_files_thread.start()
         self._recent_full_cleanup_time = float("-inf")
+        self._num_tries = num_tries
 
     # Note on lock_timeout: It will check whether a potentially existing lock file is older than this timeout,
     # and if so, then it would delete the existing lock file, assuming it is from a crashed previous run.
@@ -99,7 +106,16 @@ class FileCache:
         :return: cached file path (in the cache directory)
         """
         dst_filename = self._get_dst_filename(src_filename)
-        self._copy_file_if_needed(src_filename, dst_filename)
+        last_error = None
+        for num_try in range(1, self._num_tries + 1):
+            try:
+                self._copy_file_if_needed(src_filename, dst_filename)
+                break
+            except Exception as e:
+                print(f"FileCache: {e} while copying file in try {num_try} of {self._num_tries}", file=log.v5)
+                last_error = e
+        if last_error is not None:
+            raise last_error
         self._touch_files_thread.files_extend([dst_filename])
         return dst_filename
 
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -66,7 +66,7 @@ class FileCache:
         :param cleanup_files_wanted_older_than_days: if cleanup_disk_usage_wanted_free_ratio not reached,
             cleanup files older than this.
         :param cleanup_disk_usage_wanted_free_ratio: try to free at least this ratio of disk space.
-        :param num_retries: how many times to try caching a file before giving up
+        :param num_tries: how many times to try caching a file before giving up
         """
         self.cache_directory = expand_env_vars(cache_directory)
         self._cleanup_files_always_older_than_days = cleanup_files_always_older_than_days
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -327,23 +327,22 @@ class FileCache:
 
 
 def _copy_with_prealloc(src: str, dst: str):
-    """
-    copies from src to dst preallocating the disk space at dst before copying
-
-    :param src: source file
-    :param dst: destination file
-    """
     dst_size = os.stat(src).st_size
     with open(dst, "wb") as dst_file:
+        # we prealloc size + 1 to avoid having a file that has the exact same size as
+        # its original to be able to detect whether it has been incompletely copied
+        #
+        # see also _check_existing_copied_file_maybe_cleanup
         if dst_size > 0:
             if os.name == "posix":
-                os.posix_fallocate(dst_file.fileno(), 0, dst_size)
+                os.posix_fallocate(dst_file.fileno(), 0, dst_size + 1)
             else:
-                dst_file.seek(dst_size - 1)
+                dst_file.seek(dst_size)
                 dst_file.write(b"\0")
                 dst_file.seek(0)
         with open(src, "rb") as src_file:
             shutil.copyfileobj(src_file, dst_file)
+        dst_file.truncate(dst_size)
 
 
 @dataclass
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -279,22 +279,20 @@ class FileCache:
             os.utime(dst_filename, None)  # touch
             return
 
-        # Make sure we have enough disk space.
-        self.cleanup(need_at_least_free_space_size=os.stat(src_filename).st_size)
-
-        print(f"FileCache: Copy file {src_filename} to cache")
-
         # Create dirs.
         dst_dir = os.path.dirname(dst_filename)
         os.makedirs(dst_dir, exist_ok=True)
 
         # Copy the file, while holding a lock. See comment on lock_timeout above.
-        with LockFile(
-            directory=dst_dir, name=os.path.basename(dst_filename) + ".lock", lock_timeout=self._lock_timeout
-        ) as lock:
+        with LockFile(directory=self.cache_directory, name="dir.lock", lock_timeout=self._lock_timeout) as lock:
             # Maybe it was copied in the meantime, while waiting for the lock.
-            if os.path.exists(dst_filename):
+            if self._check_existing_copied_file_maybe_cleanup(src_filename, dst_filename):
                 return
+
+            print(f"FileCache: Copy file {src_filename} to cache")
+
+            # Make sure we have enough disk space.
+            self.cleanup(need_at_least_free_space_size=os.stat(src_filename).st_size)
 
             dst_tmp_filename = dst_filename + ".copy"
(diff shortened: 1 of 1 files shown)


returnn/util/file_cache.py
```diff
@@ -99,7 +106,16 @@ def get_file(self, src_filename: str) -> str:
         :return: cached file path (in the cache directory)
         """
         dst_filename = self._get_dst_filename(src_filename)
-        self._copy_file_if_needed(src_filename, dst_filename)
+        last_error = None
+        for num_try in range(1, self._num_tries + 1):
+            try:
+                self._copy_file_if_needed(src_filename, dst_filename)
+                break
+            except Exception as e:
```
albertz, line 114: This is way too generic. You should only catch very specific errors, e.g. out-of-disk space or so.

returnn/util/file_cache.py
```diff
@@ -64,14 +66,19 @@ def __init__(
         :param cleanup_files_wanted_older_than_days: if cleanup_disk_usage_wanted_free_ratio not reached,
             cleanup files older than this.
         :param cleanup_disk_usage_wanted_free_ratio: try to free at least this ratio of disk space.
+        :param num_retries: how many times to try caching a file before giving up
         """
+
+        assert num_tries > 0
```
albertz, line 72: This should be moved below to where it is assigned. Also without the empty lines.

returnn/util/file_cache.py
NeoLegends: Are you sure? This is just for retries anyways, and if the error is not transient but persists will be re-thrown at the end.

NeoLegends: I'm also not sure we know the set of errors upfront or if the errors we get are even meaningful. Out of disk space, for example, is just an `OSError`.

albertz: But we should not retry if there is some other exception. There could be all kind of strange things happening otherwise.



> Out of disk space, for example, is just an `OSError`.

But you can just check the error code for out-of-disk-space. Should be `errno.ENOSPC`.

NeoLegends: ... and then there are a bunch of other transient errors that we want to retry on, no? I'm just arguing trying to make this list exhaustive is not worth it when we're re-raising on the last try anyway.

albertz: > there are a bunch of other transient errors that we want to retry on, aren't there?

No, that's exactly my point. We should not just retry on whatever problem there is.

returnn/util/file_cache.py
```diff
@@ -99,7 +104,16 @@ def get_file(self, src_filename: str) -> str:
         :return: cached file path (in the cache directory)
         """
         dst_filename = self._get_dst_filename(src_filename)
-        self._copy_file_if_needed(src_filename, dst_filename)
+        last_error = None
+        for num_try in range(1, self._num_tries + 1):
```
albertz, line 108: So as said, I would not do this retrying at all currently. I would first try with the other suggestions.

albertz:
And what about the idea with reserving the space? This would reduce the race conditions also a lot. Although not sure if this is easily doable.

returnn/util/file_cache.py
NeoLegends: Ok I reduced the retry logic to out of disk space errors and then I'd leave it in for now. I also changed the lock to work on the directory level itself.

NeoLegends:
> And what about the idea with reserving the space? This would reduce the race conditions also a lot. Although not sure if this is easily doable.

Problem is copyfile replace the destination, so if we preallocated the file there it's just going to be removed before the copy. We could copy manually though?

returnn/util/file_cache.py
albertz: But not sure. I think I would avoid the retry logic. I don't see a good reason for doing it. But I see many problems with it.

albertz:
> Problem is copyfile replaces the destination, so if we preallocated the file there it's just going to be removed before the copy. We could copy manually though?

We don't need to use `copyfile`. We can use whatever we want. But I'm in any case not sure if the preallocation is easily doable.

returnn/util/file_cache.py
```diff
@@ -327,6 +327,20 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
         return True
 
 
+def _copy_with_prealloc(src: str, dst: str):
+    """
+    copies from src to dst preallocating the disk space at dst before copying
```
albertz, line 331: There should always be an empty line after the description and the params in the docstring.

Or you leave the params away anyway. I think in this case the params are not really needed to be described. If you think`src` and `dst` are too ambiguous, then call them `source_file` and `destination_file` instead. That makes the param docstring obsolete.

returnn/util/file_cache.py
```diff
@@ -327,6 +327,20 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
+    :param src: source file
+    :param dst: destination file
+    """
+    stat = os.stat(src)
+    with open(dst, "wb") as dst_file:
+        dst_file.truncate(stat.st_size)
+        dst_file.seek(0)
```
albertz, line 339: I think this `seek` is not necessary?

returnn/util/file_cache.py
```diff
@@ -327,6 +327,20 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
 
 
+def _copy_with_prealloc(src: str, dst: str):
+    """
+    copies from src to dst preallocating the disk space at dst before copying
+    :param src: source file
+    :param dst: destination file
+    """
+    stat = os.stat(src)
+    with open(dst, "wb") as dst_file:
```
albertz, line 337: I think calling `os.posix_fadvise` with `POSIX_FADV_SEQUENTIAL | POSIX_FADV_NOREUSE` or so might also be a good idea?

returnn/util/file_cache.py
```diff
@@ -327,6 +327,20 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
+        dst_file.truncate(stat.st_size)
```
albertz, line 338: I'm not 100% sure this will always really allocate the space. Some filesystems could support sparse files? Actually maybe most?

Note, I have also see other possibilities. You should also add some comment shortly describing the possible options for the preallocation.

E.g.:

* `os.seek(size-2)` + `file.write(b"\0")` + `os.seek(0)`. ([via](https://stackoverflow.com/a/68104312/133374))
* `os.posix_fallocate`: Probably the best option, as it is specifically intended for this.

returnn/util/file_cache.py
```diff
@@ -333,10 +333,15 @@ def _copy_with_prealloc(src: str, dst: str):
...
     """
-    stat = os.stat(src)
+    dst_size = os.stat(src).st_size
     with open(dst, "wb") as dst_file:
-        dst_file.truncate(stat.st_size)
-        dst_file.seek(0)
+        if dst_size > 0:
+            try:
+                os.posix_fallocate(dst_file.fileno(), 0, dst_size)
+            except:
```
albertz, line 341: We should never make an `except` so generic. But also, when would this fail? Why? I don't think we should have the fallback code unless we think we can realistically hit it.

returnn/util/file_cache.py
```diff
@@ -333,10 +333,15 @@ def _copy_with_prealloc(src: str, dst: str):
...
+                dst_file.seek(dst_size - 1)
```
albertz, line 342: I think `dst_size - 2`?

albertz, line 342: Why do you think this fallback is better than the `truncate`?

returnn/util/file_cache.py
NeoLegends: yeah I removed this meanwhile

returnn/util/file_cache.py
NeoLegends: why? `len(b"\0") == 1`

albertz: Ah no sorry, `dst_size - 1` should be correct.

NeoLegends: less magic

returnn/util/file_cache.py
```diff
@@ -312,6 +326,26 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
         return True
 
 
+def _copy_with_prealloc(src: str, dst: str):
+    """
+    copies from src to dst preallocating the disk space at dst before copying
+
+    :param src: source file
+    :param dst: destination file
```
albertz, line 334: I don't like such redundant docstrings. If you think the variable name is not self-explanatory, why not just change the variable names?

albertz:
Actually, now I wonder whether the preallocation is a good idea together with our check in `_check_existing_copied_file_maybe_cleanup` which just checks for the file size... If the job is killed during the copying, then at next run, the file size will be the correct one, and the copying will be skipt, although the file was not fully copied yet... Not sure how to solve this. Maybe allocate one byte more than necessary and truncate that away in the end?

returnn/util/file_cache.py
NeoLegends: No honestly I added this to keep the linter happy. I don't usually work in PyCharm so I don't get the pyCharm lints for free. I've since removed them.

albertz: There is a check that all public functions/methods/classes/modules must have a docstring. But:

* It does not apply for private functions. Those don't need to have any docstring.
* It does not mean that args must be documented.

However, when you document args, then it is a warning when you only document some of them but not all of them, or when you document some non-existing one.

albertz:
> So what about `cleanup_files_wanted_older_than_days`? Or even introduce another one? I think we should try to free more than just `need_at_least_free_space_size` if possible. This should also be handled here. As discussed in the issue.

So what about this?

And maybe in any case also try to free `need_at_least_free_space_size * 1.2` or so.

returnn/util/file_cache.py
```diff
@@ -312,6 +326,26 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
 
+def _copy_with_prealloc(src: str, dst: str):
+    dst_size = os.stat(src).st_size
+    with open(dst, "wb") as dst_file:
+        if dst_size > 0:
+            # We prealloc size + 1 to avoid having a file that has the exact same
+            # size as its original. After copying, we then truncate the additional
+            # byte away. This allows detecting incompletely copied files.
+            #
+            # See also `_check_existing_copied_file_maybe_cleanup`.
```
albertz, line 337:
```suggestion
            # We prealloc size + 1 to avoid having a file that has the exact same size
            # as its original, to allow detecting incompletely copied files,
            # as done in `_check_existing_copied_file_maybe_cleanup`.
            # After copying, we then truncate the additional byte away.
```

NeoLegends:
> So what about this?

I think we should touch that logic in a follow-up. There should've been log lines when cleanup is happening, but I didn't see any.

returnn/util/file_cache.py
```diff
@@ -312,6 +326,26 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
+            if os.name == "posix":
+                os.posix_fallocate(dst_file.fileno(), 0, dst_size + 1)
```
albertz, line 339: Actually, I realize now, we anyway copy to `dst_tmp_filename`, not directly to the target file, so maybe the `dst_size + 1` was not necessary?

NeoLegends: yeah perhaps, but still doesn't hurt

albertz: Ok but you should extend the comment then, otherwise someone might get confused and wonder about this.

returnn/util/file_cache.py
```diff
@@ -99,7 +104,18 @@ def get_file(self, src_filename: str) -> str:
         :return: cached file path (in the cache directory)
         """
         dst_filename = self._get_dst_filename(src_filename)
-        self._copy_file_if_needed(src_filename, dst_filename)
+        last_error = None
+        for _ in range(self._num_tries):
```
albertz, line 108:
```suggestion
        for try_nr in range(self._num_tries):
            if try_nr > 0:
                print(f"FileCache: Ignoring error while copying {dst_filename}: {type(last_error).__name__}: {last_error}")
```

returnn/util/file_cache.py
NeoLegends: there is a comment on lines 337-341

returnn/util/file_cache.py
NeoLegends: Note: I did not set `POSIX_FADV_NOREUSE` because we actually access the file range targeted by the advise multiple times (once per chunk copied). Alternative would be one advise call per chunk, but this seems like a lot of syscall overhead and also would require us to stop using `copyfileobj` and hand-roll the copying loop, which isn't worth it.

albertz: But my understanding of `POSIX_FADV_NOREUSE` is that it is per byte, not per range. You just specify the range to avoid calling `fadvise` size-times. So it's actually exactly fitting for our use case.

Otherwise, `POSIX_FADV_NOREUSE` would not really make sense to me this way. I think almost all existing file copy code (e.g. check `/bin/cp`) use exactly this logic. (But we can actually check that...)

albertz: Ok, in GNU coretuils, there is a comment saying `POSIX_FADV_NOREUSE` does nothing since Linux kernel version 2.6.31?
https://github.com/coreutils/coreutils/blob/bbc972bc827aa5b34fce697ed79a36c70c4fabe3/src/sort.c#L936-L941

Ok, so then maybe we can really ignore this...

returnn/util/file_cache.py
albertz: Yea but it is relevant, as I would otherwise think this is a bug in the code and unnecessary. I actually got confused that we do this now as it's not really needed.

NeoLegends: I extended the docstring in the meantime.

albertz: Yea but it does not say anything about that this is actually the temp file, so that the `dst_size + 1` is not necessary, and why we still do it anyway.

albertz (approved):

PR #1545, opened 2024-06-20
HDFDataset: add support for CachedFile

michelwi:
As discussed in Slack, add support to use `CachedFile(...)` in the HDFDataset. e.g.
```
train = {"class": "HDFDataset", "files": CachedFile("/path/to/hdf")}
```

PR #1546, opened 2024-06-20
`FileCache`: add generic support for all datasets

### LINKED ISSUE #1544: Ideas for generic `CachedFile` support across all datasets
WDYT about adding a global `FileCache` to RETURNN and passing every dataset kwargs through `FileCache.handle_cached_files_in_config` in `init_dataset`?

The support for this would also have to ensure we de-register the relevant files when a dataset is de-initialized. I'm not yet sure how to best do this generically. Maybe we can remember the cached files we have gotten from a dataset and then store them in the dataset somewhere for deregistration in `__del__`?

cc @albertz @michelwi @Icemole

### LINKED ISSUE #1549: `DistributeFilesDataset`: copying files blocks `init_seq_order`
When `DistributeFilesDataset.init_seq_order` is called, it will schedule the copying (in the sub proc `_FileCacheProc`), but it will wait for this to complete, thus this will block. But there is no reason for this to block. `init_seq_order` could directly return. Even the `init_seq_order` on the sub dataset can run async.

PR #1546 actually partially fixes this.

NeoLegends:
* Fix #1544.
* Fix #1549.
* Closes #1545

Let me know your thoughts about the implementation here. I'm not a big fan yet about having to register "from the outside" the files to be released later, but I'm also not sure if there's a better way to unregister the files when the dataset is deinitialized.

One advantage realising support this way is that cache operations for the preloaded subepoch in `DistributeFilesDataset` are now actually happening asynchronously, because `init_dataset`, which invokes caching, is called in the subprocess. Previously this caching would happen in the parent process, blocking training.

### DIFF
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -6,7 +6,7 @@ https://github.com/rwth-i6/returnn/issues/1519
 
 from __future__ import annotations
 
-from typing import Optional, Any, Dict, Tuple, List, Sequence, Collection, Callable, Union
+from typing import Optional, Any, Dict, Tuple, List, Sequence, Callable, Union
 import os
 import sys
 import numpy
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -136,7 +136,6 @@ class DistributeFilesDataset(CachedDataset2):
         get_sub_epoch_dataset: Callable[[List[FileTree]], Dict[str, Any]],
         preload_next_n_sub_epochs: int = 1,
         buffer_size: int = 1,
-        file_cache_opts: Optional[Dict[str, Any]] = None,
         distrib_shard_files: bool = False,
         _meta_info_cache: Optional[Dict[str, Any]] = None,
         **kwargs,
@@ -157,13 +156,11 @@ class DistributeFilesDataset(CachedDataset2):
         assert preload_next_n_sub_epochs >= 0
         self.preload_next_n_sub_epochs = preload_next_n_sub_epochs
         self.buffer_size = buffer_size
-        self.file_cache_opts = file_cache_opts or {}
         self._file_sizes: Optional[Dict[str, int]] = None  # key -> size. for equal distribution across sub epochs
         self._data_keys: Optional[List[str]] = None
         self._num_seqs: Optional[int] = None
         self._shard_index, self._num_shards = _get_rank_and_size() if distrib_shard_files else (0, 1)
 
-        self._file_cache: Optional[_FileCacheProc] = None
         self._workers: Dict[int, _WorkerProcParent] = {}  # epoch -> worker
         self._files_order_cache: Dict[int, List[List[FileTree]]] = {}  # full epoch (0-indexed) -> files order
 
@@ -205,15 +202,12 @@ class DistributeFilesDataset(CachedDataset2):
             return
         # First, we need to know the num_inputs, num_outputs, total_num_seqs, labels.
         # Init the dataset with the first file.
-        dataset_dict, exit_hook = self._get_sub_dataset_dict(files=[self.files[0]])
-        try:
-            dataset = init_dataset(dataset_dict, extra_kwargs={"seq_ordering": "default"}, parent_dataset=self)
-            self.num_inputs = dataset.num_inputs
-            self.num_outputs = dataset.num_outputs
-            self.labels = dataset.labels
-            self._data_keys = dataset.get_data_keys()
-        finally:
-            exit_hook()
+        dataset_dict = self._get_sub_dataset_dict(files=[self.files[0]])
+        dataset = init_dataset(dataset_dict, extra_kwargs={"seq_ordering": "default"}, parent_dataset=self)
+        self.num_inputs = dataset.num_inputs
+        self.num_outputs = dataset.num_outputs
+        self.labels = dataset.labels
+        self._data_keys = dataset.get_data_keys()
 
     def _lazy_init_file_sizes(self):
         import tree
@@ -224,17 +218,7 @@ class DistributeFilesDataset(CachedDataset2):
             _get_key_for_file_tree(t): sum((os.path.getsize(fn) for fn in tree.flatten(t)), 0) for t in self.files
         }
 
-    def _lazy_init_file_cache_proc(self):
-        if self._file_cache:
-            return
-        self._file_cache = _FileCacheProc(
-            name=f"{self.__class__.__name__} {self.name}", file_cache_opts=self.file_cache_opts
-        )
-
     def __del__(self):
-        if self._file_cache:
-            try_run(self._file_cache.exit, kwargs={"join": False})
-            self._file_cache = None
         for k, worker in self._workers.items():
             try_run(worker.exit, kwargs={"join": False})
         self._workers.clear()
@@ -256,7 +240,6 @@ class DistributeFilesDataset(CachedDataset2):
             return True
 
         self._lazy_init_file_sizes()
-        self._lazy_init_file_cache_proc()
 
         full_epoch_0idx = (epoch - 1) // self.partition_epoch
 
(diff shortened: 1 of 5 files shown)


returnn/datasets/distrib_files.py
```diff
@@ -324,9 +306,7 @@ def _get_sub_dataset_dict(self, files: List[FileTree]) -> Tuple[Dict[str, Any],
                 f"{self}: sub dataset should have explicit seq_ordering "
                 f"(or seq_order_control_dataset for MetaDataset), got: {dataset_dict}"
             )
-        self._lazy_init_file_cache_proc()
-        dataset_dict, exit_hook = self._file_cache.handle_cached_files_in_config(dataset_dict)
```
NeoLegends, line 328: This is where the existing code would block the master process.

returnn/util/file_cache.py
```diff
@@ -53,11 +103,11 @@ class FileCache:
...
-        cache_directory: str = "$TMPDIR/$USER/returnn/file_cache",
-        cleanup_files_always_older_than_days: float = 31.0,
-        cleanup_files_wanted_older_than_days: float = 7.0,
-        cleanup_disk_usage_wanted_free_ratio: float = 0.2,  # try to free at least 20% disk space
-        num_tries: int = 3,  # retry twice by default
+        cache_directory: str = DEFAULT_CACHE_DIR,
+        cleanup_files_always_older_than_days: float = DEFAULT_CLEANUP_FILES_ALWAYS_OLDER_THAN_DAYS,
+        cleanup_files_wanted_older_than_days: float = DEFAULT_CLEANUP_FILES_WANTED_OLDER_THAN_DAYS,
+        cleanup_disk_usage_wanted_free_ratio: float = DEFAULT_CLEANUP_DISK_USAGE_WANTED_FREE_RATIO,
+        num_tries: int = DEFAULT_NUM_TRIES,
```
albertz, line 110: I don't see the benefit in rewriting the code like this. This just makes it unnecessarily more complicated. I would keep the old code.

returnn/util/file_cache.py
```diff
@@ -19,9 +19,59 @@
...
+
+
+DEFAULT_CACHE_DIR = "$TMPDIR/$USER/returnn/file_cache"
+DEFAULT_CLEANUP_FILES_ALWAYS_OLDER_THAN_DAYS = 31.0
+DEFAULT_CLEANUP_FILES_WANTED_OLDER_THAN_DAYS = 7.0
+DEFAULT_CLEANUP_DISK_USAGE_WANTED_FREE_RATIO = 0.2  # try to free at least 20% disk space
+DEFAULT_NUM_TRIES = 3  # retry twice by default
+
+
+cache_instance: Optional["FileCache"] = None
```
albertz, line 35:
```suggestion
cache_instance: Optional[FileCache] = None
```

returnn/util/file_cache.py
```diff
@@ -19,9 +19,59 @@
...
+
+
+def init_by_config(config: Config):
+    """Initializes the global file cache using values from `config`."""
+    global cache_instance
+    if cache_instance:
+        return
+    # TODO: just pass along all `file_cache_*` config values as kwargs?
```
albertz, line 43: Yes of course. There should not be single config option for every single arg, but instead just sth like `file_cache_opts`.

returnn/datasets/distrib_files.py
```diff
@@ -647,95 +624,3 @@ def _get(seq_idx: int) -> Optional[DatasetSeq]:
                 raise Exception(f"unknown msg {msg!r}")
     except KeyboardInterrupt:  # when parent dies
         pass
-
-
-class _FileCacheProc:
```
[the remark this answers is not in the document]
NeoLegends: See https://github.com/rwth-i6/returnn/issues/1544#issuecomment-2180485412. I'm flexible on this though. Currently (current master) we block the main thread for caching anyways.

returnn/util/file_cache.py
```diff
@@ -331,6 +332,17 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
 
 
+def get_instance(config: Optional[Config] = None) -> FileCache:
+    """
+    Returns a file cache instance potentially initialized by the global config.
+
+    Uses defaults if no global config is set.
+    """
+    config = config or get_global_config(raise_exception=False)
+    kwargs = config.value("file_cache_opts", {}) if config is not None else {}
```
albertz, line 342:
```suggestion
    config = config or get_global_config(return_empty_if_none=True)
    kwargs = config.typed_value("file_cache_opts") or {}
```

NeoLegends: Ah, nice!

returnn/datasets/distrib_files.py
albertz: For reference, I made #1549 to keep track of this issue.

But I think we can resolve this thread here.

albertz:
Also, this here is missing some logic to not do the full cleanup scan too frequently, e.g. via touching a `.recent_full_cleanup` file as mentioned [here](https://github.com/rwth-i6/returnn/issues/1544#issuecomment-2180755966).

returnn/util/file_cache.py
```diff
@@ -73,7 +74,6 @@ def __init__(
         self._cleanup_files_wanted_older_than_days = cleanup_files_wanted_older_than_days
         self._cleanup_disk_usage_wanted_free_ratio = cleanup_disk_usage_wanted_free_ratio
         self._touch_files_thread = _TouchFilesThread(cache_base_dir=self.cache_directory)
-        self._touch_files_thread.start()
```
albertz, line 76: I would simply set `self._touch_files_thread: Optional[_TouchFilesThread] = None`, i.e. not create the thread object at all.

albertz: Well ok, maybe not so relevant, maybe your current code is also fine.

NeoLegends: If we leave this none we need a check+init everywhere. I think the implementation as it is now is simpler in that regard. Also, shouldn't the thread object itself be cheap, and only allocate memory for the OS thread once started?

albertz: > only allocate memory for the OS thread once started?

That was exactly the part I was not sure about. I guess it's also implementation specific to CPython. But probably it should be fine.

returnn/util/file_cache.py
```diff
@@ -381,6 +392,7 @@ def __init__(self, *, interval: float = 1.0, cache_base_dir: str):
         self.files = defaultdict(int)  # usage counter
         self.interval = interval
         self.cache_base_dir = cache_base_dir
+        self._started = False
```
albertz, line 395: This is an attrib you are overwriting from the underlying `Thread` object. See failing test. I think the underlying `Thread` actually already does exactly this logic?

NeoLegends: ohhh damn! the docs say there will be a runtimeerror when you invoke start multiple times

NeoLegends:
```
if self._started.is_set():
    raise RuntimeError("threads can only be started once")
```

albertz:
Test case needs to be fixed:
```
  File "/home/runner/work/returnn/returnn/tests/test_Util.py", line 873, in test_file_cache
    line: print("Recent cleanup is ago:", time.monotonic() - cache._recent_full_cleanup_time, "secs")
    locals:
      print = <builtin> <built-in function print>
      time = <global> <module 'time' (built-in)>
      time.monotonic = <global> <built-in function monotonic>
      cache = <local> <returnn.util.file_cache.FileCache object at 0x7f9b7acbe650>
      cache._recent_full_cleanup_time = <local> !AttributeError: 'FileCache' object has no attribute '_recent_full_cleanup_time'
AttributeError: 'FileCache' object has no attribute '_recent_full_cleanup_time'
```

albertz (approved):

albertz:
I just noticed that now some i6_core tests fail, because this PR here now makes `tree` (`pip install dm-tree`) a mandatory dependency for RETURNN. Before it was optional (only needed when the user used some code where it was used).

So is this ok?

cc @curufinwe @michelwi @NeoLegends @JackTemaki @Atticus1806 @christophmluscher (not sure who else)

michelwi:
The dm-tree package is installed in all of our singularity images since it became an (optional) requirement back then. Also in all cases where returnn is used, a fixed commit is attached to the setup. Therefore non of our old setups should break.

> So is this ok?

yes from my side.

PR #1551, opened 2024-06-21
FileCache reliability improvements

### LINKED ISSUE #1547: `FileCache`: better cleaning, free more than just the minimum
Freeing exactly `need_at_least_free_space_size` can be too little. E.g.:

* Multiple procs copy at the same time (e.g. RETURNN `FileCache` itself, see #1542, or other procs). After the space was freed, when the copying starts, or maybe during the copying, maybe some of the space was already taken again.
* We actually need one more byte due to the current logic in `_copy_with_prealloc`.
* I'm not sure if the filesystem maybe needs a few more bytes, e.g. for creating directories or so.

Another problem is that the current `cleanup_files_wanted_older_than_days` default (7 days) is maybe way too high and is rarely used when it matters (to free enough space). Although, maybe some of the other suggestions below already are enough. Not sure.

Some suggestions:

- Fix `need_at_least_free_space_size = os.stat(src_filename).st_size + 1`.
- Always try to free more than `need_at_least_free_space_size`, maybe sth like `2 * need_at_least_free_space_size` or so. However, if this fails, it's only a warning. Only if we fail to free `need_at_least_free_space_size`, it would be an error.

Other suggestions?

### LINKED ISSUE #1548: `FileCache`: avoid cache-wide dir lock
It increases the startup time. E.g. in distributed training with multiple workers, all will try to copy for the current and the next epoch, and the copying for the next epoch could block the copying for the current epoch in some other workers.

### LINKED ISSUE #1550: `FileCache`: Race condition when removing empty directories
The empty dir check causes issues w/ file caching. I'm adding this condition to the retryable exceptions.

```
Process DistributeFilesDataset train file cache:
  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/returnn/returnn/datasets/distrib_files.py", line 734, in _file_cache_proc_loop
    res, files = file_cache.handle_cached_files_in_config(**kwargs)
  File "/returnn/returnn/util/file_cache.py", line 272, in handle_cached_files_in_config
    return tree.map_structure(_handle_value, config), res_files
  File "/usr/local/lib/python3.10/dist-packages/tree/__init__.py", line 435, in map_structure
    [func(*args) for args in zip(*map(flatten, structures))])
  File "/usr/local/lib/python3.10/dist-packages/tree/__init__.py", line 435, in <listcomp>
    [func(*args) for args in zip(*map(flatten, structures))])
  File "/returnn/returnn/util/file_cache.py", line 267, in _handle_value
    res = self.get_file(value.filename)
  File "/returnn/returnn/util/file_cache.py", line 121, in get_file
    raise e
  File "/returnn/returnn/util/file_cache.py", line 115, in get_file
    self._copy_file_if_needed(src_filename, dst_filename)
  File "/returnn/returnn/util/file_cache.py", line 316, in _copy_file_if_needed
    _copy_with_prealloc(src_filename, dst_tmp_filename)
  File "/returnn/returnn/util/file_cache.py", line 348, in _copy_with_prealloc
    with open(dst, "wb") as dst_file:
FileNotFoundError: [Errno 2] No such file or directory: '[redacted]'
Waiting for lock-file: /tmp/wmichel/returnn/file_cache/dir.lock
Waiting for lock-file: /tmp/wmichel/returnn/file_cache/dir.lock
FileCache: Copy file [redacted] to cache
FileCache: Remove empty dir [redacted]
EOFError:
This exception is thrown by __iter__ of ChunkingIterDataPipe()
Unhandled exception <class 'EOFError'> in thread <_MainThread(MainThread, started 139974150349248)>, proc 584170.
```

NeoLegends:
* Closes #1547
* Closes #1548 
* Closes #1550

### DIFF
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -283,6 +283,7 @@ class FileCache:
         Copy the file to the cache directory.
         """
         if self._check_existing_copied_file_maybe_cleanup(src_filename, dst_filename):
+            print(f"FileCache: using existing file {dst_filename}")
             os.utime(dst_filename, None)  # touch
             return
 
@@ -291,7 +292,9 @@ class FileCache:
         os.makedirs(dst_dir, exist_ok=True)
 
         # Copy the file, while holding a lock. See comment on lock_timeout above.
-        with LockFile(directory=self.cache_directory, name="dir.lock", lock_timeout=self._lock_timeout) as lock:
+        with LockFile(
+            directory=dst_dir, name=os.path.basename(dst_filename) + ".lock", lock_timeout=self._lock_timeout
+        ) as lock:
             # Maybe it was copied in the meantime, while waiting for the lock.
             if self._check_existing_copied_file_maybe_cleanup(src_filename, dst_filename):
                 return
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -114,6 +114,8 @@ class FileCache:
             try:
                 self._copy_file_if_needed(src_filename, dst_filename)
                 break
+            except FileNotFoundError as e:
+                last_error = e
             except OSError as e:
                 if e.errno == errno.ENOSPC:
                     last_error = e
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -54,8 +54,8 @@ class FileCache:
         self,
         *,
         cache_directory: str = "$TMPDIR/$USER/returnn/file_cache",
-        cleanup_files_always_older_than_days: float = 31.0,
-        cleanup_files_wanted_older_than_days: float = 7.0,
+        cleanup_files_always_older_than_days: float = 7.0,
+        cleanup_files_wanted_older_than_days: float = 1.0,
         cleanup_disk_usage_wanted_free_ratio: float = 0.2,  # try to free at least 20% disk space
         cleanup_disk_usage_wanted_multiplier: float = 2.0,  # try to free 2x the required space for a file
         num_tries: int = 3,  # retry twice by default
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -57,6 +57,7 @@ class FileCache:
         cleanup_files_always_older_than_days: float = 31.0,
         cleanup_files_wanted_older_than_days: float = 7.0,
         cleanup_disk_usage_wanted_free_ratio: float = 0.2,  # try to free at least 20% disk space
+        cleanup_disk_usage_wanted_multiplier: float = 2.0,  # try to free 2x the required space for a file
         num_tries: int = 3,  # retry twice by default
     ):
         """
@@ -66,12 +67,16 @@ class FileCache:
         :param cleanup_files_wanted_older_than_days: if cleanup_disk_usage_wanted_free_ratio not reached,
             cleanup files older than this.
         :param cleanup_disk_usage_wanted_free_ratio: try to free at least this ratio of disk space.
+        :param cleanup_disk_usage_wanted_multiplier: when making space for a new file, try to free at
+            least this times as much space.
         :param num_tries: how many times to try caching a file before giving up
         """
         self.cache_directory = expand_env_vars(cache_directory)
         self._cleanup_files_always_older_than_days = cleanup_files_always_older_than_days
         self._cleanup_files_wanted_older_than_days = cleanup_files_wanted_older_than_days
         self._cleanup_disk_usage_wanted_free_ratio = cleanup_disk_usage_wanted_free_ratio
+        assert cleanup_disk_usage_wanted_multiplier >= 1.0
+        self._cleanup_disk_usage_wanted_multiplier = cleanup_disk_usage_wanted_multiplier
         self._touch_files_thread = _TouchFilesThread(cache_base_dir=self.cache_directory)
         self._touch_files_thread.start()
         self._recent_full_cleanup_time = float("-inf")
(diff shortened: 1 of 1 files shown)


returnn/util/file_cache.py
```diff
@@ -54,9 +54,10 @@ def __init__(
         self,
         *,
         cache_directory: str = "$TMPDIR/$USER/returnn/file_cache",
-        cleanup_files_always_older_than_days: float = 31.0,
-        cleanup_files_wanted_older_than_days: float = 7.0,
+        cleanup_files_always_older_than_days: float = 7.0,
```
albertz, line 57: I think I would keep the existing default for `cleanup_files_always_older_than_days`.
```suggestion
        cleanup_files_always_older_than_days: float = 31.0,
```
If we have otherwise enough space (and all the other logic should be responsible for that), there is no real reason to delete it. Maybe it becomes useful. 31 days is also somewhat arbitrary but I thought it to not be too long but also not too short (from own experience, how long I could maybe leave some data untouched after I would use it again).

Note that there is also tmpreaper running anyway. I'm not sure what settings we use there.

returnn/util/file_cache.py
```diff
@@ -114,6 +119,8 @@ def get_file(self, src_filename: str) -> str:
             try:
                 self._copy_file_if_needed(src_filename, dst_filename)
                 break
+            except FileNotFoundError as e:
```
albertz, line 122: I don't understand why you catch this. This sounds like an exception which should immediately lead to an error. We should not repeat when this happens.

NeoLegends: See #1550. This error occurs when the target directory is removed. If the source file is missing, the retries don't hurt (and will simply delay the process for 3s, but 🤷🏼‍♂️).

returnn/util/file_cache.py
```diff
@@ -282,24 +290,24 @@ def _copy_file_if_needed(self, src_filename: str, dst_filename: str):
...
 
         # Copy the file, while holding a lock. See comment on lock_timeout above.
-        with LockFile(directory=self.cache_directory, name="dir.lock", lock_timeout=self._lock_timeout) as lock:
+        with LockFile(
+            directory=dst_dir, name=os.path.basename(dst_filename) + ".lock", lock_timeout=self._lock_timeout
+        ) as lock:
             # Maybe it was copied in the meantime, while waiting for the lock.
             if self._check_existing_copied_file_maybe_cleanup(src_filename, dst_filename):
+                print(f"FileCache: using existing file {dst_filename}")
+                os.utime(dst_filename, None)  # touch
```
albertz, line 304: I was not doing this touch in my earlier code because I thought when some other proc has just copied it (before the lock, it was not there...), it should already be up-to-date. But it probably doesn't matter.

NeoLegends: exactly

returnn/util/file_cache.py
NeoLegends: I'm fine with this.

returnn/util/file_cache.py
albertz: Ah, but then I think I would make this more explicit. Check for this exception in `_copy_with_prealloc` when opening the `dst` file, and if you get it there, rethrow a custom exception (`_CannotOpenDirWasRemovedException`or whatever), and catch this here.

But despite this, I wonder a bit how this could even happen. The empty dir cleanup should not clean dirs which were recently created.

albertz: I think I would not do this logic at all currently. To me it sounds like there was a different bug that the directory was cleaned (removed) although it should not have been removed. If this is done correctly, we never should run into the problem here, thus I would also not add this logic, but rather fix the bug itself.

The question is, why was it removed. I think we did not properly update its time. Why? I think the cache dir lock was not doing this anymore. Before, via `with self._touch_files_thread.files_added_context([dst_dir, lock.lockfile]):`, we kept this dir alive and updated.

Maybe also the check `cur_time - os.stat(root).st_mtime <= cur_used_time_threshold` when cleaning up empty dirs is really a bit too aggressive.

NeoLegends: I'm actually not a big fan of removing the retry logic. I get your point that we should look at the origin of the issue, but adding multiple layers of resilience to a piece of code that is inherently very prune to race conditions and messy behavior does not hurt IMO.

albertz: There are multiple problems with it:

* It adds further complexity.
* It hides other bugs and makes debugging harder.

Without really having any benefit. I don't see how it can possibly get this wrong when there is no bug in our code. I don't like to add workarounds for bugs in our code just because we don't want to fix the bugs. After we fixed the bug, this really should never ever be necessary. There is no race condition here.

And the bug fix is probably very simple also. As said, I think already by doing the lock again on file level, the `files_added_context` on the lock file should handle this already (it will also update the dirs, see `_TouchFilesThread`).

NeoLegends: Then we should be good now. The lock is back at file level after this PR.

albertz: Yes exactly. Unless we miss something. But then there is a bug we should fix. Because I think we really should not run into this problem otherwise.

Maybe just to be a bit saver, we should increase the threshold in `cur_time - os.stat(root).st_mtime <= cur_used_time_threshold` when cleaning the dirs.

albertz: So, choosing between (1) some more crashes now but overall more stable code after solving this vs (2) less crashes now but maybe some bugs hidden and we forget about it, I would choose (1). (2) leads to the situation that RETURNN is buggier over time, and this over medium/long term is even more annoying to people.

albertz (approved):

albertz:
Regarding the discussion on retrying on `FileNotFoundError`, if we hit this again, we can reconsider. I would also like the opinion of @michelwi, @Icemole, @curufinwe, maybe others on this.

PR #1561, opened 2024-07-01
Add warning for not using `num_workers > 0` in torch

### LINKED ISSUE #1560: PyTorch Distributed Training: File descriptors opened and never closed
When using distributed training w/ PyTorch v2.30, RETURNN opens file descriptors every subepoch that are then never closed.

Comparing the output of `lsof -l` for the same process w/ a few subepochs difference one gets diffs like this:
image

with blocks of FDs like below repeating:

```
pt_main_t 13339 17144239  100r     FIFO               0,13      0t0             2311492 pipe
pt_main_t 13339 17144239  101w     FIFO               0,13      0t0             2311492 pipe
pt_main_t 13339 17144239  102u  a_inode               0,14        0               13849 [eventpoll]
pt_main_t 13339 17144239  103u     IPv4            2311493      0t0                 TCP g-02.apptek.local:33695 (LISTEN)
pt_main_t 13339 17144239  104u     IPv4            2311497      0t0                 TCP g-02.apptek.local:33695->localhost:34500 (ESTABLISHED)
```

Using the pipe FD 101 in the example above and some quick and dirty tracing code (see below) that I wrapped around various functions in the Pytorch engine I was able to narrow down the pipe FD's construction to

```
-> Engine.train_epoch
-> DataLoader.__iter__
-> DataLoader._get_iterator
-> _SingleProcessDataLoaderIter.__init__
-> _BaseDataLoaderIter.__init__
-> torch.distributed.new_group(backend="gloo")
-> torch.distributed._new_group_with_tag
-> torch.distributed._new_process_group_helper
-> backend_class = ProcessGroupGloo(backend_prefix_store, group_rank, group_size, timeout=timeout)
-> native code
```

The only reason for existence of this process group is for distributing a random seed between the worker group members. Note, in this example it does not make a difference if we use `num_workers > 0` or not as both the `_SingleProcessDataLoaderIter` and `_MultiProcessDataLoaderIter` inherit from `_BaseDataLoaderIter`. What I'm wondering about is why it's not destructed and the FDs closed once the data loader iter goes out of scope.

Torch allows re-using workers using the "persistent_workers"-flag, and in that case the workers are not recreated every subepoch but just reset, which means Torch won't create new process groups and the FDs won't accumulate. This can be a workaround for now.

lsof-Diff with `num_workers=1, persistent_workers=True` (notice the missing green blocks of added lines visible in the other diff): 
image

EDIT: I found out we already default to setting `persistent_workers=True` when `num_workers` is set. So in the case with `torch_dataloader_opts = {"num_workers": 1}` we are good.

----

Tracing code:

```python
from .basic import is_valid_fd
from returnn.log import log

num_fd = 101


def fd(name):

    def decorator(func):
        name_ = name or func.__name__

        def unfold(val):
            for v in val:
                yield v
                print(f"after value {name_}, FD {num_fd} is {is_valid_fd(num_fd)}", file=log.v3)

        def wrapped(*args, **kwargs):
            print(f"before {name_}, FD {num_fd} is {is_valid_fd(num_fd)}", file=log.v3)
            try:
                val = func(*args, **kwargs)
                return unfold(val) if hasattr(val, "__next__") else val
            finally:
                print(f"after {name_}, FD {num_fd} is {is_valid_fd(num_fd)}", file=log.v3)

        return wrapped

    return decorator


from contextlib import contextmanager


@contextmanager
def fd_c(name):
    try:
        print(f"before {name}, FD {num_fd} is {is_valid_fd(num_fd)}", file=log.v3)
        yield
    finally:
        print(f"after {name}, FD {num_fd} is {is_valid_fd(num_fd)}", file=log.v3)
```

NeoLegends:
Not sure if I need to mention the specific issue this works around, but I think it's worth having mentioned that this not only improves performance, but also works around bugs.

I made this a PR and not a direct commit to master to bikeshed the text.

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -27,6 +27,7 @@ import numpy
 import torch
 import torch.utils.data
 
+from returnn.log import log
 from returnn.util.basic import NumbersDict
 
 
@@ -306,7 +307,7 @@ def create_data_loader_from_batches(
     if loader_opts is None:
         loader_opts: Dict[str, Any] = {}
 
-    if loader_opts.get("num_workers"):
+    if loader_opts.get("num_workers", 0) > 0:
         loader_opts = loader_opts.copy()
         loader_opts.setdefault("persistent_workers", True)
         loader_opts["worker_init_fn"] = _DataLoaderWorkerInitFunc(
@@ -328,6 +329,12 @@ def create_data_loader_from_batches(
                 # See _DataLoaderWorkerPreInitFunc below, https://github.com/rwth-i6/returnn/issues/1495.
                 process_pre_init_func=SubProcCopyGlobalConfigPreInitFunc()
             )
+    else:
+        log.print_warning(
+            "Not using dedicated worker processes for torch data loading.\n"
+            'It is strongly recommended to set torch_dataloader_opts = {"num_workers": 1}\n'
+            "to improve GPU utilization (and to work around issues wrt. leaking file descriptors)."
+        )
 
     return torch.utils.data.DataLoader(
         batches_dataset,
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -307,7 +307,7 @@ def create_data_loader_from_batches(
     if loader_opts is None:
         loader_opts: Dict[str, Any] = {}
 
-    if loader_opts.get("num_workers", 0) > 0:
+    if loader_opts.get("num_workers"):
         loader_opts = loader_opts.copy()
         loader_opts.setdefault("persistent_workers", True)
         loader_opts["worker_init_fn"] = _DataLoaderWorkerInitFunc(
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -332,8 +332,10 @@ def create_data_loader_from_batches(
     else:
         log.print_warning(
             "Not using dedicated worker processes for torch data loading.\n"
-            'It is strongly recommended to set torch_dataloader_opts = {"num_workers": 1}\n'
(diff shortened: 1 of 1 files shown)


albertz:
I'm not sure we really need this warning. There are always countless of bugs/issues in older versions of PyTorch or whatever other library, and we are not really printing warnings when the user have not installed the latest version.

Also, most users don't read warnings anyway.

Also, this warning is without really giving any context to the user. There should be some links to PyTorch issues (or RETURNN issues) with further details on the reasoning.

returnn/torch/data/pipeline.py
```diff
@@ -306,7 +307,7 @@ def create_data_loader_from_batches(
     if loader_opts is None:
         loader_opts: Dict[str, Any] = {}
 
-    if loader_opts.get("num_workers"):
+    if loader_opts.get("num_workers", 0) > 0:
```
albertz, line 310: Why this change?

NeoLegends: If for some reason num_workers is set, but set to 0.

NeoLegends: Ah, 0 is falsey. Never mind.

albertz: Well but it's slightly different. E.g. when `num_workers` is set to a string, this crashes now. Or when `num_workers` is negative, it would not apply the logic anymore now. However, I specifically wanted that it would also apply the logic in those cases. Who knows what `DataLoader` might support in the future for `num_workers`. If it is some invalid value for `num_workers`, this is up to `DataLoader` to handle. RETURNN should mostly ignore it.

NeoLegends: I see, then I'll revert.

albertz (approved):

PR #1563, opened 2024-07-02
Default torch DataLoader num_workers to 1

### LINKED ISSUE #1560: PyTorch Distributed Training: File descriptors opened and never closed

NeoLegends:
Closes #1560 

This is a pretty small, but significant change, so I did make this into a PR and did not directly commit to master.

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -306,9 +306,19 @@ def create_data_loader_from_batches(
     """
     if loader_opts is None:
         loader_opts: Dict[str, Any] = {}
+    else:
+        loader_opts = loader_opts.copy()
+
+    # Make sure to use workers unless specified otherwise to ensure reasonable GPU
+    # utilization and to work around some issues wrt. overflowing ulimits when
+    # workers are non-persistent.
+    #
+    # See for context:
+    # - https://github.com/rwth-i6/returnn/issues/1560
+    # - https://github.com/pytorch/pytorch/issues/129868
+    loader_opts.setdefault("num_workers", 1)
 
     if loader_opts.get("num_workers"):
-        loader_opts = loader_opts.copy()
         loader_opts.setdefault("persistent_workers", True)
         loader_opts["worker_init_fn"] = _DataLoaderWorkerInitFunc(
             other_worker_init_fn=loader_opts.get("worker_init_fn")
@@ -329,14 +339,6 @@ def create_data_loader_from_batches(
                 # See _DataLoaderWorkerPreInitFunc below, https://github.com/rwth-i6/returnn/issues/1495.
                 process_pre_init_func=SubProcCopyGlobalConfigPreInitFunc()
             )
-    else:
-        log.print_warning(
-            "Not using dedicated worker processes for torch data loading.\n"
-            'It is strongly recommended to set torch_dataloader_opts = {"num_workers": 1} '
-            "to improve GPU utilization and to work around some issues.\n"
-            "See e.g. https://github.com/rwth-i6/returnn/issues/1560 for an issue that is fixed by "
-            "using dedicated worker processes."
-        )
 
     return torch.utils.data.DataLoader(
         batches_dataset,
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -27,7 +27,6 @@ import numpy
 import torch
 import torch.utils.data
 
-from returnn.log import log
 from returnn.util.basic import NumbersDict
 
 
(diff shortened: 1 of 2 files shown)


albertz (approved):

albertz:
Ok, two failing tests. Can you check what's the problem there?

albertz (changes requested):
Check/fix tests.

NeoLegends:
Sure, asap

albertz:
I meant `test_torch_engine_train` and `test_torch_engine_train_exception`. I think sth with serialization fails, probably due to the change in this PR. So this is important. (The question is also whether potentially some real existing setups will break now for similar reasons.)

NeoLegends:
So, it seems like setups that have `nn.Module` classes declared in a non-global scope will break. I don't think many people have setups like that though, as we normally use imports to declare `get_model`.

NeoLegends:
Can't (un)pickle local classes or functions.

albertz:
> including the config's `get_model`, which in the test case points to a locally (i.e. in the scope of the test function) declared class.

Ah that's how it got the reference, I see. I wonder though if this might be problematic at some point. Because for the dataset, we don't care about `get_model` or many other things from the config. (Actually, why do we care at all about the config?)

PR #1569, opened 2024-07-02
Torch: optionally apply `cleanup_old_models` logic to optimizer states

### LINKED ISSUE #1565: PyTorch/RF (?): choosing on which epochs to save optimizer state
Currently we keep two optimizer states around, AFAIK from the last two epochs. Being able to choose which epochs to additionally save the optimizer state of enables trying out e.g. different LR cooldown strategies starting from the same training.

NeoLegends:
Closes #1565 

This PR adds a config flag to set that the existing logic for keeping checkpoints around is applied to the optimizer states as well, i.e. the optimizer states are kept just like the checkpoints. This makes working w/ long-running trainings more consistent where we want to branch off at intermediate points.

Not sure if this works yet, hence draft.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -929,7 +929,7 @@ class Engine(EngineBase):
 
         # keep only the last two optimizer states (two in case one file gets corrupted)
         clean_epoch = self.epoch - 2
-        apply_keep_logic = self.config.bool("torch_apply_keep_epoch_to_optim_states", False)
+        apply_keep_logic = self.config.bool("apply_cleanup_old_models_to_optim_states", False)
         if clean_epoch > 0 and not apply_keep_logic:
             filename = self.get_epoch_model_filename(epoch=clean_epoch) + ".opt" + util.get_model_filename_postfix()
             if os.path.isfile(filename):
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -1079,12 +1079,14 @@ class Engine(EngineBase):
         # They consist of a file with the extension ".pt" and potentially an optimizer state with extension ".opt.pt"
 
         count_bytes = 0
-        assert os.path.exists(filename + ".pt")
-        count_bytes += os.stat(filename + ".pt").st_size
-        os.remove(filename + ".pt")
-        if os.path.exists(filename + "opt.pt"):
-            count_bytes += os.stat(filename + "opt.pt").st_size
-            os.remove(filename + "opt.pt")
+        fname = filename + ".pt"
+        assert os.path.exists(fname)
+        count_bytes += os.stat(fname).st_size
+        os.remove(fname)
+        opt_fname = filename + ".opt.pt"
+        if os.path.exists(opt_fname):
+            count_bytes += os.stat(opt_fname).st_size
+            os.remove(opt_fname)
         assert count_bytes > 0
         return count_bytes
 
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -929,7 +929,8 @@ class Engine(EngineBase):
 
         # keep only the last two optimizer states (two in case one file gets corrupted)
         clean_epoch = self.epoch - 2
-        if clean_epoch > 0:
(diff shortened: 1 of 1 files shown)


NeoLegends:
This works now.

returnn/torch/engine.py
```diff
@@ -1078,12 +1079,14 @@ def delete_model(filename):
...
-            count_bytes += os.stat(filename + "opt.pt").st_size
-            os.remove(filename + "opt.pt")
+        fname = filename + ".pt"
+        assert os.path.exists(fname)
+        count_bytes += os.stat(fname).st_size
+        os.remove(fname)
+        opt_fname = filename + ".opt.pt"
+        if os.path.exists(opt_fname):
+            count_bytes += os.stat(opt_fname).st_size
+            os.remove(opt_fname)
```
albertz, line 1089: There is no change here?

albertz: Oh wait, there is, it's `".opt.pt"` vs `"opt.pt"`?

albertz:
Is such option `torch_apply_keep_epoch_to_optim_states` really what the users want? Or maybe you only want a `keep_epochs_optim_states: list[int]`?

I feel like you rarely would want to apply exactly the same logic from `cleanup_old_models` also to the optimizer states, but instead you maybe want to have a subset of that? 

But as I'm currently not a user of this, I think those people who want to use this should decide.

Btw, I would also rename this config option. First of all, I think we don't need the `torch_` prefix. And then, why `apply_keep_epoch`? You mean that the `cleanup_old_models` logic is applied here, right? So the canonical name would be `apply_cleanup_old_models_to_optim_states`?

returnn/torch/engine.py
NeoLegends: Yes. This was actually a bug. If it wasn‘t for the original cleanup logic those files would have been missed.

michelwi:
> Or maybe you only want a `keep_epochs_optim_states: list[int]`?

or maybe a `cleanup_old_optim_states` with all the options as `cleanup_old_models`?

albertz:
> or maybe a `cleanup_old_optim_states` with all the options as `cleanup_old_models`?

I think either the user want to have the same logic as `cleanup_old_models` (exactly same settings then), i.e. like `apply_cleanup_old_models_to_optim_states`, or the user want to have a subset which is just a list of epochs, like `keep_epochs_optim_states`.

I don't really see a valid use case for having all the flexibility there to support e.g. `keep_best_n` but to use different settings than in `cleanup_old_models`?

Maybe `keep_last_n`.

W.r.t. implementing this, while not difficult, this would require probably most changes to the code, as our `cleanup_old_models` method is not intended to be applied only to optim states.

NeoLegends:
> Maybe keep_last_n.

I don't think that this is what is wanted, at least in the case when we run cooldowns of intermediate epochs before the main training is done.

E.g. for LR schedules like this: A09F5E5C-E4ED-4A9B-838B-EC698171645F_1_105_c
... it makes sense to be able to specify the epochs of which to keep the optimizer state, i.e. we want the same logic as `cleanup_old_models`.

The subset approach could also be valid in this case, too, although I'm not sure whether the additional complexity is really worth it. And then we miss the case where we want to e.g. continue training from the best N checkpoints (i.e. same logic as `keep_best_n`).

I'll rename the flag first.

albertz (approved):
I don't care too much whether we use `apply_cleanup_old_models_to_optim_states` or the different approach like `keep_epochs_optim_states`. I'm fine with both, i.e. also with the current PR.

JackTemaki:
> I don't care too much whether we use `apply_cleanup_old_models_to_optim_states` or the different approach like `keep_epochs_optim_states`. I'm fine with both, i.e. also with the current PR.

I would leave this to @michelwi, I have no clear opinion on that.

michelwi:
> > I don't care too much whether we use `apply_cleanup_old_models_to_optim_states` or the different approach like `keep_epochs_optim_states`. I'm fine with both, i.e. also with the current PR.
> 
> I would leave this to @michelwi, I have no clear opinion on that.

I am fine with the current implementation. It seems to me as the simplest solution to the problem.

michelwi (approved):

PR #1572, opened 2024-07-03
DistributeFilesDataset: Distribute files more evenly

### LINKED ISSUE #1570: DistributeFilesDataset: _distribute_evenly_by_size suboptimal for multi-gpu sharding
I observed a suboptimal behavior for the [_distribute_evenly_by_size](https://github.com/rwth-i6/returnn/blob/bd0a83f75ebc849dcddf01a9e5082542ea0b8188/returnn/datasets/distrib_files.py#L332) of DistributeFilesDataset when using `distrib_shard_files` option.

Let the number of equally sized HDF files be `4x7+1 = 29` and the number of GPUs be `8`. 
Then the `avg_size_per_sub_epoch` is `29/8 = 3.625`.
Now every time 3 Files are added to each bin [(because 3 is smaller than the average)](https://github.com/rwth-i6/returnn/blob/bd0a83f75ebc849dcddf01a9e5082542ea0b8188/returnn/datasets/distrib_files.py#L374) and then a 4th file is added [because 4 is closer to 3.6 than 3](https://github.com/rwth-i6/returnn/blob/bd0a83f75ebc849dcddf01a9e5082542ea0b8188/returnn/datasets/distrib_files.py#L380). 
This now leaves me with 7 GPUs that each have 4 HDF files and one GPU with just 1 file. 

And once the last GPU finishes its single HDF file, all other GPUs stop processing the remaining files and out of the 29 files only 8 have been processed.

NeoLegends:
We now take the `avg_size_per_sub_epoch` over the remaining data in every iteration, improving the accuracy of the heuristics for placing the current item in the current bin or in the next.

Note that this changes the behavior of how files are distributed over shards and/or num epochs. But as discussed in #1570, this should be still ok.

I also added a smoke test to make sure the function distributes reasonably in simple cases where the files are all roughly the same size.

Fix #1570


tests/test_Dataset.py
```diff
@@ -640,7 +640,25 @@ def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]]):
     _test([1, 1, 78, 120], 4, [[1], [1], [78], [120]])
     _test([1, 1, 1, 56, 141], 5, [[1], [1], [1], [56], [141]])
     _test([1, 1, 1, 56, 141], 4, [[1, 1], [1], [56], [141]])
-    _test([5, 5] + [10] * 7, 5, [[5, 5, 10], [10, 10], [10, 10], [10], [10]])
+    _test([5, 5] + [10] * 7, 5, [[5, 5, 10], [10, 10], [10], [10, 10], [10]])
+    _test(
+        [1] * 29,
+        8,
+        [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1], [1, 1, 1, 1], [1, 1, 1], [1, 1, 1, 1], [1, 1, 1]],
+    )
```
NeoLegends, line 648: @michelwi This is a much more reasonable output for your edge case.

returnn/datasets/distrib_files.py
```diff
@@ -355,39 +355,45 @@ def _distribute_evenly_by_size(
...
         size_taken = 0
+        total_size_taken = 0
         for i, f_tree in enumerate(files_order):
             size = file_sizes[_get_key_for_file_tree(f_tree)]
             num_remaining = len(files_order) - i
-            if num_remaining <= num_bins - bin_idx - 1:
-                # All remaining sub epochs must be filled.
-                assert files_per_bin[bin_idx]
+
+            try:
```
albertz, line 363: Why this `try` block?

NeoLegends: Yeah, I'm not so sure about it. I'm using it to DRY the update to the average via the finally block. Otherwise with the `continue`s within the loop I'd have to repeat that code multiple times.

tests/test_Dataset.py
```diff
@@ -640,7 +640,25 @@ def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]]):
...
+        files = [f"file-{i}" for i in range(len(sizes))]
+        file_sizes = {f: s for f, s in zip(files, sizes)}
+        res = DistributeFilesDataset._distribute_evenly_by_size(
+            num_bins=partition_epoch, file_sizes=file_sizes, files_order=files
+        )
+        sizes = numpy.sum([[file_sizes[s] for s in v] for v in res], axis=-1)
+        assert numpy.std(sizes) <= max_stddev_percent * avg_per_bin
+
+    # This test verifies the algorithm distributes evenly in easy cases
+    _test_stddev(numpy.random.uniform(low=97, high=103, size=(100,)), 20, 0.03)
```
albertz, line 661: All tests should be deterministic. I.e. you should explicitly create `numpy.random.RandomState(42)` or whatever.

returnn/datasets/distrib_files.py
albertz: No we should definitely not do this. It's like an obfuscated goto what you do here...

returnn/datasets/distrib_files.py
```diff
@@ -355,39 +355,45 @@ def _distribute_evenly_by_size(
...
                 bin_idx += 1
-                files_per_bin[bin_idx].append(f_tree)
-                size_taken = size
-                continue
-            if bin_idx == num_bins - 1:
-                # We are done. Just add the rest to the last sub epoch.
-                files_per_bin[bin_idx].append(f_tree)
-                size_taken += size
-                continue
-            assert size_taken <= avg_size_per_sub_epoch
```
albertz, line 373: You could simply add `avg_size_per_sub_epoch = (total_size - total_size_taken + size_taken) / (num_bins - bin_idx)` here.

(Btw., I wonder whether the `assert` is still true in general now.)

michelwi: I would update the avg value only upon reaching a new bin idx. Then size_taken should be zero and thereby smaller than the new avg.

returnn/datasets/distrib_files.py
```diff
@@ -6,7 +6,7 @@
 
 from __future__ import annotations
 
-from typing import Optional, Any, Dict, Tuple, List, Sequence, Callable, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
```
albertz, line 9: Both the old order but also the new order are very inconsistent to how we usually order these in RETURNN.
We usually start with generic ones (`Optional`, `Union`, `Any`), and it gets more specific to the right, i.e. things like `List`, `Tuple`, `Dict` come last. `Sequence` before those. `Callable`, `TypeVar` etc usually also before.

albertz:
```suggestion
from typing import Union, Optional, Any, Callable, Sequence, Tuple, List, Dict
```

albertz: I would say, let's move this to a separate issue/discussion and not discuss here. So here in this PR, I would either not change the import order at all now, or make it consistent to remaining RETURNN code.

albertz (approved):

michelwi (changes requested):

michelwi (approved):

Issue #1573, opened 2024-07-04
ConcatSeqsDataset with extended functionality


Stefanwuu:
Since it turns out helpful for some datasets to concatenate sequences during the training, some new functionality might be desired for future training usage.
- More generic implementation for alignment. The current implementation supports padding frames between concatenated sequences to a multiple of the frame rate to assure valid alignment after concatenation. But for alignments with raw WAV, the situation is more complicated. Therefore, I propose a strict length constraint for alignment.
- On-the-fly decision of whether to concatenate two sequences or not. The current implementation allows us to pre-define whether we have a single sequence or concatenated sequences. But in training, this means the model always sees some sequences as single and some as concatenated together. An on-the-fly decision of whether to concatenate or not might add more regularization. But I'm still considering how to do this within the current framework, any discussion is welcome.

albertz:
A couple of related thoughts (but but more) are in #292. (This is for TF, but most of the discussion is generic and can be applied in the same way to PT, or even be backend independent.)

In case of PT, I think an easy way right now is to implement that as another `IterDataPipe`, e.g. like `ChunkingIterDataPipe`. Currently in our PT `Engine._create_data_loader`, it is somewhat hardcoded what pipe transformations we construct, but we could make this more configurable/flexible, that the user could put arbitrary own things in between.

Note that there are specific assumptions on what kind of data flows through the pipe (e.g. how seq lens are stored), and this is currently not well defined, and we planned to replace that by the well defined `TensorDict` at some point (#1302). So if you depend on the current behavior, then this might break in the future.

Another approach would be to just extend `ConcatSeqsDataset`, or implement another more flexible dataset, and then the logic applies on the dataset, independent from the backend. I personally would maybe leave `ConcatSeqsDataset` untouched and make some generic `DynConcatSeqsDataset` or so where the user gives a subdataset (just like for `ConcatSeqsDataset`) and then some generic function which decided what sequences to concatenate, and another function where the user can do the concatenation in whatever way he/she wants.

Stefanwuu:
#1574

Stefanwuu:
[...]

Issue #1564, opened 2024-07-02
Dataset ctx_left/ctx_right extension: ctx_clip_to_valid option


albertz:
When you have some feature extraction or downsampling module inside your network with padding=valid (standard stft is also padding=valid), it means that for a kernel size K, K-1 frames left/right will be gone. Ignoring the downsampling aspect, you have:

$T_{\textrm{in}} - (K - 1) = T_{\textrm{out}}$

Now, with chunking, you would get:

$T_{\textrm{in}} / N - (K - 1) \ne T_{\textrm{out}} / N$

Using ctx_left/ctx_right cannot fix this. Assuming $K$ uneven, if you set ctx_left=ctx_right=(K-1)/2, then you get:

$T_{\textrm{in}} / N + (K - 1) - (K - 1) = T_{\textrm{in}} / N \ne T_{\textrm{out}} / N$

Additionally to ctx_left/ctx_right, you want to remove some context on the sequence level, not on chunk level, such that you have:

$(T_{\textrm{in}} - (K - 1)) / N = T_{\textrm{out}} / N$

We (@robin-p-schmitt) thought about adding some `global_ctx_left`/`global_ctx_right`, and you would set global_ctx_left=global_ctx_right=-(K-1)/2 for this example.

But then we realized, this effectively means that at the left boundary on sequence level and on the right boundary on sequence level, the local (chunk level) ctx and global ctx would cancel out. Currently we have this code:
```python
for seq_idx, t_start, t_end in self.iterate_seqs(recurrent_net=recurrent_net, used_data_keys=used_data_keys):
    ...
    t_start -= self.ctx_left
    t_end += self.ctx_right
```
For this example, we could also introduce a `ctx_clip_to_valid` option which would clip `t_start`/`t_end` to valid frames, like so:
```python
if ctx_clip_to_valid:
    t_start = max(t_start, 0)
    t_end = min(t_end, T)  # t_end is exclusive
```
`ctx_clip_to_valid` is conceptually a bit simpler than global_ctx_left/global_ctx_right, and solves at least our problem, so we suggest to go with this simpler solution.

Note, the code I posted above is currently only used for TF. For PyTorch, we have the independent `ChunkingIterDataPipe` implementation, which even does not currently support ctx_left/ctx_right. So I would suggest to add `ctx_left`/`ctx_right` and `ctx_clip_to_valid` there.

cc @michelwi, @JackTemaki, @mmz33, @Icemole, @kuacakuaca

[...]

PR #1574, opened 2024-07-04
ConcatSeqsDataset pad_narrow_data_to_multiple_of_target_len

Stefanwuu:
To allow training with alignments of raw wav data, we make sure every concatenated sequence has strictly aligned length

### DIFF
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -1536,10 +1536,18 @@ class ConcatSeqsDataset(CachedDataset2):
                     key,
                     sub_dataset_keys,
                 )
-            for key in self.force_align:
+            for key in self.repeat_in_between_last_frame_up_to_multiple_of:
                 assert (
                     key in sub_dataset_keys
-                ), "%s: force_align key %r not in sub dataset data-keys %r" % (
+                ), "%s: repeat_in_between_last_frame_up_to_multiple_of key %r not in sub dataset data-keys %r" % (
+                    self,
+                    key,
+                    sub_dataset_keys,
+                )
+            for key in self.pad_truncate_data_to_multiple_of:
+                assert (
+                    key in sub_dataset_keys
+                ), "%s: pad_truncate_data_to_multiple_of key %r not in sub dataset data-keys %r" % (
                     self,
                     key,
                     sub_dataset_keys,
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -1580,7 +1580,7 @@ class ConcatSeqsDataset(CachedDataset2):
                     len_diff = data.shape[0] - target_data.shape[0] * multiple
                     if len_diff > 0:
                         # if data longer than ref_data * frame_rate, truncate
-                        data = data[:-len_diff,]
+                        data = data[:-len_diff]
                     elif len_diff < 0:
                         # if data shorter than ref_data * frame_rate, pad by repeating last frame
                         data = numpy.concatenate([data] + [data[-1:]] * -len_diff, axis=0)
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -1417,7 +1417,7 @@ class ConcatSeqsDataset(CachedDataset2):
         self.seq_tag_delim = seq_tag_delim
         self.remove_in_between_postfix = remove_in_between_postfix or {}
         self.repeat_in_between_last_frame_up_to_multiple_of = repeat_in_between_last_frame_up_to_multiple_of or {}
-        self.pad_truncate_data_to_multiple_of = pad_truncate_data_to_target_length or {}
+        self.pad_truncate_data_to_multiple_of = pad_truncate_data_to_multiple_of or {}
         self.epoch_wise_filter = EpochWiseFilter(epoch_wise_filter) if epoch_wise_filter else None
         if isinstance(dataset, dict):
             dataset = dataset.copy()
@@ -1633,7 +1633,7 @@ class ConcatSeqsDataset(CachedDataset2):
     def get_total_num_seqs(self):
         """
         :rtype: int
-        """ 
+        """
         return len(self.full_seq_list)
 
 
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -1391,7 +1391,7 @@ class ConcatSeqsDataset(CachedDataset2):
         seq_tag_delim=";",
         remove_in_between_postfix=None,
         repeat_in_between_last_frame_up_to_multiple_of=None,
-        pad_truncate_data_to_multiple_of=None,
+        pad_truncate_data_to_multiple_of_target_len=None,
         use_cache_manager=False,
         epoch_wise_filter=None,
         **kwargs,
@@ -1407,7 +1407,7 @@ class ConcatSeqsDataset(CachedDataset2):
           Now it could happen that ceildiv(data_len1 + data_len2, 6) < align_len1 + align_len2.
           This option would repeat intermediate ending frames such that data_len1 % 6 == 0,
           by setting it to {"data": 6}.
-        :param dict[str,(str,int)]|None pad_truncate_data_to_multiple_of: data_key -> (target_key, multiple).
+        :param dict[str,(str,int)]|None pad_truncate_data_to_multiple_of_target_len: data_key -> (target_key, multiple).
             Force data_len == target_len * multiple for specified data and target.
             Similar to repeat_in_between_last_frame_up_to_multiple_of but also truncates the data.
             Mainly for raw wav alignment, where data_len could be slightly longer than target_len * multiple
@@ -1419,7 +1419,7 @@ class ConcatSeqsDataset(CachedDataset2):
         self.seq_tag_delim = seq_tag_delim
         self.remove_in_between_postfix = remove_in_between_postfix or {}
         self.repeat_in_between_last_frame_up_to_multiple_of = repeat_in_between_last_frame_up_to_multiple_of or {}
-        self.pad_truncate_data_to_multiple_of = pad_truncate_data_to_multiple_of or {}
+        self.pad_truncate_data_to_multiple_of_target_len = pad_truncate_data_to_multiple_of_target_len or {}
         self.epoch_wise_filter = EpochWiseFilter(epoch_wise_filter) if epoch_wise_filter else None
         if isinstance(dataset, dict):
             dataset = dataset.copy()
@@ -1546,9 +1546,9 @@ class ConcatSeqsDataset(CachedDataset2):
                     key,
                     sub_dataset_keys,
                 )
-            for key in self.pad_truncate_data_to_multiple_of:
+            for key in self.pad_truncate_data_to_multiple_of_target_len:
                 assert key in sub_dataset_keys, (
-                    f"{self}: pad_truncate_data_to_multiple_of key {key}"
+                    f"{self}: pad_truncate_data_to_multiple_of_target_len key {key}"
                     f" not in sub dataset data-keys {sub_dataset_keys}"
                 )
         for sub_seq_idx, sub_seq_tag in zip(sub_seq_idxs, sub_seq_tags):
@@ -1574,8 +1574,8 @@ class ConcatSeqsDataset(CachedDataset2):
                     if data.shape[0] % multiple != 0:
                         data = numpy.concatenate([data] + [data[-1:]] * (multiple - data.shape[0] % multiple), axis=0)
                         assert data.shape[0] % multiple == 0
-                if key in self.pad_truncate_data_to_multiple_of and sub_seq_idx != sub_seq_idxs[-1]:
-                    target_key, multiple = self.pad_truncate_data_to_multiple_of[key]
+                if key in self.pad_truncate_data_to_multiple_of_target_len and sub_seq_idx != sub_seq_idxs[-1]:
+                    target_key, multiple = self.pad_truncate_data_to_multiple_of_target_len[key]
                     target_data = self.sub_dataset.get_data(sub_seq_idx, target_key)
                     len_diff = data.shape[0] - target_data.shape[0] * multiple
                     if len_diff > 0:
--- a/returnn/datasets/meta.py
+++ b/returnn/datasets/meta.py
@@ -1391,6 +1391,7 @@ class ConcatSeqsDataset(CachedDataset2):
         seq_tag_delim=";",
         remove_in_between_postfix=None,
         repeat_in_between_last_frame_up_to_multiple_of=None,
+        pad_truncate_data_to_multiple_of=None,
         use_cache_manager=False,
         epoch_wise_filter=None,
         **kwargs,
@@ -1406,6 +1407,9 @@ class ConcatSeqsDataset(CachedDataset2):
           Now it could happen that ceildiv(data_len1 + data_len2, 6) < align_len1 + align_len2.
           This option would repeat intermediate ending frames such that data_len1 % 6 == 0,
           by setting it to {"data": 6}.
+        :param dict[str,(str,int)]|None pad_truncate_data_to_multiple_of: data_key -> (target_key, multiple).
+            to make sure the alignment is still valid after concatenation, we force length of every seq in data_key
+            to be length of seq in target_key * multiple
         :param bool use_cache_manager:
         :param dict[(int,int),dict] epoch_wise_filter: see :class:`EpochWiseFilter`
         """
@@ -1413,6 +1417,7 @@ class ConcatSeqsDataset(CachedDataset2):
         self.seq_tag_delim = seq_tag_delim
         self.remove_in_between_postfix = remove_in_between_postfix or {}
         self.repeat_in_between_last_frame_up_to_multiple_of = repeat_in_between_last_frame_up_to_multiple_of or {}
+        self.pad_truncate_data_to_multiple_of = pad_truncate_data_to_target_length or {}
         self.epoch_wise_filter = EpochWiseFilter(epoch_wise_filter) if epoch_wise_filter else None
         if isinstance(dataset, dict):
             dataset = dataset.copy()
@@ -1562,6 +1567,17 @@ class ConcatSeqsDataset(CachedDataset2):
                     if data.shape[0] % multiple != 0:
                         data = numpy.concatenate([data] + [data[-1:]] * (multiple - data.shape[0] % multiple), axis=0)
                         assert data.shape[0] % multiple == 0
+                if key in self.pad_truncate_data_to_multiple_of and sub_seq_idx != sub_seq_idxs[-1]:
+                    target_key, multiple = self.pad_truncate_data_to_multiple_of[key]
+                    target_data = self.sub_dataset.get_data(sub_seq_idx, target_key)
+                    len_diff = data.shape[0] - target_data.shape[0] * multiple
+                    if len_diff > 0:
+                        # if data longer than ref_data * frame_rate, truncate
+                        data = data[:-len_diff,]
+                    elif len_diff < 0:
+                        # if data shorter than ref_data * frame_rate, pad by repeating last frame
+                        data = numpy.concatenate([data] + [data[-1:]] * -len_diff, axis=0)
+                    assert data.shape[0] == target_data.shape[0] * multiple
                 features[key].append(data)
         features = {key: numpy.concatenate(values, axis=0) for (key, values) in features.items()}
         return DatasetSeq(seq_idx=seq_idx, seq_tag=seq_tag, features=features)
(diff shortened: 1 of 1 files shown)


returnn/datasets/meta.py
```diff
@@ -1390,7 +1390,7 @@ def __init__(
         seq_len_file,
         seq_tag_delim=";",
         remove_in_between_postfix=None,
-        repeat_in_between_last_frame_up_to_multiple_of=None,
+        force_align=None,
```
NeoLegends, line 1393: Renaming/changing the purpose of this parameter is going to break existing setups that use this dataset. Usually we try to stay backwards compatible.

One solution here is to keep both parameters (and their defaults), and then assert that only one of them is set at a time and then apply the corresponding logic.

I'm also not sure `force_align` is a good name for this key, as AFAIK the term "forced alignment" is usually used for producing an alignment from a trained model. I'd rather go with something like `force_pad_truncate_data_to_target_length`.

returnn/datasets/meta.py
```diff
@@ -1557,11 +1555,17 @@ def _collect_single_seq(self, seq_idx):
...
+                if key in self.force_align and sub_seq_idx != sub_seq_idxs[-1]:
+                    ref_key, frame_rate = self.force_align[key]
+                    ref_data = self.sub_dataset.get_data(sub_seq_idx, ref_key)
+                    len_diff = data.shape[0] - ref_data.shape[0] * frame_rate
+                    if len_diff > 0:
+                        # if data longer than ref_data * frame_rate, cut down
+                        data = data[:-len_diff,]
+                    elif len_diff < 0:
+                        # if data shorter than ref_data * frame_rate, pad by repeating last frame
+                        data = numpy.concatenate([data] + [data[-1:]] * -len_diff, axis=0)
```
NeoLegends, line 1567: Should there be a sanity limit how much can be truncated or padded? Perhaps to allow padding/truncating only within one multiple of the target length vs. feature length? In my own experiments w/ subsampled MFCCs the mismatch was generally within this range (but then it was also MFCCs and not raw WAV...).

NeoLegends (changes requested):
Thank you for filing this PR! I believe there are backwards incompatible changes which should probably be implemented differently.

returnn/datasets/meta.py
michelwi: Compared to features, for raw wav the mismatch will generally be larger by a factor of sampling_rate * frame_time e.g. 160 for 16kHz audio and 10ms frames.

returnn/datasets/meta.py
Stefanwuu: Thank you for your information, I did this because (to my best knowledge) nobody used this in training before:) But you are right, such potential issue should be avoided, I will add back the parameter before and change the name

returnn/datasets/meta.py
Stefanwuu: Do you mean sth like `assert len_diff <= frame_rate`? I think there is no clear upper bound if raw WAV is used. The mismatch is partially setup dependent(window_size - window_step in feature extraction) and can be more than one frame(10 ms). Maybe it would make sense to throw a warning instead of a hard limit?

albertz:
So this PR here is to fix issue #1573?

returnn/datasets/meta.py
```diff
@@ -1486,7 +1491,7 @@ def get_seq_len(i):
             sub_seq_list.extend(sub_seq_tags)
         assert sub_seq_idx == len(sub_seq_list) and len(seq_list) == len(sub_seq_idxs)
         self.cur_sub_seq_idxs = sub_seq_idxs
-        return self.sub_dataset.init_seq_order(seq_list=sub_seq_list)
+        return self.sub_dataset.init_seq_order(epoch=epoch, seq_list=sub_seq_list)
```
albertz, line 1494: I wonder a bit, why this change? It should be ok, but actually also should not matter? Or was there a problem? I'm just curious.

returnn/datasets/meta.py
```diff
@@ -1539,6 +1544,14 @@ def _collect_single_seq(self, seq_idx):
...
                     sub_dataset_keys,
                 )
+            for key in self.pad_truncate_data_to_multiple_of:
+                assert (
+                    key in sub_dataset_keys
+                ), "%s: pad_truncate_data_to_multiple_of key %r not in sub dataset data-keys %r" % (
+                    self,
+                    key,
+                    sub_dataset_keys,
+                )
```
albertz, line 1554: Better use f-strings.
```suggestion
                assert key in sub_dataset_keys, (
                    f"{self}: pad_truncate_data_to_multiple_of key {key}"
                    f" not in sub dataset data-keys {sub_dataset_keys}"
                )
```

returnn/datasets/meta.py
```diff
@@ -1606,6 +1630,12 @@ def get_data_shape(self, key):
         """
         return self.sub_dataset.get_data_shape(key)
 
+    def get_total_num_seqs(self):
+        """
+        :rtype: int
+        """
```
albertz, line 1636:
```suggestion
    def get_total_num_seqs(self) -> int:
        """total num seqs"""
```

returnn/datasets/meta.py
albertz: Note that I know that people did use `ConcatSeqsDataset` in training, even recently. :) (Although not sure if this specific option was used.)

albertz:
Failing tests?

returnn/datasets/meta.py
```diff
@@ -1406,13 +1407,17 @@ def __init__(
           Now it could happen that ceildiv(data_len1 + data_len2, 6) < align_len1 + align_len2.
           This option would repeat intermediate ending frames such that data_len1 % 6 == 0,
           by setting it to {"data": 6}.
+        :param dict[str,(str,int)]|None pad_truncate_data_to_multiple_of: data_key -> (target_key, multiple).
+            to make sure the alignment is still valid after concatenation, we force length of every seq in data_key
+            to be length of seq in target_key * multiple
```
albertz, line 1412: I find the documentation a bit misleading. Specifically "to make sure the alignment is still valid after concatenation". Specifically, the option `repeat_in_between_last_frame_up_to_multiple_of` is also there to handle this issue already. So you should better explain, why is this new option needed. What would be a situation where `repeat_in_between_last_frame_up_to_multiple_of` would not work.

albertz:
On the method itself: The first question is, how is the actual relation between input len and output len?

Note e.g. in `repeat_in_between_last_frame_up_to_multiple_of`, I mention the relation $\lceil T_{\textrm{in}} / F \rceil = T_{\textrm{out}}$ where $F$ is the downsampling rate. This is what you get when you have padding=same.

For padding=valid (i.e. no padding allowed for the window, the window must always cover real frames), you would have $\lceil ( T_{\textrm{in}} - (W - 1) ) / F \rceil = T_{\textrm{out}}$, where $W$ is the window/kernel size.

With feature extraction on raw wav, we have actually various different variants. Most implementations in TF and/or PT would use `stft` in some way, which usually is padding=valid.

Note that any custom padding schemes are possible.

Padding=same is actually just designed in a way to get this nice $\lceil T_{\textrm{in}} / F \rceil = T_{\textrm{out}}$ relation (or with $F=1$ you simply have $T_{\textrm{in}} = T_{\textrm{out}}$, thus "same"). Padding=same means it will pad $W - 1$ frames, evenly distributed left/right (right one more if uneven).

The `padding` option in Torch (`Conv` and co) specifies usually how many frames are padded left and right (so it means for `padding=P`, it will actually pad 2*P frames in total).

RASR feature extraction, for some strange reasons, does pad some totally custom number of frames: It pads in total $F-1$ frames. I.e. you have the relation $\lceil ( T_{\textrm{in}} - (W - 1) + (F - 1)) / F \rceil = \lceil ( T_{\textrm{in}} - W + F) / F \rceil = T_{\textrm{out}}$.

Btw, also see https://github.com/rwth-i6/returnn/issues/1564 for a recent discussion on exactly the same topic.

I already commented on improving the documentation. I think these different cases should become clear in the documentation.

`repeat_in_between_last_frame_up_to_multiple_of` is right when you have padding=same.

`pad_truncate_data_to_multiple_of` is actually also only right when you have padding=same, or not?

returnn/datasets/meta.py
Stefanwuu: The problem is it could be incompatible with some sub_dataset like `MetaDataset`, where init_seq_order would not be actually executed if `epoch` stays `None`, maybe there are other workarounds?

albertz: Ah yea, good example. Yes this is definitely more correct.

returnn/datasets/meta.py
```diff
@@ -1406,13 +1407,19 @@ def __init__(
           Now it could happen that ceildiv(data_len1 + data_len2, 6) < align_len1 + align_len2.
           This option would repeat intermediate ending frames such that data_len1 % 6 == 0,
           by setting it to {"data": 6}.
+        :param dict[str,(str,int)]|None pad_truncate_data_to_multiple_of: data_key -> (target_key, multiple).
+            Force data_len == target_len * multiple for specified data and target.
+            Similar to repeat_in_between_last_frame_up_to_multiple_of but also truncates the data.
```
albertz, line 1412: > Similar to repeat_in_between_last_frame_up_to_multiple_of but also truncates the data.

This does not really explain it. In what case would it be relevant to truncate the data?

returnn/datasets/meta.py
```diff
@@ -1406,13 +1407,19 @@ def __init__(
...
+            Mainly for raw wav alignment, where data_len could be slightly longer than target_len * multiple
+            due to mismatch in feature extraction.
```
albertz, line 1414: "raw wav alignment" doesn't really say anything / is not relevant here.

What is the case you have specifically, i.e. what type of padding was done? This is what is relevant.

Why is `repeat_in_between_last_frame_up_to_multiple_of` not already enough?

returnn/datasets/meta.py
Stefanwuu: I know, this is not the final version as mentioned in the commit comment. I want to first consider your long comment above that I haven't read through, maybe there is indeed situations(especially with padding) that is not covered by my code.

returnn/datasets/meta.py
```diff
@@ -1391,6 +1391,7 @@ def __init__(
         seq_tag_delim=";",
         remove_in_between_postfix=None,
         repeat_in_between_last_frame_up_to_multiple_of=None,
+        pad_truncate_data_to_multiple_of=None,
```
albertz, line 1394: Shouldn't there also be `target_len` or so in the name?

returnn/datasets/meta.py
Stefanwuu: The main problem I want to tackle here is indeed relevant to "raw wav alignment"(but of course also to the padding scheme). The main problem with raw wav is that even with same padding, we could have $\left \lceil{T_{in}/F}\right \rceil \neq T_{out}$. With most common `stft` setting, we have `frame_size` of 25ms and `frame_shift` of 10 ms. This means a perfectly aligned 3-frame text corresponds to 45ms audio data, and you can see $\left \lceil{45/10}\right \rceil \neq 3$. In chunked training, we solve this by setting min_chunk_size to a magic number(which is not elegant, but we all do this, see [more explanation](https://omnifluent.atlassian.net/wiki/spaces/AS/pages/2636709904/Deep+dive+into+NN+training#Chunking)). But this gets wrong if such mismatch adds up with sequence concatenation. That's why I truncate such mismatches for all sub-sequences except the last one.

returnn/datasets/meta.py
```diff
@@ -1406,13 +1407,19 @@ def __init__(
...
+            Mainly for raw wav alignment, where data_len could be slightly longer than target_len * multiple
+            due to mismatch in feature extraction.
         :param bool use_cache_manager:
         :param dict[(int,int),dict] epoch_wise_filter: see :class:`EpochWiseFilter`
         """
         super(ConcatSeqsDataset, self).__init__(**kwargs)
         self.seq_tag_delim = seq_tag_delim
         self.remove_in_between_postfix = remove_in_between_postfix or {}
         self.repeat_in_between_last_frame_up_to_multiple_of = repeat_in_between_last_frame_up_to_multiple_of or {}
+        self.pad_truncate_data_to_multiple_of_target_len = pad_truncate_data_to_multiple_of_target_len or {}
```
NeoLegends, line 1422: I think this should have an assertion that not both properties are set?

returnn/datasets/meta.py
albertz: > The main problem I want to tackle here is indeed relevant to "raw wav alignment"(but of course also to the padding scheme).

This doesn't make sense. It can not be relevant that you operate on raw wav. It only matters what kind of downsampling/padding scheme you have. Why don't you say what case you have actually? What code are you using? You can simply look that up.

> The main problem with raw wav is that even with same padding, we could have $\left \lceil{T_{in}/F}\right \rceil \neq T_{out}$

No. Padding=same implies that you have $\left \lceil{T_{\textrm{in}}/F}\right \rceil = T_{\textrm{out}}$, always. That follows directly from the definition of padding=same. That's actually what it means. If you don't have $\left \lceil{T_{\textrm{in}}/F}\right \rceil = T_{\textrm{out}}$, then it means, you don't have padding=same but you have some different padding scheme.

So I get back to my question: What padding scheme do you have?

> With most common `stft` setting, ...

I assume you have padding=valid then?

> In chunked training, we solve this by setting min_chunk_size to a magic number

Was this related to the discussion here? https://github.com/rwth-i6/returnn/issues/1564

But again, the main question: What type of padding do you use? Or if you don't know and cannot tell from your code, can you at least say what code you use?

Stefanwuu: Hmm, I thought I knew which padding you were talking about but now I'm not sure. Do you refer to the padding/downsampling scheme in RETURNN? E.g. `pad_last_frame` in `MultiChannelStftLayer` or same padding in conv layer if conformer was used for downsampling?

albertz: No, I mean whatever you did to get your original input len <-> output len relation. I don't care about what software you used. I just want to know your exact input len <-> output len relation, i.e. the type of padding scheme you have used there.

Stefanwuu: But that's why raw wav is relevant here. In raw wav, the first frame is longer than the rest of frames due to mismatched window size and window shift step in feature extraction(see my example above). A padding can't fix this because it's a problem of too long data instead of too short data. That's why I propose to truncate in ConcatSeqsDataset in addition to the padding you implemented before.

Stefanwuu: Sorry for the confusion, I had a bit discussion with Moritz and it turns out I didn't understand your point on padding. 
The main issue here is the padding logic in my RETURNN in-network feature extraction, as you described
> For padding=valid (i.e. no padding allowed for the window, the window must always cover real frames), you would have ⌈(Tin−(W−1))/F⌉=Tout, where W is the window/kernel size.

does not fit the padding logic in RASR alignment generation. So yes, it is possible to modify the alignment somewhere else to avoid the modification of RETURNN. Then maybe I should close this pull request?

albertz: I still don't understand your padding scheme. What *exactly* is your input len <-> output len relation of the given data you have? There must be an *exact* formula to describe that relation. If you don't know the formula, please find out. It would also help when you say what code in your case determines this relation. E.g. what is your feature extraction? I didn't really understand whether you use RETURNN, RASR, or sth else to generate this data, and in any case, what exactly.

Whatever the formula is, from the formula, you can infer your padding scheme. Nothing else is relevant. It's still not relevant that you operate on raw wav.

Stefanwuu: Yes, I understand your point that operating on raw WAV is not relevant now, only the alignment logic applied there is relevant. Sorry again for the misunderstanding before. The exact formula in my setup(slightly different from most common setup): 
$T_{out}=\left \lfloor{(T_{in}-(W-F))/F} \right \rfloor$
, which is different from the alignment(or padding) scheme of RASR-generated alignment.

albertz: Really floordiv, not ceildiv? Can you give a reference to the code which leads to such formula?

albertz: Assuming the formula is correct: You can translate this into the standard scheme of the formula:
$T_{out} = \lfloor (T_{in}-(W-F))/F \rfloor = \lceil (T_{in} - (W - F) - (F - 1)) / F \rceil = \lceil (T_{in} - (W - 1)) / F \rceil$
I.e. it means you have exactly the padding=valid case. That's what I wanted to know.

Stefanwuu: Yes, padding=valid is the case I have. Then what would be your suggestion in this case? Set up the alignment so that padding=valid also holds there instead of modifying RETURNN? Thanks!

albertz:
After some clarification, as I understand you know, I think this new option is intended for the padding=valid case. At least that is your current use case.

Did you actually test the code? Check also my earlier comment here: https://github.com/rwth-i6/returnn/pull/1574#issuecomment-2209637554

I assume you have F=160 (10ms) and W=400 (25ms). And you have padding=valid.

Some example sequences you can have then:
* $T_{in} = 399 \rightarrow T_{out} = 0$
* $T_{in} = 400 \rightarrow T_{out} = 1$
* $T_{in} = 559 \rightarrow T_{out} = 1$
* $T_{in} = 560 \rightarrow T_{out} = 2$
* $T_{in} = 719 \rightarrow T_{out} = 2$
* $T_{in} = 720 \rightarrow T_{out} = 3$

Now when you set `pad_truncate_data_to_multiple_of = {"data": ("targets", 160)`}:

Assume you concat two seqs with $T_{in} = 400, T_{out} = 1$: The first one will be truncated to 160 frames. You get a sequence with $T_{in} = 560, T_{out}=2$. Ok.

Ah, I see now, this should always be ok, no matter the padding. Assume you have a padding scheme like this:
$T_{out} = \lceil (T_{in} - (W + 1) + P) / F \rceil$ for any amount of padded frames $P$. You have $P=0$ (padding=valid).
Now, if you concatenate other seqs, but with your scheme, you make sure that $\Delta T_{out} \cdot F = \Delta T_{in}$ for all but the right outermost seq, so you get for the concatenated seq: $T_{out} + \Delta T_{out} = \lceil (T_{in} + \Delta T_{in} - (W+1) + P) / F \rceil$.

returnn/datasets/meta.py
albertz: No. I was simply saying, you should clarify the documentation: you should make clear, for what cases the user would want to use `pad_truncate_data_to_multiple_of`. Why is `repeat_in_between_last_frame_up_to_multiple_of` not enough?

See my last comment in the main thread. The answer is basically: `repeat_in_between_last_frame_up_to_multiple_of` only works for padding=same. But your option works for any padding scheme, e.g. for padding=valid.

Additionally, as you misunderstood for a while this concept of padding scheme (no matter what type of input/output, no matter the software, the input len <-> output len relation can always be expressed as a formula, and that always corresponds to a specific padding scheme), it probably would be helpful to add another sentence or so explaining this.

returnn/datasets/meta.py
```diff
@@ -1391,6 +1391,7 @@ def __init__(
         seq_tag_delim=";",
         remove_in_between_postfix=None,
         repeat_in_between_last_frame_up_to_multiple_of=None,
+        pad_truncate_data_to_multiple_of_target_len=None,
```
albertz, line 1394: I still don't like the name. `truncate` is misleading here. `truncate` usually means sth different (e.g. see [`torch.trunc`](https://pytorch.org/docs/stable/generated/torch.trunc.html)).

In RETURNN, we have the function `slice_pad_zeros` which does what you want here. So we use the term `slice`. Instead of `slice`, you might also want to call this `narrow` ([`torch.narrow`](https://pytorch.org/docs/stable/generated/torch.narrow.html#torch.narrow)).

So maybe:
```suggestion
        pad_narrow_data_to_multiple_of_target_len=None,
```

Stefanwuu:
> Assume you have a padding scheme like this: Tout=⌈(Tin−(W+1)+P)/F⌉ for any amount of padded frames P. You have P=0 (padding=valid). Now, if you concatenate other seqs, but with your scheme, you make sure that ΔTout⋅F=ΔTin for all but the right outermost seq, so you get for the concatenated seq: Tout+ΔTout=⌈(Tin+ΔTin−(W+1)+P)/F⌉.

Yes, that was precisely my motivation to do this. Sorry for wasting time on the bad communication before, I always considered padding as a specific operation instead of a general formula of alignment. I will update the comment and do some last tests.

albertz (approved):
See remaining comment by @NeoLegends. Also let's wait for tests. But otherwise ok from my side.

returnn/datasets/meta.py
Stefanwuu: There is no unexpected behavior if both parameters are set(but probably redundant), so I don't know whether that would make sense.

albertz: Yea, for me, also not strictly necessary.
(@NeoLegends, can you re-review?)

NeoLegends (approved):
🚢

PR #1576, opened 2024-07-05
Torch: print model at log verbosity 3

### LINKED ISSUE #1575: Torch: print model at log verbosity 3
I find log verbosity 3 a reasonable verbosity for "daily" work/trainings, but I'm missing the model structure in the log. This is because it's only printed at v4 level. Is there any good reason the log is not printed at v3?

See https://github.com/rwth-i6/returnn/blob/2c0b75c1c0c61732a12a175ac3cef433f0bea271/returnn/torch/engine.py#L865C53-L865C53

NeoLegends:
Closes #1575

This probably didn't need a PR but 🤷🏼‍♂️

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -862,7 +862,7 @@ class Engine(EngineBase):
         else:
             raise TypeError(f"get_model returned {model} of type {type(model)}, expected rf.Module or torch.nn.Module")
         assert isinstance(self._pt_model, torch.nn.Module)
-        print("Model:", self._pt_model, file=log.v4)
+        print("Model:", self._pt_model, file=log.v3)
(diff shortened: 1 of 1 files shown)


returnn/torch/engine.py
```diff
@@ -862,7 +862,7 @@ def _create_model(self, *, epoch: int, step: int):
         else:
             raise TypeError(f"get_model returned {model} of type {type(model)}, expected rf.Module or torch.nn.Module")
         assert isinstance(self._pt_model, torch.nn.Module)
-        print("Model:", self._pt_model, file=log.v4)
+        print("Model:", self._pt_model, file=log.v3)
```
albertz, line 865:
```suggestion
        print("Model:", self._pt_model, file=log.v5)
```

PR #1578, opened 2024-07-11
RF: weight parametrization

### LINKED ISSUE #1518: RF weight dropout and variational noise

albertz:
Fix #1518.

* Add `rf.gradient_checkpoint_scope`.
* Add `rf.register_parametrization`, similarly as in PyTorch.
* Add `rf.weight_dropout` and `rf.weight_noise`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -360,6 +360,11 @@ jobs:
           - TEST=rf_rec
           - TEST=rf_reduce
           - TEST=rf_signal
+        include:
+          - action: TEST=rf_base
+            python-version: "3.10"
+            torch-version: 2.0.0  # needs PyTorch >=2.0 for grad scope test
+            tf-version: 2.10.0
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -397,8 +397,10 @@ jobs:
         # Install nose anyway because we currently use some of its helper functions.
         pip install --user nose
 
-        pip install --user --progress-bar=off -r requirements.txt
+        pip install --user dm-tree h5py typing
+        pip install --user --progress-bar=off "numpy<2"
         pip install --user --progress-bar=off scipy  # for some tests
+
         pip install --user --upgrade torch==${{ matrix.torch-version }}
 
         if [[ "${{matrix.tf-version}}" == 2.10.* ]]; then
--- a/returnn/frontend/__init__.py
+++ b/returnn/frontend/__init__.py
@@ -41,6 +41,7 @@ from .math_ import *
 from .matmul import *
 from .normalization import *
 from .parameter import *
+from .parametrizations import *
 from .parametrize import *
 from .piecewise_linear import *
 from .rand import *
--- a/returnn/frontend/__init__.py
+++ b/returnn/frontend/__init__.py
@@ -41,6 +41,7 @@ from .math_ import *
 from .matmul import *
 from .normalization import *
 from .parameter import *
+from .parametrize import *
 from .piecewise_linear import *
 from .rand import *
 from .rec import *
--- a/returnn/frontend/_backend.py
+++ b/returnn/frontend/_backend.py
@@ -377,6 +377,11 @@ class Backend(Generic[T]):
         raise NotImplementedError
 
     @staticmethod
+    def gradient_checkpoint_scope():
+        """gradient checkpoint scope"""
+        raise NotImplementedError
+
+    @staticmethod
     def merge_dims(
         source: Tensor,
         *,
--- a/returnn/frontend/gradient.py
+++ b/returnn/frontend/gradient.py
@@ -5,9 +5,16 @@ Utilities which affect the gradient
 from __future__ import annotations
 from typing import Optional, Union
 from returnn.tensor import Tensor, Dim
+from ._backend import global_backend
 
-
-__all__ = ["set_requires_gradient", "gradient", "stop_gradient", "scaled_gradient", "scaled_gradient_ext"]
+__all__ = [
+    "set_requires_gradient",
+    "gradient",
+    "stop_gradient",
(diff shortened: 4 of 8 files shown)


albertz:
Still WIP. Well, the parametrization API is basically already ready. But I want to add some tests for it. Also I want to add weight dropout and variational noise already. Maybe also weight norm. Similar as here:
https://pytorch.org/docs/stable/_modules/torch/nn/utils/parametrizations.html
https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.weight_norm.html#torch.nn.utils.parametrizations.weight_norm

I'm probably not squashing all commits together but leave them as separate commits. This PR here is mostly just for easier reviewing, and maybe also discussions. (Although the main design questions are already being discussed in #1518.)

albertz:
Ok, all done, at least `weight_dropout` and `weight_noise`. I think this is enough for this PR.

albertz:
I already merged this because I want to use this in some experiments, and I think we can just directly fix any issues in the code in follow-up commits/PRs. I think we even can still do changes to the API for a short while, as long as no-one else is using this.

So, please anyway review/comment.

PR #1586, opened 2024-07-17
Optimize RF PT pack_padded

albertz:
Fix #1584.

Issue #1584, opened 2024-07-17
`rf.pack_padded` with PyTorch takes a lot of memory


albertz:
At least from some initial profiling it seems so, and also when comparing it to a naive simple implementation.

Version with `rf.pack_padded`:
```python
def _get_packed_logits_v2():
  logits_packed, pack_dim = rf.pack_padded(
    logits,
    dims=batch_dims + [enc_spatial_dim, non_blank_targets_spatial_dim_ext],
    enforce_sorted=False
  )  # [B * T * (S+1), D]
  logits_packed_raw = logits_packed.raw_tensor
```

Naive version:
```python
def _get_packed_logits_v1():
  logits_raw = logits.copy_transpose(batch_dims + [enc_spatial_dim, non_blank_targets_spatial_dim_ext, model.align_target_dim]).raw_tensor
  enc_lens = enc_spatial_dim.dyn_size_ext.raw_tensor
  non_blank_lens = non_blank_targets_spatial_dim_ext.dyn_size_ext.raw_tensor
  vocab_len = model.align_target_dim.dimension

  batch_tensors = []

  for b in range(logits_raw.shape[0]):
    enc_len = enc_lens[b]
    non_blank_len = non_blank_lens[b]
    combined_len = enc_len * non_blank_len
    logits_single = logits_raw[b, :enc_len, :non_blank_len]
    logits_single = torch.reshape(logits_single, (combined_len, vocab_len))
    batch_tensors.append(logits_single)

  return torch.cat(batch_tensors, dim=0)
```

**cc** @robin-p-schmitt

albertz:
See the analysis I did in the comments of the PR:
https://github.com/rwth-i6/returnn/pull/1586#issuecomment-2236874473
https://github.com/rwth-i6/returnn/pull/1586#issuecomment-2238726490

It seems that `torch.masked_select` is really not optimal. It does a lot of extra allocations and is also very inefficient. Note that `torch.masked_select` behaves different on CPU or with CUDA, but it's bad in both cases.

I found some related upstream PyTorch issues:
https://github.com/pytorch/pytorch/issues/30246
https://github.com/pytorch/pytorch/issues/56896

PR #1593, opened 2024-07-19
Torch masked_select, no custom nonzero for now

albertz:
Follow-up to #1586, #1584.
<s>Specifically https://github.com/rwth-i6/returnn/pull/1586#issuecomment-2239162301. Specifically, add this to `rf.pack_padded`:
```python
if not out_dim:
    out_dim = Dim(rf.num_elements_of_shape(dims), name="packed")
```
This allows to use a more efficient `masked_select` and custom `nonzero` variant where we can avoid a CUDA host-device synchronization.</s>

Actually, we don't do this. We even removed this code path from the RF PT `masked_select`, because it seems to be slightly slower in my benchmarks.

We might reconsider this later. Let's also see here: https://github.com/pytorch/pytorch/issues/131256

### DIFF
--- a/returnn/frontend/array_.py
+++ b/returnn/frontend/array_.py
@@ -584,11 +584,15 @@ def pack_padded(
     assert not enforce_sorted  # not implemented yet...
     mask = rf.sequence_mask(dims, device=source.device)
     assert mask.dims_set == set(dims)
-    if not out_dim:
-        # Note: Already calculating out_dim here can trigger a more efficient calculation path in masked_select,
-        # where we can avoid a CUDA host-device synchronization, e.g. in the PyTorch backend.
-        # See https://github.com/rwth-i6/returnn/pull/1593.
-        out_dim = Dim(rf.num_elements_of_shape(dims), name="packed")
+    # Note: We could already calculate out_dim here, as follows:
+    #   out_dim = Dim(rf.num_elements_of_shape(dims), name="packed")
+    # This could trigger a more efficient calculation path in masked_select,
+    # where we can avoid a CUDA host-device synchronization, e.g. in the PyTorch backend.
+    # However, in our benchmarks so far, it seems it's not helping so far,
+    # so we don't do this, to also avoid the (minor) overhead of num_elements_of_shape here.
+    # See: https://github.com/rwth-i6/returnn/pull/1593
+    # This might change in the future when we have this:
+    # https://github.com/pytorch/pytorch/issues/131256
     return rf.masked_select(source, mask=mask, dims=dims, out_dim=out_dim)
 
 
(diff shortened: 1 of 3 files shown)


albertz:
While playing around with the benchmark, I'm not sure to add this. While it avoids one CUDA host-device synchronization, on the other side, the `torch.argsort` is probably way more costly than `torch.nonzero`.

Here are the current benchmark results:
```
func: <function _benchmark_pack_padded.<locals>._no_op_test at 0x7f562102f7e0>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f56222b2f50>
func()
  Median: 25.68 ns
  2 measurements, 10000000 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_logits at 0x7f562102f880>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f562125dd50>
func()
  Median: 30.66 ns
  2 measurements, 10000000 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_rf_pack_packed at 0x7f562102f920>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f56222b2f50>
func()
  Median: 3.34 ms
  2 measurements, 100 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_rf_pack_padded_known_lens at 0x7f562102f9c0>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f562148bf90>
func()
  Median: 3.33 ms
  2 measurements, 100 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_rf_pack_padded_no_known_lens at 0x7f562102fa60>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f5621f49110>
func()
  Median: 3.31 ms
  2 measurements, 100 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_torch_masked_select_pack_padded at 0x7f562102fb00>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f562148bf90>
func()
  Median: 37.87 ms
  2 measurements, 10 runs per measurement, 1 thread
func: <function _benchmark_pack_padded.<locals>._get_naive_pack_padded at 0x7f562102fba0>
<torch.utils.benchmark.utils.common.Measurement object at 0x7f5621f1d390>
func()
  Median: 6.26 ms
  IQR:    0.01 ms (6.26 to 6.27)
  8 measurements, 10 runs per measurement, 1 thread
```
`_get_rf_pack_padded_no_known_lens` uses `torch.nonzero`, and `_get_rf_pack_padded_known_lens` (and `_get_rf_pack_packed`) use our custom `nonzero`, which uses `torch.argsort`. The difference is always small but always slightly better for the `torch.nonzero`.

I'm not sure though if there benchmark really covers the problems of CUDA host-device synchronization. It will actually explicitly do some CUDA synchronization.

albertz:
The custom `nonzero` which uses `torch.argsort` is maybe not optimal. I opened an issue about that: https://github.com/pytorch/pytorch/issues/131256

I think I will actually remove this code path in our custom `masked_select` for now, as it seems always slower in the benchmarks.

PR #1594, opened 2024-07-22
Torch: allow setting custom batch_size for each dataset

### LINKED ISSUE #1567: Make batch_size configurable for cross validation
There is currently one batch_size defined for both training and cross validation. 
Since we do not have to keep track of gradients etc during CV, we could increase the batch_size for higher throughput.

michelwi:
Fix #1567.

For cross-validation the batch_size can be much larger compared to training.
This allows to define a custom batch_size for each dataset.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -185,9 +185,9 @@ class Engine(EngineBase):
             for dataset_name, dataset_opts in config.typed_value("eval_datasets", {}).items():
                 self.eval_datasets[dataset_name] = init_dataset(dataset_opts, default_kwargs={"name": dataset_name})
 
-        self._train_dataloader = self._create_data_loader(train_data) if train_data else None
+        self._train_dataloader = self._create_data_loader(train_data, train=True) if train_data else None
         for dataset_name, dataset in self.eval_datasets.items():
-            self._eval_dataloaders[dataset_name] = self._create_data_loader(dataset)
+            self._eval_dataloaders[dataset_name] = self._create_data_loader(dataset, train=False)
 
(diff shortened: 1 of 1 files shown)


albertz (approved):

returnn/torch/engine.py
```diff
@@ -611,8 +611,12 @@ def _create_data_loader(self, dataset: Dataset) -> DataLoader:
                 wrapped_dataset, chunking, min_chunk_size=min_chunk_size
             )
 
-        assert self.config.typed_value("batch_size") is not None, "batch_size not defined in config"
+        assert self.config.typed_value("batch_size") is not None or (
+            self.config.typed_value("batch_size_train") is not None
+            and self.config.typed_value("batch_size_dev") is not None
+        ), "batch_size not defined in config"
```
michelwi, line 617: there could be cases e.g. forwarding only where it would be ok to only set `batch_size_dev` and not set (`batch_size` and `batch_size_train`). Is the assert like this ok or should we rather do

```
batch_size = self.config.typed_value("batch_size", -1)
batch_size = self.config.typed_value(f"batch_size_{'train' if train else 'dev'}", batch_size)
assert batch_size != -1, "batch_size not defined in config"

NeoLegends (approved):

NeoLegends (approved):

returnn/torch/engine.py
albertz: Yeah, `assert batch_size != -1` is probably easier. It should also report the `train` flag then.

albertz (approved):

PR #1596, opened 2024-07-25
PostprocessingDataset

### LINKED ISSUE #1505: Dataset post processing
Examples of post-processing:

- Raw audio is stored in the HDFDataset, do feature extraction on-the-fly (but not in the network, but instead as part of the dataset).
- Ogg is stored in the HDFDataset, do decoding on-the-fly into raw audio.
- Apply speed perturbation on-the-fly.
- Text (UTF8 bytes) is stored in the HDFDataset, apply BPE (or any `Vocabulary`) on-the-fly.
- Concatenate sequences on-the-fly (#1573). I.e. the post-processing is not just on a single sequence.
- Apply mixup, i.e. mix in data from previous sequences. But this buffer would only be kept local, and reset after each epoch.

Some datasets already have partial support for post processing. Examples:

- `OggZipDataset` `targets`, can be any type of `Vocabulary` (e.g. `BytePairEncoding`, but also `SamplingBytePairEncoding`, or `SentencePieces`). Similarly `ExternSprintDataset` `orth_vocab`.
- `ExtractAudioFeatures` is used in a couple of places, e.g. by `OggZipDataset` `audio`. It supports also `pre_process` (on raw audio) and `post_process` (on audio features).

There was the idea about storing generic raw audio (or maybe even Ogg) inside the HDFDataset. And similarly, there was also the idea about storing the text (UTF8 bytes) inside the HDFDataset. In both cases, you would then maybe want to transform those into audio features or BPE labels on-the-fly as part of the dataset.

There are multiple options how to implement this:

- Similarly as in `OggZipDataset`, extend some other dataset (e.g. `HDFDataset`) by such functionality. But how to do it in a somewhat generic and flexible way? One aspect to keep in mind is that this might also change the dimension or shape of the data. E.g. raw audio to audio features will add one dimension.
- Make another dedicated dataset just for such transformation, similarly as `MetaDataset`. Or maybe make it part of `MetaDataset`?
- Make this a separate mechanism in RETURNN, which we can apply after the dataset logic? Which e.g. operates on `TensorDict` (before batching), or on individual data streams. But the distinction when something should be done as part of the dataset and when it would be done as such post-processing would be kind of arbitrary.

NeoLegends:
Closes #1505


returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@dataclass
+class LoadedData:
+    tensors: TensorDict
+    seq_idx: int
+
+
+Postprocessor = Callable[[TensorDict], TensorDict]
+StreamingPostprocessor = Callable[[Iterator[TensorDict]], Iterator[TensorDict]]
```
albertz, line 20: It's only used once? Inline this below. (Even when used in multiple places, I think I would still inline this.)

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,195 @@
+__all__ = ["PostprocessingDataset"]
```
albertz, line 1: Move after the imports (consistency to other code).

albertz, line 1: Module docstring missing.

albertz, line 1: `from __future__ import annotations` missing.

returnn/datasets/postprocessing.py
NeoLegends: I found the type gets super long and not very readable. That's why I extracted them out.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,195 @@
...
+from returnn.tensor import TensorDict
+from returnn.util.basic import NumbersDict
+from .basic import Dataset, init_dataset
+from .util.strings import str_to_numpy_array
+
+
+@dataclass
+class LoadedData:
+    tensors: TensorDict
+    seq_idx: int
```
albertz, line 16: This should be moved to the end of the file. The public class should come first. Also, prefix it with `_` to make clear this is only used internally in the dataset.

albertz:
Some simple test cases would be good as well.

returnn/datasets/postprocessing.py
albertz: But it's still easier to read. Otherwise a user needs to lookup another place just to know the type.

albertz:
Btw, lot's of missing docstrings, some PEP8 violations, etc.

albertz:
I see that you copied a lot of code of `CachedDataset2`. Wouldn't it be easier to just derive from that?

NeoLegends:
> I see that you copied a lot of code of CachedDataset2. Wouldn't it be easier to just derive from that?

Yes, probably. Initially you suggested not to use it, that's why I didn't end up deriving from it in the first place. Good point though, now that the code turned out to be very close, I'll derive and refactor. :)

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,174 @@
...
+    data processing work across multiple CPU cores using `MultiProcDataset` and in
+    turn frees the GPU from data preprocessing tasks.
+
+    Both functions can be specified at the same time, and their results will be
+    combined by first applying the segment-level function and then the
+    multiple-segments-level function.
+
+    Example usage:
+        train = {
+            "class": "PostprocessingDataset", "dataset": { "class": "..." },
```
albertz, line 38:
```suggestion
            "class": "PostprocessingDataset",
            "dataset": {"class": "...", ...},
            # One of the following (but not both):
            "map_seq": my_map_seq_func,
            "map_seq_stream": my_map_seq_stream_func,
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,174 @@
...
+        }
+    """
+
+    def __init__(
+        self,
+        dataset: Dict[str, Any],
+        map_seq: Optional[Callable[[TensorDict], TensorDict]] = None,
+        map_seq_stream: Optional[Callable[[Iterator[TensorDict]], Iterator[TensorDict]]] = None,
```
albertz, line 46: I think it might make sense that you also specify the expected output (to derive `num_outputs` from it, but also to make the automatic conversion easier, and also as an extra verification).

Maybe even the expected input?

(Similar as `extern_data` in the config.)

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,205 @@
...
+        self.labels = dataset.labels
+        self._estimated_num_seqs = dataset.estimated_num_seqs
+        if self._map_outputs is not None:
+            tdict = TensorDict(self._map_outputs)
+            # TODO: What if feature_dim == None?
+            self.num_inputs = tdict.data["data"].feature_dim.size
+            self.num_outputs = {
+                k: (t.sparse_dim.size, 1) if t.sparse_dim is not None else (t.feature_dim.size, t.ndim)
+                for k, t in tdict.data.items()
+            }
```
NeoLegends, line 131: @albertz I'm not sure how to do the inference properly here. I think the Tensor class w/ Dim behavior v2 leaves the feature_dim often unspecified, but the code seems to want a value here. It's probably not correct to assume `feature_dim = tensor.dims[-1]`, but I'm not sure what to infer otherwise.

albertz: > I think the Tensor class w/ Dim behavior v2 leaves the feature_dim often unspecified

No. It's just optional but it often is used.

But anyway, for datasets, basically feature dim always either refers to the sparse_dim when it is sparse, or otherwise to the last dim. I think datasets don't really allow anything else currently. The structure is basically always [T] or [T,D]. I'm not really sure what other cases we handle properly. With `get_data_shape`, `is_data_sparse`, `get_data_dim`, it's slightly more flexible.

`num_inputs`/`num_outputs` don't really allow to describe all different cases properly. But that doesn't matter. They shouldn't be used for any critical code, I think. They are very much deprecated.

albertz, line 131:
```suggestion
            self.num_outputs = {
                k: (t.sparse_dim.size if t.sparse_dim else t.shape[-1] or 1, t.ndim)
                for k, t in tdict.data.items()
            }
            self.num_inputs = self.num_outputs[self._default_input][0]
```

albertz: You must also derive `self._default_input`. It's ok to use some heuristics, it shouldn't matter.

albertz: Btw, see `AnythingDataset`. There I do this:
```python
def __init__(self, *, data_keys: Dict[str, Dict[str, Any]], **kwargs):
    """
    :param data_keys: similar like extern_data. defines shape, dtype, sparse, dim, etc
    """
    super().__init__(**kwargs)
    self._data_keys = data_keys
    self._input_key = next(iter(data_keys))
    self.num_inputs = data_keys[self._input_key]["dim"]
    self.num_outputs = {k: (v["dim"], len(v["shape"])) for (k, v) in data_keys.items()}
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+
+        def _make_tensor(name: str, data: ndarray) -> Tensor:
+            dims = None
+            sparse_dim = None
+            dtype = data.dtype
+            if data.dtype.name.startswith("str"):
+                # TODO: is this correct?
+                dims = []
+                dtype = "string"
```
NeoLegends, line 198: @albertz Is this how we'd handle string tensors?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,239 @@
...
+            if dims is None:
+                dims, sparse_dim = self._dim_cache.get(name, (None, None))
+            if dims is None:
+                dims = [Dim(dimension=seq_len[name], name=f"{name}_num_frames")]
```
NeoLegends: And how we'd deal with seq lengths?

albertz: If it is dynamic, it is `dimension=None`.

returnn/datasets/postprocessing.py
albertz: The dataset doesn't really have well defined support for string tensors. It doesn't really matter, as long as it's somewhat reasonable.

returnn/datasets/audio.py
```diff
@@ -469,7 +469,7 @@ def get_corpus_seq(self, corpus_seq_idx: int) -> DatasetSeq:
             with self._open_audio_file(corpus_seq_idx) as audio_file:
                 features = self.feature_extractor.get_audio_features_from_raw_bytes(audio_file, seq_name=seq_tag)
         else:
-            features = numpy.zeros((), dtype=numpy.float32)  # currently the API requires some dummy values...
+            features = numpy.zeros((0, 0), dtype=numpy.float32)  # currently the API requires some dummy values...
```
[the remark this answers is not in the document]
albertz: Why do you need this? You should not care at all about keys which are not defined by the dataset (not in `get_data_keys()`).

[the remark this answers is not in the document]
NeoLegends: The issue is that the OggZipDataset does in fact output `data` in `get_data_keys()` for historical reasons, I think.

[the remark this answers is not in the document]
albertz: Ah that's an error then in `OggZipDataset`.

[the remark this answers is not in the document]
albertz: So, as said, this change should not be done here. You should fix `get_data_keys` instead.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        }
+    """
+
+    def __init__(
+        self,
+        dataset: Dict[str, Any],
+        map_seq: Optional[
+            Union[
+                Callable[[TensorDict], TensorDict],
+                Callable[[TensorDict], Iterator[TensorDict]],
```
albertz, line 56: I don't understand this. When operating on a single seq, it should not return an iterator?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+    multiple-segments-level function.
+
+    Example usage:
+        train = {
+            "class": "PostprocessingDataset",
+            "dataset": {
+                "class": "HDFDataset",
+                "files": ["/path/to/data.hdf"],
+            },
+            # at least one of them:
```
albertz, line 44:
```suggestion
            # one of the two but not both:
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            "map_seq": map_seq,  # (data: TensorDict) -> TensorDict
+            "map_seq_stream": map_seqs,  # (iter: Iterator[TensorDict]) -> Iterator[TensorDict]
+        }
```
albertz, line 47: Und auch `map_outputs` im Beispiel.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        :param map_seq: post processor function operating on the single-segment level.
+        :param map_seq_stream: post processor function operating on the multiple segment level via an iterator.
+            Allows merging multiple segments into one, or generating multiple output segments from one input segment.
+            When both functions are specified, the segment-level function is applied first to every segment, and the
+            results are passed to the iterator-level function.
+        :param output: Type and axis specification of the outputs of the mapping functions,
+            like extern_data and model_outputs.
+            To simplify the common case when no shapes change, this value can be left unspecified. The dataset then
+            assumes the same data layout as returned by the wrapped dataset.
+        :param _meta_info_cache: internal usage
```
albertz, line 75: `map_outputs` fehlt.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            "labels": self.labels,
+            "num_inputs": self.num_inputs,
+            "num_outputs": self.num_outputs,
+        }
+
+    def _lazy_init_num_outputs(self):
+        if self.num_outputs is not None:
+            return
+
+        dataset = init_dataset(self._dataset_def, parent_dataset=self)
```
albertz, line 129:
```suggestion
        if self._dataset is None:
            self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
        dataset = self._dataset
```
?

albertz: We should avoid creating the subdataset instance twice.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            tdict = TensorDict(self._map_outputs)
+            self.num_outputs = {
+                k: (t.sparse_dim.size if t.sparse_dim else t.shape[-1] or 1, t.ndim) for k, t in tdict.data.items()
+            }
+            self.num_inputs = self.num_outputs[self._default_input][0]
+        else:
+            self.num_inputs = dataset.num_inputs
+            self.num_outputs = dataset.num_outputs
+
+    def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=None):
```
albertz, line 146: Typing of `seq_list`, `seq_order` missing.

returnn/tensor/tensor_dict.py
```diff
@@ -79,6 +79,7 @@ def as_raw_tensor_dict(
         include_scalar_dyn_sizes: bool = True,
         exclude_duplicate_dims: bool = False,
         expected_value_type: Union[Type, Sequence[Type]] = object,
+        exclude_sequence_lengths: bool = False,
```
albertz, line 82: Do not add this here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        self._data_iter_seq_idx = seq_idx
+
+        while True:
+            try:
+                loaded_seq_idx, tensor_dict = next(self._data_iter)
+            except StopIteration:
+                return None
+            if loaded_seq_idx != seq_idx:
+                continue
+            seq = DatasetSeq(features=tensor_dict.as_raw_tensor_dict(exclude_sequence_lengths=True), seq_idx=seq_idx)
```
albertz, line 178: Do not use `as_raw_tensor_dict` here.

returnn/tensor/tensor_dict.py
albertz: `as_raw_tensor_dict` is specifically only intended for the purpose to add seq lens. Otherwise the whole function is not necessary at all.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            return seq
+
+    def _build_dataset_iter(self, dataset: Dataset) -> Iterator[TensorDict]:
```
albertz, line 181: I don't understand the purpose to have `dataset` as an arg here. Either move this function completely outside of the dataset (so no `self`), then it makes sense. Or otherwise, just access `self._dataset`?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+                if isinstance(mapped_item, Iterator):
+                    yield from mapped_item
+                else:
+                    yield mapped_item
+
+        map_iter = self._map_seq_stream(_wrap_map_seq())
+        assert isinstance(map_iter, Iterator), "map_seq_stream must produce an iterator"
+        return map_iter
+
+    def _data_dict_to_tensor_dict(self, dataset: Dataset, data_dict: Dict[str, ndarray]) -> TensorDict:
```
albertz, line 198: Same about `dataset` as arg here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            except Exception as exc:
+                raise Exception(
+                    "Could not convert from mapping function output to `TensorDict`, "
+                    f"do the data shapes {data.shape} match up with the ones declared "
+                    f"in `map_output` (inferred to: {dims}, sparse={sparse_dim})?"
+                ) from exc
+
+        return TensorDict({k: _make_tensor(k, v) for k, v in data_dict.items()})
+
+    def _iterate_dataset(self, dataset: Dataset) -> Iterator[TensorDict]:
```
albertz, line 235: Again same about `dataset` as arg.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            seq = DatasetSeq(features=tensor_dict.as_raw_tensor_dict(exclude_sequence_lengths=True), seq_idx=seq_idx)
+            return seq
+
+    def _build_dataset_iter(self, dataset: Dataset) -> Iterator[TensorDict]:
+        """
+        :return: an iterator applying both the segment level and across-segment transformations on the given dataset
+        """
+
+        def _wrap_map_seq():
+            for item in self._iterate_dataset(dataset):
```
albertz, line 187: It seems `_iterate_dataset` is only used once here? Why to obfuscate the code then? Just inline `_iterate_dataset` here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+                mapped_item = self._map_seq(item)
+                if isinstance(mapped_item, Iterator):
+                    yield from mapped_item
+                else:
+                    yield mapped_item
```
albertz, line 192: I don't understand how you could get the first case?
```suggestion
                yield mapped_item
```

returnn/datasets/postprocessing.py
albertz: Ah, actually, you really intend to allow both together? Does it make sense? Then it also is a bit ambiguous in which order they are executed? But also, I don't really understand why the user would want this. If the user already provides `map_seq_stream`, there should really be no need for `map_seq` anymore. So I would not allow both, to avoid any ambiguity/confusion here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+
+        map_iter = self._map_seq_stream(_wrap_map_seq())
```
albertz, line 194:
```suggestion
        map_iter = self._map_seq_stream(self._iterate_dataset()) if self._map_seq_stream else _wrap_map_seq()
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        assert (
+            self.seq_ordering == "default"
+        ), f"specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}"
+        assert (
+            map_seq is not None or map_seq_stream is not None
+        ), "need to either define map_seq or map_seq_stream functions"
+
+        self._dataset_def = dataset
+        self._map_seq = map_seq or self._identity
+        self._map_seq_stream = map_seq_stream or self._identity
```
albertz, line 89:
```suggestion
        self._map_seq = map_seq
        self._map_seq_stream = map_seq_stream
```
See also my comments above on allowing both (I would not allow both), and comments below on `_wrap_map_seq`.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+                self._dim_cache[name] = (dims, sparse_dim)
+
+            try:
+                return Tensor(name, dims=dims, dtype=dtype, sparse_dim=sparse_dim, raw_tensor=data)
+            except Exception as exc:
+                raise Exception(
+                    "Could not convert from mapping function output to `TensorDict`, "
+                    f"do the data shapes {data.shape} match up with the ones declared "
+                    f"in `map_output` (inferred to: {dims}, sparse={sparse_dim})?"
+                ) from exc
```
albertz, line 231: Don't do such generic all exception catching. Your code should already catch any errors, and then you should expect that there is no way that any further exception is raised (i.e. an exception here would be unexpected).
```suggestion
            return Tensor(name, dims=dims, dtype=dtype, sparse_dim=sparse_dim, raw_tensor=data)
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+            return seq
+
+    def _build_dataset_iter(self, dataset: Dataset) -> Iterator[TensorDict]:
+        """
+        :return: an iterator applying both the segment level and across-segment transformations on the given dataset
+        """
+
+        def _wrap_map_seq():
+            for item in self._iterate_dataset(dataset):
+                mapped_item = self._map_seq(item)
```
albertz, line 188: You could actually move this to `_iterate_dataset`. Then you don't need this `_wrap_map_seq` at all here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        :return: generator providing data samples in the form of a TensorDict
+        """
+        data_keys = dataset.get_data_keys()
+
+        seq_index = 0
+        while dataset.is_less_than_num_seqs(seq_index):
+            dataset.load_seqs(seq_index, seq_index + 1)
+            data = {data_key: dataset.get_data(seq_index, data_key) for data_key in data_keys}
+            data["seq_tag"] = str_to_numpy_array(dataset.get_tag(seq_index))
+            yield self._data_dict_to_tensor_dict(dataset, data)
```
albertz, line 246:
```suggestion
            item = self._data_dict_to_tensor_dict(dataset, data)
            if self._map_seq is not None:
                item = self._map_seq(item)
            yield item
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+                dtype = "string"
+            else:
+                dims, sparse_dim = self._dim_cache.get(name, (None, None))
+                dtype = data.dtype
+
+            if dims is None:
+                feature_dims = [
+                    Dim(dimension=v, name=f"{name}_dim{i + 1}") for i, v in enumerate(dataset.get_data_shape(name))
+                ]
+                dims = [Dim(dimension=None, name=f"{name}_num_frames"), *feature_dims]
```
albertz, line 216:
```suggestion
                dims = [Dim(dimension=None, name=f"{name}_frame"), *feature_dims]
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+                if dataset.is_data_sparse(name):
+                    sparse_dim = Dim(
+                        dimension=dataset.get_data_dim(name) if dataset.is_data_sparse(name) else None,
+                        name=f"{name}_sparse",
+                    )
```
albertz, line 221:
```suggestion
                    )
                    if name in dataset.labels:
                        sparse_dim.vocab = Vocabulary.create_vocab_from_labels(dataset.labels[name])
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        self._dataset.init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
+        self._data_iter = enumerate(self._build_dataset_iter(self._dataset))
+        self._data_iter_seq_idx = -1
+        self._lazy_init_num_outputs()
+        return True
+
+    def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
+        assert seq_idx > self._data_iter_seq_idx, "collecting seqs must be done strictly monotonically"
+        self._data_iter_seq_idx = seq_idx
+
```
albertz, line 170: You don't need this check, or `_data_iter_seq_idx` at all. `_collect_single_seq` is a function only used internally and we already do this check.
```suggestion
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+        self._dataset_def = dataset
+        self._map_seq = map_seq or self._identity
+        self._map_seq_stream = map_seq_stream or self._identity
+        self._map_outputs = map_outputs
+
+        self._dataset: Optional[Dataset] = None
+        self._data_keys: Optional[List[str]] = None
+        self._data_iter: Optional[Iterator[Tuple[int, TensorDict]]] = None
+        self._data_iter_seq_idx = -1
```
albertz, line 95:
```suggestion
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if epoch is None:
+            self._num_seqs = 0
+            return True
+
+        if self._dataset is None:
+            self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self._dataset.init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
+        self._data_iter = enumerate(self._build_dataset_iter(self._dataset))
+        self._data_iter_seq_idx = -1
```
albertz, line 163:
```suggestion
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,251 @@
...
+    def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
+        assert seq_idx > self._data_iter_seq_idx, "collecting seqs must be done strictly monotonically"
+        self._data_iter_seq_idx = seq_idx
+
+        while True:
+            try:
+                loaded_seq_idx, tensor_dict = next(self._data_iter)
+            except StopIteration:
+                return None
+            if loaded_seq_idx != seq_idx:
```
albertz, line 176: Here you could add such sanity check.
```suggestion
            assert loaded_seq_idx <= seq_idx
            if loaded_seq_idx != seq_idx:
```

returnn/datasets/postprocessing.py
NeoLegends: Well, depends on how map_seq is defined we could allow it to return an iterator.

returnn/datasets/postprocessing.py
NeoLegends: The idea is to help the user catch any dim mismatch errors more quickly that occur due to the mapping function returning data in the wrong shape. I think the issue here is that `Tensor`, during its internal shape validation, just raises a generic `Exception`, and not a more specific one, that we could catch and extend here. I think making `Tensor` throw a more specific exception on shape mismatch, and than catching and re-raising /that one/ here instead makes a lot of sense from the users perspective.

albertz: But there should never be any exception here. If there is, your code is wrong. If something is not matching, you should have already checked that before.

returnn/datasets/postprocessing.py
albertz: We could allow for a million different things. But this seems here very arbitrary and not necessary and not what was discussed before. Let's first only implement the minimal required features that we discussed and make it as simple as possible, definitely not add any complexity like this here. If it turns out that this is a feature some people want, then we can add it later.

returnn/datasets/postprocessing.py
NeoLegends: thanks!

returnn/datasets/postprocessing.py
NeoLegends: Ok, that makes sense to me as well!

returnn/datasets/postprocessing.py
NeoLegends: Yeah, I figured that it doesn't hurt to allow both, if we specify the order they run in.

returnn/datasets/postprocessing.py
NeoLegends: I agree, then the code is wrong. But people are going to realize that their code is wrong by running into the error, at least at first and until they pull out a debugger. And then the error message can help them find the issue.

I changed the code to use a specific exception type, so that only dim mismatches are caught. WDYT about the implementation now?

returnn/datasets/audio.py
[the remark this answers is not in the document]
NeoLegends: Why is there that comment then though? It seems that it is somehow required in other parts of the code that there is a data entry?

NeoLegends:
>   returnn/datasets/postprocessing.py:14: WEAK WARNING PyProtectedMemberInspection: 'ShapeMismatchException' is not declared in __all__

I don't understand this lint. Why should that exception be in `__all__`? Does the lint want all raised exceptions in `__all__`?

returnn/datasets/postprocessing.py
albertz: No, I'm not speaking about the user code. I'm speaking about the code here in PostprocessingDataset. There should never be an exception raised here.

returnn/datasets/audio.py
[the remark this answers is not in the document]
albertz: This comment is about the API of `DatasetSeq`.

Actually, I think it's a bit outdated. You don't need to pass `targets` anymore, you can directly pass `features` as a dict, and then can also omit `"data"` completely. But otherwise, `features` must be some `ndarray`, but otherwise it does not matter.

returnn/datasets/postprocessing.py
albertz: But as argued, do not allow that. That's ambiguous and confusing.

returnn/datasets/postprocessing.py
NeoLegends: Ah, but the constructor of Tensor _does_ indeed trigger the shape validation error. Do you think I should create the tensor first, w/o the raw data, and then assign it after the fact in a second statement to make the validation step clearer? Or is there any other good way to validate that the data matches the shape set in the tensor?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+                self._dim_cache[name] = (dims, sparse_dim)
+
+            try:
+                return Tensor(name, dims=dims, dtype=dtype, sparse_dim=sparse_dim, raw_tensor=data)
+            except ShapeMismatchException as exc:
+                raise ShapeMismatchException(
+                    "Could not convert from mapping function output to `TensorDict`, "
+                    f"do the data shapes {data.shape} match up with the ones declared "
+                    f"in `map_output` (inferred to: {dims}, sparse={sparse_dim})?"
+                ) from exc
```
albertz, line 251:
```suggestion
            return Tensor(name, dims=dims, dtype=dtype, sparse_dim=sparse_dim, raw_tensor=data)
```

returnn/datasets/postprocessing.py
NeoLegends: I mean I can always just extract & re-use the logic from the tensor class to make the validation explicit

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+                dims = [Dim(dimension=None, name=f"{name}_frame"), *feature_dims]
+                if self._dataset.is_data_sparse(name):
+                    sparse_dim = Dim(
+                        dimension=self._dataset.get_data_dim(name) if self._dataset.is_data_sparse(name) else None,
+                        name=f"{name}_sparse",
+                    )
+                    if name in self._dataset.labels:
+                        sparse_dim.vocab = Vocabulary.create_vocab_from_labels(self._dataset.labels[name])
+                self._dim_cache[name] = (dims, sparse_dim)
+
```
albertz, line 243: Here is some check missing on the shape/dims.

returnn/datasets/postprocessing.py
albertz: There should never be an exception here, so there is no point in catching something which can never happen.

returnn/datasets/postprocessing.py
albertz: But PostprocessingDataset should already check it. You cannot rely that some other code does some checks for you.

tests/test_Dataset.py
```diff
@@ -962,6 +962,60 @@ def _get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+    with create_ogg_zip_txt_only_dataset_opts(text=_demo_txt) as sub_ds_opts:
+        ds_opts = {
+            "class": "PostprocessingDataset",
+            "dataset": sub_ds_opts,
+            "map_seq": _add_1337_to_classes,
+        }
+        dataset = init_dataset(ds_opts)
+        dataset.init_seq_order(epoch=1)
+        assert dataset.have_seqs()
+        dataset.init_seq_order(epoch=1)
```
albertz, line 983: What is the purpose to call this twice?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+            self._default_input = _meta_info_cache["_default_input"]
+            self._dim_cache = _meta_info_cache["_dim_cache"]
+            self._estimated_num_seqs = _meta_info_cache["_estimated_num_seqs"]
+            self.labels = _meta_info_cache["labels"]
+            self.num_inputs = _meta_info_cache["num_inputs"]
+            self.num_outputs = _meta_info_cache["num_outputs"]
+
+    def initialize(self):
+        """init"""
+        self._lazy_init_dataset()
```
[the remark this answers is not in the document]
NeoLegends: What is the contract wrt. the properties `num_output`, `labels`, etc.? After which calls are they expected to be initialized/available?

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+            To simplify the common case when no shapes change, this value can be left unspecified. The dataset then
+            assumes the same data layout as returned by the wrapped dataset.
+        :param _meta_info_cache: internal usage
+        :param kwargs: see :class:`CachedDataset2`, :class:`Dataset`
+        """
+        super().__init__(**kwargs)
+
+        assert (
+            self.seq_ordering == "default"
+        ), f"specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}"
```
albertz, line 88: I think we should start to be a bit more correct about the type of exceptions and the use of `assert`. `assert` is intended to check internal logical errors, which should never happen if our code is correct.

For anything else, internal or also based on user input, there are other exception types, usually `ValueError` or `TypeError` when it is about some input/args, or otherwise just using `Exception` is also fine.

```suggestion
        if self.seq_ordering != "default":
            raise ValueError(f"{self}: specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}")
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+        assert (
+            map_seq is not None or map_seq_stream is not None
+        ), "need to either define map_seq or map_seq_stream functions"
```
albertz, line 91: Also check that both are not defined together.
```suggestion
        if map_seq is None and map_seq_stream is None:
            raise ValueError(f"{self}: need to either define map_seq or map_seq_stream functions")
        if map_seq is not None and map_seq_stream is not None:
            raise ValueError(f"{self}: must not define map_seq and map_seq_stream together")        
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+            "dataset": {
+                "class": "HDFDataset",
+                "files": ["/path/to/data.hdf"],
+            },
+            # at least one of them:
+            "map_seq": map_seq,  # (data: TensorDict) -> TensorDict
+            "map_seq_stream": map_seqs,  # (iter: Iterator[TensorDict]) -> Iterator[TensorDict]
+            # only required when data shapes change wrt. the wrapped dataset:
+            "map_outputs": {
+                "data": { "dims": [time_dim, new_data_dim] },
```
albertz, line 56: Always use properly PEP8 formatted code in examples.
```suggestion
                "data": {"dims": [time_dim, new_data_dim]},
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def _lazy_init_dataset(self):
+        if self._dataset is not None:
+            return
+
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self.labels = self._dataset.labels
+        self._default_input = (
+            "data" if not self._map_outputs or "data" in self._map_outputs else next(self._map_outputs.keys())
+        )
```
albertz, line 138: That's wrong when `self._map_outputs` is not defined. `"data"` is not necessarily defined. Check `num_outputs` instead.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
+
+        if self._map_outputs is not None:
+            tdict = TensorDict(self._map_outputs)
```
albertz, line 142: Better also store this in the attrib.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+            "num_inputs": self.num_inputs,
+            "num_outputs": self.num_outputs,
+        }
+
+    def _lazy_init_dataset(self):
+        if self._dataset is not None:
+            return
+
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self.labels = self._dataset.labels
```
albertz, line 135: That's also wrong in general. You should read it from `map_outputs`.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+            else:
+                dims, sparse_dim = self._dim_cache.get(name, (None, None))
+                dtype = data.dtype
+
+            if dims is None:
+                feature_dims = [
+                    Dim(dimension=v, name=f"{name}_dim{i + 1}")
+                    for i, v in enumerate(self._dataset.get_data_shape(name))
+                ]
+                dims = [Dim(dimension=None, name=f"{name}_frame"), *feature_dims]
```
albertz, line 234: This here is not necessary. You should create the `Dim`s only once. You already did that before when you created the `TensorDict` object.

returnn/datasets/postprocessing.py
albertz: The `Dim`s are already created here.

tests/test_Dataset.py
```diff
@@ -962,6 +962,60 @@ def _get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+            "map_seq_stream": _repeat2,
+        }
+        dataset = init_dataset(ds_opts)
+        dataset.init_seq_order(epoch=1)
+        assert dataset.have_seqs()
+        dataset.init_seq_order(epoch=1)
+
+        dataset.load_seqs(0, 2)
+        for i in range(2):
+            dataset.get_data(i, "classes")
```
albertz, line 1015: Also check the output?

tests/test_Dataset.py
```diff
@@ -962,6 +962,60 @@ def _get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+        ds_opts = {
+            "class": "PostprocessingDataset",
+            "dataset": sub_ds_opts,
+            "map_seq": _count,
+            "map_seq_stream": _repeat2,
+        }
+        dataset = init_dataset(ds_opts)
+        dataset.init_seq_order(epoch=1)
+        assert dataset.have_seqs()
+        dataset.init_seq_order(epoch=1)
```
albertz, line 1011: Why do you call this twice?

tests/test_Dataset.py
```diff
@@ -962,6 +962,60 @@ def _get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+
+        dataset.load_seqs(0, 2)
+        for i in range(2):
+            dataset.get_data(i, "classes")
+        assert i == 1
```
albertz, line 1016: I don't understand this check. Are you testing the Python stdlib `range`? You can assume that `range` is correct.

returnn/datasets/postprocessing.py
NeoLegends: But only if `_map_outputs` is defined in `extern_data`-shape/type. If not, and we want to use the shape from the sub dataset we need to transform the sub-dataset's `num_output` into a form that is usable to initialize the `TensorDict` with. Not sure this is a good way to go, unless we switch over _all_ datasets to `TensorDict` now?

returnn/datasets/postprocessing.py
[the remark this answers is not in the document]
albertz: They are expected right after `__init__`.

So if the user passes `map_outputs`, you can infer all that information from there. If not, you anyway must init the subdataset right away.

To keep things simple, maybe just always init the subdataset right away. The subdataset itself should be efficient to create usually, most of the other datasets also have some lazy init logic, i.e. creating is cheap, and only reading the first data is expensive.

returnn/datasets/postprocessing.py
albertz: No, `Dim`s are always created.

NeoLegends: > always 

I don't understand why, the code initializes the `TensorDict` there only when `_map_outputs` is defined. Where is the other place?

albertz: I would actually suggest that if `_map_outputs` is not defined, you also create a `TensorDict` here, but derived from the subdataset. Then you have the code more consistent and simple and all `Dim`s already created here.

NeoLegends: Yes, okay, then I agree this would be nice. 

> If not, and we want to use the shape from the sub dataset we need to transform the sub-dataset's num_output into a form that is usable to initialize the TensorDict with.

I don't like that we need to transform the info from `self._dataset.num_outputs` though. Is there even enough info in there?

albertz: > I don't understand why, the code initializes the `TensorDict` there only when `_map_outputs` is defined. Where is the other place?

I meant for the case that `map_outputs` is defined. Then you even *must* reuse the `Dim`s from there.

But I was also suggesting to create the `TensorDict` just in all cases right here. Then the code is also simpler.

albertz: > I don't like that we need to transform the info from `self._dataset.num_outputs` though. Is there even enough info in there?

But that's exactly what you do already right now in `_make_tensor`, using `get_data_dim` and `get_data_shape`? I just say to move that logic right here, so you always have the template `TensorDict` already ready, and you can just get the `Dim`s from there.

NeoLegends: Yes, it's probably all wrong already. Imagine the case when one PostprocessingDataset wraps the other, the outer one then accesses `num_outputs` of the inner one, and thereby loses the `Dim` information potentially declared in the inner `map_outputs`-dict.

albertz: But `num_outputs` does not expose any `Dim`s?

returnn/tensor/tensor.py
```diff
@@ -180,7 +186,7 @@ def raw_tensor(self, value: Optional[RawTensorType]):
                 if dim.dimension is None:
                     continue  # we allow anything in the raw_tensor dim
                 if raw_shape[i] != dim.dimension:
-                    raise Exception(
+                    raise ShapeMismatchException(
```
albertz, line 189: Undo this change here. This does not belong into this PR. And this is not consistent. There are many other places throughout the `Tensor` code where shapes are checked, where it does not use this exception type consistently now. This requires much more code changes though, and they definitely do not belong here in this PR. And also not sure if really needed.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,253 @@
...
+
+    The integration into the data loading pipeline makes it easy to distribute the
+    data processing work across multiple CPU cores using `MultiProcDataset` and in
+    turn frees the GPU from data preprocessing tasks.
+
+    Both functions can be specified at the same time, and their results will be
+    combined by first applying the segment-level function and then the
+    multiple-segments-level function.
+
+    Example usage:
```
albertz, line 39:
```suggestion
    Example usage::
    
```

albertz: (Rest syntac.)

returnn/datasets/postprocessing.py
NeoLegends: How?

albertz: Assuming you have `self._map_outputs_tensor_dict` or so:
```suggestion
        self.labels = {}
        for k, v in self._map_outputs_tensor_dict.data.items():
            if v.vocab:
                self.labels[k] = v.vocab.labels
            elif v.sparse_dim:  # sparse_dim but not vocab
                self.labels[k] = list(map(str, range(v.sparse_dim.dimension)))  # dummy labels
```

returnn/datasets/audio.py
```diff
@@ -151,8 +151,6 @@ def __init__(
             self.num_outputs["classes"] = [self.targets.num_labels, 1]
         if self.feature_extractor:
             self.num_outputs["data"] = [self.num_inputs, 2]
-        else:
-            self.num_outputs["data"] = [0, 2]
```
albertz, line 155: Ideally we should do these `OggZipDataset` changes in a separate PR.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,226 @@
...
+    ):
+        """
+        :param dataset: inner dataset to be post-processed
+        :param map_seq: post processor function operating on the single-segment level.
+        :param map_seq_stream: post processor function operating on the multiple segment level via an iterator.
+            Allows merging multiple segments into one, or generating multiple output segments from one input segment.
+            When both functions are specified, the segment-level function is applied first to every segment, and the
+            results are passed to the iterator-level function.
+        :param map_outputs: Type and axis specification of the outputs of the mapping functions,
+            like extern_data and model_outputs.
```
albertz, line 75: A minimal example (`{"data": {"dim": 13}}` or so) would be good here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,226 @@
...
+        :param kwargs: see :class:`CachedDataset2`, :class:`Dataset`
+        """
+        super().__init__(**kwargs)
+
+        if self.seq_ordering != "default":
+            raise ValueError(f"specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}")
+        if map_seq is None and map_seq_stream is None:
+            raise ValueError("need to either set map_seq or map_seq_stream")
+        if map_seq and map_seq_stream:
+            raise ValueError("cannot set both map_seq and map_seq_stream")
```
albertz, line 88:
```suggestion
        if self.seq_ordering != "default":
            raise ValueError(f"{self}: specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}")
        if map_seq is None and map_seq_stream is None:
            raise ValueError(f"{self}: need to either set map_seq or map_seq_stream")
        if map_seq and map_seq_stream:
            raise ValueError(f"{self}: cannot set both map_seq and map_seq_stream")
```
Btw, when I make code suggestions here, usually you could (or even should) just overtake them directly (there is a button here for that, just click on it), unless you have good reasons to do it differently.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,226 @@
...
+        if self.num_outputs is None:
+            return None
+        return {
+            "_default_input": self._default_input,
+            "_dim_template_dict": self._dim_template_dict,
+            "_estimated_num_seqs": self._estimated_num_seqs,
+            "labels": self.labels,
+            "num_inputs": self.num_inputs,
+            "num_outputs": self.num_outputs,
+        }
```
albertz, line 120: If you do all the init of the `TensorDict` and thus also `labels`, `num_outputs` etc already in `__init__`, as we discussed, then I think this `_meta_info_cache` is not really needed.

albertz:
> Sometimes when applying suggestions they contain formatting errors, which is why I sometimes apply them manually.

You can also just fix the formatting error it in a follow-up commit then.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,211 @@
...
+        :param map_seq: post processor function operating on the single-segment level.
+        :param map_seq_stream: post processor function operating on the multiple segment level via an iterator.
+            Allows merging multiple segments into one, or generating multiple output segments from one input segment.
+            When both functions are specified, the segment-level function is applied first to every segment, and the
+            results are passed to the iterator-level function.
+        :param map_outputs: Type and axis specification of the outputs of the mapping functions,
+            like extern_data and model_outputs.
+            To simplify the common case when no shapes change, this value can be left unspecified. The dataset then
+            assumes the same data layout as returned by the wrapped dataset.
+        :param _meta_info_cache: internal usage
```
albertz, line 77:
```suggestion
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,211 @@
...
+    ):
+        """
+        :param epoch:
+        :param seq_list:
+        :param seq_order:
+        :return: whether the order changed (True is always safe to return)
+        """
+        super().init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
+
+        if epoch is None:
```
albertz, line 162:
```suggestion
        if epoch is None and seq_list is None and seq_order is None:
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
+        self._data_iter: Optional[Iterator[Tuple[int, TensorDict]]] = None
+        self._dim_template_dict: Optional[TensorDict]
+
+        if self._map_outputs is not None:
+            self._dim_template_dict = TensorDict(self._map_outputs)
+        else:
+            # Some sub datasets need their seq order initialized for the shape/dtype accessor functions to work
+            self._dataset.init_seq_order(epoch=1)
```
albertz, line 102: Definitely do not do this here. We must not call `init_seq_order` at this point on the subdataset. We definitely do not want to trigger their lazy init.

Also, in general, do not add workarounds for other buggy code, if we can just fix the other buggy code.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+        self._map_seq_stream = map_seq_stream
+        self._map_outputs = map_outputs
+
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
+        self._data_iter: Optional[Iterator[Tuple[int, TensorDict]]] = None
+        self._dim_template_dict: Optional[TensorDict]
+
+        if self._map_outputs is not None:
+            self._dim_template_dict = TensorDict(self._map_outputs)
```
albertz, line 99: Rename this.
```suggestion
            self._tensor_dict_template = TensorDict(self._map_outputs)
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+                        sparse_dim.vocab = Vocabulary.create_vocab_from_labels(self._dataset.labels[name])
+                dtype = self._dataset.get_data_dtype(name)
+                return Tensor(name, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
+
+            tensor_dict = {
+                name: _make_tensor_template(name)
+                for name, num_output_def in self._dataset.num_outputs.items()
+                if isinstance(
+                    num_output_def, (tuple, list)
+                )  # Some datasets (e.g. OggZipDataset) have data that does not conform to the typings in num_outputs
```
albertz, line 123:
```suggestion
                for name in self._dataset.get_data_keys()
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+            }
+            self._dim_template_dict = TensorDict(tensor_dict)
```
albertz, line 125: I would inline this with the prev line.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+        self._default_input = "data" if "data" in self.num_outputs else next(iter(self.num_outputs.keys()))
+        self.num_inputs = self.num_outputs[self._default_input][0]
+
+        self.labels = {}
+        for k, t in self._dim_template_dict.data.items():
+            if t.vocab:
+                self.labels[k] = t.vocab.labels
+            elif t.sparse_dim:  # sparse_dim but not vocab
+                self.labels[k] = list(map(str, range(t.sparse_dim.dimension)))  # dummy labels
+
```
albertz, line 139: You should also define most of the other accessor functions when possible (`get_data_shape`, `get_data_dtype`, etc).

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+        :return: generator providing data samples in the form of a TensorDict
+        """
+        data_keys = self._dataset.get_data_keys()
+
+        seq_index = 0
+        while self._dataset.is_less_than_num_seqs(seq_index):
+            self._dataset.load_seqs(seq_index, seq_index + 1)
+            data = {data_key: self._dataset.get_data(seq_index, data_key) for data_key in data_keys}
+            tensor_dict = self._dim_template_dict.copy_template()
+            tensor_dict.assign_from_raw_tensor_dict_(data)
```
albertz, line 194: This is not the right function here. `assign_from_raw_tensor_dict_` expects also seq sizes usually. Also, there is no real check on the shape.

tests/test_Dataset.py
albertz: Ah, I think I understand your intention now. You mix up the variable `i` here multiple times. You probably want a different name:
```suggestion
        assert count == 1
```

tests/test_Dataset.py
```diff
@@ -962,6 +962,56 @@ def _get_sub_epoch_dataset(files_subepoch: List[str]) -> Dict[str, Any]:
...
+    i = 0
+
+    def _repeat2(input_iter: Iterator[TensorDict]) -> Iterator[TensorDict]:
+        nonlocal i
+
+        for tdict in input_iter:
+            i += 1
+
+            yield tdict
+            yield tdict
```
albertz, line 997:
```suggestion
    count = 0

    def _repeat2(input_iter: Iterator[TensorDict]) -> Iterator[TensorDict]:
        nonlocal count
        for tdict in input_iter:
            count += 1
            yield tdict
            yield tdict
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+        :return: generator providing data samples in the form of a TensorDict
+        """
+        data_keys = self._dataset.get_data_keys()
+
+        seq_index = 0
+        while self._dataset.is_less_than_num_seqs(seq_index):
+            self._dataset.load_seqs(seq_index, seq_index + 1)
+            tensor_dict = self._tensor_dict_template.copy_template()
+            for data_key in data_keys:
+                tensor_dict.data[data_key].raw_tensor = self._dataset.get_data(seq_index, data_key)
```
albertz, line 177: There should be some check on the shape.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+    def _iterate_dataset(self) -> Iterator[TensorDict]:
+        """
+        :return: generator providing data samples in the form of a TensorDict
+        """
+        data_keys = self._dataset.get_data_keys()
+
+        seq_index = 0
+        while self._dataset.is_less_than_num_seqs(seq_index):
+            self._dataset.load_seqs(seq_index, seq_index + 1)
+            tensor_dict = self._tensor_dict_template.copy_template()
```
albertz, line 175: Hm wait, `self._tensor_dict_template` is about the output (maybe we should rename it further to also reflect that, like `self._out_tensor_dict_template`), but here you have this for the input. We should have a separate `self._in_tensor_dict_template`.

returnn/datasets/postprocessing.py
albertz: Ah or maybe this is fine here now, and we expect at this point that it's correct, and checks should have been done already beforehand.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+
+        self._dataset_def = dataset
+        self._map_seq = map_seq
+        self._map_seq_stream = map_seq_stream
+        self._map_outputs = map_outputs
+
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
+        self._data_iter: Optional[Iterator[Tuple[int, TensorDict]]] = None
+        self._tensor_dict_template: Optional[TensorDict]
```
albertz, line 99: This line is not necessary and wrong (it's not optional).
```suggestion
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if self._map_outputs is not None:
+            self._tensor_dict_template = TensorDict(self._map_outputs)
```
albertz, line 102: As said, should maybe be renamed further to `_out_tensor_dict_template` or so:
```suggestion
            self._out_tensor_dict_template = TensorDict(self._map_outputs)
```

returnn/datasets/postprocessing.py
NeoLegends: Ohh, good catch.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+        if map_seq and map_seq_stream:
+            raise ValueError(f"{self}: cannot set both map_seq and map_seq_stream")
+
+        self._dataset_def = dataset
+        self._map_seq = map_seq
+        self._map_seq_stream = map_seq_stream
+        self._map_outputs = map_outputs
+
+        self._dataset = init_dataset(self._dataset_def, parent_dataset=self)
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
```
albertz, line 97: I wonder if we really should overtake this. At least not in the case when `map_seq_stream` is used.

returnn/datasets/postprocessing.py
albertz: Or maybe not necessarily `self._in_tensor_dict_template` but you can directly create it here.

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+                tensor_dict.data[data_key].raw_tensor = self._dataset.get_data(seq_index, data_key)
+            if self._map_seq is not None:
+                tensor_dict = self._map_seq(tensor_dict)
+                assert isinstance(
+                    tensor_dict, TensorDict
+                ), f"map_seq must produce a {TensorDict.__name__}, but produced {type(tensor_dict).__name__}"
+            yield tensor_dict
+            seq_index += 1
+
+    def _make_tensor_template(self, name: str) -> Tensor:
```
albertz, line 186: Better?
```suggestion
    def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+
+    The integration into the data loading pipeline makes it easy to distribute the
+    data processing work across multiple CPU cores using `MultiProcDataset` and in
+    turn frees the GPU from data preprocessing tasks.
+
+    Both functions can be specified at the same time, and their results will be
+    combined by first applying the segment-level function and then the
+    multiple-segments-level function.
+
+    Example usage::
```
albertz, line 36:
```suggestion
    Example usage::
    
```

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,201 @@
...
+            `map_seq` and `map_seq_stream` can be specified at a time.
+        :param map_seq_stream: post processor function operating on the multiple segment level via an iterator.
+            Allows merging multiple segments into one, or generating multiple output segments from one input segment.
+            To avoid confusion on the order of how the processing functions are applied to the data, only one of
+            `map_seq` and `map_seq_stream` can be specified at a time.
+        :param map_outputs: Type and axis specification of the outputs of the mapping functions,
+            like extern_data and model_outputs.
+            To simplify the common case when no shapes change, this value can be left unspecified. The dataset then
+            assumes the same data layout as returned by the wrapped dataset.
+            Example: `map_outputs={"data":{"dim": 42}}`
```
albertz, line 79:
```suggestion
            Example: `map_outputs={"data": {"dim": 42}}`
```

returnn/datasets/postprocessing.py
albertz: It's important though that we use the same for the case it is the same (and thus also the same `Dim`s), e.g. when the user did not specify `map_outputs`

returnn/datasets/postprocessing.py
```diff
@@ -0,0 +1,202 @@
...
+        self._estimated_num_seqs = self._dataset.estimated_num_seqs
+        self._data_iter: Optional[Iterator[Tuple[int, TensorDict]]] = None
+        self._tensor_dict_template: Optional[TensorDict]
+
+        if self._map_outputs is not None:
+            self._tensor_dict_template = TensorDict(self._map_outputs)
+        else:
+            self._tensor_dict_template = TensorDict(
+                {name: self._make_tensor_template(name) for name in self._dataset.get_data_keys()}
+            )
```
albertz, line 107: This code basically should always be done first like this:
```suggestion
            self._in_tensor_dict_template = TensorDict(
                {name: self._make_tensor_template_from_input(name) for name in self._dataset.get_data_keys()}
            )
```
And then if `map_outputs` is not defined, you do:
```python
self._out_tensor_dict_template = self._in_tensor_dict_template
```

albertz (approved):

PR #1598, opened 2024-07-26
Add nose to requirements-dev

NeoLegends:
Any objections? Should we specify the version?


albertz:
No real objections, but also, I don't really use this anyway. I would not specify the version.

albertz (approved):

albertz:
Btw, one thing to add: nose is already deprecated, and we actually want to remove it, and it should not be used for new code.

PR #1599, opened 2024-07-30
OggZipDataset various fixes

NeoLegends:
- No longer relies on `init_seq_order` for data shape/type accessors to be callable
- No longer output dummy data w/ wrong shape
- No longer define `num_outputs["data"]` when no features are extracted.

Pulled out from #1596

### DIFF
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -422,23 +422,25 @@ class OggZipDataset(CachedDataset2):
             keys.append("classes")
         return [*keys, "orth", "raw"]
 
-    def get_data_shape(self, key):
+    def get_data_shape(self, key: str):
         """
         :returns get_data(*, key).shape[1:], i.e. num-frames excluded
         :rtype: list[int]
         """
+        assert key in self.get_data_keys()
         if key == "data" and self.feature_extractor is not None:
-            if self.feature_extractor.num_channels is not None:
-                return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
-        if key == "raw":
+            assert self.feature_extractor is not None
+            assert self.feature_extractor.num_channels is not None
+            return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
+        elif key in ["classes", "orth", "raw"]:
             return []
-        return super(OggZipDataset, self).get_data_shape(key)
+        else:
+            raise ValueError(f"{self}: unknown data key {key}")
 
-    def is_data_sparse(self, key) -> bool:
+    def is_data_sparse(self, key: str) -> bool:
         """:return: whether data entry with `key` is sparse"""
-        if key == "raw":
-            return False
-        return super().is_data_sparse(key)
+        assert key in self.get_data_keys()
+        return key == "classes"
 
     def _get_transcription(self, corpus_seq_idx: int):
         """
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -151,8 +151,6 @@ class OggZipDataset(CachedDataset2):
             self.num_outputs["classes"] = [self.targets.num_labels, 1]
         if self.feature_extractor:
             self.num_outputs["data"] = [self.num_inputs, 2]
-        else:
-            self.num_outputs["data"] = [0, 2]
         self._data: Optional[List[Dict[str, Any]]] = None  # lazily loaded
         self._fixed_random_subset = fixed_random_subset
         self._fixed_random_subset_seed = fixed_random_subset_seed
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -427,8 +427,7 @@ class OggZipDataset(CachedDataset2):
         :returns get_data(*, key).shape[1:], i.e. num-frames excluded
         :rtype: list[int]
         """
-        assert key in self.get_data_keys()
-        if key == "data" and self.feature_extractor is not None:
+        if key == "data":
             assert self.feature_extractor is not None
             assert self.feature_extractor.num_channels is not None
             return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
@@ -439,7 +438,6 @@ class OggZipDataset(CachedDataset2):
 
     def is_data_sparse(self, key: str) -> bool:
         """:return: whether data entry with `key` is sparse"""
-        assert key in self.get_data_keys()
         return key == "classes"
 
     def _get_transcription(self, corpus_seq_idx: int):
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -490,11 +490,11 @@ class OggZipDataset(CachedDataset2):
         """
         self._lazy_init()
         seq_tag = self._get_tag_from_info_dict(self._data[corpus_seq_idx])
+        data_features = {}
         if self.feature_extractor:
             with self._open_audio_file(corpus_seq_idx) as audio_file:
-                features = self.feature_extractor.get_audio_features_from_raw_bytes(audio_file, seq_name=seq_tag)
-        else:
-            features = numpy.zeros((), dtype=numpy.float32)  # currently the API requires some dummy values...
+                data = self.feature_extractor.get_audio_features_from_raw_bytes(audio_file, seq_name=seq_tag)
+            data_features = {"data": data}
         targets, txt = self._get_transcription(corpus_seq_idx)
         targets = numpy.array(targets, dtype="int32")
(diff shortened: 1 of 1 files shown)


returnn/datasets/audio.py
```diff
@@ -402,6 +400,24 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
         self._lazy_init()
         return len(self._data)
 
+    def get_data_dtype(self, key):
```
albertz, line 403: Docstring and typing is missing.

returnn/datasets/audio.py
```diff
@@ -402,6 +400,24 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+        elif key == "classes":
+            return "int32"
+        elif key == "raw":
+            return "string"
+        elif key == "orth":
+            return "uint8"
+        else:
+            raise ValueError(f"{self}: unknown data key: {key}")
+
+    def get_data_keys(self):
```
albertz, line 415: Docstring and typing.

returnn/datasets/audio.py
```diff
@@ -402,6 +400,24 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+        keys = ["classes", "orth", "raw"]
```
albertz, line 416: I think if `self.targets is None`, we should also omit `classes` here.

returnn/datasets/audio.py
```diff
@@ -402,6 +400,24 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+        if self.feature_extractor is not None:
+            keys.append("data")
```
albertz, line 418: I'm not sure how relevant this is, but I would put `"data"` first, `"classes"` next, and then the remaining ones.

returnn/datasets/audio.py
NeoLegends: It is always available in the `DatasetSeq` though. hmm

returnn/datasets/audio.py
```diff
@@ -483,8 +506,7 @@ def get_corpus_seq(self, corpus_seq_idx: int) -> DatasetSeq:
             orth = list(map(ord, orth))
         orth = numpy.array(orth, dtype="uint8")
         return DatasetSeq(
-            features=features,
-            targets={"classes": targets, "raw": raw_txt, "orth": orth},
+            features={**data_features, "classes": targets, "raw": raw_txt, "orth": orth},
```
albertz, line 512: Also here maybe exclude `targets` when `self.targets is None`.

returnn/datasets/audio.py
NeoLegends: Ah, but empty.

albertz: Yes it should neither be there as well.

returnn/datasets/audio.py
```diff
@@ -410,8 +430,16 @@ def get_data_shape(self, key):
...
                 return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
+        if key == "raw":
+            return []
         return super(OggZipDataset, self).get_data_shape(key)
 
+    def is_data_sparse(self, key) -> bool:
+        """:return: whether data entry with `key` is sparse"""
+        if key == "raw":
+            return False
+        return super().is_data_sparse(key)
```
albertz, line 441: Why do you rely on `super().is_data_sparse()` here? I would simply return it directly for all the possible cases.

returnn/datasets/audio.py
```diff
@@ -410,8 +430,16 @@ def get_data_shape(self, key):
         if key == "data" and self.feature_extractor is not None:
             if self.feature_extractor.num_channels is not None:
                 return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
+        if key == "raw":
+            return []
```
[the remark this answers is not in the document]
NeoLegends: yep, good point

returnn/datasets/audio.py
```diff
@@ -402,15 +400,47 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+        assert key in self.get_data_keys()
         if key == "data" and self.feature_extractor is not None:
-            if self.feature_extractor.num_channels is not None:
-                return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
-        return super(OggZipDataset, self).get_data_shape(key)
+            assert self.feature_extractor is not None
+            assert self.feature_extractor.num_channels is not None
+            return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
+        elif key in ["classes", "orth", "raw"]:
+            return []
```
NeoLegends, line 436: Is this correct for `classes`?

returnn/datasets/audio.py
```diff
@@ -402,15 +400,47 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+        if self.targets is not None:
+            keys.append("classes")
+        return [*keys, "orth", "raw"]
+
+    def get_data_shape(self, key: str):
         """
         :returns get_data(*, key).shape[1:], i.e. num-frames excluded
         :rtype: list[int]
         """
+        assert key in self.get_data_keys()
```
albertz, line 430: Not necessary. Instead, add the missing `else:` branch below.

returnn/datasets/audio.py
```diff
@@ -402,15 +400,47 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
         if key == "data" and self.feature_extractor is not None:
-            if self.feature_extractor.num_channels is not None:
```
albertz, line 411: I would change the `if` check here to just `if key == "data":` and then make an `assert self.feature_extractor is not None`.

returnn/datasets/audio.py
albertz: Yes.

returnn/datasets/audio.py
NeoLegends: Oh, yes. This is an oversight, I meant it to be that way.

returnn/datasets/audio.py
```diff
@@ -402,15 +400,47 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
+            assert self.feature_extractor.num_channels is not None
+            return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
+        elif key in ["classes", "orth", "raw"]:
+            return []
+        else:
+            raise ValueError(f"{self}: unknown data key {key}")
+
+    def is_data_sparse(self, key: str) -> bool:
+        """:return: whether data entry with `key` is sparse"""
+        assert key in self.get_data_keys()
```
albertz, line 442: Not necessary here.

albertz, line 442:
```suggestion
```

returnn/datasets/audio.py
albertz, line 430: Not necessary here.
```suggestion
```

NeoLegends: Why is it not necessary? Do you think it's appropriate to return values here for data keys that don't exist in the instance, e.g. when feature extraction is turned off and `data` does not exist?

albertz: The behavior of `get_data_shape` (and basically all others) is just undefined for unknown data keys. There is no need to verify that it is correct. It just adds additional complexity (an extra line of code) and also slows down the code.

But also, you anyway already have the `else:` branch below, so it's really also redundant here.

returnn/datasets/audio.py
```diff
@@ -402,15 +400,45 @@ def get_total_num_seqs(self, *, fast: bool = False) -> int:
...
         :rtype: list[int]
         """
-        if key == "data" and self.feature_extractor is not None:
-            if self.feature_extractor.num_channels is not None:
-                return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
-        return super(OggZipDataset, self).get_data_shape(key)
+        if key == "data":
+            assert self.feature_extractor is not None
+            assert self.feature_extractor.num_channels is not None
+            return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
```
albertz, line 433: That's wrong now. I think the correct code should be:
```suggestion
            if self.feature_extractor.num_channels is not None:
                return [self.feature_extractor.num_channels, self.feature_extractor.get_feature_dimension()]
            return [self.feature_extractor.get_feature_dimension()]
```

NeoLegends: Ah, I see. Thanks for the catch!

albertz (approved):

PR #1603, opened 2024-08-01
`FileCache`: detect modified source files

### LINKED ISSUE #1602: Make `FileCache` able to detect updated remote files
Sometimes corrupt or old files are updated/replaced on the remote file system. If a file is already cached on the local machine, the `FileCache` will not detect this situation and happily use the locally cached variant. I think we can improve on this by moving the cache-keepalive `mtime`-logic to a proxy file, e.g. `original-file.name.keepalive` and comparing the `mtime` of the actual data files.

NeoLegends:
...by storing the `mtime`s in dedicated keepalive files next to the cached files themselves.

Closes #1602 

@Icemole Do you want to test this in your setup?

### DIFF
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -312,7 +312,7 @@ class FileCache:
         return f"{filename}.returnn-keepalive"
 
     @staticmethod
-    def _is_keepalive_filename(filename: str) -> str:
+    def _is_keepalive_filename(filename: str) -> bool:
         """:return: whether `filename` points to a keepalive file."""
         return filename.endswith(".returnn-keepalive")
 
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -9,7 +9,7 @@ See https://github.com/rwth-i6/returnn/issues/1519 for initial discussion.
 Main class is :class:`FileCache`.
 """
 
-from typing import Any, Collection, List, Optional, Tuple
+from typing import Any, Collection, Iterable, List, Optional, Tuple, Union
 import errno
 import os
 import pathlib
@@ -35,16 +35,14 @@ class FileCache:
 
     Some assumptions we depend on:
 
-    - When a cached file is available, and its size matches the original file,
+    - When a cached file is available,
+      and its size matches the original file
+      and its mtime is not older than the remove file,
       we can use it.
-      (We might want to extend this logic later, e.g. by storing and checking the original source mtime.
-       But for simplicity, we don't do this for now.)
-    - We will update the cached file mtime frequently (every second) via a background thread
+    - We will update the mtime of a proxy file frequently (every second) via a background thread
       of used cached files, to mark that they are used.
       (We would maybe want to use atime, but we don't expect that atime can be relied on.)
       Note that updating mtime might influence the behavior of some external tools.
-      (If this becomes a problem at some point, instead of abusing mtime this way,
-       we might store the most recent usage time e.g. in a separate file.)
     - :func:`os.utime` will update mtime, and mtime is somewhat accurate (up to 10 secs maybe),
       mtime compares to time.time().
     - :func:`shutil.disk_usage` can be relied on.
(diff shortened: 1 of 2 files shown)


michelwi:
But this would not keep the nodes tmp cleanup scripts from deleting the data files.. would it?

NeoLegends:
No, this is not something the `FileCache` deals with anyway, though. What do the temp reaper scripts look at to determine if a file is to be cleaned up?

returnn/util/file_cache.py
```diff
@@ -337,8 +363,12 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
             return False
         src_stat = os.stat(src_filename)
         dst_stat = os.stat(dst_filename)
-        if src_stat.st_size != dst_stat.st_size:
+        if src_stat.st_size != dst_stat.st_size or src_stat.st_mtime > dst_stat.st_mtime:
             os.remove(dst_filename)
+            try:
+                os.remove(FileCache._get_keepalive_filename(dst_filename))
```
Icemole, line 369: In which situations can a keepalive file not be along with the file itself?

returnn/util/file_cache.py
```diff
@@ -337,8 +363,12 @@ def _check_existing_copied_file_maybe_cleanup(src_filename: str, dst_filename: s
...
+            except FileNotFoundError:
+                pass
```
Icemole, line 371: Above there's the following code when the keepalive file can't be found:
```
print(
    f"FileCache: Ignoring error file removing keepalive file of {fn}: {type(exc).__name__}: {exc}"
)
```

I'd suggest to uniformize the calls by either setting the print here (recommended), abstracting this into a function (also recommended, as you prefer) or getting rid of the original print and adding a `pass` instead.

Requesting changes for this.

returnn/util/file_cache.py
```diff
@@ -35,16 +35,14 @@ class FileCache:
 
     Some assumptions we depend on:
 
-    - When a cached file is available, and its size matches the original file,
+    - When a cached file is available,
+      and its size matches the original file
+      and its mtime is not older than the remove file,
```
Icemole, line 40:
```suggestion
      and its mtime is not older than the original file,
```

Icemole (changes requested):

Icemole:
Yes, I can test this with a small setup I had lying around. Let me know once you fix all tests via slack so that I test the final functionality.

albertz:
> ...by storing the `mtime`s in dedicated keepalive files next to the cached files themselves.

As discussed in #1602, wouldn't it make more sense the other way around?

I think it would cause some other unexpected problems. E.g. the [tmpreaper](https://manpages.ubuntu.com/manpages/trusty/man8/tmpreaper.8.html) would cleanup the files (it looks at mtime).

NeoLegends:
I like this idea.

> No, this is not something the FileCache deals with anyway, though

... and I stand corrected. Apparently since we update the mtimes of the files that are alive, they are indeed protected from the tmpreaper.

returnn/util/file_cache.py
```diff
@@ -397,6 +436,29 @@ class CachedFile:
...
+
+    We currently save the last known mtime of the source file.
+    """
+
+    mtime_ns: int
+
+    def save(self, filename: str):
+        """Saves this instance to `filename`."""
+        with open(filename, "wt") as f:
+            f.write(json.dumps(dataclasses.asdict(self)))
```
NeoLegends, line 452: JSON and not pickle, btw, because it is human readable.

albertz: Note, in most other cases, we store as Python string (repr), and then use eval (or ast eval). But Json is also fine.

Icemole (approved):
I've started a small training with this patch and the dev frame error rates are practically equivalent to the original ones. I'm approving.

michelwi (approved):

PR #1604, opened 2024-08-15
`PostprocessingDataset`: implement laplace sequence ordering

NeoLegends:
This PR implements laplace sequence ordering based on the number of segments per bin for the `PostprocessingDataset` via a buffer. The implementation is thereby compatible w/ all the different mapping variants and is designed to reduce padding as much as possible by always completing one full up-and-down cycle wrt. the sequence lengths in the resulting stream of segments.

@Stefanwuu Would you like to test this on your CTC model w/o chunking? :)

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -78,8 +78,20 @@ class PostprocessingDataset(CachedDataset2):
         """
         super().__init__(**kwargs)
 
-        if self.seq_ordering != "default":
-            raise ValueError(f"{self}: specify seq_ordering in wrapped dataset, not in {self.__class__.__name__}")
+        if self.seq_ordering != "default" and not self.seq_ordering.startswith("laplace:."):
+            raise ValueError(
+                f"{self}: can only specify default or laplace:.num_seqs-style seq ordering "
+                f"in {self.__class__.__name__}. "
+                "Any other seq_ordering must be set on wrapped dataset."
+            )
+        if self._seq_list_filter_file is not None:
+            raise ValueError(
+                f"{self}: specify seq_list_filter_file in wrapped dataset, not in {self.__class__.__name__}"
+            )
+        if self._seq_order_seq_lens_file is not None:
+            raise ValueError(
+                f"{self}: specify seq_order_seq_lens_file in wrapped dataset, not in {self.__class__.__name__}"
+            )
         if map_seq is None and map_seq_stream is None:
             raise ValueError(f"{self}: need to either set map_seq or map_seq_stream")
         if map_seq and map_seq_stream:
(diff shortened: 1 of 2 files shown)


albertz:
I'm not sure that we should implement this as part of the PostprocessingDataset. Why? That seems like a somewhat arbitrary choice to m. We could also have choose get MetaDataset. But I would also not do that for the same reason. Shouldn't we maybe just move this logic to a new separate dataset just dedicated for this? This is much more decoupled then.

albertz:
> I feel like having an entire dataset dedicated just to implementing Laplace sequence ordering (or maybe also other kinds of seq order) introduces too much boilerplate for the end user

Well we had the same discussion for `PostprocessingDataset`, where the exact same argument was made to include this as part of `MetaDataset` (the people who made that argument had a setup where they anyway used `MetaDataset`).



> But I could also see this on the `CachedDataset2`-level, since it conceptually also works on a dataset seq iterator (just not using the iterator interface).

Also here I think this mixes up too much unrelated logic. It just adds complexity to `CachedDataset2`. We should not add any complexity there. We should never add complexity when we can avoid it (by leaving things decoupled).


> Since all the ordering work is done by the generator it works equally well in RETURNN "userland".

I'm wondering about this. So why do we need to have any modification in RETURNN at all? RETURNN maybe could provide some predefined `map_seq_stream` functions, like for Laplace ordering, and/or chaining multiple `map_seq_stream` functions.

NeoLegends:
> I'm wondering about this. So why do we need to have any modification in RETURNN at all?

Ah, I forgot about `get_seq_order_for_epoch`, which is a function on the dataset itself. Maybe we need to extend the function's signature to take the dataset as parameter as well, then it could call the function to obtain the indices w/o having to rely on `self`.

> RETURNN maybe could provide some predefined map_seq_stream functions, like for Laplace ordering, and/or chaining multiple map_seq_stream functions.

Yes, such a library might be quite useful.

albertz:
> > I'm wondering about this. So why do we need to have any modification in RETURNN at all?
> 
> Ah, I forgot about `get_seq_order_for_epoch`, which is a function on the dataset itself. Maybe we need to extend the function's signature to take the dataset as parameter as well, then it could call the function to obtain the indices w/o having to rely on `self`.

I'm not sure I understand. Which function signature? `get_seq_order_for_epoch` already gets the dataset (as `self`). But why would you want to extend that? Or you mean `map_seq_stream`? But for what? It doesn't need anything from the dataset, and should not access it directly. In any case, why is that needed? When you want to pass some custom laplace ordering as a `map_seq_stream` function, that is not needed at all? It would look sth like:
```python
from returnn.datasets.postprocessing import LaplaceOrdering

dataset = PostprocessingDataset(
    ...,
    map_seq_stream=LaplaceOrdering(num_seqs=1000),
)

PR #1608, opened 2024-08-28
`PostprocessingDataset`: add laplace ordering `map_seq_stream` iterator

NeoLegends:
Alternative implementation to #1604 based solely on "RETURNN userland". The iterator can be imported into a RETURNN config like `from returnn.datasets.postprocessing import LaplaceOrdering` and then used.

Closes #1604

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -215,7 +215,7 @@ class PostprocessingDataset(CachedDataset2):
 
 class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
     """
-    Iterator compatible with `PostprocessingDataset`'s `map_seq_stream` applying
+    Iterator compatible with :class:`PostprocessingDataset`'s ``map_seq_stream`` applying
     laplace sequence ordering based on the number of segments per bin.
     """
 
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -4,6 +4,7 @@ Provides :class:`PostprocessingDataset`.
 
 from __future__ import annotations
 
+from itertools import islice
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


michelwi (approved):

returnn/datasets/postprocessing.py
```diff
@@ -210,3 +211,46 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+
+    def __call__(self, iterator: Iterator[TensorDict]) -> Iterator[TensorDict]:
+        """:return: generator applying laplace sequence ordering on the data"""
+        iterator = iter(iterator)
+        num_full_cycle = 2 * self.num_seqs_per_bin
+
+        while True:
+            # Fill two bins to complete one full upwards and downwards cycle wrt. the
+            # resulting seq lens and avoid a step/break.
+            seq_buffer = list(islice(iterator, num_full_cycle))
```
albertz, line 239: We could even do it more online by operating only on a half cycle. I think the more online, i.e. the less buffering it needs, the better?

Also, that is more consistent to the original laplace ordering behavior. (Specifically also for the last laplace bin.)

returnn/datasets/postprocessing.py
```diff
@@ -210,3 +211,46 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
             if data_key in self._dataset.labels:
                 sparse_dim.vocab = Vocabulary.create_vocab_from_labels(self._dataset.labels[data_key])
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
+
+
+class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
+    """
+    Iterator compatible with `PostprocessingDataset`'s `map_seq_stream` applying
```
albertz, line 218:
```suggestion
    Iterator compatible with :class:`PostprocessingDataset`'s ``map_seq_stream`` applying
```

albertz (approved):

Issue #1326, opened 2023-05-16
Drop older Python support


albertz:
Via #487, we are now supporting Python >=3.7.

Do we want to bump this and require Python >=3.8 or later? See also the Python feature list in #487.

Similarly, Python >=3.8 is now required for i6_core: https://github.com/rwth-i6/i6_core/pull/409, https://github.com/rwth-i6/i6_core/issues/408

Same discussion for Python >=3.8 for i6_models: https://github.com/rwth-i6/i6_models/pull/11

christophmluscher:
I would say at least 3.8. Ubuntu 22 defaults to 3.10, so maybe move to 3.10?!

curufinwe:
3.8 is ok (as it's the default for ubuntu 20.04). Anything higher I would delay for now until we (Apptek) have firmly tested our new ubuntu 22.04 images. Thanks!

albertz:
Btw, I think we don't really have any place where we explicitly write the min required Python version? We should maybe add that to the README or so.

NeoLegends:
Do you plan to make any changes to RETURNN during the version bump (other than bumping the min version)? I don't see many changes in 3.7 vs. 3.8 that would warrant large code changes? Maybe some arguments can be made positional-only for better forwards-compatibility, but this is at the same a backwards-compatibility risk, so I'm not sure it should be done to any existing code.

NeoLegends:
Ah, by dropping support for 3.7 we also drop support for TF1. See https://github.com/rwth-i6/returnn/pull/1668#issuecomment-2541204819.

PR #1609, opened 2024-08-29
`PostprocessingDataset`: add composition function

NeoLegends:
Makes a common pattern much easier.

@albertz Any comments on the added docs?

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -303,7 +303,7 @@ class Sequential(Generic[T]):
     iterators operating on multiple segments. Just make sure not to mix both styles.
     """
 
-    def __init__(self, *postprocessing_funcs: Callable[[T], T]):
+    def __init__(self, *postprocessing_funcs: Callable):
         """
         :param postprocessing_funcs: Postprocessing functions to compose.
         """
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from itertools import islice
 from numpy.random import Generator, PCG64
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 from returnn.datasets.basic import DatasetSeq
 from returnn.datasets.util.vocabulary import Vocabulary
@@ -53,6 +53,30 @@ class PostprocessingDataset(CachedDataset2):
             "map_outputs": {
                 "data": {"dims": [time_dim, new_data_dim]},
             },
+        }
+
+    The dataset itself does not support its own seq ordering and relies on the wrapped
+    dataset for seq ordering instead. Specifying a ``seq_ordering`` other than ``default``
+    results in an error.
+
+    However, we provide an iterator that implements the common `laplace:.NUM_SEQS_PER_BIN`-variant
+    of seq ordering that any custom ``map_seq_stream``-style postprocessing iterator can be composed
+    with to implement the ordering.
+
+    Like this::
+
+        from returnn.datasets.postprocessing import compose, LaplaceOrdering
+
+        def my_map_seq_stream(iterator):
+            ...
+
+        train = {
+            "class": "PostprocessingDataset",
+            # ...
+            "map_seq_stream": compose(
+                LaplaceOrdering(num_seqs_per_bin=1000),
+                my_map_seq_stream,
+            ),
         }
     """
 
@@ -228,10 +252,44 @@ class PostprocessingDataset(CachedDataset2):
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
+def compose(*postprocessing_funcs: Callable):
+    """
+    Composes multiple postprocessing functions into one by sequential application,
+    i.e. compose(f, g)(x) = (f ∘ g)(x) = f(g(x)).
+
+    Can either compose ``map_seq``-style single-segment processor functions or ``map_seq_stream``-style
+    iterators operating on multiple segments.
+
+    The functions are applied in reverse order, i.e. last argument first.
+
+    :return: composite function applying :param:``postprocessing_funcs`` in reverse order.
+    """
+
+    def wrapper(arg: Union[TensorDict, Iterator[TensorDict]], **kwargs) -> Union[TensorDict, Iterator[TensorDict]]:
+        """composite postprocessing function"""
+
+        # If we are passed an iterator do not check for the concrete type (which may be some generator),
+        # but for parent `Iterator` type instead.
+        arg_type = Iterator if isinstance(arg, Iterator) else type(arg)
+
+        for i, callable in enumerate(reversed(postprocessing_funcs)):
+            arg = callable(arg, **kwargs)
+            assert isinstance(arg, arg_type), (
+                f"Function {i} returned a value of type {type(arg)}, "
+                f"but must return the same type as the input ({arg_type}) to ensure valid composition. "
+                "Did you mix map_seq and map_seq_stream-style postprocessing functions?"
+            )
+        return arg
+
+    return wrapper
+
+
 class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
     """
     Iterator compatible with :class:`PostprocessingDataset`'s ``map_seq_stream`` applying
     laplace sequence ordering based on the number of segments per bin.
+
+    To be composed with any custom data postprocessing logic via ``compose``.
     """
 
     def __init__(self, num_seqs_per_bin: int, length_key: str = "data"):
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from itertools import islice
 from numpy.random import Generator, PCG64
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, TypeVar
 
 from returnn.datasets.basic import DatasetSeq
 from returnn.datasets.util.vocabulary import Vocabulary
@@ -252,6 +252,9 @@ class PostprocessingDataset(CachedDataset2):
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
+T = TypeVar("T", TensorDict, Iterator[TensorDict])
+
+
 def compose(*postprocessing_funcs: Callable):
     """
     Composes multiple postprocessing functions into one by sequential application,
@@ -265,7 +268,7 @@ def compose(*postprocessing_funcs: Callable):
     :return: composite function applying :param:``postprocessing_funcs`` in reverse order.
     """
 
-    def wrapper(arg: Union[TensorDict, Iterator[TensorDict]], **kwargs) -> Union[TensorDict, Iterator[TensorDict]]:
+    def wrapper(arg: T, **kwargs) -> T:
         """composite postprocessing function"""
 
         # If we are passed an iterator do not check for the concrete type (which may be some generator),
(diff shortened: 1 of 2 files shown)


returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,44 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+
+    Can either compose ``map_seq``-style single-segment processor functions or ``map_seq_stream``-style
+    iterators operating on multiple segments.
+
+    The functions are applied in reverse order, i.e. last argument first.
+
+    :return: composite function applying :param:``postprocessing_funcs`` in reverse order.
+    """
+
+    def wrapper(arg: Union[TensorDict, Iterator[TensorDict]], **kwargs) -> Union[TensorDict, Iterator[TensorDict]]:
```
michelwi, line 268: This looks like it should be a typevar bound to either of those as the return type does not change.

And then do we really need to manually typecheck the arguments via isinstance checks?

returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,44 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+        """composite postprocessing function"""
+
+        # If we are passed an iterator do not check for the concrete type (which may be some generator),
+        # but for parent `Iterator` type instead.
+        arg_type = Iterator if isinstance(arg, Iterator) else type(arg)
```
michelwi, line 273: what else is left if not `Iterator` except for `TensorDict`?

Or do we want that a specific subclass of `TensorDict` does not become a different subclass?

NeoLegends: It's about making sure users don't mix the iterator-form with the single-segment form.

returnn/datasets/postprocessing.py
```diff
@@ -54,6 +54,30 @@ class PostprocessingDataset(CachedDataset2):
...
             },
         }
+
+    The dataset itself does not support its own seq ordering and relies on the wrapped
+    dataset for seq ordering instead. Specifying a ``seq_ordering`` other than ``default``
+    results in an error.
+
+    However, we provide an iterator that implements the common `laplace:.NUM_SEQS_PER_BIN`-variant
+    of seq ordering that any custom ``map_seq_stream``-style postprocessing iterator can be composed
+    with to implement the ordering.
```
albertz, line 64:
```suggestion
    with to implement the ordering via :class:`LaplaceOrdering`.
```

returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,47 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
+T = TypeVar("T", TensorDict, Iterator[TensorDict])
+
+
+def compose(*postprocessing_funcs: Callable):
```
albertz, line 258: It's inconsistent that this is here a function (`compose`) vs having a class (`LaplaceOrdering`) otherwise.
I would suggest to make this here a class as well.

returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,47 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+    """
+
+    def wrapper(arg: T, **kwargs) -> T:
+        """composite postprocessing function"""
+
+        # If we are passed an iterator do not check for the concrete type (which may be some generator),
+        # but for parent `Iterator` type instead.
+        arg_type = Iterator if isinstance(arg, Iterator) else type(arg)
+
+        for i, func in enumerate(reversed(postprocessing_funcs)):
```
albertz, line 278: Don't do it reversed. That just adds confusion.
```suggestion
        for i, func in enumerate(postprocessing_funcs):
```

returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,47 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
+T = TypeVar("T", TensorDict, Iterator[TensorDict])
+
+
+def compose(*postprocessing_funcs: Callable):
+    """
+    Composes multiple postprocessing functions into one by sequential application,
+    i.e. compose(f, g)(x) = (f ∘ g)(x) = f(g(x)).
```
albertz, line 261:
```suggestion
    i.e. compose(f, g)(x) = (g ∘ f)(x) = g(f(x)).
```

returnn/datasets/postprocessing.py
```diff
@@ -228,10 +252,47 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+
+    Can either compose ``map_seq``-style single-segment processor functions or ``map_seq_stream``-style
+    iterators operating on multiple segments.
+
+    The functions are applied in reverse order, i.e. last argument first.
+
```
albertz, line 267:
```suggestion
```

returnn/datasets/postprocessing.py
NeoLegends: But this is how its done in math, and how I'd expect it to work.

albertz: What do you mean by "in math"? You mean when you write "f ∘ g"? But why does that matter?

For some algorithm/code, you would always assume the most simple logic, which is not reversing it but just using the original order. Only because some specific notation in math has the reverse order is irrelevant for this? Despite "f ∘ g", there are other notations being used which have the original order, not reversed order (see [here](https://en.wikipedia.org/wiki/Function_composition#Alternative_notations)).

And also, all other similar functions in common frameworks also don't reverse the order but keep the original order, e.g. [`torch.nn.Sequential`](https://pytorch.org/docs/stable/generated/torch.nn.Sequential.html).

But maybe instead of calling it `compose`, call it `Sequential`, to make that relation even more obvious.

michelwi: I would also find the "reversed" order more natural (because of writing "f ∘ g", but also the order how one would write it in code `foo(bar(baz(x))) =: compose(foo, bar, baz)(x)` ). 
So I would call it "reversed order" now and "in order" before.

I do not care much which order is implemented, though,  as long as it is documented well.

returnn/datasets/postprocessing.py
```diff
@@ -263,3 +289,29 @@ def _get_seq_len(self, tdict: TensorDict) -> int:
         :return: segment length of the segment in `tdict` as measured by `self.length_key` for comparison.
         """
         return tdict.data[self.length_key].raw_tensor.shape[0]
+
+
+T = TypeVar("T", TensorDict, Iterator[TensorDict])
+
+
+class Sequential(Callable):
```
michelwi, line 297:
```suggestion
class Sequential(Generic[T]):
```
Do we need `Callable` here? My IDE complains about Callable not being a base class.. anyway, otherwise inherit from both.

returnn/datasets/postprocessing.py
```diff
@@ -263,3 +289,29 @@ def _get_seq_len(self, tdict: TensorDict) -> int:
...
+    """
+    Callable that composes multiple postprocessing functions into one by sequential application,
+    i.e. Sequential(f, g)(x) = (g ∘ f)(x) = g(f(x)).
+
+    Can either compose ``map_seq``-style single-segment processor functions or ``map_seq_stream``-style
+    iterators operating on multiple segments. Just make sure not to mix both styles.
+    """
+
+    def __init__(self, *postprocessing_funcs: Callable):
```
michelwi, line 306:
```suggestion
    def __init__(self, *postprocessing_funcs: Callable[[T], T]):
```
just to better express that the type T that the callables use here is the same as the one expected and returned by the `__call__`.

returnn/datasets/postprocessing.py
albertz: Right, `Callable` is not needed, this is anyway implicit.

albertz: But I think `Generic[T]` is also not needed?

michelwi: > But I think Generic[T] is also not needed?

iirc if one does not inherit from `Generic[T]` then the T of the `__init__` could assume a different type value from the T in `__call__`

returnn/datasets/postprocessing.py
michelwi: Ah that is true unfortunately then we leave it, correct is better than specific.

returnn/datasets/postprocessing.py
albertz: Ah right.

returnn/datasets/postprocessing.py
```diff
@@ -263,3 +289,29 @@ def _get_seq_len(self, tdict: TensorDict) -> int:
...
+class Sequential(Generic[T]):
+    """
+    Callable that composes multiple postprocessing functions into one by sequential application,
+    i.e. Sequential(f, g)(x) = (g ∘ f)(x) = g(f(x)).
+
+    Can either compose ``map_seq``-style single-segment processor functions or ``map_seq_stream``-style
+    iterators operating on multiple segments. Just make sure not to mix both styles.
+    """
+
+    def __init__(self, *postprocessing_funcs: Callable[[T], T]):
```
albertz, line 306:
```suggestion
    def __init__(self, *postprocessing_funcs: Callable):
```
As discussed. Also, see the warning you currently get.

NeoLegends:
What is the syntax for the comments that turn off the linter for a specific line?

returnn/datasets/postprocessing.py
NeoLegends: ah, hm, that's another way.

returnn/datasets/postprocessing.py
albertz: But as discussed in another thread, you would also anyway not have the `T` in the `__init__`. I.e. then also not need the `Generic[T]` here.

albertz:
> What is the syntax for the comments that turn off the linter for a specific line?

Why? Where do you want to turn it off? I'm pretty sure there are no valid cases here to turn it off. All warnings given by the linter (see failing test) are totally correct, and you should fix the code instead.

returnn/datasets/postprocessing.py
NeoLegends: Then the `T` can change from invocation to invocation, which is also not correct.

returnn/datasets/postprocessing.py
albertz: I think with the rename to `Sequential`, this ambiguity is resolved.

returnn/datasets/postprocessing.py
```diff
@@ -15,7 +15,7 @@
 from .basic import init_dataset
 from .cached2 import CachedDataset2
 
-__all__ = ["PostprocessingDataset"]
+__all__ = ["PostprocessingDataset", "LaplaceOrdering", "T", "Sequential"]
```
albertz, line 18: The `T` does not need to be exported.
```suggestion
__all__ = ["PostprocessingDataset", "LaplaceOrdering", "Sequential"]
```

returnn/datasets/postprocessing.py
albertz: > Then the `T` can change from invocation to invocation, which is also not correct.

There are many possible other incorrect things which are not covered by the typechecker. I don't really understand the argument.

albertz: Note, if we want to have this type-checked, we need to use `Protocol` instead of `Callable`, but that requires Python 3.8, i.e. it depends on #1326, i.e. currently we cannot do this cleanly. I would say to just not type-check this currently, and once #1326 is solved, maybe it can be added. (But really not so important.)

returnn/datasets/postprocessing.py
albertz: As this code was removed now, I guess this here is not relevant anymore?

returnn/datasets/postprocessing.py
albertz: I guess this is not relevant anymore?

michelwi (approved):

PR #1610, opened 2024-08-30
PT preload_from_files ext: random init, external part

albertz:
Two features here (maybe we can also leave the individual commits as is):

### 1. `init_for_train=True` and `filename=None` and `pattern`
 
Explicitly specify some parameters which should be randomly initialized (thus `filename=None`) in training.

Note: Currently, if you start training (initial epoch 1), and there is `preload_from_files`, then if that does not cover all model parameters, it will exit with the error: "Missing key(s) in state_dict: ... Any missing key is an error.". I.e. the expectation is, once `preload_from_files` is used, that covers all model parameters.

In further training epochs, when an existing checkpoint is loaded, it will also do the check, but only for those params which are not already in the previous checkpoint. The same logic also applies also for recognition.

So, when you want to use `preload_from_files` for training to preload some of the parameters, and you don't cover all parameters, but you want to leave some parameters randomly initialized, then you need this feature.

### 2. `init_for_train="always"`

Specify some parameters which are always loaded, no matter if it is the initial first training epoch, some later training epoch, or recognition. This also will exclude those parameters from training and exclude those parameters when saving the checkpoint, as it is assumed that they will later again be loaded when they are needed. Or maybe you don't need them anyway for recognition, as they are only used for some auxiliary loss, or for knowledge distillation or so.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -10,6 +10,8 @@ import gc
 import os
 import time
 import socket
+import fnmatch
+import re
 
 import torch
 import torch.distributed
@@ -760,6 +762,20 @@ class Engine(EngineBase):
                 else:  # default: init for recog
                     if is_training:
                         continue
+                if opts["filename"] is None:
+                    print(f"Pre-load (initialize) weights for key '{preload_key}'", file=log.v3)
+                    pattern = opts["pattern"]
+                    match = re.compile(fnmatch.translate(pattern)).match
+                    remove = []
+                    for name in self._pt_model.state_dict().keys():
+                        if match(name) and name in missing_keys:
+                            remove.append(name)
+                    if remove:
+                        print(f"Randomly initialize params: {remove}", file=log.v3)
+                        missing_keys.difference_update(remove)
+                    else:
+                        print("(No relevant parameters matching.)", file=log.v3)
+                    continue
                 print(f"Pre-load weights for key '{preload_key}' from {opts['filename']}", file=log.v3)
                 preload_model_state = torch.load(opts["filename"])
                 if opts.get("checkpoint_key", "model") is not None:
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -3,7 +3,7 @@ Main engine for PyTorch
 """
 
 from __future__ import annotations
-from typing import Optional, Any, Union, Callable, Dict
+from typing import Optional, Any, Union, Callable, Dict, Set
 from contextlib import nullcontext
 
 import gc
@@ -43,6 +43,7 @@ from .data import extern_data as extern_data_util
 from .data.queued_data_iter import QueuedDataIter
 from .frontend.bridge import rf_module_to_pt_module
 from .util import diagnose_gpu
+from .util import module as util_module
 from .distributed import DistributedContext, get_ctx as dist_get_ctx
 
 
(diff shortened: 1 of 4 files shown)


Atticus1806:
>So, when you want to use preload_from_files for training to preload some of the parameters, and you don't cover all parameters, but you want to leave some parameters randomly initialized, then you need this feature.

I am wondering if this is true. I am currently doing training where I preload a model, but it does not contain all the parameters of the network (since I am only loading the teacher). It does not raise an exception, but instead prints all the missing keys. 

The code I am using (removed filename): 
```
"teacher": {
    "filename": ...,
    "ignore_missing": False,
    "ignore_params_prefixes": [],
    "init_for_train": True,
    "prefix": "teacher.",
}
```

Or what is the relation to `filename=None`? Where from would I preload without a file? 
I am probably missing the point of this PR?

albertz:
I mean this exception, which I get without this:
```python
if missing_keys:
    raise Exception(
        "\n".join(
            [
                f"While loading model {filename}, preload_from_files {bool(preload_from_files)}:",
                "Unexpected key(s) in state_dict: " + ", ".join(map(repr, sorted(unexpected_keys))),
                "Missing key(s) in state_dict: " + ", ".join(map(repr, sorted(missing_keys))),
                "Any missing key is an error.",
            ]
        )
    )
```
You don't get this because you have a `prefix` here, and the preload is *complete* for that prefix.

I have some preload which is *not complete*. Specifically, I have:
```python
preload_from_files = {
    ...,  # also some other things
    "03_encoder": {
        "init_for_train": "always",  # or True
        # this checkpoint here does not cover the full model, just the encoder
        "filename": enc.checkpoint,
        # no prefix here
    },
}
```

> Or what is the relation to `filename=None`? Where from would I preload without a file?

With `filename=None`, it would load from nowhere. It would not preload from anything but instead randomly initialize the parameters. But that will then not have `missing_keys`.

returnn/torch/util/module.py
```diff
@@ -0,0 +1,80 @@
...
+
+    parameter = getattr(module, parameter_name)
+    if not isinstance(parameter, torch.nn.Parameter):
+        raise ValueError(f"{parameter_name} is not a torch.nn.Parameter, got type {type(parameter).__name__}")
+    delattr(module, parameter_name)
+    parameter.requires_grad = False
+    module.register_buffer(parameter_name, parameter, persistent=persistent)
+
+
+def unregister_module_or_parameter_or_buffer(module: torch.nn.Module, attrib_name: str, *, deep: bool = True):
```
NeoLegends, line 46: Is this function relevant for the PR?

NeoLegends (approved):

returnn/torch/util/module.py
albertz: Ah yea I think I used it in an earlier version. Can be removed.

returnn/torch/engine.py
```diff
@@ -913,9 +951,16 @@ def _save_model(self):
         tmp_filename = filename + ".tmp_write"
         if os.path.exists(tmp_filename):
             os.unlink(tmp_filename)
+        state_dict = self._pt_model.state_dict()
+        if self._ignore_param_set:
+            # Do some extra check that we don't save the ignored parameters.
```
NeoLegends, line 956: I was wondering if this would cause downstream ergonomics issues if the resulting model was used outside of RETURNN, where the preloading logic was less handily available, e.g. in ONNX, and if you then do suddenly end up depending on their values. I suppose this is a weird edge case though, because if so, just load these parameters normally in training and freeze them.

albertz: Yea I would say this is up to the user. RETURNN should not restrict the user to do what he/she wants.

PR #1613, opened 2024-09-04
Use util function to generate forwards compat kwargs

NeoLegends:
Some DRY-ness.

Based on #1612, only merged afterwards.


NeoLegends:
No longer relevant.

PR #1623, opened 2024-09-10
`PostprocessingDataset`: pass along `seq_tag`

### LINKED ISSUE #1622: Sequence tags in Postprocessing dataset
It appears to me that the sequence tags in are dropped in the Postprocessing dataset. In the traceback one can see that the sequence tags look as follows:

    seq_tags = <local> Tensor{'seq_tag', [B?], dtype='string'}
    seq_tags.raw_tensor = <local> array(['seq-0', 'seq-1', 'seq-2', 'seq-3', 'seq-4', 'seq-5', 'seq-6',
                                         'seq-7', 'seq-8', 'seq-9', 'seq-10', 'seq-11', 'seq-12', 'seq-13',
                                         'seq-14', 'seq-15', 'seq-16', 'seq-17', 'seq-18', 'seq-19',
                                         'seq-20', 'seq-21', 'seq-22', 'seq-23', 'seq-24', 'seq-25',
                                         'seq-26', 's...

Here's a part of the dataset configuration

<details>
<summary>Dataset configuration</summary>
  
  ```
  train = {
      "class": "PostprocessingDataset",
      "dataset": {
          "class": "HDFDataset",
          "files": [...]
          "partition_epoch": 25,
          "seq_list_filter_file": "/path/to/file",
          "seq_ordering": "random",
      },
      "map_seq_stream": Sequential(
          apply_speed_perturbation, LaplaceOrdering(num_seqs_per_bin=1000)
      ),
      "seq_ordering": "default",
  }
  ```
</details>

NeoLegends:
Closes #1622

I added an assertion to make sure the seq_tags are not lost. Do we want this, or should we be lenient here? I think the breakage's impact to people from adding this assertion is rather limited.

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -202,7 +202,6 @@ class PostprocessingDataset(CachedDataset2):
             assert loaded_seq_idx <= seq_idx, "_collect_single_seq must be done monotonically"
             if loaded_seq_idx != seq_idx:
                 continue
-            assert "seq_tag" in tensor_dict.data, "seq_tag dropped from TensorDict in postprocessing pipeline"
             seq = DatasetSeq(
                 features={k: t.raw_tensor for k, t in tensor_dict.data.items()},
                 seq_idx=seq_idx,
@@ -217,6 +216,7 @@ class PostprocessingDataset(CachedDataset2):
 
         def _validate_tensor_dict_iter(inner: Iterator[TensorDict]) -> Iterator[TensorDict]:
             for t_dict in inner:
+                assert "seq_tag" in t_dict.data, "seq_tag dropped from TensorDict in postprocessing pipeline"
                 for data_key, out_t in self._out_tensor_dict_template.data.items():
                     in_t = t_dict.data[data_key]
                     assert (
@@ -258,6 +258,12 @@ class PostprocessingDataset(CachedDataset2):
                 assert isinstance(
                     tensor_dict, TensorDict
                 ), f"map_seq must produce a {TensorDict.__name__}, but produced {type(tensor_dict).__name__}"
+
+                # Re-adding the seq tag here causes no harm in case it's dropped since we don't
+                # add/drop any segments w/ the non-iterator postprocessing function.
+                if not "seq_tag" in tensor_dict.data:
+                    tensor_dict.data["seq_tag"].raw_tensor = str_to_numpy_array(self._dataset.get_tag(seq_index))
+
             yield tensor_dict
             seq_index += 1
 
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -203,7 +203,7 @@ class PostprocessingDataset(CachedDataset2):
             if loaded_seq_idx != seq_idx:
                 continue
             seq = DatasetSeq(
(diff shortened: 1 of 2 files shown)


returnn/datasets/postprocessing.py
```diff
@@ -237,9 +245,12 @@ def _iterate_dataset(self) -> Iterator[TensorDict]:
         seq_index = 0
         while self._dataset.is_less_than_num_seqs(seq_index):
             self._dataset.load_seqs(seq_index, seq_index + 1)
+
             tensor_dict = self._in_tensor_dict_template.copy_template()
             for data_key in data_keys:
                 tensor_dict.data[data_key].raw_tensor = self._dataset.get_data(seq_index, data_key)
+            tensor_dict.data["seq_tag"].raw_tensor = str_to_numpy_array(self._dataset.get_tag(seq_index))
```
albertz, line 252: You can also add this to `if self._map_seq...` below, in case the user did not add it to the `tensor_dict`.

returnn/datasets/postprocessing.py
```diff
@@ -199,7 +202,12 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
             assert loaded_seq_idx <= seq_idx, "_collect_single_seq must be done monotonically"
             if loaded_seq_idx != seq_idx:
                 continue
-            seq = DatasetSeq(features={k: t.raw_tensor for k, t in tensor_dict.data.items()}, seq_idx=seq_idx)
+            assert "seq_tag" in tensor_dict.data, "seq_tag dropped from TensorDict in postprocessing pipeline"
```
albertz, line 205: I think this check is better in `_validate_tensor_dict_iter`.

returnn/datasets/postprocessing.py
NeoLegends: So you would re-add it, in case `_map_seq` dropped it? I'd add the tag to the tensor dict in any case before the callback is invoked, too, in case it depends on the seq tag for e.g. FSA/alignment lookup or so.

albertz: Yea, I think that's safe here, and a reasonable thing to do.

NeoLegends: Alright, done! Then this PR is good for review/merge!

returnn/datasets/postprocessing.py
```diff
@@ -258,6 +258,12 @@ def _iterate_dataset(self) -> Iterator[TensorDict]:
                 assert isinstance(
                     tensor_dict, TensorDict
                 ), f"map_seq must produce a {TensorDict.__name__}, but produced {type(tensor_dict).__name__}"
+
+                # Re-adding the seq tag here causes no harm in case it's dropped since we don't
+                # add/drop any segments w/ the non-iterator postprocessing function.
+                if not "seq_tag" in tensor_dict.data:
```
albertz, line 264:
```suggestion
                if "seq_tag" not in tensor_dict.data:
```

NeoLegends:
Oh, should we drop the seq tag from the `DatasetSeq` features?

albertz:
> Oh, should we drop the seq tag from the `DatasetSeq` features?

Ah yes.

PR #1624, opened 2024-09-10
`OggZipDataset`: normalize // to / when reading files from archive

NeoLegends:
The dataset implementation joins some (potentially empty) paths, sometimes leading to double slashes in the resulting paths. This can break lookup in the zip file, even though there is a file at that path. By normalizing // to / the file can be read from the archive properly. On normal file systems this usually isn't a problem, because they also do this.

### DIFF
--- a/returnn/datasets/audio.py
+++ b/returnn/datasets/audio.py
@@ -179,7 +179,7 @@ class OggZipDataset(CachedDataset2):
 
                 return gzip.open(self._separate_txt_files[name], "rb").read()
         if self._zip_files is not None:
-            return self._zip_files[zip_index].read(filename)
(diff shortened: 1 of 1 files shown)


albertz:
This sounds like a bug in the dataset preparation pipeline?

And there are also much more cases how it could be wrong, e.g. having `///` in it, or having `..` in it, or whatever else. And I'm not sure to what degree we should add workarounds on RETURNN side just because you don't want to fix the bug in the dataset preparation pipeline?

albertz:
What is the status here? Was this just a bug in the dataset preparation pipeline, as I assumed? So then just fix the dataset preparation pipeline, and this here is not needed?

albertz:
Why is this still open?

PR #1630, opened 2024-09-23
Datasets: implement support for within-dataset sharding

NeoLegends:
This PR adds support for within-dataset sharding, allowing e.g. the use of `MultiProcDataset` wrapping a `PostprocessingDataset`.

### DIFF
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -647,7 +647,7 @@ class Dataset(object):
         :param shard_index: index of the current data shard
         :return: partition of seq_index for current epoch
         """
-        assert num_shards > shard_index
+        assert 0 <= shard_index < num_shards
         num_seqs = len(seq_index)
         current_partition = ((epoch or 1) - 1) % partition_epoch
         seqs_per_epoch = num_seqs // partition_epoch
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -111,8 +111,8 @@ class Dataset(object):
         min_chunk_size=0,
         chunking_variance=0,
         estimated_num_seqs=None,
-        _num_data_shards=1,
-        _data_shard_index=0,
+        _num_shards=1,
+        _shard_index=0,
     ):
         """
         :param str name: e.g. "train" or "eval"
@@ -136,8 +136,8 @@ class Dataset(object):
         :param str|None seq_order_seq_lens_file: for seq order, use the seq length given by this file
         :param int shuffle_frames_of_nseqs: shuffles the frames. not always supported
         :param None|int estimated_num_seqs: for progress reporting in case the real num_seqs is unknown
-        :param int _num_data_shards: number of shards the data is split into
-        :param int _data_shard_index: local shard index, when sharding is enabled
+        :param int _num_shards: number of shards the data is split into
+        :param int _shard_index: local shard index, when sharding is enabled
         """
         self.name = name or ("dataset_id%s" % id(self))
         self.lock = None  # type: Optional[RLock]  # Used when manipulating our data potentially from multiple threads.
@@ -171,9 +171,9 @@ class Dataset(object):
         self._chunking = chunking
         self.chunk_size, self.chunk_step, self.custom_chunking_func = self._parse_chunking(chunking)
         self._context_window = context_window
-        assert 0 <= _data_shard_index < _num_data_shards
-        self._num_data_shards = _num_data_shards
-        self._data_shard_index = _data_shard_index
+        assert 0 <= _shard_index < _num_shards
+        self._num_shards = _num_shards
+        self._shard_index = _shard_index
         if isinstance(context_window, (tuple, list)):
             assert len(context_window) == 2
             for elem in context_window:
@@ -604,9 +604,9 @@ class Dataset(object):
             seq_index = [
                 i for i in seq_index if (all_seq_tags[i] not in used_seq_tags, used_seq_tags.add(all_seq_tags[i]))[0]
             ]
-        if partition_epoch > 1 or self._num_data_shards > 1:
+        if partition_epoch > 1 or self._num_shards > 1:
             seq_index = self._apply_partition_epoch_and_sharding(
-                seq_index, partition_epoch, epoch, self._num_data_shards, self._data_shard_index
+                seq_index, partition_epoch, epoch, self._num_shards, self._shard_index
             )
         if repeat_epoch > 1:
             seq_index = list(seq_index) * repeat_epoch
@@ -698,8 +698,8 @@ class Dataset(object):
         self.epoch = epoch
         self.rnd_seq_drop = Random(self._get_random_seed_for_epoch(epoch=epoch))
         assert (
-            self._num_data_shards == 1 or self.supports_sharding()
-        ), f"{self}: does not support sharding, but got num_data_shards == {self._num_data_shards}"
+            self._num_shards == 1 or self.supports_sharding()
+        ), f"{self}: does not support sharding, but got num_data_shards == {self._num_shards}"
         return False
 
     def finish_epoch(self, *, free_resources: bool = False):
@@ -1513,9 +1513,9 @@ def _dataset_extend_default_kwargs_from_parent_dataset(
     default_kwargs = default_kwargs.copy() if default_kwargs else {}
     default_kwargs.setdefault("random_seed_offset", parent_dataset.random_seed_offset)
     # noinspection PyProtectedMember
-    default_kwargs.setdefault("_num_data_shards", parent_dataset._num_data_shards)
+    default_kwargs.setdefault("_num_shards", parent_dataset._num_shards)
     # noinspection PyProtectedMember
-    default_kwargs.setdefault("_data_shard_index", parent_dataset._data_shard_index)
+    default_kwargs.setdefault("_shard_index", parent_dataset._shard_index)
     return default_kwargs
 
 
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1439,7 +1439,6 @@ def init_dataset(
     default_kwargs: Optional[Dict[str, Any]] = None,
     *,
     parent_dataset: Optional[Dataset] = None,
-    forward_sharding_config: bool = False,
 ) -> Dataset:
     """
     :param kwargs:
@@ -1459,7 +1458,6 @@ def init_dataset(
             extra_kwargs=extra_kwargs,
             default_kwargs=default_kwargs,
             parent_dataset=parent_dataset,
-            forward_sharding_config=forward_sharding_config,
         )
     if isinstance(kwargs, str):
         if kwargs.startswith("{"):
@@ -1474,14 +1472,11 @@ def init_dataset(
                 extra_kwargs=extra_kwargs,
                 default_kwargs=default_kwargs,
                 parent_dataset=parent_dataset,
-                forward_sharding_config=forward_sharding_config,
             )
         else:
             config_str = kwargs
             kwargs = {}
-            default_kwargs = _dataset_extend_default_kwargs_from_parent_dataset(
-                default_kwargs, parent_dataset, forward_sharding_config=forward_sharding_config
-            )
+            default_kwargs = _dataset_extend_default_kwargs_from_parent_dataset(default_kwargs, parent_dataset)
             if default_kwargs:
                 kwargs.update(default_kwargs)
             if extra_kwargs:
@@ -1500,9 +1495,7 @@ def init_dataset(
     clazz = get_dataset_class(clazz_name)
     if not clazz:
         raise Exception("Dataset class %r not found" % clazz_name)
-    default_kwargs = _dataset_extend_default_kwargs_from_parent_dataset(
-        default_kwargs, parent_dataset, forward_sharding_config=forward_sharding_config
-    )
+    default_kwargs = _dataset_extend_default_kwargs_from_parent_dataset(default_kwargs, parent_dataset)
     if default_kwargs:
         for key, value in default_kwargs.items():
             kwargs.setdefault(key, value)
@@ -1519,9 +1512,7 @@ def init_dataset(
 
 
 def _dataset_extend_default_kwargs_from_parent_dataset(
-    default_kwargs: Optional[Dict[str, Any]],
-    parent_dataset: Optional[Dataset],
-    forward_sharding_config: bool = False,
+    default_kwargs: Optional[Dict[str, Any]], parent_dataset: Optional[Dataset]
 ) -> Optional[Dict[str, Any]]:
     """
     :param default_kwargs:
@@ -1531,9 +1522,8 @@ def _dataset_extend_default_kwargs_from_parent_dataset(
         return default_kwargs
     default_kwargs = default_kwargs.copy() if default_kwargs else {}
     default_kwargs.setdefault("random_seed_offset", parent_dataset.random_seed_offset)
-    if forward_sharding_config:
-        default_kwargs["num_data_shards"] = parent_dataset.num_data_shards
-        default_kwargs["data_shard_index"] = parent_dataset.data_shard_index
+    default_kwargs.setdefault("num_data_shards", parent_dataset.num_data_shards)
+    default_kwargs.setdefault("data_shard_index", parent_dataset.data_shard_index)
     return default_kwargs
 
 
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -647,23 +647,20 @@ class Dataset(object):
         :param shard_index: index of the current data shard
         :return: partition of seq_index for current epoch
         """
-        assert 0 <= shard_index < num_shards
+        assert num_shards > shard_index
         num_seqs = len(seq_index)
-        current_partition = ((epoch or 1) - 1) % partition_epoch
-        seqs_per_epoch = num_seqs // partition_epoch
-        partition_sizes = [seqs_per_epoch + 1] * (num_seqs % partition_epoch) + [seqs_per_epoch] * (
-            partition_epoch - num_seqs % partition_epoch
+        num_partitions = partition_epoch * num_shards
+        current_partition = (((epoch or 1) - 1) % partition_epoch) * num_shards + shard_index
+        seqs_per_epoch = num_seqs // num_partitions
+        partition_sizes = [seqs_per_epoch + 1] * (num_seqs % num_partitions) + [seqs_per_epoch] * (
+            num_partitions - num_seqs % num_partitions
         )
-        assert sum(partition_sizes) == num_seqs and len(partition_sizes) == partition_epoch
+        assert sum(partition_sizes) == num_seqs and len(partition_sizes) == num_partitions
         partitions = functools.reduce(lambda a, x: a + [a[-1] + x], partition_sizes, [0])  # cumulative sum
-        assert len(partitions) == partition_epoch + 1
-        seq_index = seq_index[shard_index:][::num_shards][
-            partitions[current_partition] : partitions[current_partition + 1]
-        ]
-        assert len(seq_index) in [
-            (partition_sizes[current_partition] // num_shards),
-            (partition_sizes[current_partition] // num_shards) + 1,
-        ]
+        assert len(partitions) == num_partitions + 1
+        seq_index = seq_index[partitions[current_partition] : partitions[current_partition + 1]]
+        assert len(seq_index) == partition_sizes[current_partition]
+
         return seq_index
 
     def supports_sharding(self) -> bool:
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1446,6 +1446,7 @@ def init_dataset(
     :param extra_kwargs:
     :param default_kwargs:
     :param parent_dataset: if given, will adapt some of the default_kwargs (when not set)
+    :param forward_sharding_config: whether to forward the sharding config from the parent dataset
     """
     assert kwargs
     if isinstance(kwargs, Dataset):
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -647,7 +647,7 @@ class Dataset(object):
         :param shard_index: index of the current data shard
         :return: partition of seq_index for current epoch
         """
-        assert num_shards > shard_index
+        assert 0 <= shard_index < num_shards
         num_seqs = len(seq_index)
         num_partitions = partition_epoch * num_shards
         current_partition = (((epoch or 1) - 1) % partition_epoch) * num_shards + shard_index
--- a/returnn/datasets/basic.py
+++ b/returnn/datasets/basic.py
@@ -1330,22 +1330,6 @@ class Dataset(object):
             shape = [max(shape[0], batch.max_num_frames_per_slice[data_key]), shape[1] + batch.num_slices]
         return tuple(shape)
 
-    @classmethod
-    def extend_default_kwargs_from_parent_dataset(
-        cls, default_kwargs: Optional[Dict[str, Any]], parent_dataset: Optional[Dataset]
-    ) -> Optional[Dict[str, Any]]:
-        """
-        :param default_kwargs:
-        :param parent_dataset:
-        """
-        if not parent_dataset:
-            return default_kwargs
-        default_kwargs = default_kwargs.copy() if default_kwargs else {}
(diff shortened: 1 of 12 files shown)


returnn/datasets/postprocessing.py
```diff
@@ -186,6 +186,11 @@ def init_seq_order(
                 pass  # some datasets don't know their num_seqs
         return True
 
+    def get_current_seq_order(self):
+        """:return: current seq order of wrapped dataset"""
+        assert self._dataset is not None
+        return self._dataset.get_current_seq_order()
```
albertz, line 192: That's only correct when `map_seq_stream` is not used.

albertz (changes requested):

NeoLegends:
What do we want from seq_order + map_seq_stream support?

- We want to shard the returned segments returned from the dataset across processes w/ MultiProcDS
- We preferably want to pass the seq order info on to the sub dataset, so it can apply a similar logic
- We can emulate sharding by discarding every element but the `n-th + i` from the iterator. This is suboptimal because we lose the efficiency benefit of MultiProc sharding

Idea: we add a flag that, when set, asserts `map_seq_stream` behaves in a compatible way to the seq_order passed from MultiProcDS. I.e. it will not invent any additional elements that are not themselves based on the seqs from the underlying DS. We then pass on the seq_order normally to the sub dataset, where the sharding actually happens.

albertz:
Just to note what we discussed offline:

* `get_current_seq_order`/`get_all_tags` must always be correct. Thus it cannot be used together with `map_seq_stream`, as the main purpose of `map_seq_stream` (over just `map_seq`) is to modify the sequences (e.g. concat them or so).
* `MultiProcDataset` must have another mechanism to do the sharding. This new mechanism would be enabled by some option, or flag (e.g. `sharding_method` or so).
* This new way would extend the current `partition_epoch` logic, by another `num_shards` and `shard_idx`. This would be implemented in `get_seq_order_for_epoch`. Any datasets which are not using this would just throw an error when `partition_epoch/`num_shards` are not handled properly.
* `num_shards`/`shard_idx` are always passed by the parent dataset, or maybe by RETURNN itself (e.g. we could also use this for multi-GPU training).
* `num_shards`/`shard_idx` (and thus maybe also `partition_epoch`, when set, for consistency) are passed on to child datasets for all the wrapper datasets (like `PostprocessingDataset`).

returnn/datasets/basic.py
```diff
@@ -111,6 +111,8 @@ def __init__(
         min_chunk_size=0,
         chunking_variance=0,
         estimated_num_seqs=None,
+        num_data_shards=1,
+        data_shard_index=0,
```
albertz, line 115: As discussed, should be:
```suggestion
        _num_data_shards=1,
        _data_shard_index=0,
```

returnn/datasets/postprocessing.py
```diff
@@ -186,6 +186,13 @@ def init_seq_order(
                 pass  # some datasets don't know their num_seqs
         return True
 
+    def get_current_seq_order(self):
+        """:return: current seq order of wrapped dataset, if map_seq_stream is not used"""
+        if self._map_seq_stream is not None:
+            raise util.OptionalNotImplementedError
```
albertz, line 192: That's wrong. That's not something which is not implemented (also optionally not implemented does anyway not make sense). That's something which is just not allowed / not possible.
```suggestion
            raise Exception(f"{self} get_current_seq_order not allowed with map_seq_stream")
```

returnn/datasets/multi_proc.py
```diff
@@ -308,8 +361,6 @@ def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
             msg, num_seqs = self._seq_order_proc_parent_conn.recv()
             assert msg == "num_seqs"
             self._num_seqs = num_seqs
-        else:
-            self._num_seqs = 0
```
albertz, line 312: There is an `else:` missing here.

returnn/datasets/multi_proc.py
```diff
@@ -33,13 +33,18 @@ def __init__(
...
         _meta_info_cache: Optional[Dict[str, Any]] = None,
         **kwargs,
     ):
         """
         :param dataset: the dataset to use
         :param num_workers: number of workers to use
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
+        :param sharding_method: which method to use for sharding the data across the worker procs.
+            Default is ``seq_order``, which fetches the full list of seq indices from worker 0, and then
+            distributes shards of that to the other workers.
```
albertz, line 46:
```suggestion
            Default is ``seq_order``, which fetches the full list of seq indices,
            and then distributes shards of that to the other workers.
```
"from worker 0" is an implementation detail, and could also change (has already been changed in the past), and not relevant/important.

albertz:
There are failing tests. Please always check that.

returnn/datasets/multi_proc.py
```diff
@@ -87,61 +95,74 @@ def _meta_info_cache(self):
...
+
+        _mp = NonDaemonicSpawnContext(process_pre_init_func=SubProcCopyGlobalConfigPreInitFunc())
+
+        # Seq order proc (first worker) directly sends the seq order to each (other) worker.
+        seq_order_to_worker = []  # type: List[mpConnection]
+        worker_from_seq_order = []  # type: List[mpConnection]
+        for i in range(self.num_workers - 1):
+            reader, writer = _mp.Pipe(duplex=False)
+            seq_order_to_worker.append(writer)
+            worker_from_seq_order.append(reader)
```
albertz, line 109: These are only needed for the `seq_order` sharding method, right?

returnn/datasets/sprint.py
```diff
@@ -1235,6 +1235,10 @@ def supports_seq_order_sorting(self) -> bool:
         """supports sorting"""
         return True
 
+    def supports_sharding(self) -> bool:
+        """:return: whether this dataset supports sharding"""
+        return True
```
albertz, line 1240:
```suggestion
        return False
```
Or not?

returnn/datasets/meta.py
```diff
@@ -445,6 +445,14 @@ def supports_seq_order_sorting(self) -> bool:
             return True
         return False
 
+    def supports_sharding(self) -> bool:
+        """:return: whether this dataset supports sharding"""
+        return (
+            self.datasets[self.seq_order_control_dataset].supports_sharding()
+            if self.seq_order_control_dataset is not None
+            else True
```
albertz, line 453:
```suggestion
            else False
```
?

returnn/datasets/meta.py
```diff
@@ -1142,7 +1150,7 @@ def _get_random_dataset_seq_order(self, total_num_seqs):
         assert sum(counters) == total_num_seqs
 
         if self.partition_epoch:
```
albertz, line 1152: Shouldn't this `if` be extended by `num_shards`?

returnn/datasets/basic.py
```diff
@@ -622,14 +631,23 @@ def get_seq_order_for_epoch(
...
-        :param int|None epoch: current epoch
+        :param seq_index: full list of ordered sequence indices
+        :param partition_epoch: number of partitions seq_index should be split into
+        :param epoch: current epoch
+        :param num_shards: how many shards the data is split into
+        :param shard_index: index of the current data shard
         :return: partition of seq_index for current epoch
-        :rtype: typing.Sequence[int]
         """
+        assert num_shards > shard_index
```
albertz, line 650:
```suggestion
        assert 0 <= shard_index < num_shards
```

returnn/datasets/basic.py
```diff
@@ -639,11 +657,19 @@ def _apply_partition_epoch(cls, seq_index, partition_epoch, epoch):
         assert sum(partition_sizes) == num_seqs and len(partition_sizes) == partition_epoch
         partitions = functools.reduce(lambda a, x: a + [a[-1] + x], partition_sizes, [0])  # cumulative sum
         assert len(partitions) == partition_epoch + 1
-        seq_index = seq_index[partitions[current_partition] : partitions[current_partition + 1]]
-        assert len(seq_index) == partition_sizes[current_partition]
-
+        seq_index = seq_index[shard_index:][::num_shards][
```
albertz, line 660: Why do you do the logic twice, and also in a different way?

Just do the logic once, but instead of `partition_epoch`, use `partition_epoch * num_shards`. And then the index is `current_partition * num_shards + shard_index`.

There are almost no code changes here in this function (except maybe rename some of the vars).

returnn/datasets/basic.py
```diff
@@ -1410,12 +1439,14 @@ def init_dataset(
     default_kwargs: Optional[Dict[str, Any]] = None,
     *,
     parent_dataset: Optional[Dataset] = None,
+    forward_sharding_config: bool = False,
```
albertz, line 1442: I don't understand this. Why is such an argument needed? It should just always be forwarded.

And even if there is a case when this is not wanted, you would not need such an option here. You can simply put `num_shards=0` to the `kwargs` or `extra_kwargs` to explicitly disable it.

returnn/datasets/basic.py
```diff
@@ -1476,7 +1519,9 @@ def init_dataset(
 
 
 def _dataset_extend_default_kwargs_from_parent_dataset(
-    default_kwargs: Optional[Dict[str, Any]], parent_dataset: Optional[Dataset]
+    default_kwargs: Optional[Dict[str, Any]],
+    parent_dataset: Optional[Dataset],
+    forward_sharding_config: bool = False,
```
albertz, line 1524: Same here.

returnn/datasets/basic.py
```diff
@@ -1486,6 +1531,9 @@ def _dataset_extend_default_kwargs_from_parent_dataset(
         return default_kwargs
     default_kwargs = default_kwargs.copy() if default_kwargs else {}
     default_kwargs.setdefault("random_seed_offset", parent_dataset.random_seed_offset)
+    if forward_sharding_config:
+        default_kwargs["num_data_shards"] = parent_dataset.num_data_shards
+        default_kwargs["data_shard_index"] = parent_dataset.data_shard_index
```
albertz, line 1536: Why do you have that inconsistent? In the other code, I was using `setdefault`.

returnn/datasets/distrib_files.py
```diff
@@ -283,11 +283,11 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
             else:
                 raise ValueError(f"{self}: seq_ordering {self.seq_ordering!r} not supported")
             file_bins = self._distribute_evenly_by_size(
-                num_bins=self._num_shards * self.partition_epoch,
+                num_bins=self._num_file_shards * self.partition_epoch,
```
albertz, line 286: Maybe also?
```suggestion
                num_bins=self._num_file_shards * self.num_data_shards * self.partition_epoch,
```

returnn/datasets/multi_proc.py
```diff
@@ -254,23 +278,31 @@ def _get(seq_idx: int) -> Optional[DatasetSeq]:
...
+                            parent_conn.send(("num_seqs", len(seq_order)))
+                            # Now reset seq order for ourself (as the role of a normal worker).
+                            kwargs["seq_order"] = seq_order[0 :: len(other_worker_conns) + 1]
+                            kwargs.pop("seq_list", None)
+                            dataset.init_seq_order(**kwargs)
+                        else:
+                            assert seq_order_conn is not None
+                            msg_, seq_order = seq_order_conn.recv()
+                            assert msg_ == "seq_order_shard"
+                            dataset.init_seq_order(seq_order=seq_order, **kwargs)
```
albertz, line 305: There is an `else` branch missing here.

returnn/datasets/sprint.py
NeoLegends: This is the `SprintCacheDataset`, which reads sprint cache files w/ RETURNN's own implementation, I think that can do it. At least it uses `get_seq_order_for_epoch`.

returnn/datasets/basic.py
NeoLegends: If you split the data like you said you get consecutive seq IDs for the different sub workers. This is ok for the workers themselves, but makes it difficult (or does it?) in MultiProcDataset to look up which worker to query for which seq, because `seq_idx % num_workers` no longer gives the correct worker index.

I think `seq_idx // num_seqs_per_worker` would then give the correct worker index again, but if you don't know `num_seqs_per_worker` you're out of luck. Moreover, if you iterate the through the segments front to back, `worker_idx = seq_idx // num_seqs_per_worker` means you first query worker 1 for all the segments it has, then worker 2, and so on. This means you lose the benefit from having more than one worker in the first place, because the workers do not get any time to fill their buffers in the background while other workers are queried for segments.

Currently MultiProcDS distributes the `seq_order` shards therefore also in a strided way between the workers and I think it should stay that way. Or am I wrong here somewhere?

returnn/datasets/multi_proc.py
NeoLegends: 🤔

returnn/datasets/sprint.py
albertz: Ah sorry, I did not see that in the diff.

returnn/datasets/meta.py
albertz: Ah right.

returnn/datasets/basic.py
albertz: > They cannot have a leading underscore, because the `Dataset.reduce` logic will treat them special if they do.

I don't exactly understand this argument. We can just fix `Dataset.reduce` if there is some problem? But also, do you mean `Dataset.__reduce__`? And how does it treat them special?

`_dataset_extend_default_kwargs_from_parent_dataset` can surely read internal things from the `Dataset`, that's not really a problem. (We should not blindly follow the rules by the linting, but think about what makes sense.)

In any case, they really should also be private attributes (I forgot to comment that), not public.

tests/test_MultiProcDataset.py
```diff
@@ -64,6 +64,22 @@ def test_MultiProcDataset_n3_b5_shuffle():
...
+    hdf_dataset_seqs = dummy_iter_dataset(hdf_dataset)
+
+    with timeout():
+        mp_dataset = MultiProcDataset(
+            dataset=hdf_dataset_dict, num_workers=3, buffer_size=5, sharding_method="dedicated"
+        )
+        mp_dataset.initialize()
+        mp_dataset_seqs = dummy_iter_dataset(mp_dataset)
+
+        compare_dataset_seqs(hdf_dataset_seqs, mp_dataset_seqs)
```
albertz, line 80: This test is wrong. `compare_dataset_seqs` assumes same order. But with `MultiProcDataset` and `sharding_method="dedicated"`, you will not have the same order.

returnn/datasets/basic.py
albertz: I still don't exactly understand where you see the problem. The test is just wrong. But you added the test?

NeoLegends: Oh, then ok. Yeah I copied and adapted an existing test case, and I assumed we should leave the order as-is.

returnn/datasets/multi_proc.py
```diff
@@ -254,23 +287,33 @@ def _get(seq_idx: int) -> Optional[DatasetSeq]:
...
+                            num_seqs = dataset.num_seqs
+                        except NotImplementedError:
+                            num_seqs = None
+                        parent_conn.send(("num_seqs", num_seqs))
+                    elif sharding_method == "seq_order":
+                        if worker_index == 0:
+                            # We are responsible to get the seq order and distrib it to all the other workers.
+                            assert other_worker_conns is not None
+                            dataset.init_seq_order(**kwargs)
+                            seq_order = dataset.get_current_seq_order()
```
NeoLegends, line 302: WDYT: should we add a log line here if this fails w/ `NotImplementedError` to direct the user towards the other sharding option, which may be better supported?

albertz: You can maybe wrap this in a `try`/`except Exception as exc` and throw some new `Exception(f"{self}: sub dataset {dataset} get_current_seq_order failed. Maybe try sharding_method='dedicated'") from exc` or so.

returnn/datasets/basic.py
```diff
@@ -1304,6 +1330,22 @@ def index_shape_for_batches(cls, batches, data_key="data"):
             shape = [max(shape[0], batch.max_num_frames_per_slice[data_key]), shape[1] + batch.num_slices]
         return tuple(shape)
 
+    @classmethod
+    def extend_default_kwargs_from_parent_dataset(
```
albertz, line 1334: This should be private, as this is only intended for internal purpose.

But also, I think I prefer to not have this as a member function. We already have too much here.

returnn/datasets/multi_proc.py
```diff
@@ -308,14 +379,10 @@ def init_seq_order(self, epoch=None, seq_list=None, seq_order=None):
             msg, num_seqs = self._seq_order_proc_parent_conn.recv()
             assert msg == "num_seqs"
             self._num_seqs = num_seqs
-        else:
```
albertz, line 311: This `else` is missing now.

returnn/datasets/distrib_files.py
```diff
@@ -283,13 +287,15 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
...
                 file_sizes=self._file_sizes,
                 files_order=files_order_flat,
             )
-            self_index_base = self.partition_epoch * self._shard_index
-            self_index_end = self_index_base + self.partition_epoch
-            self._files_order_cache[full_epoch_0idx_] = file_bins[self_index_base:self_index_end]
+            self_index_base = self._num_data_shards * self.partition_epoch * self._file_shard_index
+            self_index_end = self_index_base + (self.partition_epoch * self._num_data_shards)
+            self._files_order_cache[full_epoch_0idx_] = file_bins[self_index_base:self_index_end][
+                self._data_shard_index :: self._num_data_shards
```
albertz, line 297: Here do again do this inconsistently / more complex than it needs to be. Same case as in `_apply_partition_epoch`, we can just reuse the existing logic.

albertz (changes requested):

returnn/datasets/basic.py
```diff
@@ -1424,7 +1450,10 @@ def init_dataset(
         return data
     if callable(kwargs):
         return init_dataset(
-            kwargs(), extra_kwargs=extra_kwargs, default_kwargs=default_kwargs, parent_dataset=parent_dataset
+            kwargs(),
+            extra_kwargs=extra_kwargs,
+            default_kwargs=default_kwargs,
+            parent_dataset=parent_dataset,
```
albertz, line 1456: Note: It makes it a bit unnecessary extra effort to look through the diff when there are such cosmetic changes in a bigger PR. Everytime I review again, I look through that again, only to notice that there is actually no change here. I also wonder if this looks really nicer now.

returnn/datasets/basic.py
```diff
@@ -1435,12 +1464,15 @@ def init_dataset(
...
+                data,
+                extra_kwargs=extra_kwargs,
+                default_kwargs=default_kwargs,
+                parent_dataset=parent_dataset,
             )
         else:
             config_str = kwargs
             kwargs = {}
-            default_kwargs = _dataset_extend_default_kwargs_from_parent_dataset(default_kwargs, parent_dataset)
+            default_kwargs = _extend_default_kwargs_from_parent_dataset(default_kwargs, parent_dataset)
```
albertz, line 1475: Same here, now this function was renamed. Was this really necessary?

returnn/datasets/basic.py
NeoLegends: Oh, yeah. I should probably keep the diff in mind myself. These changes happen over time. I think in this case it was because it had an additional parameter at first, forcing the breakup into multiple lines, and after the parameter was removed again, the trailing comma remained, forcing the lines still to be broken. I'll remove those.

returnn/datasets/basic.py
```diff
@@ -1486,6 +1518,8 @@ def _dataset_extend_default_kwargs_from_parent_dataset(
         return default_kwargs
     default_kwargs = default_kwargs.copy() if default_kwargs else {}
     default_kwargs.setdefault("random_seed_offset", parent_dataset.random_seed_offset)
+    default_kwargs.setdefault("_num_data_shards", parent_dataset._num_data_shards)  # noqa
+    default_kwargs.setdefault("_data_shard_index", parent_dataset._data_shard_index)  # noqa
```
albertz, line 1522: I don't like the `noqa` so much. I prefer to disable only the specific check:

```suggestion
    # noinspection PyProtectedMember
    default_kwargs.setdefault("_num_data_shards", parent_dataset._num_data_shards)
    # noinspection PyProtectedMember
    default_kwargs.setdefault("_data_shard_index", parent_dataset._data_shard_index)
```

returnn/datasets/distrib_files.py
```diff
@@ -283,11 +287,14 @@ def init_seq_order(self, epoch: Optional[int] = None, seq_list=None, seq_order=N
             else:
                 raise ValueError(f"{self}: seq_ordering {self.seq_ordering!r} not supported")
             file_bins = self._distribute_evenly_by_size(
-                num_bins=self._num_shards * self.partition_epoch,
+                num_bins=self._num_data_shards * self._num_file_shards * self.partition_epoch,
```
albertz, line 290: I wonder a bit about this. Does it make sense actually to have these two types of sharding? Isn't that the same? I think we should just reuse the existing sharding. I.e. undo all the changes in `DistribFilesDataset`, just rename `_num_shards` to `_num_data_shards`. Or why do you actually call it `_num_data_shards` at all? I initially suggested to just call it `_num_shards`. I think that's a better name. Then `DistribFilesDataset` does not need any changes.

albertz (changes requested):

returnn/datasets/distrib_files.py
NeoLegends: It is a better name, I made the distinction to have this conflict-free wrt. DistribFilesDataset's own implementation of sharding. But yeah maybe this should just be unified, in the same way MultiProc sets the shards for the wrapped ones DistribFilesDataset can fetch them from the distributed context.

albertz: I don't see the point of this distinction. I think specifically we don't want that these are separate types of shardings. These are the same type of sharding. Maybe we can actually generalize it a bit, and let this be an option in RETURNN, similar to `horovod_dataset_distribution`, e.g. just `distrib_dataset_sharding = True` or so, and if that is enabled, it would pass rank/size as the sharding opts for the top level train dataset.

NeoLegends: I have an implementation for this generalistic approach now, let's discuss it in a follow-up PR. Can you review this PR as-is again to see if we can merge it now? :)

returnn/datasets/distrib_files.py
```diff
@@ -151,8 +153,27 @@ def __init__(
             distributed training scenaria
         :param _meta_info_cache: for internal use
         :param _distrib_info: for internal use
+        :param _num_shards: how many data shards there are
+        _param _shard_index: the index of the local shard
         """
-        super().__init__(**kwargs)
+        self.distrib_shard_files = distrib_shard_files
+        if distrib_shard_files:
+            assert _num_shards is None and _shard_index is None, (
```
albertz, line 161: I think this check is wrong. When this is already some sub dataset (e.g. from DistribFilesDataset), this would not be None but 1/0.

Maybe move this code after the `super` and just check whether this is different from 1/0.

returnn/datasets/distrib_files.py
```diff
@@ -151,8 +153,27 @@ def __init__(
...
+                # See also Dataset.__reduce__.
+                shard_index = _distrib_info["_shard_index"]
+                num_shards = _distrib_info["_num_shards"]
+            else:
+                shard_index, num_shards = _get_rank_and_size()
+        else:
+            num_shards = _num_shards or 1
+            shard_index = _shard_index or 0
+        super().__init__(**kwargs, _num_shards=num_shards, _shard_index=shard_index)
+        assert not distrib_shard_files or (_num_shards == 1 and _shard_index == 0), (
```
albertz, line 173: The check is still wrong. You need to check `self._num_shards`. But actually I meant to move the whole code block, so there is nothing above `super().__init__()`.

albertz (approved):

albertz:
Sorry for the delay, I was distracted with other things. I looked at it again and it all looks good from what I can see here in the diff. I think it makes sense to merge now and improve/extend/fix then in follow-up commits/PRs if needed.

PR #1632, opened 2024-09-25
`LaplaceOrdering`: avoid spiky CPU utilization

### LINKED ISSUE #1631: `LaplaceOrdering` interacts badly w/ MultiProcDataset
I'm observing the following behavior in training, where I use the postprocessing dataset for speed perturbation and segment concatenation, followed by laplace ordering:

- The MP worker tries to buffer 10 segments into its internal small buffer
- The laplace iterator kicks in, fills its huge buffer w/ 1000 segments, and sorts them by length
- The MP worker then gradually pulls from that buffer, leading to almost no work being done in the subprocesses in the mean time
- Eventually the large buffer is empty, and the laplace ordering iterator attempts to fill it again
- This takes a while and maxes out the CPU on the worker processes
- In particularly, this blocks the training until the laplace buffer is full again

It would be better if the subprocesses would gradually chew through their work to avoid blocking the training. This requires a change to the laplace iterator: it needs to gradually fill the next buffer while depleting/yielding the current one. This means the work of the sub iterators can overlap w/ the work of the trainer process, leading to more efficient training.

NeoLegends:
Closes #1631

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -323,7 +323,7 @@ class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
                     has_ended = True
 
             if len(seq_buffer) < self.num_seqs_per_bin:
-                assert has_ended
+                assert has_ended and not next_seq_buffer
                 break
 
             is_down_phase = not is_down_phase
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -304,15 +304,30 @@ class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
         iterator = iter(iterator)
         is_down_phase = False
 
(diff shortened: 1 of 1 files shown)


NeoLegends:
I've ran a test, the training is no longer being blocked by spiky CPU usage, it's much more even now.

returnn/datasets/postprocessing.py
```diff
@@ -304,16 +304,31 @@ def __call__(self, iterator: Iterator[TensorDict], **kwargs) -> Iterator[TensorD
...
+                yield item
+
+                try:
+                    if not has_ended:
+                        next_seq_buffer.append(next(iterator))
+                except StopIteration:
+                    has_ended = True
 
             if len(seq_buffer) < self.num_seqs_per_bin:
+                assert has_ended
```
albertz, line 326:
```suggestion
                assert has_ended and not next_seq_buffer
```

albertz (approved):

Issue #1643, opened 2024-11-09
FileCache assertion on previous copy attempt age triggered


NeoLegends:
```
Waiting for lock-file: /ssd/[...]/returnn/file_cache/nas/data/[...]/data[...].lock
Waiting for lock-file: /ssd/[...]/returnn/file_cache/nas/data/[...]/data[...].lock
Waiting for lock-file: /ssd/[...]/returnn/file_cache/nas/data/[...]/data[...].lock
Removing old lockfile '/ssd/[...]/returnn/file_cache/nas/data/[...]/data[...].lock' (probably crashed proc).

[...]

  File "/nas/models/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.SdBYnl2SGErb/output/returnn/returnn/util/file_cache.py", line 294, in FileCache.handle_cached_files_in_config.<locals>._handle_value
    line: res = self.get_file(value.filename)
    locals:
      res = <not found>
      self = <local> <returnn.util.file_cache.FileCache object at 0x7efe22723b80>
      self.get_file = <local> <bound method FileCache.get_file of <returnn.util.file_cache.FileCache object at 0x7efe22723b80>>
      value = <local> CachedFile(filename='/nas/data/[...]/data[...].hdf')
      value.filename = <local> '/nas/data/[...]/data[...].hdf', len = 100
  File "/nas/models/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.SdBYnl2SGErb/output/returnn/returnn/util/file_cache.py", line 120, in FileCache.get_file
    line: self._copy_file_if_needed(src_filename, dst_filename)
    locals:
      self = <local> <returnn.util.file_cache.FileCache object at 0x7efe22723b80>
      self._copy_file_if_needed = <local> <bound method FileCache._copy_file_if_needed of <returnn.util.file_cache.FileCache object at 0x7efe22723b80>>
      src_filename = <local> '/nas/data/[...]/data[...].hdf', len = 100
      dst_filename = <local> '/ssd/[...]/returnn/file_cache/nas/data/[...]/data[...].hdf', len = 138
  File "/nas/models/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.SdBYnl2SGErb/output/returnn/returnn/util/file_cache.py", line 346, in FileCache._copy_file_if_needed
    line: assert dst_tmp_file_age > self._lock_timeout * 0.8, (
[...]

PR #1644, opened 2024-11-09
FileCache: update mtime of lockfile immediately after acquiring it

NeoLegends:
After this change there are no more intermediate FS operations after we acquire the lock file and before we start updating its mtime. If the FS is slow this could help the lockfile stay non-stale.


albertz:
Ah ok. But did you really ran into this?

Is #1643 related or independent of this?

albertz (approved):

NeoLegends:
Yes, somebody ran into the issue. #1643 is the issue. I'm not sure if this PR really resolves it, I kind of doubt it tbh.

PR #1646, opened 2024-11-13
`MPDataset`: make compatible with being wrapped in `PPDataset`

NeoLegends:
This PR makes wrapping a MultiProcDataset into a PPDataset possible

Previously the least you'd get is this crash:

```
EXCEPTION
  File "/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.L2lZBnhLY2SW/output/returnn/rnn.py", line 11, in <module>
    line: main()
    locals:
      main = <local> <function main at 0x7f8117bb7010>
  File "/[...]/recipe/returnn/returnn/__main__.py", line 740, in main
    line: init(command_line_options=argv[1:])
    locals:
      init = <global> <function init at 0x7f8117bb6d40>
      command_line_options = <not found>
      argv = <local> ['/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.L2lZBnhLY2SW/output/returnn/rnn.py', '/[...]/work/i6_core/returnn/training/ReturnnTrainingJob.v52PWHM5t44W/output/returnn.config'], _[0]: {len = 129}
  File "/[...]/recipe/returnn/returnn/__main__.py", line 486, in init
    line: init_data()
    locals:
      init_data = <global> <function init_data at 0x7f8117bb69e0>
  File "/[...]/recipe/returnn/returnn/__main__.py", line 218, in init_data
    line: train_data, extra_train = load_data(config, train_cache_bytes, "train")
    locals:
      train_data = <global> None
      extra_train = <not found>
      load_data = <global> <function load_data at 0x7f8117bb6950>
      config = <global> <returnn.config.Config object at 0x7f8117b8bf10>
      train_cache_bytes = <local> 1513975971841
  File "/[...]/recipe/returnn/returnn/__main__.py", line 190, in load_data
    line: data = init_dataset(kwargs)
    locals:
      data = <not found>
      init_dataset = <global> <function init_dataset at 0x7f8117b7f640>
      kwargs = <local> {'name': 'train', 'class': 'PostprocessingDataset', 'dataset': {'buffer_size': 100, 'class': 'MultiProcDataset', 'dataset': {'class': 'PostprocessingDataset', 'dataset': {'audio': {'features': 'raw', 'peak_normalization': True, 'pre_process': None, 'preemphasis': None, 'sample_rate': 8000}, 'clas..., len = 6
  File "/[...]/recipe/returnn/returnn/datasets/basic.py", line 1496, in init_dataset
    line: obj = clazz(**kwargs)
    locals:
      obj = <not found>
      clazz = <local> <class 'returnn.datasets.postprocessing.PostprocessingDataset'>
      kwargs = <local> {'name': 'train', 'dataset': {'buffer_size': 100, 'class': 'MultiProcDataset', 'dataset': {'class': 'PostprocessingDataset', 'dataset': {'audio': {'features': 'raw', 'peak_normalization': True, 'pre_process': None, 'preemphasis': None, 'sample_rate': 8000}, 'class': 'OggZipDataset', 'epoch_wise_f...
  File "/[...]/recipe/returnn/returnn/datasets/postprocessing.py", line 138, in PostprocessingDataset.__init__
    line: {name: self._make_tensor_template_from_input(name) for name in self._dataset.get_data_keys()}
    locals:
      name = <not found>
      self = <local> <PostprocessingDataset 'train' epoch=None>
      self._make_tensor_template_from_input = <local> <bound method PostprocessingDataset._make_tensor_template_from_input of <PostprocessingDataset 'train' epoch=None>>
      self._dataset = <local> <MultiProcDataset 'dataset_id140187802594336' epoch=None>
      self._dataset.get_data_keys = <local> <bound method CachedDataset2.get_data_keys of <MultiProcDataset 'dataset_id140187802594336' epoch=None>>
  File "/[...]/recipe/returnn/returnn/datasets/cached2.py", line 218, in CachedDataset2.get_data_keys
    line: self._load_something()
    locals:
      self = <local> <MultiProcDataset 'dataset_id140187802594336' epoch=None>
      self._load_something = <local> <bound method CachedDataset2._load_something of <MultiProcDataset 'dataset_id140187802594336' epoch=None>>
  File "/[...]/recipe/returnn/returnn/datasets/cached2.py", line 172, in CachedDataset2._load_something
    line: assert self.added_data, f"{self}: _load_something: could not load anything"
    locals:
      self = <local> <MultiProcDataset 'dataset_id140187802594336' epoch=None>
      self.added_data = <local> []
AssertionError: <MultiProcDataset 'dataset_id140187802594336' epoch=None>: _load_something: could not load anything
```

### DIFF
--- a/returnn/datasets/multi_proc.py
+++ b/returnn/datasets/multi_proc.py
@@ -461,7 +461,7 @@ class MultiProcDataset(CachedDataset2):
         """data keys"""
         if self._data_keys is None:
             self._lazy_init()
-            assert self._data_keys  is not None
+            assert self._data_keys is not None
         return self._data_keys
 
     def get_data_dtype(self, key: str):
--- a/returnn/datasets/multi_proc.py
+++ b/returnn/datasets/multi_proc.py
@@ -66,6 +66,7 @@ class MultiProcDataset(CachedDataset2):
             )
         self._sharding_method = sharding_method
         self._data_keys = None
+        self._data_dtypes: Optional[Dict[str, str]] = None
         self._num_seqs = None
         self._total_num_seqs = None
         self._all_tags = None
@@ -83,6 +84,8 @@ class MultiProcDataset(CachedDataset2):
             self.num_outputs = _meta_info_cache["num_outputs"]
             self._total_num_seqs = _meta_info_cache["total_num_seqs"]
             self.labels = _meta_info_cache["labels"]
+            self._data_keys = _meta_info_cache["data_keys"]
+            self._data_dtypes = _meta_info_cache["data_dtypes"]
 
     def initialize(self):
         """init"""
@@ -100,6 +103,8 @@ class MultiProcDataset(CachedDataset2):
             "labels": self.labels,
             "total_num_seqs": self._total_num_seqs,
             "all_tags": self._all_tags,
+            "data_keys": self._data_keys,
+            "data_dtypes": self._data_dtypes,
         }
 
     def _lazy_init(self):
@@ -176,6 +181,10 @@ class MultiProcDataset(CachedDataset2):
         assert msg == "num_outputs"
         msg, self.labels = self._seq_order_proc_parent_conn.recv()
         assert msg == "labels"
+        msg, self._data_keys = self._seq_order_proc_parent_conn.recv()
+        assert msg == "data_keys"
+        msg, self._data_dtypes = self._seq_order_proc_parent_conn.recv()
+        assert msg == "data_dtypes"
 
     def __del__(self):
         if self._worker_procs:
@@ -282,6 +291,9 @@ class MultiProcDataset(CachedDataset2):
                     parent_conn.send(("num_inputs", dataset.num_inputs))
                     parent_conn.send(("num_outputs", dataset.num_outputs))
                     parent_conn.send(("labels", dataset.labels))
+                    data_keys = dataset.get_data_keys()
+                    parent_conn.send(("data_keys", data_keys))
+                    parent_conn.send(("data_dtypes", {k: dataset.get_data_dtype(k) for k in data_keys}))
                 elif msg == "get_total_num_seqs":
                     assert dataset
                     try:
@@ -444,3 +456,24 @@ class MultiProcDataset(CachedDataset2):
         super().finish_epoch(free_resources=free_resources)
         for worker_parent_conn in self._worker_parent_conns:
             worker_parent_conn.send(("finish_epoch", {"free_resources": free_resources}))
+
+    def get_data_keys(self) -> List[str]:
+        """data keys"""
+        if self._data_keys is None:
+            self._lazy_init()
+            assert self._data_keys  is not None
+        return self._data_keys
+
+    def get_data_dtype(self, key: str):
+        """:return: dtype of `key`"""
+        if self._data_dtypes is None:
+            self._lazy_init()
+            assert self._data_dtype is not None
+        return self._data_dtypes[key]
+
+    def is_data_sparse(self, key: str):
+        """:return: whether `key` is sparse"""
+        if self.num_outputs is None:
+            self._lazy_init()
+            assert self.num_outputs is not None
+        return super().is_data_sparse(key)
--- a/returnn/datasets/multi_proc.py
+++ b/returnn/datasets/multi_proc.py
@@ -181,6 +181,15 @@ class MultiProcDataset(CachedDataset2):
         assert msg == "num_outputs"
         msg, self.labels = self._seq_order_proc_parent_conn.recv()
         assert msg == "labels"
+
+    def _lazy_init_data_keys_and_dtypes(self):
+        if self._data_keys is not None:
+            assert self._data_dtypes is not None
+            return
+
+        self._lazy_init()  # indempotent, will not run twice if not needed
+
+        self._seq_order_proc_parent_conn.send(("get_data_keys_and_dtypes", {}))
         msg, self._data_keys = self._seq_order_proc_parent_conn.recv()
         assert msg == "data_keys"
         msg, self._data_dtypes = self._seq_order_proc_parent_conn.recv()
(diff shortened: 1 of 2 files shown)


returnn/datasets/postprocessing.py
```diff
@@ -155,7 +155,7 @@ def __init__(
         for k, t in self._out_tensor_dict_template.data.items():
             if t.vocab:
                 self.labels[k] = t.vocab.labels
-            elif t.sparse_dim:  # sparse_dim but not vocab
+            elif t.sparse_dim and t.sparse_dim.dimension is not None:  # sparse_dim but not vocab
```
albertz, line 158: I don't understand this. You would never have a sparse_dim with `sparse_dim.dimension == None`.

returnn/datasets/multi_proc.py
```diff
@@ -444,3 +456,24 @@ def finish_epoch(self, *, free_resources: bool = False):
...
             worker_parent_conn.send(("finish_epoch", {"free_resources": free_resources}))
+
+    def get_data_keys(self) -> List[str]:
+        """data keys"""
+        if self._data_keys is None:
+            self._lazy_init()
+            assert self._data_keys is not None
+        return self._data_keys
+
+    def get_data_dtype(self, key: str):
```
albertz, line 467:
```suggestion
    def get_data_dtype(self, key: str) -> str:
```

returnn/datasets/multi_proc.py
```diff
@@ -444,3 +456,24 @@ def finish_epoch(self, *, free_resources: bool = False):
...
+        """:return: dtype of `key`"""
+        if self._data_dtypes is None:
+            self._lazy_init()
+            assert self._data_dtype is not None
+        return self._data_dtypes[key]
+
+    def is_data_sparse(self, key: str):
```
albertz, line 474:
```suggestion
    def is_data_sparse(self, key: str) -> bool:
```

returnn/datasets/postprocessing.py
NeoLegends: Then the dim construction in the PP Dataset from the underlying dataset's data info has a bug. I ran into this as error on a dim by the dataset as part of `_out_tensor_dict_template`. Let me see if I can improve things...

returnn/datasets/multi_proc.py
```diff
@@ -282,6 +291,9 @@ def _get(seq_idx: int) -> Optional[DatasetSeq]:
                     parent_conn.send(("num_inputs", dataset.num_inputs))
                     parent_conn.send(("num_outputs", dataset.num_outputs))
                     parent_conn.send(("labels", dataset.labels))
+                    data_keys = dataset.get_data_keys()
+                    parent_conn.send(("data_keys", data_keys))
+                    parent_conn.send(("data_dtypes", {k: dataset.get_data_dtype(k) for k in data_keys}))
```
albertz, line 296: Same for `get_data_dtype`.

albertz (changes requested):
See my comments, specifically about `get_data_keys`/`get_data_dtype`. Those must be done lazily.

returnn/datasets/postprocessing.py
albertz: Yes that's possible. How does your `sparse_dim` look like? And where does it come from?

NeoLegends: No wait, this carries raw UTF-8 bytes and has a well-known dimension. This is a bug in PPDataset then.

NeoLegends: Yeah, it was the orth in OggZipDataset. No bug in PPDataset.

returnn/datasets/multi_proc.py
```diff
@@ -298,6 +316,11 @@ def _get(seq_idx: int) -> Optional[DatasetSeq]:
                     except NotImplementedError as exc:
                         all_tags = NotImplementedError(f"{exc} in {dataset}")
                     parent_conn.send(("all_tags", all_tags))
+                elif msg == "get_data_keys_and_dtypes":
```
albertz, line 319: Why do you have made this a combined getter? This looks a bit weird to me.

albertz: Also, I don't really see any advantage of this, but then, there are several disadvantages:
* it is inconsistent to the remaining code
* there could be cases where you only need one of them but not both, and the other one would add maybe some non-trivial overhead. e.g. maybe get_data_keys is still efficient for some dataset, but now it will also trigger get_data_dtype which is slow

returnn/datasets/postprocessing.py
albertz: So then undo the change here?

albertz: And also, fix it in `OggZipDataset`?

albertz: Also, again the question: Where does it come from? We should never even allow to construct a `sparse_dim` with unknown dim here.

NeoLegends: `OggZipDataset.get_data_shape("orth") -> []` and `OggZipDataset.is_data_sparse("orth") -> false` lead to the only dim being the batch dim created in the line `dims = [Dim(dimension=None, ...)]`:

```python
def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
    dtype = self._dataset.get_data_dtype(data_key)
    if dtype == "string":
        dims = []
    else:
        feature_dims = [
            Dim(dimension=dim, name=f"{data_key}_dim{i + 1}")
            for i, dim in enumerate(self._dataset.get_data_shape(data_key))
        ]
        dims = [Dim(dimension=None, name=f"{data_key}_frame"), *feature_dims]
    sparse_dim = None
    if self._dataset.is_data_sparse(data_key):
        sparse_dim = Dim(dimension=self._dataset.get_data_dim(data_key), name=f"{data_key}_sparse")
        if data_key in self._dataset.labels:
            sparse_dim.vocab = Vocabulary.create_vocab_from_labels(self._dataset.labels[data_key])
```

So we never constructed a sparse Dim with `dimension=None`. The bug in OggZipDataset is fixed in master already.

returnn/datasets/multi_proc.py
NeoLegends: I made this a combined getter because you need to call one to compute the other, and then it made more sense to me to always do both at the same time. But I can also split this up since we usually cache the information returned from the getters, which means calling `get_data_keys` twice should usually be not much slower than calling it once.

albertz (approved):

PR #1650, opened 2024-11-22
MultiEpochDataset: implement get_current_seq_order

dorian-K:


dorian-K:
I just realized this patch is wrong, it only works correctly when multi_epoch=1

albertz (changes requested):
But this is wrong. The sub dataset does not have the same seq order but only a subset. The MultiEpochDataset will go over multiple epochs in the sub dataset.

albertz:
> I just realized this patch is wrong, it only works correctly when multi_epoch=1

Yes exactly.

I don't think there is an easy way (or any way at all in general) to implement `get_current_seq_order`. So I think we can only close this PR without having this implemented.

Btw, why do you need this?

dorian-K:
MetaDataset needs this method when setting a MultiEpochDataset as the `seq_order_control_dataset`.

albertz:
> MetaDataset needs this method when setting a MultiEpochDataset as the `seq_order_control_dataset`.

Yea, don't do that. :)

PR #1653, opened 2024-11-26
Add extra_labels to SimpleHDFWriter

dorian-K:


albertz (changes requested):
Please use black. (See also failing test.)

albertz (approved):

Issue #1506, opened 2024-01-31
Dataset batching like ESPnet support


albertz:
ESPnet basically does it like this:

- Sort the whole dataset. (The dataset could maybe be directly stored in a sorted way. This would speed up the random access later.)
- Creates batches for the whole dataset (or just some meta information on what sequences would come together in a batch). This has only minimal padding now. And also, we know the number of batches in advance.
- Shuffle the order of the batches, i.e. randomly sample from the batches.

It's not clear whether this is really better than what we do, but in any case, it would be good if we could support this scheme as well, just for comparison.

The question is, how? Our shuffling logic currently happens before the batching, not after.

(*TODO* Also check ESPnet code again, to see how exactly they do it, i.e. whether this is a pre-processing step, etc.)

Some options:

- Some offline processing of the dataset, which builds such table for the list of batches.
  - The dataset sequence order would use this table, and shuffle only across batches, i.e. keep seqs of the same batch together.
  - The batching logic would also use this table.
- The dataset sequence ordering has another mode specifically for this. But then this must know exactly about the batching logic/parameters (batch size etc). In the batching afterwards, we probably also should have some sanity checks that the batching is as it was expected by the sequence ordering.
- Maybe some new way for the user to provide custom sequence ordering + batching in a combined way, via some custom user function, where it is all up to the user.

NeoLegends:
1. Does any existing dataset offer custom sorting of the segments based on a dataset key currently?
2. If we assume we can build batches w/ minimal padding from the entire sorted dataset, do you think the online shuffling is "enough" and adds sufficient randomness vs. true (full) batch shuffling?

NeoLegends:
1. This batching style does not seem very scalable to me. What do they do in ESPnet when the dataset grows really large? Maybe w/ offline preprocessing it can be made scalable, but I still struggle to see how this would work w/ `DistributeFilesDataset`. Maybe if you ensure the data that is supposed to go into one batch always stays within one HDF/primitive file that is distributed by the dataset... Still shuffling will be slightly different.
2. Does any existing dataset offer custom sorting of the segments based on a dataset key currently? This `batch_idx` could be created in preprocessing, and if used as sort key the batching can be done efficiently. Having the key does not solve the batch shuffling issue though, if we want true shuffling and not online shuffling... which brings me to the next question:
[...]

PR #1655, opened 2024-11-27
`PPDataset`: implement `BucketOrdering`

NeoLegends:
Let's see how well this works for the models.

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -298,8 +298,8 @@ class PostprocessingDataset(CachedDataset2):
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
-class SeqOrdering:
-    """Base class for defining seq ordering iterators compatible w/ the :class:`PostprocessingDataset`."""
+class _GetSeqLenByKey:
+    """Base class that defines a seq length accessor based on a data key"""
 
     def __init__(self, length_key: str):
         """:param length_key: data key to determine the segment length from for ordering."""
@@ -312,7 +312,7 @@ class SeqOrdering:
         return tdict.data[self._length_key].raw_tensor.shape[0]
 
 
-class BucketOrdering(SeqOrdering, Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
+class BucketOrdering(_GetSeqLenByKey, Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
     """
     Iterator compatible with :class:`PostprocessingDataset`'s ``map_seq_stream`` applying
     an ordering that creates batches consisting of segments falling into a few distinct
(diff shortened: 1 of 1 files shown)


albertz:
Please check failing tests.
```
  returnn/datasets/postprocessing.py:301: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/datasets/postprocessing.py:329: WEAK WARNING Pep8CodeStyle: E741: ambiguous variable name 'l'
  returnn/datasets/postprocessing.py:359: WEAK WARNING PyShadowingBuiltinsInspection: Shadows built-in name 'len'
```

returnn/datasets/postprocessing.py
```diff
@@ -296,7 +298,70 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
         return Tensor(data_key, dims=dims, dtype=dtype, sparse_dim=sparse_dim)
 
 
-class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
+class SeqOrdering:
```
albertz, line 301: I don't see that this is such generic base class that I would call/use it that way. Maybe better:
```suggestion
class _GetSeqLenByLengthKey:
```

albertz:
I don't really understand the purpose of this. This is not at the batch level here but still at the sequence level? So the batching comes after that and is independent? So this does not make sense? Bucketing is usually a way to do the batching. I.e. this is an alternative to our current `BatchingIterDataPipe`. Or we introduce sth like `batch_idx` and follow some logic as discussed in #1506.

returnn/datasets/postprocessing.py
```diff
@@ -296,7 +298,73 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+    Iterator compatible with :class:`PostprocessingDataset`'s ``map_seq_stream`` applying
+    an ordering that creates batches consisting of segments falling into a few distinct
+    length buckets.
+
+    To be composed with any custom data postprocessing logic via :class:`Sequential`.
+
+    See also :class:`LaplaceOrdering`.
+    """
+
+    def __init__(self, seq_lens: Sequence[int], length_key: str = "data"):
```
albertz, line 326: For every max_seq_len, you should define the max_num_seqs. The option here should be more like:
```python
bucket_sizes: List[Tuple[int, int]]
```
representing the (max_seq_len,max_num_seqs) pair.

NeoLegends:
The idea is that this yields the segments so they group up to batches in the desired way later in the pipeline w/o explicitly doing the batching. This is kind of a hack and I agree, this could also be done on the level of `BatchingIterDataPipe` or via `batch_idx`. I think I'll wait until I have results for the models and then we can decide whether to integrate this functionality deeper into the pipeline (which would make sense) or if we drop this PR completely.

NeoLegends:
So maybe let's not merge this yet.

PR #1656, opened 2024-11-27
PT: print padding amount per batch and subepoch

NeoLegends:
I first had an impl based on another data pipe, but that was cumbersome to integrate w/ per-step and per-subepoch logging, so the logic resides in the engine now. It's not too much anyway.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -386,9 +386,11 @@ class Engine(EngineBase):
                     break
 
                 keys_w_seq_len = [k for k in extern_data_raw if f"{k}:seq_len" in extern_data_raw]
-                total_data_size_packed += NumbersDict({k: sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len})
+                total_data_size_packed += NumbersDict(
+                    {k: sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len},
+                )
                 total_data_size_padded += NumbersDict(
-                    {k: util.prod(extern_data_raw[k].shape[:2]) for k in keys_w_seq_len}
+                    {k: util.prod(extern_data_raw[k].shape[:2]) for k in keys_w_seq_len},
                 )
 
                 num_seqs_ = (
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -367,8 +367,8 @@ class Engine(EngineBase):
         num_seqs = None
         last_seq_idx = 0
 
-        ep_data_len = 0
-        ep_data_space = 0
+        ep_data_len = NumbersDict()
+        ep_data_space = NumbersDict()
 
         try:
             while True:
@@ -386,9 +386,9 @@ class Engine(EngineBase):
                     break
 
                 keys_w_seq_len = [k for k in extern_data_raw if f"{k}:seq_len" in extern_data_raw]
-                data_len = sum(torch.sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len)
-                data_space = sum(torch.numel(extern_data_raw[k]) for k in keys_w_seq_len)
-                padding_ratio = 1 - (data_len / data_space)
+                data_len = NumbersDict({k: torch.sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len})
+                data_space = NumbersDict({k: torch.prod(extern_data_raw[k].shape[:2]) for k in keys_w_seq_len })
+                padding_ratio = NumbersDict.constant_like(1.0, data_len) - (data_len / data_space)
                 ep_data_len += data_len
                 ep_data_space += data_space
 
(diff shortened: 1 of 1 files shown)


returnn/torch/engine.py
```diff
@@ -366,6 +366,10 @@ def train_epoch(self):
         extern_data = None
         num_seqs = None
         last_seq_idx = 0
+
+        ep_data_len = 0
+        ep_data_space = 0
```
albertz, line 371: You should better make this a `NumberDict` and keep separate entries per data. At least I also want to know about this. Also, you cannot just sum them up as you do now, as the length might have totally different scales, but anyway different implications on memory usage.

returnn/torch/engine.py
```diff
@@ -380,6 +384,14 @@ def train_epoch(self):
                     torch.distributed.all_reduce(_has_data, op=torch.distributed.ReduceOp.MIN)
                 if not _has_data[0]:
                     break
+
+                keys_w_seq_len = [k for k in extern_data_raw if f"{k}:seq_len" in extern_data_raw]
+                data_len = sum(torch.sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len)
+                data_space = sum(torch.numel(extern_data_raw[k]) for k in keys_w_seq_len)
```
albertz, line 390: Despite the `sum` (over the keys), which you should not do as discussed above, `numel` is also wrong here? That would also count the feature dim(s)? You want to compare sum(seq_len[k]) vs prod(data[k].shape[:2]).

returnn/torch/engine.py
```diff
@@ -1361,6 +1388,8 @@ def _print_process(
             info += ["%s %s" % (k, _format_score_value(v)) for k, v in eval_info.items()]
         if batch_size_info:
             info += ["%s %s" % (k, _format_score_value(v)) for k, v in batch_size_info.items()]
+        if padding_ratio is not None:
+            info += [f"pad {padding_ratio:.1%}"]
```
albertz, line 1392: The problem here per batch is: As discussed above, we should separate this per data key. But then printing this info per batch really makes it too spammy.

I currently don't really know a good solution for this. So I think I would just not print this per batch for now, and only report at the end of the epoch.

albertz:
Please check failing tests. (I think you should use our own `returnn.util.basic.prod`, not `torch.prod`.)

returnn/torch/engine.py
```diff
@@ -366,6 +366,10 @@ def train_epoch(self):
         extern_data = None
         num_seqs = None
         last_seq_idx = 0
+
+        ep_data_len = NumbersDict()
+        ep_data_space = NumbersDict()
```
albertz, line 371: The var names are not good. I don't understand the meaning of "len" or "space". Also, the `ep_` prefix is inconsistent to the other var names here (e.g. `num_seqs`).
```suggestion
        total_data_size_packed = NumbersDict()
        total_data_size_padded = NumbersDict()
```

returnn/torch/engine.py
```diff
@@ -380,6 +384,13 @@ def train_epoch(self):
                     torch.distributed.all_reduce(_has_data, op=torch.distributed.ReduceOp.MIN)
                 if not _has_data[0]:
                     break
+
+                keys_w_seq_len = [k for k in extern_data_raw if f"{k}:seq_len" in extern_data_raw]
+                data_len = NumbersDict({k: torch.sum(extern_data_raw[f"{k}:seq_len"]).item() for k in keys_w_seq_len})
+                data_space = NumbersDict({k: torch.prod(extern_data_raw[k].shape[:2]).item() for k in keys_w_seq_len})
```
albertz, line 390:
```suggestion
                data_len = NumbersDict({k: sum(extern_data_raw[f"{k}:seq_len"]) for k in keys_w_seq_len})
                data_space = NumbersDict({k: prod(extern_data_raw[k].shape[:2]) for k in keys_w_seq_len})
```
For `prod`, do `from returnn.util.basic import prod`.

returnn/torch/engine.py
```diff
@@ -466,6 +477,10 @@ def train_epoch(self):
                 accumulated_losses_dict += losses_dict
                 accumulated_inv_norm_factors_dict += inv_norm_factors_dict
                 eval_info = self._maybe_extend_losses_info(losses_dict / inv_norm_factors_dict)
+                if self.config.bool("log_stepwise_padding_ratio", False):
```
albertz, line 480: Do you really want to use this? I will not use this. So then I would just remove this.

NeoLegends: Yes, I am going to use this for the bucketing experiments.

NeoLegends: Yeah I can also keep the changes local to my RETURNN copy.

returnn/torch/engine.py
```diff
@@ -366,6 +366,10 @@ def train_epoch(self):
         extern_data = None
         num_seqs = None
         last_seq_idx = 0
+
+        ep_data_size_packed = NumbersDict()
+        ep_data_size_padded = NumbersDict()
```
albertz, line 371:
```suggestion
        total_data_size_packed = NumbersDict()
        toal_data_size_padded = NumbersDict()
```

returnn/torch/engine.py
albertz: You can also do this just inside your `train_step` function.

albertz (approved):

PR #1659, opened 2024-11-29
Allow skipping sequences in forward config option

dorian-K:


albertz (approved):

albertz:
Yes I just wanted to say that.

Regarding the warning in better_exchook, this seems unrelated here. The new caching now might have caused some different IPython version, where you get this warning. I also get the warning locally in my env. So we can ignore this here. The whole related code in better_exchook really seems somewhat outdated. I will try to fix this. But this is all independent of this PR here.

PR #1660, opened 2024-11-29
PT: allow custom batching, add bucket batching

NeoLegends:
Support bucket batching, and other custom batching schemes, via the config option `torch_batching`.

E.g. put this into the config for bucket batching:
```python
torch_batching = {
    "class": "BucketOrderingIterDataPipe",
    "buckets": [(1000, 100), (2000, 50), (4000, 25), (10_000, 10), (100_000, 1)],
    "length_key": "data"
}
```
This will have 5 buckets (specified as (max_seq_len, num_seqs) tuples), using seq length from "data" key.

### DIFF
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -341,7 +341,7 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
         assert buckets, "empty bucket batching configuration"
         if not all(size > 0 and max_seqs > 0 for size, max_seqs in buckets):
             raise ValueError(f"bucket sizes and max seqs in bucket must be positive")
-        self._max_seq_lens, self._max_bucket_sizes = zip(*sorted(buckets, key=lambda b: b[0]))
+        self._max_seq_lens, self._max_bucket_sizes = zip(*sorted(buckets))
         assert len(set(self._max_seq_lens)) == len(self._max_seq_lens), "seq len boundaries must all be unique"
 
     def __iter__(self):
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -326,12 +326,7 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
     """
 
     def __init__(
-        self,
-        dataset: torch.utils.data.IterableDataset,
-        *,
-        buckets: Sequence[Tuple[int, int]],
-        length_key: str,
-        **_kwargs,
+        self, dataset: torch.utils.data.IterableDataset, *, buckets: Sequence[Tuple[int, int]], length_key: str
     ):
         """
         :param dataset: dataset to apply bucket batching to
@@ -339,7 +334,6 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
             Segments longer than the largest size limit configured in the buckets are dropped. To avoid dropping
             any segments make sure your largest bucket allows segments larger than your longest training segment.
         :param length_key: data key to take as length measure
-        :param _kwargs: unused args
         """
         self._dataset = dataset
         self._length_key = length_key
@@ -370,8 +364,8 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
         yield from non_empty_buckets
 
 
-def init_batching(
-    dataset: torch.utils.data.IterableDataset, config: Config, train: bool
+def get_batching_iterable_dataset_from_config(
+    *, dataset: torch.utils.data.IterableDataset, config: Config, train: bool
 ) -> torch.utils.data.IterableDataset:
     """
     Batches the segments in the given dataset given the config settings.
@@ -389,18 +383,30 @@ def init_batching(
         batches_dataset = BatchingIterDataPipe(dataset, batch_size=batch_size, max_seqs=max_seqs)
         return batches_dataset
 
-    assert isinstance(custom_batching, dict)
-    assert "class" in custom_batching
-    custom_batching = custom_batching.copy()
-    clazz = custom_batching.pop("class")
-    if isinstance(clazz, str):
-        batchers = {c.__name__: c for c in [BatchingIterDataPipe, BucketOrderingIterDataPipe]}
-        make_batcher = batchers[clazz]
-    elif isinstance(clazz, type):
-        make_batcher = clazz
+    if isinstance(custom_batching, dict):
+        assert "class" in custom_batching
+        batching_args = custom_batching.copy()
+        type_or_name = batching_args.pop("class")
+        if isinstance(type_or_name, str):
+            cls = globals()[type_or_name]
+        elif isinstance(type_or_name, (type, Callable)):
+            # custom types need to be forward compatible
+            batching_args.update(get_fwd_compat_kwargs())
+            batching_args["train"] = train
+            cls = type_or_name
+        else:
+            raise ValueError(f"Custom batching class key must either be a string naming a type or a class.")
+    elif isinstance(custom_batching, (type, Callable)):  # short form w/o a dict
+        # callables need to be forward compatible
+        batching_args = get_fwd_compat_kwargs()
+        batching_args["train"] = train
+        cls = custom_batching
     else:
-        raise ValueError(f"Custom batching class must either be a string or a type.")
-    batches_dataset = make_batcher(dataset, train=train, **custom_batching)
+        raise ValueError(
+            f"custom_batching must either be a dict containing a `class` key naming a type, a type or a callable."
+        )
+    batches_dataset = cls(dataset, train=train, **batching_args)
+    assert isinstance(batches_dataset, torch.utils.data.IterableDataset)
     return batches_dataset
 
 
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -372,6 +372,13 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
 def init_batching(
     dataset: torch.utils.data.IterableDataset, config: Config, train: bool
 ) -> torch.utils.data.IterableDataset:
+    """
+    Batches the segments in the given dataset given the config settings.
+
+    :param dataset: dataset whose segments to group into batches
+    :param config: RETURNN config
+    :param train: whether we are in training or in inference mode
+    """
     custom_batching = config.typed_value("custom_batching", None)
     if custom_batching is None:
         batch_size = config.typed_value("batch_size", -1)
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -311,6 +311,7 @@ class BatchingIterDataPipe(torch.utils.data.IterDataPipe):
             yield current_batch
 
 
+# noinspection PyAbstractClass
 class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
     """
     Converts a dataset yielding sequences (dict data_key -> array per sequence) into
--- a/returnn/torch/data/pipeline.py
+++ b/returnn/torch/data/pipeline.py
@@ -325,7 +325,12 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
     """
 
     def __init__(
-        self, dataset: torch.utils.data.IterableDataset, *, buckets: Sequence[Tuple[int, int]], length_key: str
+        self,
+        dataset: torch.utils.data.IterableDataset,
+        *,
+        buckets: Sequence[Tuple[int, int]],
+        length_key: str,
+        **kwargs,
     ):
         """
         :param dataset: dataset to apply bucket batching to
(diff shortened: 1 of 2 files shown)


NeoLegends:
The test failure seems unrelated to the changes in this PR. After the fix for the number overflows this works well for me.

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
             yield current_batch
 
 
+# noinspection PyAbstractClass
```
albertz, line 314: Why is this needed? Is this an abstract class or not? I assume it is not, but then why do you get this warning?

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+            buckets[bucket_idx].append(data_dict)
+            if len(buckets[bucket_idx]) >= self._max_bucket_sizes[bucket_idx]:
+                yield buckets[bucket_idx]
+                buckets[bucket_idx] = []
+
+        non_empty_buckets = [b for b in buckets if b]
+        yield from non_empty_buckets
+
+
+def init_batching(
```
albertz, line 373: The name is bad.
```suggestion
def get_batching_iterable_dataset_from_config(
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    dataset: torch.utils.data.IterableDataset, config: Config, train: bool
```
albertz, line 374:
```suggestion
    *, dataset: torch.utils.data.IterableDataset, config: Config, train: bool
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+        batch_size = config.typed_value(f"batch_size_{'train' if train else 'dev'}", batch_size)
+        assert batch_size != -1, f"batch_size or batch_size_{'train' if train else 'dev'} not defined in config"
+        max_seqs = config.typed_value("max_seqs", -1)
+        batches_dataset = BatchingIterDataPipe(dataset, batch_size=batch_size, max_seqs=max_seqs)
+        return batches_dataset
+
+    assert isinstance(custom_batching, dict)
+    assert "class" in custom_batching
+    custom_batching = custom_batching.copy()
+    clazz = custom_batching.pop("class")
```
albertz, line 395: Use more standard var name.
```suggestion
    cls = custom_batching.pop("class")
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    if isinstance(clazz, str):
+        batchers = {c.__name__: c for c in [BatchingIterDataPipe, BucketOrderingIterDataPipe]}
+        make_batcher = batchers[clazz]
```
albertz, line 398:
```suggestion
        cls = globals()[cls]
        assert issubclass(cls, torch.utils.data.IterDataPipe)
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    elif isinstance(clazz, type):
+        make_batcher = clazz
```
albertz, line 400:
```suggestion
        pass
```
(Maybe move the `assert issubclass(cls, torch.utils.data.IterDataPipe)` that I suggested after this `if`.)

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    this happens in 'collate_batch()'.
+    """
+
+    def __init__(
+        self,
+        dataset: torch.utils.data.IterableDataset,
+        *,
+        buckets: Sequence[Tuple[int, int]],
+        length_key: str,
+        **_kwargs,
```
albertz, line 334:
```suggestion
```
You should never just ignore any kwargs.

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    ):
+        """
+        :param dataset: dataset to apply bucket batching to
+        :param buckets: Bucket configuration as tuples of seq length and max number of seqs in that bucket.
+            Segments longer than the largest size limit configured in the buckets are dropped. To avoid dropping
+            any segments make sure your largest bucket allows segments larger than your longest training segment.
+        :param length_key: data key to take as length measure
+        :param _kwargs: unused args
```
albertz, line 342:
```suggestion
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    custom_batching = custom_batching.copy()
+    clazz = custom_batching.pop("class")
+    if isinstance(clazz, str):
+        batchers = {c.__name__: c for c in [BatchingIterDataPipe, BucketOrderingIterDataPipe]}
+        make_batcher = batchers[clazz]
+    elif isinstance(clazz, type):
+        make_batcher = clazz
+    else:
+        raise ValueError(f"Custom batching class must either be a string or a type.")
+    batches_dataset = make_batcher(dataset, train=train, **custom_batching)
```
albertz, line 403:
```suggestion
    batches_dataset = cls(dataset, train=train, **custom_batching)
```

returnn/torch/data/pipeline.py
albertz: > Maybe move the `assert issubclass(cls, torch.utils.data.IterDataPipe)` that I suggested after this `if`.

Maybe we can also just let it be any `callable`, not just subclass of `IterDataPipe`. In case it is some generic function, you can also pass `train` and `**get_fwd_compat_kwargs()` as kwargs to this callable.

And then after it was called, check:
```python
assert isinstance(batches_dataset, torch.utils.data.IterDataPipe)
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,99 @@ def __iter__(self):
...
+    custom_batching = config.typed_value("custom_batching", None)
+    if custom_batching is None:
+        batch_size = config.typed_value("batch_size", -1)
+        batch_size = config.typed_value(f"batch_size_{'train' if train else 'dev'}", batch_size)
+        assert batch_size != -1, f"batch_size or batch_size_{'train' if train else 'dev'} not defined in config"
+        max_seqs = config.typed_value("max_seqs", -1)
+        batches_dataset = BatchingIterDataPipe(dataset, batch_size=batch_size, max_seqs=max_seqs)
+        return batches_dataset
+
+    assert isinstance(custom_batching, dict)
```
albertz, line 392: You could already here make another case for `if callable(custom_batching): ...` and then do what I suggested below.

returnn/torch/engine.py
```diff
@@ -386,12 +386,13 @@ def train_epoch(self):
                 if not _has_data[0]:
                     break
 
+                # convert values from torch int32 to Python ints to prevent overflow
```
albertz, line 389: Oh, good catch. This is not really related to this PR though, right? Maybe move it out and push separately directly to master?

returnn/torch/data/pipeline.py
NeoLegends: I get the warning that this class does not implement all abstract methods, probably because we're not implementing `reset()` (which we don't need to, since we're overriding `__iter__`). The same suppression is applied to the `BatchingIterDataPipe`.

returnn/torch/data/pipeline.py
NeoLegends: `clazz` is also used in `init_dataset`, but ok

returnn/torch/engine.py
NeoLegends: Ah, oh, yes!

returnn/torch/data/pipeline.py
NeoLegends: I have both forms (dict containing class key, where class can be a str, type or callable, and `custom_batching = type or callable`) now, but I'm not sure both are needed. WDYT?

returnn/torch/data/pipeline.py
albertz: What PyTorch version? `reset()` is not an abstract method for me. It seems it's also not needed for `BatchingIterDataPipe`?

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,105 @@ def __iter__(self):
...
+            any segments make sure your largest bucket allows segments larger than your longest training segment.
+        :param length_key: data key to take as length measure
+        """
+        self._dataset = dataset
+        self._length_key = length_key
+
+        assert buckets, "empty bucket batching configuration"
+        if not all(size > 0 and max_seqs > 0 for size, max_seqs in buckets):
+            raise ValueError(f"bucket sizes and max seqs in bucket must be positive")
+        self._max_seq_lens, self._max_bucket_sizes = zip(*sorted(buckets, key=lambda b: b[0]))
```
albertz, line 344:
```suggestion
        self._max_seq_lens, self._max_bucket_sizes = zip(*sorted(buckets))
```

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,105 @@ def __iter__(self):
...
+    if isinstance(custom_batching, dict):
+        assert "class" in custom_batching
+        batching_args = custom_batching.copy()
+        type_or_name = batching_args.pop("class")
+        if isinstance(type_or_name, str):
+            cls = globals()[type_or_name]
+        elif isinstance(type_or_name, (type, Callable)):
+            # custom types need to be forward compatible
+            batching_args.update(get_fwd_compat_kwargs())
+            batching_args["train"] = train
```
albertz, line 395: Remove this. If the user wants the custom callable, it can be used directly without a dict.
```suggestion
```
Even having support for `type` here is maybe also not really necessary, not sure.

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,105 @@ def __iter__(self):
...
+    elif isinstance(custom_batching, (type, Callable)):  # short form w/o a dict
+        # callables need to be forward compatible
+        batching_args = get_fwd_compat_kwargs()
+        batching_args["train"] = train
+        cls = custom_batching
+    else:
+        raise ValueError(
+            f"custom_batching must either be a dict containing a `class` key naming a type, a type or a callable."
+        )
+    batches_dataset = cls(dataset, train=train, **batching_args)
```
albertz, line 408:
```suggestion
    batches_dataset = cls(dataset, **batching_args)
```

returnn/torch/data/pipeline.py
NeoLegends: The one in CI that is used when the PyCharm inspections run.

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,105 @@ def __iter__(self):
...
+    *, dataset: torch.utils.data.IterableDataset, config: Config, train: bool
+) -> torch.utils.data.IterableDataset:
+    """
+    Batches the segments in the given dataset given the config settings.
+
+    :param dataset: dataset whose segments to group into batches
+    :param config: RETURNN config
+    :param train: whether we are in training or in inference mode
+    """
+    custom_batching = config.typed_value("custom_batching", None)
```
albertz, line 377: I thought a bit about the config option name. I don't really like the `custom_` prefix. This is not more custom than anything else really. E.g. this is just like `optimizer`. We also don't call this `custom_optimizer`. It's up to the user how "custom" this is.

So, just `"batching"` makes more sense. However, for `"batching"`, we have the problem that this conflicts with the already existing `"batching"` option, used by `Dataset`:
```python
        set_or_remove("seq_ordering", config.value("batching", None))
```
E.g. when the user has `batching = "laplace:..."` in the config.
This is config option is actually quite related.
But I don't think we can reuse this option here. This would be too confusing (just to figure out where this config option is actually used...).

So, how else could this be called? `batching_method`?

returnn/torch/data/pipeline.py
albertz: Ok... I don't like this. But let's clean this up in a followup. First check that we maybe can increase the PyTorch version in CI.

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,98 @@ def __iter__(self):
...
+        batches_dataset = BatchingIterDataPipe(dataset, batch_size=batch_size, max_seqs=max_seqs)
+        return batches_dataset
+
+    if isinstance(custom_batching, dict):
+        assert "class" in custom_batching
+        batching_args = custom_batching.copy()
+        type_name = batching_args.pop("class")
+        assert isinstance(type_name, str)
+        cls = globals()[type_name]
+    elif isinstance(custom_batching, (type, Callable)):
```
albertz, line 392:
```suggestion
    elif callable(custom_batching):
```

returnn/torch/data/pipeline.py
albertz: Ah, also, this here is now PyTorch specific. So maybe call this `torch_batching`?

returnn/torch/data/pipeline.py
```diff
@@ -309,6 +311,102 @@ def __iter__(self):
...
+            buckets[bucket_idx].append(data_dict)
+            if len(buckets[bucket_idx]) >= self._max_bucket_sizes[bucket_idx]:
+                yield buckets[bucket_idx]
+                buckets[bucket_idx] = []
+
+        non_empty_buckets = [b for b in buckets if b]
+        yield from non_empty_buckets
+
+    def __getitem__(self, index):
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
albertz: (Also `black`.)

albertz (approved):

PR #1661, opened 2024-12-05
PT: add uniform likelihood bucket batching

NeoLegends:
This is a first attempt at automatically optimizing the bucket limits during training. After every subepoch the limits are adjusted to make every bucket catch a roughly equal number of segments.

I am debating on two more things:

1. Keeping only a fixed number of seq lens that have passed through the batching (e.g. 10k + upper + lower value) to have an upper bound on the memory consumption. Whenever the list is extended by another batch and crosses said limit, we re-sample randomly from that list to keep it at a fixed size. As long as the number of seq lens we keep is large enough this shouldn't have a big influence on the accuracy of the results.
2. Keeping seq len statistics even across subepochs to make the statistics better.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
       with:
-        python-version: 3.7
+        python-version: 3.8
         cache: 'pip'
         cache-dependency-path: '.github/workflows/black.yml'
     - run: pip install black==22.3.0
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -14,8 +14,8 @@ jobs:
   check-black-formatting:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
       with:
         python-version: 3.7
         cache: 'pip'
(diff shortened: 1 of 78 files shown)


albertz:
When I mention that I would try to optimize this, I was more thinking about writing a dedicated script just to do that.

But this here could also be interesting.

I'm not sure thought that I would put this into RETURNN yet, while you are still experimenting with it. You can rather put this somewhere into your `i6_experiments.users.` and then just use it from there. (I think we might need to extend the class resolution slightly, `if "." in cls_name: ...`; see similar code in `rf.build_from_dict` or `get_optimizer_class` or other such functions.)

albertz:
> This is a first attempt at automatically optimizing the bucket limits during training. After every subepoch the limits are adjusted to make every bucket catch a roughly equal number of segments.

With "limit", you mean the max num seqs of a bucket?

I still don't really understand why this is something you want to optimize for. Why does it matter that they are evenly distributed?

This is to get better speed or better model performance or both? I don't see how this affects speed. I also don't see/understand how this affects model performance.

I thought that optimizing the max seq lens of each bucket, i.e. the boundaries, would be a more reasonable thing to optimize w.r.t. minimizing the amount of padding. That would improve at least the speed.

albertz:
It seems this PR includes all the epoch_continuous changes? Why? Are those in any way related here? I don't see how. If they are not related, can you remove them from this PR?

NeoLegends:
I don‘t think this PR is going to be merged in the current form.

PR #1663, opened 2024-12-06
`LockFile`: inspect other processes to check whether lockfile is held

NeoLegends:
For #1643.


NeoLegends:
Probably no longer needed.

PR #1665, opened 2024-12-11
RF cum_concat_step simplify and other RF things

albertz:
(This is just here to at least run the test once before I merge it into main.)

(The other commits accidentally went into here, but it cannot hurt. Let's just not squash them together but rebase then.)

### DIFF
--- a/returnn/frontend/array_.py
+++ b/returnn/frontend/array_.py
@@ -385,6 +385,8 @@ def concat(
             assert src.dims_set - {dim} == dims, f"concat {sources}, need allow_broadcast=True"
     if not out_dim:
         out_dim = sum(d for _, d in sources)
+    for src, dim in sources[:-1]:
+        assert dim.is_static(), f"concat {sources}, dim {dim} is not static"
     # noinspection PyProtectedMember
     return sources[0][0]._raw_backend.concat(*sources, allow_broadcast=allow_broadcast, out_dim=out_dim), out_dim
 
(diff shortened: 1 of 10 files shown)


returnn/frontend/array_.py
```diff
@@ -507,13 +512,18 @@ def cum_concat_step(
...
     if not out_spatial_dim:
         out_spatial_dim = axis + 1
-    # noinspection PyProtectedMember
-    return (
-        source._raw_backend.cum_concat_step(source, prev_accum=prev_accum, axis=axis, out_spatial_dim=out_spatial_dim),
-        out_spatial_dim,
+    out, (out_spatial_dim,) = rf.pad(
+        prev_accum, axes=[axis], padding=[(0, 1)], out_dims=[out_spatial_dim], value=source, handle_dynamic_dims=True
     )
+    return out, out_spatial_dim
```
NeoLegends, line 526: I approve the idea here. Seems like `rf.pad` does not like padding w/ entire tensors though. CI says `value` needs to be scalar.

albertz:
> What about the other backends? It seems strange to me that only one backend can pad w/ non-scalar values. If we only implement this for torch I'd rather adjust the RF `cum_conat_step` impl to not rely on `rf.pad` but a more general operation that works across all backends, no?.

There is not really any other backend currently which fully works and is actively used, so I don't think this has such a high priority now.

NeoLegends:
But is it much work to change the impl to be generic by default?

albertz:
I don't exactly understand? Via this PR, `rf.cum_concat_step` becomes generic, so this PR here fixes this?

NeoLegends:
I mean wrt. `rf.pad` of non-scalar values.

albertz:
But I still don't understand? `rf.pad` never supported non-scalar values? Now it does for the PT backend.

PR #1672, opened 2025-01-08
Fix _distribute_evenly_by_size for duplicate entries in files_order

dorian-K:
I ran into this bug when I added the same file into the "files" list of DistributeFilesDataset multiple times, which messed up the distribution over the subepochs. Subepochs 1-299 contained 1-2 files, and subepoch 300 contained over 50.

### DIFF
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -364,7 +364,7 @@ class DistributeFilesDataset(CachedDataset2):
         Distributes the files from files_order into ``num_bins`` while attempting
         to make every bin as evenly sized (based on ``file_sizes``) as possible.
         """
-        total_size = sum([file_sizes[_get_key_for_file_tree(f_tree)] for f_tree in files_order])
+        total_size = sum(file_sizes[_get_key_for_file_tree(f_tree)] for f_tree in files_order)
         avg_size_per_sub_epoch = total_size / num_bins
         # Now evenly distribute the files over the bins.
         # Note that many one-pass variants of algorithms to evenly distribute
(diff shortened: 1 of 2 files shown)


returnn/datasets/distrib_files.py
```diff
@@ -364,8 +364,7 @@ def _distribute_evenly_by_size(
         Distributes the files from files_order into ``num_bins`` while attempting
         to make every bin as evenly sized (based on ``file_sizes``) as possible.
         """
-
-        total_size = sum(file_sizes.values())
+        total_size = sum([file_sizes[_get_key_for_file_tree(f_tree)] for f_tree in files_order])
```
NeoLegends, line 367:
```suggestion
        total_size = sum(file_sizes[_get_key_for_file_tree(f_tree)] for f_tree in files_order)
```

NeoLegends (approved):
Interesting use case. Are you trying to mix multiple datasets in a weighed way?

Just a very minor style nit, no strong need to fix this from my side.

tests/test_Dataset.py
```diff
@@ -905,8 +905,10 @@ def __getitem__(self, item):
 def test_DistributeFilesDataset_distribute_evenly_by_size():
     from returnn.datasets.distrib_files import DistributeFilesDataset
 
-    def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]]):
-        files = [f"file-{i}" for i in range(len(sizes))]
+    def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]], files_order=None):
```
NeoLegends, line 908: Oh another small nit (probably requires reformatting w/ black):
```suggestion
    def _test(sizes: List[int], partition_epoch: int, expected: List[List[int]], files_order: Optional[List[str]] = None):
```

dorian-K:
> Interesting use case. Are you trying to mix multiple datasets in a weighed way?

Yes, exactly. Although I assume there are better ways to properly mix datasets, this felt like a quick solution (until my training failed catastrophically at the very last epoch...)

albertz (approved):

PR #1673, opened 2025-01-09
Dataset: allow LR scheduling based on `get_complete_frac`

NeoLegends:
I want to be able to use `epoch_continuous` w/ `PostprocessingDataset` when `map_seq_stream` is applied (which makes knowing how many seqs there are complicated). This PR in its form does not accomplish that yet.

We could:
1. Let the user define an additional function that estimates (in a more precise way than we can do ourselves) how many segments there will be. We will then use this number to base the `epoch_continuous` calculation from. But: what will the inputs to that function be? Just how many seqs there are in the underlying dataset? This will work w/ reasonable accuracy for relatively `map_seq_stream` (like my random sequence concatenation), but what if the postprocessing function looks at the individual segments to decide how many there are going to be? When will it be invoked?
2. Make the dataset let the training know about the share of data that is "already consumed" for the current (sub)epoch. PPDataset could simply pass on this information. This will result in slight inaccuracies/non-smooth `epoch_continuous` values for my case though (because `map_seq_stream` groups seqs together), but maybe that's okay. We can also name the resulting key something other than `epoch_continuous` to make this clear. This functionality would be very similar to `get_complete_frac`, but we wouldn't pass in a `seq_idx` as parameter since that would not be a valid/accurate `seq_idx` in the wrapped dataset.

I'm leaning towards implementing 2). 1) seems to have many footguns. WDYT @dorian-K @albertz?


NeoLegends:
Ah, one issue that I see coming up w/ this is that whenever there is buffering in place (like when `LaplaceOrdering` is used) there will be a constant offset in the true `epoch_continuous` vs. the one the wrapped dataset will report. Not sure if we can work around that. Maybe it just needs to be documented well enough. 🤔

albertz:
`PostprocessingDataset` with `LaplaceOrdering` (or anything which would possibly reorder the seqs) anyway also needs to be fixed then, right? Currently `LaplaceOrdering` operates on `TensorDict`, and the `epoch_continuous` (maybe `complete_frac`) would just be in there, just like `seq_tag`? So then should `LaplaceOrdering` be extended to fix that? But should we expect that for every possible other user `map_seq_stream` function, that the user properly handles that? Maybe it's ok if the user ignores it, but then we should fix it up somehow? E.g. making sure it's at least monotonic?

I don't see any problem for `MultiProcDataset`? Why would the buffer cause some offset? No, that should not be the case.


> but we wouldn't pass in a `seq_idx` as parameter since that would not be a valid/accurate `seq_idx` in the wrapped dataset.

I don't really understand this. Of course we would pass `seq_idx` in there. The function would behave very much like `get_seq_tag` or `get_data`, which needs `seq_idx` as well. The `seq_idx` is always w.r.t. the dataset itself. So if there is some meta dataset like `PostprocessingDataset`, this `seq_idx` of course would not be just passed on to the subdataset. You would treat the `complete_frac` just in the same way as you treat `seq_tag` and other things.

returnn/datasets/basic.py
```diff
@@ -799,6 +801,7 @@ def get_data(self, seq_idx, key) -> numpy.ndarray:
         :param str key: data-key, e.g. "data" or "classes"
         :return: features or targets: format 2d (time,feature) (float)
         """
+        self._last_cur_seq_idx = max(self._last_cur_seq_idx, seq_idx)
```
albertz, line 804: Why do you need that?

NeoLegends:
> PostprocessingDataset with LaplaceOrdering (or anything which would possibly reorder the seqs) anyway also needs to be fixed then, right? Currently LaplaceOrdering operates on TensorDict, and the epoch_continuous (maybe complete_frac) would just be in there, just like seq_tag?

Ah yeah, this is a good idea.

> So if there is some meta dataset like PostprocessingDataset, this seq_idx of course would not be just passed on to the subdataset. You would treat the complete_frac just in the same way as you treat seq_tag and other things.

How do you map the `seq_idx` passed to PostprocessingDataset to the `seq_idx` the wrapped dataset would expect, without keeping an explicit map in the dataset (which seems too costly)?

returnn/datasets/basic.py
NeoLegends: I'm still wondering how to map the indices between their respective spaces. Keeping the index inside the dataset circumvents any mapping. But I also wrote this in the other comment.

NeoLegends:
Maybe we don't need the full map, just the latest value.

albertz:
> How do you map the `seq_idx` passed to PostprocessingDataset to the `seq_idx` the wrapped dataset would expect, without keeping an explicit map in the dataset (which seems too costly)?

No need for an explicit map. It's just the same as you handle any other data, e.g. also specifically how you handle `seq_tag`.

NeoLegends:
Yeah, I now know how to do it.

albertz:
Ok. Still what I raised before remains a bit the question, how to handle reordering like `LaplaceOrdering`. `LaplaceOrdering` can itself easily fix this. But I think we also need to allow that some `map_seq_stream` from the user ignores this, and then we should reasonably fix this up (e.g. making it monotonic at least? not sure...). (But then, maybe we should also not overthink this now. This case is even not really relevant for our current use cases.)

NeoLegends:
Yes I'm on it right now, just not pushed yet. :)

returnn/datasets/basic.py
```diff
@@ -967,6 +967,25 @@ def num_seqs(self) -> int:
...
 
+    def get_epoch_continuous(self, sorted_seq_idx: int) -> float:
+        """
+        Calculates the position of ``sorted_seq_idx`` in the current epoch as relative value in [0, 1].
+        ``sorted_seq_idx`` cannot be less than the seq index of the previously loaded seq.
+
+        This value is used to calculate ``epoch_continuous`` for any dynamic learning rate scheduling.
+
+        :param sorted_seq_idx: sorted seq idx
+        :return: continuous value in [0, 1] which represents the position in the current epoch
```
albertz, line 978:
```suggestion
        :return: continuous value in [0, 1) which represents the position in the current epoch
```
It's actually important that this never reaches 1, so that it can easily be used to calculate `epoch_continuous` based on this, and such that we can guarantee that `int(epoch_continuous) == epoch`, which would otherwise not be the case for the last seq of the dataset.

NeoLegends: We actually do this differently in the torch engine right now. There the value goes from `[epoch - 1, epoch]` inclusive.

NeoLegends: In the torch engine we have (as you correctly say):
```
epoch_continuous = self.epoch - 1 + (last_seq_idx + 1) / num_seqs
```
Why does this mean this value can never reach 1? It must, eventually, if `seq_idx == num_seqs - 1`.

albertz: Ah yes sorry, I confused this. So it's actually (0,1]. The interpretation is: How much have we finished of the epoch including the seqs of the current batch. We start with the first seq, which is close to 0.0, and the last minibatch of the first epoch (epoch 1) would then finally reach 1.0.

We should not change how it is in the engine now. So what we do here should be consistent to that.

NeoLegends: It is consistent to the existing behavior the way it is now!

NeoLegends:
>But I think we also need to allow that some map_seq_stream from the user ignores this

I think the convention should be that you simply leave untouched/pass on/copy over any entries in the tensor dict that you do not know about or wish to manipulate. That way we have forward compatibility.

returnn/datasets/postprocessing.py
```diff
@@ -57,6 +58,14 @@ class PostprocessingDataset(CachedDataset2):
             },
         }
 
+    The postprocessor functions operate on ``TensorDict``s, which have entries for
+    all data keys in the underlying dataset.
+
+    There may also be additional "meta" entries in the tensor dicts, like ``epoch_continuous``
+    and ``seq_tag``. These may be modified, but not removed. Any other entries that
```
albertz, line 65:
```suggestion
    and ``seq_tag``.
    These may be modified, but not removed. Any other entries that
```

NeoLegends:
I believe those test failures are unrelated to the changes.

NeoLegends:
Ahh, the implementation is incorrect for sharding datasets. Need to fix this before merge.

returnn/datasets/basic.py
```diff
@@ -967,6 +967,26 @@ def num_seqs(self) -> int:
         """
         raise NotImplementedError
 
+    def get_epoch_continuous(self, sorted_seq_idx: int) -> float:
```
albertz, line 970: The name here is wrong. `epoch_continuous` is what you have here added to `epoch - 1`.

However, it makes sense to calculate what you do here. So do not change the logic, just change the name of this function (and similarly also below for other uses of "epoch_continuous"). I suggested "complete_frac" before. Not sure. So maybe:
```suggestion
    def get_complete_frac(self, sorted_seq_idx: int) -> float:
```

returnn/datasets/basic.py
```diff
@@ -1374,16 +1394,26 @@ class DatasetSeq:
...
     """
 
-    def __init__(self, seq_idx, features, targets=None, seq_tag=None):
+    def __init__(
+        self,
+        seq_idx: int,
+        features,
+        targets=None,
+        seq_tag: Optional[str] = None,
+        epoch_continuous: Optional[float] = None,
```
albertz, line 1403:
```suggestion
        complete_frac: Optional[float] = None,
```

albertz, line 1403: I'm a bit hesitating whether this really needs to be part of `DatasetSeq`...

returnn/datasets/cached2.py
```diff
@@ -228,6 +228,15 @@ def get_target_list(self):
...
 
+    def get_epoch_continuous(self, sorted_seq_idx: int) -> float:
+        """
+        :return: epoch continuous value for the given seq_idx
+        """
+        seq = self._get_seq(sorted_seq_idx)
+        if seq.epoch_continuous is not None:
+            return seq.epoch_continuous
+        return super().get_epoch_continuous(sorted_seq_idx)
+
```
albertz, line 239: I think this here is wrong?
```suggestion
```

NeoLegends: Hm, this is meant to add support for `complete_frac` for datasets that store the value in `DatasetSeq` in a generic fashion. If the seq does not have the `complete_frac` defined, the impl falls back to the one from the base class based on `num_seqs` (which is not defined for all cases where the value is defined in the `DatasetSeq`).

I agree this implementation is weird to have given we have so many derived implementations of `CachedDataset2` that do not store `complete_frac` in `DatasetSeq`. I don't see where this would be wrong, though?

Where else would you store the `complete_frac` if not as property on the `DatasetSeq`? As tensor inside `DatasetSeq.features`? Then the implementations would each need a special cased variant of the current functionality and we might end up with some duplicated code.

returnn/datasets/basic.py
NeoLegends: They don't /have/ to be extended, they can also choose to define `num_seqs`... But I agree having this distinction seems a bit off.

returnn/datasets/basic.py
```diff
@@ -1375,16 +1398,26 @@ class DatasetSeq:
...
+    ):
         """
-        :param int seq_idx: sorted seq idx in the Dataset
+        :param seq_idx: sorted seq idx in the Dataset
         :param numpy.ndarray|dict[str,numpy.ndarray] features: format 2d (time,feature) (float)
         :param dict[str,numpy.ndarray]|numpy.ndarray|None targets: name -> format 1d (time) (idx of output-feature)
-        :param str seq_tag: sequence name / tag
+        :param seq_tag: sequence name / tag
+        :param complete_frac: continuous value in (0, 1] which represents of much of the current epoch
+            is consumed when this seq is processed
```
dorian-K, line 1415:
```suggestion
        :param complete_frac: continuous value in (0, 1] which represents how much of the current epoch
            has been consumed when this seq is processed
```

returnn/datasets/cached2.py
```diff
@@ -228,6 +228,15 @@ def get_target_list(self):
             keys.remove("data")
         return keys
 
+    def get_exact_complete_frac(self, sorted_seq_idx: int) -> float:
+        """
+        :return: fractional completion value for the given seq_idx
+        """
+        seq = self._get_seq(sorted_seq_idx)
```
albertz, line 235: This is a bit costly. Doesn't it make more sense to first check `super().get_exact_complete_frac(sorted_seq_idx)` and if that throws `OptionalNotImplementedError`, then try this?

returnn/datasets/basic.py
```diff
@@ -960,6 +960,29 @@ def get_complete_frac(self, seq_idx):
...
+        having processed seq ``sorted_seq_idx``.
+        ``sorted_seq_idx`` cannot be less than the seq index of the previously loaded seqs.
+
+        Raises a ``NotImplementedError`` if the value cannot be calculated.
+
+        This value is used to calculate ``epoch_continuous`` for any dynamic learning rate scheduling.
+
+        :param sorted_seq_idx: sorted seq idx
+        :return: continuous value in (0, 1] which represents how much of the current epoch
+            is completed after ``sorted_seq_idx``
```
albertz, line 975: This should be extended. Specifically, you should say that we guarantee this to be monotonic (but not necessarily strictly monotonic).

returnn/datasets/multi_proc.py
```diff
@@ -182,6 +183,8 @@ def _lazy_init(self):
         msg, self.labels = self._seq_order_proc_parent_conn.recv()
         assert msg == "labels"
 
+        self._cur_max_complete_frac = 0.0
```
albertz, line 186: This should not be set here, but instead in `init_seq_order`.

returnn/torch/data/pipeline.py
```diff
@@ -215,11 +218,11 @@ def __init__(self, dataset: torch.utils.data.IterableDataset, batch_size=1, max_
             in one batch (padding included).
             If given as a dict data_key -> value, sets different individual limits per data key.
             If None, no limit.
-            Can also be a callable with kwargs epoch, seq_idx, epoch_continuous, **_other_kwargs,
+            Can also be a callable with kwargs epoch, seq_idx, complete_frac, **_other_kwargs,
```
albertz, line 221: I don't understand?

returnn/torch/data/pipeline.py
```diff
@@ -268,9 +271,13 @@ def _parse_max_seqs(
     def _get_user_func_kwargs_from_data_dict(data_dict: Dict[str, Any]) -> Dict[str, Any]:
         epoch = int(data_dict["epoch"])
         seq_idx = int(data_dict["seq_idx"])
-        num_seqs = int(data_dict["num_seqs"])  # >=1 if known, otherwise -1
-        epoch_continuous = (epoch - 1 + (seq_idx + 1) / num_seqs) if num_seqs > 0 else None
-        return {"epoch": epoch, "seq_idx": seq_idx, "epoch_continuous": epoch_continuous, **get_fwd_compat_kwargs()}
+        complete_frac = float(data_dict["complete_frac"])  # >= 0 if known, otherwise -1
+        return {
+            "epoch": epoch,
+            "complete_frac": complete_frac if complete_frac >= 0.0 else None,
```
albertz, line 277: You are breaking here user code? That doesn't make sense?

returnn/datasets/cached2.py
NeoLegends: Fine with me!

returnn/torch/data/pipeline.py
NeoLegends: Ah, I mistakenly must have search+replaced this.

returnn/torch/data/pipeline.py
NeoLegends: Same as above.

tests/test_Dataset.py
```diff
@@ -1135,14 +1135,21 @@ def _repeat2(input_iter: Iterator[TensorDict], **kwargs) -> Iterator[TensorDict]
...
             assert prev_len is None or classes.shape[0] >= prev_len
+            assert prev_complete_frac is None or complete_frac > prev_complete_frac
             prev_len = classes.shape[0]
+            prev_complete_frac = complete_frac
         for i in range(3, 6):
             classes = dataset.get_data(i, "classes")
-            assert prev_len is None or classes.shape[0] <= prev_len or i == 3
+            complete_frac = dataset.get_exact_complete_frac(i)
+            assert classes.shape[0] <= prev_len or i == 3
+            assert complete_frac > prev_complete_frac
```
albertz, line 1150:
```suggestion
            assert complete_frac >= prev_complete_frac
```

tests/test_Dataset.py
```diff
@@ -1135,14 +1135,21 @@ def _repeat2(input_iter: Iterator[TensorDict], **kwargs) -> Iterator[TensorDict]
         dataset.load_seqs(0, 6)
 
         prev_len = None
+        prev_complete_frac = None
         for i in range(3):
             classes = dataset.get_data(i, "classes")
+            complete_frac = dataset.get_exact_complete_frac(i)
             assert prev_len is None or classes.shape[0] >= prev_len
+            assert prev_complete_frac is None or complete_frac > prev_complete_frac
```
albertz, line 1143:
```suggestion
            assert prev_complete_frac is None or complete_frac >= prev_complete_frac
```

returnn/torch/engine.py
```diff
@@ -398,22 +399,30 @@ def train_epoch(self):
...
-                        num_seqs = num_seqs_
-                elif num_seqs_ >= 0:
-                    assert num_seqs_ == num_seqs
-                del num_seqs_
-                if num_seqs is not None:
-                    assert last_seq_idx < num_seqs
-                epoch_continuous = (self.epoch - 1 + (last_seq_idx + 1) / num_seqs) if num_seqs is not None else None
+                complete_frac = float(extern_data_raw["complete_frac"])
+                # Note: can only use complete_frac from data dict if batches are not shuffled,
+                # otherwise fall back to calculation via seq_idx and num_seqs.
```
albertz, line 404: We should not do this here. Rather, in `returnn.torch.engine.Engine._create_data_loader`, we could have another `IterDataPipe` which makes sure that `complete_frac` and/or `epoch_continuous` is monotonic.

returnn/datasets/basic.py
```diff
@@ -960,6 +960,29 @@ def get_complete_frac(self, seq_idx):
                 num_seqs = None  # ignore
         return self.generic_complete_frac(seq_idx, num_seqs)
 
+    def get_exact_complete_frac(self, sorted_seq_idx: int) -> float:
```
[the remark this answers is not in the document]
albertz: > I think estimated values are not appropriate if we want to base the LR scheduling on top of that.

I don't understand. There are other estimates we do later (e.g. in `MultiProcDataset`, or `PostprocessingDataset`), which are much more inaccurate than what `get_complete_frac` is maybe doing. So I don't see why this is an issue for `get_complete_frac`.

returnn/datasets/cached2.py
albertz: With costly, I mean that `get_complete_frac` could trigger that some data would be loaded (via `_get_seq`) which was maybe not loaded yet and also there is no reason to load it, specifically not for just getting the complete_frac.

NeoLegends: I don't think `_get_seq` loads new segments unless you call `load_seqs`.

returnn/datasets/basic.py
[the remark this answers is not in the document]
NeoLegends: The existing `get_complete_frac` does the following:

```python
@classmethod
def generic_complete_frac(cls, seq_idx, num_seqs):
    """
    :param int seq_idx: idx
    :param int|None num_seqs: None if not available
    :return: Returns a fraction (float in [0,1], always > 0) of how far we have advanced
        for this seq in the dataset.
        This does not have to be exact. This is only for the user.
    """
    if num_seqs:
        return min(float(seq_idx + 1) / num_seqs, 1.0)
    else:
        # We don't know. So:
        # Some monotonic increasing function in [0,1] which never reaches 1.
        import math

        return max(1.0e-10, 1.0 - math.exp(-seq_idx * 1000))

def get_complete_frac(self, seq_idx):
    """
    :param int seq_idx:
    :return: Returns a fraction (float in [0,1], always > 0) of how far we have advanced
        for this seq in the dataset.
        This does not have to be exact. This is only for the user.
    :rtype: float
    """
    # noinspection PyBroadException
    try:
        num_seqs = self.num_seqs
    except Exception:  # num_seqs not always available
        # noinspection PyBroadException
        try:
            num_seqs = self.estimated_num_seqs
        except Exception:  # also not always available
            num_seqs = None  # ignore
    return self.generic_complete_frac(seq_idx, num_seqs)
```

This means that, whenever `num_seqs` is not known, it will report a kind of "arbitrary" monotonic value that never reaches `1`. My specific case (PPDataset + Seq Concat) neither knows `num_seqs` nor `estimated_num_seqs` (it does not set `map_seq_stream_preserves_num_seqs`), so it will always run into this approximative case, and I think this approximation is not appropriate for LR scheduling.

returnn/torch/data/pipeline.py
```diff
@@ -463,6 +466,66 @@ def __getitem__(self, index):
         raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")
 
 
+class ShufflingDataPipe(torch.utils.data.IterDataPipe):
```
albertz, line 469: This does not really belong into this PR?

albertz: Ah now I see, you rewrote that. That's also a solution. But it's different to what I suggested. Why do you do this now and not what I suggested?

NeoLegends: I know how I can make the values monotonic w/ another adapter, but I think that w/o having access to all batches in the shuffle buffer ensuring monotonicity will result in jumps and "skips" in the values (since you effectively do a `batch["complete_frac"] = max(last_complete_frac, batch["complete_frac"]`). By re-writing the adapter we can keep the same, smooth, monotonic values that we have had before. I don't know if there is a better way other than re-writing the shuffler. Do you know a better way?

albertz: Yes sure. I just thought that the monotonicity-ensuring-adapter is more generic and might be useful in other cases as well. But yes, what you did is better for this specific case.

I just wanted to make sure there was no misunderstanding. And I was wondering initially why you did it.

I think it's fine and it can stay this way.

returnn/datasets/cached2.py
albertz: Oh sorry, you are right, I remembered wrongly. So then it doesn't really matter.

returnn/datasets/basic.py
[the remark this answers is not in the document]
albertz: Ah sorry, I forgot about this behavior of `generic_complete_frac`.

But anyway, why don't you adapt this `get_complete_frac` then? I still don't really understand why we need another variant of this function with a slightly different name, even though the meaning of the function is basically the same. I don't really understand even how they are supposed to be different.

So, what is the difference? And why can the `get_complete_frac` not be adapted?

Maybe it's the type of approximation? But e.g. for `MultiProcDataset` or `PostprocessingDataset`, you also have other approximations. So when is an approximation still good enough and when not?

And what fallback do we want when the approximation becomes too inaccurate? Can we even always tell how inaccurate it is? E.g. in `PostprocessingDataset`, we would have no idea in general.

`get_complete_frac` currently is mostly used for visual feedback (e.g. in terminal progress bars), remaining time estimation. So not so critical.

So, you want that when it is too inaccurate (however you want to tell that), then there is no fallback and just `None` passed to the user (i.e. `None` for `epoch_continuous`), so that the user doesn't use this accidentally for LR scheduling without realizing that it's very inaccurate?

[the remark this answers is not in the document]
albertz: Also, consider this for code maintenance in the future: When someone will see this code, and see that there is some `get_complete_frac` function and some `get_exact_complete_frac` (which is also not really always exact), which seem to do the same thing, that would be extremely confusing.

So, if there are two possible behavior types for approximations (always return sth vs failing with error / returning `None`), we should still keep a single function, but add a flag, like `only_allow_accurate` or so.

returnn/datasets/basic.py
```diff
@@ -941,48 +941,45 @@ def generic_complete_frac(cls, seq_idx, num_seqs):
...
-        except Exception:  # num_seqs not always available
-            # noinspection PyBroadException
-            try:
-                num_seqs = self.estimated_num_seqs
-            except Exception:  # also not always available
-                num_seqs = None  # ignore
-        return self.generic_complete_frac(seq_idx, num_seqs)
-
-    def get_exact_complete_frac(self, sorted_seq_idx: int) -> float:
+    def get_complete_frac(self, sorted_seq_idx: int, *, allow_approximation=True):
```
albertz, line 944:
```suggestion
    def get_complete_frac(self, sorted_seq_idx: int, *, allow_approximation: bool = True):
```

returnn/datasets/basic.py
```diff
@@ -941,48 +941,45 @@ def generic_complete_frac(cls, seq_idx, num_seqs):
...
         """
         Tries to calculate exactly how much of the current epoch is completed when
         having processed seq ``sorted_seq_idx``.
 
-        Raises a ``NotImplementedError`` if the value cannot be calculated.
+        If ``allow_approximation=False``, raises a ``NotImplementedError`` if the value cannot be
+        calculated exactly.
```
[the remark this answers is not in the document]
NeoLegends: Ah, good point. I think `None` might be preferrable. Many downstream implementations catch the exception and convert the value to `None`, so we might just as well do this directly w/o the exception in the way.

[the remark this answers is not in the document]
albertz: We should check whether this potentially breaks some external usages (e.g. search usages through i6_experiments).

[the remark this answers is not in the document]
NeoLegends: Given the flag is set to allowing approximative values by default I think we are safe (I also checked).

returnn/torch/data/pipeline.py
```diff
@@ -455,6 +466,74 @@ def __getitem__(self, index):
...
+    """Data pipe that shuffles batches while keeping certain data keys in order."""
```
albertz, line 470: Extend the doc to explain how this differs from (or extends over) `torch.utils.data.datapipes.iter.Shuffler`.

returnn/torch/data/pipeline.py
```diff
@@ -489,11 +494,18 @@ def __init__(
...
-        self._rng = numpy.random.RandomState(seed)
+        self._rng = numpy.random.RandomState()
+        self._seed = seed
 
     def __iter__(self):
         # The implementation is very similar to the PostprocessingDataset's combinator LaplaceOrdering.
 
+        if self._seed is None:
+            # torch uses a random seed when no other seed has been set via set_seed or via __init__
+            self._seed = numpy.random.randint(0)
```
albertz: No that's not exactly correct. `torch.utils.data.datapipes.iter.Shuffler` has this:
```python
    def reset(self) -> None:
        self._buffer = []
        if self._enabled:
            if self._seed is None:
                self._seed = int(torch.empty((), dtype=torch.int64).random_().item())
            self._rng.seed(self._seed)
            self._seed = None
```
I think the `reset` function is also called when a new iterator is created (I always forget which of those funcs is called when exactly... it's also quite non-trivial to follow this code/logic...)

Specifically, this uses the Torch RNG. This is crucial, because the Torch RNG is properly initialized: In the PT engine in `_create_model`, we call `rf.set_random_seed`, which also initializes the Torch RNG. Then in `train_epoch` (and co), we create the iterator (`data_iter = iter(self._train_dataloader)`), which would call `reset` on all involved data pipes.

returnn/torch/data/pipeline.py
```diff
@@ -455,6 +466,94 @@ def __getitem__(self, index):
...
+    def set_seed(self, seed: int) -> ShufflingDataPipe:
+        """
+        Sets the seed for the next invocation of ``__iter__``, for compatibility with
+        ``torch.utils.data.graph_settings.apply_random_seed``.
+        """
+        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
+        return self
+
+    def __getitem__(self, index):
+        raise Exception(f"{self.__class__.__name__}.__getitem__ not supported")
```
albertz, line 554: Note: `torch.utils.data.datapipes.iter.Shuffler` also defines `__getstate__` and `__setstate__`. I'm not exactly sure why that is needed, i.e. whether/why the serialization breaks otherwise. But then the question is, if it breaks there, maybe it would also break for you? Or maybe it only breaks under certain circumstances? From a quick glance, I'm not exactly sure why it is needed. Maybe the RNG state is not properly preserved otherwise?

returnn/torch/data/pipeline.py
NeoLegends: Oh, I see. This is quite intricate.

dorian-K:
I find the use of the parameter name `allow_approximation` a bit misleading, because LR scheduling can still use an approximation as long as its accurate *enough*, like a relative error of a few percent (this would be my use case). 
And if someone wants to implement a dataset and use it together with LR scheduling, they would see this code and simply ignore the value of `allow_approximation` and return the approximation, which would be very bad if future code were to actually need a "perfect" estimation and assumes that setting `allow_approximation=False` would achieve this.

NeoLegends:
I am wondering: do we even need this generic fallback after this PR? The PR enables computing meaningful `complete_frac`s in more scenarios than before, maybe we have covered them all now? I will now check the datasets for support.

albertz:
> I am wondering: do we even need this generic fallback after this PR? The PR enables computing meaningful `complete_frac`s in more scenarios than before, maybe we have covered them all now? Are there any cases now that are conceptually not-covered by the new implementation (maybe infinite datasets, but these cannot ever return meaningful values anyway)? I will now check the datasets for support.

I think there are quite a number of scripts (also outside of RETURNN, e.g. in i6_experiments) which use this function. We generally don't want to break user code.

NeoLegends:
Ok, we keep get_generic_complete_frac, but maybe we can get rid of the `allow_approximation` flag altogether and make `Dataset.get_complete_frac` never take the branch where `num_seqs` is `None`. 🤔

albertz:
> To clarify what it means right now: In `generic_complete_frac`, we simply use this code to get some "complete frac":
> 
> ```python
> # Some monotonic increasing function in [0,1] which never reaches 1.
> return max(1.0e-10, 1.0 - math.exp(-seq_idx * 1000))
> ```

@dorian-K realized, this formula that we use here is 1.0 for all `seq_idx >= 1`. I think this was a mistake in the code, and it should have been `... / 1000` instead of `... * 1000`. So, it means, currently the `generic_complete_frac` is not really meaningful for the case that `num_seqs` is unknown.

albertz:
> > I find the use of the parameter name `allow_approximation` a bit misleading, because LR scheduling can still use an approximation as long as its accurate _enough_, ...
> 
> Yes, maybe we can use a better name. I agree, it's not so clear. Any suggestions?

Now @NeoLegends changed this to `allow_only_exact`, but that has exactly the same problems as before: LR scheduling can still use it even when it is not exact.

We want a flag that expresses: Return a value only if it would be _accurate enough for LR scheduling_. If this flag is enabled, and the dataset cannot guarantee that it is accurate enough, it would return None. If this flag is disabled (default), it would always return something (even if not meaningful) (to not break existing code).

So I discussed with @dorian-K: As this is what we want to use the flag for, and this is what the flag should express, we should just express this meaning in the name of the flag, and ambiguous like `allow_approximation` or `allow_only_exact`. So maybe `allow_only_when_suitable_for_lr_scheduling` or so?

Btw, when `estimated_num_seqs` is set, would this be suitable for LR scheduling? I would say, probably yes. It's anyway quite rare that we have this case.

returnn/datasets/basic.py
```diff
@@ -937,28 +938,51 @@ def generic_complete_frac(cls, seq_idx, num_seqs):
...
+
+        ``sorted_seq_idx`` cannot be less than the seq index of the previously loaded seqs.
+
+        :param sorted_seq_idx: sorted seq idx
+        :param allow_only_lr_suitable: only return a value when that value is suitable/accurate enough
+            to base LR scheduling on it. If false, this function will return an approximative value
+            when the exact value cannot be calculated (due to unknown ``num_seqs``).
+            Approximative values can be appropriate for e.g. progress bars.
+        :return: continuous value in (0, 1] which represents how much of the current epoch
+            is completed after ``sorted_seq_idx`
```
albertz, line 956:
```suggestion
            is completed after ``sorted_seq_idx``
```

returnn/datasets/basic.py
```diff
@@ -1375,16 +1399,26 @@ class DatasetSeq:
     Encapsulates all data for one sequence.
     """
 
-    def __init__(self, seq_idx, features, targets=None, seq_tag=None):
+    def __init__(
+        self,
+        seq_idx: int,
+        features,
```
albertz, line 1405:
```suggestion
        features,
        *,
```

returnn/torch/data/returnn_dataset_wrapper.py
```diff
@@ -109,11 +109,18 @@ def __iter__(self) -> Iterable[Dict[str, numpy.ndarray]]:
                 data = {data_key: self._dataset.get_data(seq_index, data_key) for data_key in data_keys}
                 data["seq_tag"] = str_to_numpy_array(self._dataset.get_tag(seq_index))
                 data["seq_idx"] = numpy.array(seq_index)
-                # It's slightly redundant to have num_seqs in each entry,
+
+                # It's slightly redundant to have the following data in each entry,
                 # but it's difficult to pass this back to the main proc otherwise.
-                data["num_seqs"] = num_seqs
-                # epoch is also redundant, but that's the cleanest/simplest way to pass it on to BatchingIterDataPipe.
```
albertz, line 115: Why did you remove this comment?

returnn/datasets/basic.py
```diff
@@ -1375,16 +1399,26 @@ class DatasetSeq:
...
+        targets=None,
+        seq_tag: Optional[str] = None,
+        complete_frac: Optional[float] = None,
```
albertz, line 1408: I think I raised this question before, but it got a bit lost: Why do we need to store it in `DatasetSeq`? Where is this used then? And any outside use of this, why could it not simply use `Dataset.get_complete_frac` instead?

I see this a bit problematic: It complicates the code here, and it complicates every dataset implementation. It's not much, but those small things add up over time. Specifically, it means that every single dataset (or other code) using DatasetSeq needs to be extended. I don't like this.

NeoLegends: Hm, the info could just not be stored here and instead fetched directly from any wrapped dataset. I think only in `PostprocessingDataset` this might become a bit more involved (when `map_seq_stream` is set), or rather, that dataset needs to store `complete_frac` somewhere after `map_seq_stream` has been applied. `PostprocessingDataset` could also just use a custom derived `DatasetSeq` that stores this information.

For `DistributeFilesDataset` and `MultiProcDataset` the wrapped datasets are in subprocesses and communicated with over pipes. Not storing `complete_frac` in `DatasetSeq` would mean these datasets need two round trips per seq into the child process, one for the data, and another one for `complete_frac`. I don't know if that's not too much overhead.

🤔 I'm torn on this. I see what you mean with the complexity creep. As a measure to reduce this, WDYT about using a class that derives from `DatasetSeq` (e.g. `DatasetSeqWithCompleteFrac`) that only the datasets use that benefit from storing `complete_frac` in there? That would limit the complexity creep to only these datasets and leave the rest as-is.

returnn/torch/data/returnn_dataset_wrapper.py
NeoLegends: The new comment refers to all the data.

returnn/datasets/basic.py
albertz: Hm, so I was checking where `DatasetSeq` is actually used. I think it was introduced for `CachedDataset2` and thus is used by all datasets deriving from there. That also includes `MultiProcDataset`, `DistributeFilesDataset`, `PostprocessingDataset`.

I see that it can make sense to also cache this value, specifically for those cases where you cannot easily calculate it otherwise (e.g. `PostprocessingDataset` with `map_seq_stream`9, and then `get_complete_frac` anyway should check the cache.

The question is, now introducing `DatasetSeqWithCompleteFrac`, would that really be simpler? I'm a bit afraid that this adds even more complexity / convoluted code.

So, I would now conclude, leave it like you did it here now, i.e. just extending `DatasetSeq`.

returnn/datasets/basic.py
NeoLegends: Now it's ok.

returnn/datasets/postprocessing.py
```diff
@@ -155,7 +165,10 @@ def __init__(
             self._out_tensor_dict_template = self._in_tensor_dict_template.copy_template()
             self.labels = self._dataset.labels.copy()
         # update only after _out_tensor_dict_template has been created from _in_tensor_dict_template
-        self._in_tensor_dict_template.update({"seq_tag": {"dims": (), "dtype": "string"}}, auto_convert=True)
+        self._in_tensor_dict_template.update(
+            {"complete_frac": {"dims": (), "dtype": "float64"}, "seq_tag": {"dims": (), "dtype": "string"}},
```
albertz, line 169: Why float64 for complete_frac? We usually have float32 everywhere else (by default). I don't see why you need such high accuracy for the complete_frac.

returnn/datasets/basic.py
```diff
@@ -1375,16 +1399,27 @@ class DatasetSeq:
     Encapsulates all data for one sequence.
     """
 
-    def __init__(self, seq_idx, features, targets=None, seq_tag=None):
+    def __init__(
+        self,
+        seq_idx: int,
+        features,
+        targets=None,
+        *,
```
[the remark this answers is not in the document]
NeoLegends: No, this will break existing code.

returnn/datasets/postprocessing.py
NeoLegends: fixed

returnn/datasets/basic.py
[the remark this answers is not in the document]
albertz: I don't understand. Why is it a problem to fix the other RETURNN code? It's only a problem if there is other external code which would break, or not?

tests/test_torch_engine.py
```diff
@@ -760,7 +760,7 @@ def _dynamic_learning_rate(
     assert epoch_continuous_diffs
     print("epoch continuous diffs:", epoch_continuous_diffs)
     # Just some sanity check. The exact number here depends on num_seqs_per_epoch, batch_size, etc.
-    assert numpy.min(epoch_continuous_diffs) >= 0.01
+    assert numpy.min(epoch_continuous_diffs) >= 0.005
```
albertz, line 763: No, the value 0.01 was relevant here. Don't just half it. It should only be off by some eps. So then adapt the check like that. `... >= 0.01 - eps`.

returnn/datasets/postprocessing.py
```diff
@@ -263,7 +293,8 @@ def _build_mapping_iter(self) -> Iterator[TensorDict]:
 
         def _validate_tensor_dict_iter(inner: Iterator[TensorDict]) -> Iterator[TensorDict]:
             for t_dict in inner:
-                assert "seq_tag" in t_dict.data, "seq_tag dropped from TensorDict in postprocessing pipeline"
+                for k in ["complete_frac", "seq_tag"]:
+                    assert k in t_dict.data, f"{k} dropped from TensorDict in postprocessing pipeline"
```
albertz, line 297: I wonder, does this break existing code? Or even if not, do we really want to require that from the user? Even in our own code, we don't always require that `complete_frac` is set for `DatasetSeq` for example.

NeoLegends: > I wonder, does this break existing code?

I know it will not break AppTek-internal code, at i6, I cannot tell.

> we don't always require that complete_frac is set for DatasetSeq for example

Hm, depends on how you want RETURNN to behave wrt. user error. It's not a problem if the postprocessing logic removes `complete_frac` if you don't rely on it for LR scheduling. But it would probably be annoying to debug (have to dig through dataset code) it if you do depend on it for LR scheduling and your postprocessing function accidentally drops that information. In that case the lint would come in very handy.

Other than that, I think it is worthwhile to extend the contract for the postprocessing function along the lines that I have noted in the docs, i.e. that it is required to copy over all tensor dict keys that it does not wish to manipulate. This way, we get automatic forwards compatibility.

albertz: > depends on how you want RETURNN to behave wrt. user error

I don't mean this as a user error. It might just be simpler for the user to write it this way.

But also, what you expect from the user. Ok, this does not break anything now, just because it's not so much used yet. So, this is then maybe fine for now. But consider once this is being more used, and then there comes another meta information we add here, e.g. "orig_seq_idx", or "timestamp", or whatever else. So then, to not break anything, you must handle that case properly.

> extend the contract ... i.e. that it is required to copy over all tensor dict keys

But this is not always well defined. Ok, for `map_seq`, yes, this is fine and trivial. But for `map_seq_stream`, what does this mean? For `map_seq_stream_preserves_num_seqs=True`, yes, maybe it's still well defined. But what about `map_seq_stream_preserves_num_seqs=False`, e.g. when you concat seqs?

NeoLegends: Hm, right, I hadn't thought about this aspect. As long as you know what each key means I think it is quite easy to decide what to do for a specific case, but yes, achieving true forwards compatibility wrt. the meaning of future tensor dict keys is difficult by this requirement...

So I guess it's best to drop this requirement for now.

NeoLegends: >We could make it a requirement for all current keys, and then say that any additional future keys are allowed to be ignored or so. But this is a bit inconsistent: Why is it required for the current keys but not for any future key. But I don't know if this inconsistency (or actually only theoretical/potential/future inconsistency) is such a problem.

Yeah, this sounds like the worst of both worlds to me. The only "issue" that can arise under the current situation is that the support for `epoch_continuous` (which is already an optional feature) depends on your postprocessor function, but nothing else breaks. And I think that `epoch_continuous` LR scheduling is actually not very heavily used (yet). What's more, the code at AppTek already handles this correctly. Seq Tags can be backfilled (which is done now) automatically, just like we already do in other situations (and AppTek code also handles this correctly).

Can you review the recent changes?

> I think we anyway would do a sanity check on it, right?

We could make sure the values are monotonous. Probably this would be best done elsewhere in a generic fashion for all datasets, and not just for PPDataset (or we do it twice).

NeoLegends: >I think we anyway would do a sanity check on it, right?

Added.

returnn/datasets/postprocessing.py
```diff
@@ -262,8 +284,20 @@ def _build_mapping_iter(self) -> Iterator[TensorDict]:
         """
 
         def _validate_tensor_dict_iter(inner: Iterator[TensorDict]) -> Iterator[TensorDict]:
+            last_complete_frac = -1
```
albertz, line 287:
```suggestion
            last_complete_frac = 0
```

returnn/torch/data/returnn_dataset_wrapper.py
```diff
@@ -102,18 +102,32 @@ def __iter__(self) -> Iterable[Dict[str, numpy.ndarray]]:
 
         try:
             data_keys = self._dataset.get_data_keys()
+            last_complete_frac = -1
```
albertz, line 105:
```suggestion
            last_complete_frac = 0
```

returnn/datasets/postprocessing.py
```diff
@@ -307,6 +347,8 @@ def _iterate_dataset(self) -> Iterator[TensorDict]:
 
                 # Re-adding the seq tag here causes no harm in case it's dropped since we don't
                 # add/drop any segments w/ the non-iterator postprocessing function.
+                if "complete_frac" not in tensor_dict.data and comp_frac_tensor is not None:
+                    tensor_dict.data["complete_frac"].raw_tensor = comp_frac_tensor
```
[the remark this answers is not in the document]
NeoLegends: Ahh, good catch.

albertz (approved):

PR #1674, opened 2025-01-09
Interrupt main thread on Exception in sub thread

### LINKED ISSUE #1667: Unhandled exceptions in threads should halt the program
I don't really like that we have threads crashing and that this just stays unnoticed (as it was the case in #1643). Shouldn't we stop the whole program when we get an unhandled exception? For the main thread, this is the case.

Is there an easy way to get this behavior? Also that we maybe have `better_exchook` installed in all threads?

Basically, we should set `sys.excepthook` in every thread to our custom handler, which then does:
* call `better_exchook`
* call `_thread.interrupt_main()`

There is [`threading.excepthook`](https://docs.python.org/3/library/threading.html#threading.excepthook) (since Python 3.8), is this what we want?

NeoLegends:
Now that we have py3.8 we can implement this.

Closes #1667

### DIFF
--- a/returnn/util/debug.py
+++ b/returnn/util/debug.py
@@ -184,7 +184,7 @@ def init_better_exchook():
 
     def threading_excepthook(args, /):
         """threading-specific excepthook"""
-        better_exchook(args["exc_type"], args["exc_value"], args["exc_traceback"], file=log.v1 or sys.stdout)
+        better_exchook(args.exc_type, args.exc_value, args.exc_traceback, file=log.v1 or sys.stdout)
         thread.interrupt_main()
 
     threading.excepthook = threading_excepthook
--- a/returnn/util/debug.py
+++ b/returnn/util/debug.py
@@ -184,7 +184,9 @@ def init_better_exchook():
 
     def threading_excepthook(args, /):
         """threading-specific excepthook"""
-        better_exchook(args.exc_type, args.exc_value, args.exc_traceback, file=log.v1 or sys.stdout)
+        better_exchook(
+            args.exc_type, args.exc_value, args.exc_traceback, autodebugshell=False, file=log.v1 or sys.stdout
+        )
         thread.interrupt_main()
 
     threading.excepthook = threading_excepthook
--- a/returnn/util/debug.py
+++ b/returnn/util/debug.py
@@ -9,8 +9,6 @@ import os
 import sys
 import signal
 
-from returnn.util.basic import interrupt_main
-
 try:
     import thread
 except ImportError:
--- a/returnn/util/debug.py
+++ b/returnn/util/debug.py
@@ -9,6 +9,8 @@ import os
 import sys
 import signal
 
+from returnn.util.basic import interrupt_main
+
 try:
     import thread
 except ImportError:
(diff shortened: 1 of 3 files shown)


returnn/util/debug.py
```diff
@@ -182,6 +182,13 @@ def excepthook(exc_type, exc_obj, exc_tb):
 
     sys.excepthook = excepthook
 
+    def threading_excepthook(args, /):
+        """threading-specific excepthook"""
+        better_exchook(args["exc_type"], args["exc_value"], args["exc_traceback"], file=log.v1 or sys.stdout)
```
albertz, line 187: Did you test this? The [doc](https://docs.python.org/3/library/threading.html#threading.excepthook) says that `exc_type` and co are *attributes* of `args`.
```suggestion
        better_exchook(args.exc_type, args.exc_value, args.exc_traceback, file=log.v1 or sys.stdout)
```

If you did not test this, can you maybe test this? Maybe even write a simple test case, where some thread is spawn, raises some random exception.

tests/test_threading.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        #!rnn.py
+
+        backend = "torch"  # just require any backend
+        log_verbosity = 5
+
+        import threading
+
+        t = threading.Thread(target=lambda: 1/0)
+        t.start()
```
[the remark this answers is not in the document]
NeoLegends: oh good catch

tests/test_threading.py
```diff
@@ -0,0 +1,94 @@
...
+    env_update = os.environ.copy()
+    return env_update
+
+
+def run(args, input=None):
+    """run subproc"""
+    args = list(args)
+    print("run:", args)
+    # RETURNN by default outputs on stderr, so just merge both together
+    p = Popen(args, stdout=PIPE, stderr=STDOUT, stdin=PIPE, env=build_env())
```
albertz, line 36: I think you don't need `env`?
```suggestion
    p = Popen(args, stdout=PIPE, stderr=STDOUT, stdin=PIPE)
```

tests/test_threading.py
```diff
@@ -0,0 +1,94 @@
...
+py = sys.executable
+PY3 = sys.version_info[0] >= 3
+
+
+def build_env():
+    """build env"""
+    env_update = os.environ.copy()
+    return env_update
+
+
```
albertz, line 30:
```suggestion
```

tests/test_threading.py
```diff
@@ -0,0 +1,94 @@
...
+import tempfile
+import textwrap
+from returnn.util import better_exchook
+
+
+__my_dir__ = os.path.dirname(os.path.abspath(__file__))
+__base_dir__ = os.path.dirname(__my_dir__)
+__main_entry__ = __base_dir__ + "/rnn.py"
+py = sys.executable
+PY3 = sys.version_info[0] >= 3
```
albertz, line 22:
```suggestion
```

returnn/util/debug.py
```diff
@@ -182,6 +182,13 @@ def excepthook(exc_type, exc_obj, exc_tb):
 
     sys.excepthook = excepthook
 
+    def threading_excepthook(args, /):
+        """threading-specific excepthook"""
+        better_exchook(args.exc_type, args.exc_value, args.exc_traceback, file=log.v1 or sys.stdout)
```
albertz, line 187:
```suggestion
        better_exchook(args.exc_type, args.exc_value, args.exc_traceback, autodebugshell=False, file=log.v1 or sys.stdout)
```

albertz (approved):

PR #1676, opened 2025-01-16
Dataset: implement global sharding option

### LINKED ISSUE #1634: Sharding for multi-GPU training
Now, as a follow-up of #1630: a very nice next step/feature would be if we can use this sharding feature in general for any kind of multi GPU training. Similar as the `horovod_dataset_distribution="shard"` option we had for TF (which was implemented very inefficiently though, by just selecting every Nth seq from the dataset, i.e. the dataset still iterated through all the data). So maybe, to distinguish or making it more explicit where the sharding is done, we should not just call this `"shard"`, but `"dataset_sharding"` or so.

We should also not reuse the `horovod_dataset_distribution` option (which is intended only for Horovod), but maybe generically `distributed_dataset_distribution` or so? Or it could be part of `torch_distributed`, just `dataset_distribution` in it? (In principle, we could reuse the feature later also for TF or other backend engines. But having it currently in `torch_distributed` is also fine.)

The `dataset_distribution` default would be "random_seed_offset" (i.e. like `horovod_dataset_distribution="random_seed_offset"`), which is the current behavior of PyTorch distributed training. (We could change the default via a new behavior version if we want to...)

(Also note, similar comment as I made in #1612: There are some implicit assumptions here: That the worker index and rank is static. This might not always be the case. But it might be possible to just update the shard index / num shards dynamically for the next sub-epoch. Just to keep this in mind. I don't think we need to take care of this now.)

NeoLegends:
I think before merging this should get a dedicated test around the sharding.

Closes #1634


albertz:
> I think before merging this should get a dedicated test around the sharding.

Well then this should be a draft for now?

NeoLegends:
@albertz Do you think this needs a test around the config processing?

returnn/datasets/basic.py
```diff
@@ -249,6 +249,59 @@ def __reduce__(self):
...
+            return 0, 1
+
+    @staticmethod
+    def _get_default_shard_config():
+        """
+        :return: default shard index and number of shards based on the global config
+        """
+        from returnn.config import get_global_config
+
+        config = get_global_config(raise_exception=False)
```
albertz, line 284: I don't like that we access the global config here. I know this follows similar code as `_get_default_random_seed_offset` but I also don't like it there. Why is this needed? This should come from outside, or not? Specifically at the place where we call `init_dataset`. E.g. in the `__main__`. There we also call `Dataset.kwargs_update_from_config`.

Also, the code is wrong. Distributed training is only one possible source which defines/influences the shard index and num shards. But there are other reasons, for example the MultiProcDataset, or PyTorch DataLoader num_workers.

albertz:
> @albertz Do you think this needs a test around the config processing?

I'm not exactly sure what you mean by that.

But e.g. there could be a test with PyTorch DataLoader num_workers=2 which checks that all data from the dataset was properly covered.

returnn/datasets/basic.py
NeoLegends: > MultiProcDataset

This is already setting the num_shards and shard index for its children. The code always was designed in such a way that it would only look at the global config in case there was no value already set. But I agree, now it's better.

> PyTorch DataLoader num_workers

I think this is actually not that trivial to implement because the torch engine already is given initialized datasets and it's difficult to change the sharding config after having initialized a dataset. So factoring in the PyTorch num_workers needs to be done during the initial dataset initialization, which messes up the code a bit, I feel. Do you know a good way? Maybe this is fine after all.

NeoLegends: I think we cannot achieve both 
> But e.g. there could be a test with PyTorch DataLoader num_workers=2 which checks that all data from the dataset was properly covered.

and

> I know this follows similar code as _get_default_random_seed_offset but I also don't like it there. Why is this needed? This should come from outside, or not? 

because of

> I think this is actually not that trivial to implement because the torch engine already is given initialized datasets and it's difficult to change the sharding config after having initialized a dataset

However, I think it's worth it to have proper support for `torch_dataloader_opts = {"num_workers": n}` with `n > 1` because this makes it much simpler for the end user to have multi-process data loading and this feature can replace `MultiProcDataset` for simple use cases. So I think I need to revert back on the changes where `num_shards` and `shard_index` are set from the outside, and rather fetch them from inside the dataset, when they are needed. At that point the torch worker info is also available, which means that we can take it into account properly.

NeoLegends: Hm, I was originally not a fan of the mutability of these properties, but it seems ok now.

returnn/datasets/multi_proc.py
```diff
@@ -144,7 +148,11 @@ def _lazy_init(self):
                     self._sharding_method,
                 )
             elif self._sharding_method == "dedicated":
-                sub_dataset = {**self.dataset, "_num_shards": self.num_workers, "_shard_index": i}
+                sub_dataset = {
+                    **self.dataset,
+                    "num_shards": self.num_workers * self.num_shards,
+                    "shard_index": (self.shard_index * self.num_workers) + i,
+                }
```
NeoLegends, line 155: This is something that I overlooked before. We can trivially allow sharding for `MultiProcDataset` in this case.

NeoLegends:
@albertz Can you give another review for this PR?

returnn/datasets/basic.py
```diff
@@ -170,9 +175,9 @@ def __init__(
         self._chunking = chunking
         self.chunk_size, self.chunk_step, self.custom_chunking_func = self._parse_chunking(chunking)
         self._context_window = context_window
-        assert 0 <= _shard_index < _num_shards
-        self._num_shards = _num_shards
-        self._shard_index = _shard_index
+        assert 0 <= shard_index < num_shards
+        self.num_shards = num_shards
+        self.shard_index = shard_index
```
Icemole, line 180:
```suggestion
        self.set_shard_idx_and_num_shards(shard_index, num_shards)
```
Slightly cleaner and reuses code, but it's also fine as is. Your choice.

returnn/datasets/distrib_files.py
```diff
@@ -494,38 +485,33 @@ def get_data_keys(self) -> List[str]:
...
-    config = get_global_config(raise_exception=False)
-    if not config:
-        return 0, 1
-    if config.typed_value("torch_distributed") is not None:
-        import returnn.torch.distributed
+        if distrib_shard_files and dd_cfg is None:
+            # RETURNN will set sharding info on the dataset if the global config is set.
+            # If it's not set, however, we need to respect the existing `distrib_shard_files` property
+            # for backwards compatibility and load the sharding info ourselves.
+            return CachedDataset2._get_sharding_rank_and_size(config)
```
Icemole, line 507: Shouldn't you call the local `_get_rank_and_size` as per the previous change in https://github.com/rwth-i6/returnn/pull/1676/files#diff-44adfde5339aa94bf0770b09138330d5ea06d6b8e3f3b975bf270058f8c0c4baL188?

returnn/datasets/multi_proc.py
Icemole: I might not know the context of this comment. Isn't sharding already allowed as per your change here?

returnn/torch/data/returnn_dataset_wrapper.py
```diff
@@ -41,8 +41,22 @@ class ReturnnDatasetResetMpSharedEpochCallback:
...
+        self._sharding_config_set = False
 
     def __call__(self):
+        # Include local worker rank in the dataset sharding config
+        if not self._sharding_config_set:
+            worker_info = torch.utils.data.get_worker_info()
+            if worker_info is not None:
+                assert self.dataset.supports_sharding() or worker_info.num_workers == 1, (
+                    f"Using {worker_info.num_workers} torch data loading workers "
+                    "but dataset does not support sharding. This will result in repeated training data."
```
Icemole, line 53: If I understood correctly, maybe you can point to using the `MultiProcDataset` implementation here?

Icemole (commented):
Just a few minor comments from my side. The functionality looks good!

albertz:
To clarify: this also fixes #1678?

NeoLegends:
Yes.

returnn/datasets/basic.py
NeoLegends: That function has additional asserts that rely on correct initialization. Not for now.

albertz:
Sorry for introducing the small conflict, but my change should fix #1678 and #1737 already, and shouldn't really cause any issues to merge with the PR here.

albertz:
(Summarize) Added feature: when `torch.utils.data.DataLoader` is used with `num_workers>1`, this will set the sharding accordingly.

Btw, some questions regarding this:

What happens when this is used together with distributed training? Will it set num_shards = distrib_world_size * dataloader_num_workers then?

Is the order of seqs you get from the `DataLoader` deterministic?

Will it always be complete? E.g. if one worker returns more seqs than the other (e.g. total num seqs is 11, and 2 workers), will the `DataLoader` finish only until all the workers have finished?

returnn/torch/data/returnn_dataset_wrapper.py
```diff
@@ -41,8 +41,22 @@ class ReturnnDatasetResetMpSharedEpochCallback:
...
+                )
+                self.dataset.set_shard_idx_and_num_shards(
+                    self.dataset.shard_index + worker_info.id, self.dataset.num_shards * worker_info.num_workers
+                )
```
albertz, line 57: I don't understand. Why does this consider the existing shard_index/num_shards? I would expect that you overwrite those here.
```suggestion
                self.dataset.set_shard_idx_and_num_shards(worker_info.id, worker_info.num_workers)
```

returnn/datasets/basic.py
albertz: > That function has additional asserts that rely on correct initialization. Not for now.

I don't understand the comment. What correct initialization? What do you mean by "not for now"?

returnn/datasets/basic.py
```diff
@@ -68,6 +67,12 @@ def set_or_remove(key, value):
         set_or_remove("min_chunk_size", config.opt_typed_value("min_chunk_size", 0) or None)
         set_or_remove("chunking_variance", config.float("chunking_variance", 0))
 
+        dd_cfg = config.typed_value("dataset_distribution", "random_seed_offset")
+        assert dd_cfg in ["random_seed_offset", "shard"]
+        shard_index, num_shards = Dataset._get_sharding_rank_and_size(config) if dd_cfg == "shard" else 0, 1
+        set_or_remove("num_shards", num_shards)
+        set_or_remove("shard_index", shard_index)
```
albertz, line 74: I'm not sure whether it is a good idea to do this logic in `kwargs_update_from_config`. You don't want to apply this just for every dataset. E.g. for the dev/test/eval datasets, we would not want this logic, at least not right now.

I think this should come from the outside, not from within.

returnn/datasets/multi_proc.py
albertz: This is when sharding was set on the MultiProcDataset itself, and then MultiProcDataset also uses sharding on the subdatasets.

albertz:
The `_get_random_seed_for_epoch`, shouldn't it also consider `num_shards`/`shard_index`? Or only in the case of `dataset_distribution=="dataset_distribution"`?

albertz:
(Summary) New global config option `dataset_distribution`, which can be either set to `"random_seed_offset"` (default) or `"shard"`. This is for distributed training. "shard" will enable sharding for the dataset, so on N GPUs, processing one full epoch will only go through the data once, unlike with "random_seed_offset", where one full epoch sees all the data N times (each worker with different random seed).

returnn/torch/data/returnn_dataset_wrapper.py
albertz: We could also do the distributed logic directly here. We pass on the rank/size info anyway to the subproc children via the env var `_RETURNN_TORCH_DISTRIBUTED_INIT_INFO`.

The question is, how to handle `dataset_distribution`. We could also just check the global config here at this point (even though I don't like accessing the global config too much... maybe I get a better idea). Then all the logic about what sharding options to set (or whether to set it at all) is all here in this place, together with the dataloader num workers.

albertz, line 57: Instead of having all this logic here, we maybe should better move it to `ReturnnDatasetIterDataPipe.reset` though?

returnn/datasets/distrib_files.py
```diff
@@ -172,21 +171,16 @@ def __init__(
...
-        if distrib_shard_files:
-            assert self._num_shards == 1 and self._shard_index == 0, (  # ensure defaults are set
-                f"{self}: Cannot use both dataset-sharding via properties _num_shards and _shard index "
-                f"and {self.__class__.__name__}'s own sharding implementation based on the trainings rank and size."
+        if distrib_shard_files is not None:
+            log.print_deprecation_warning(
+                f"{self.__class__.__name__}' `distrib_shard_files` config option is set. "
+                "Use global config option `dataset_distribution` instead "
+                "for the same behavior across more types of datasets."
             )
```
albertz, line 179: I'm not sure this is really necessary to mark this as deprecated.
```suggestion
```

returnn/datasets/distrib_files.py
```diff
@@ -153,10 +152,10 @@ def __init__(
         :param get_sub_epoch_dataset: callable which returns a dataset dict for a given subset of files
         :param preload_next_n_sub_epochs: how many sub epoch datasets to preload
         :param buffer_size: buffer size for each worker, amount of seqs to prefetch
-        :param distrib_shard_files: set to true to shard the data across worker processes in
-            distributed training scenaria
+        :param distrib_shard_files: deprecated. Replaced by global config option ``dataset_distribution="shard"``.
+
+            Set to true to shard the data across worker processes in distributed training scenaria.
```
albertz, line 157: I don't think this is necessary to mark it as deprecated.
```suggestion
        :param distrib_shard_files: deprecated. set to true to shard the data across worker processes in
            distributed training scenaria
```

PR #1677, opened 2025-01-17
Add option to passthrough num_seqs in PostprocessingDataset

dorian-K:
I'm open to adjusting the name of the variable

My use case for this is to get Laplace Ordering on a Dataset, so map_seq_stream wouldn't drop any sequences.

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -88,6 +88,7 @@ class PostprocessingDataset(CachedDataset2):
         map_seq: Optional[Callable] = None,
         map_seq_stream: Optional[Callable] = None,
         map_outputs: Optional[Dict[str, Any]] = None,
+        iterator_preserves_sequences: Optional[bool] = None,
         **kwargs,
     ):
         """
@@ -111,6 +112,8 @@ class PostprocessingDataset(CachedDataset2):
             To simplify the common case when no shapes change, this value can be left unspecified. The dataset then
             assumes the same data layout as returned by the wrapped dataset.
             Example: `map_outputs={"data": {"dim": 42}}`
+        :param iterator_preserves_sequences: whether the function in map_seq_stream preserves the number of sequences,
+            i.e. for every input sequence there is exactly one output sequence.
         :param kwargs: see :class:`CachedDataset2`, :class:`Dataset`
         """
         super().__init__(**kwargs)
@@ -121,10 +124,13 @@ class PostprocessingDataset(CachedDataset2):
             raise ValueError(f"{self}: need to either set map_seq or map_seq_stream")
         if map_seq and map_seq_stream:
             raise ValueError(f"{self}: cannot set both map_seq and map_seq_stream")
+        if map_seq and iterator_preserves_sequences is not None:
+            raise ValueError(f"{self}: iterator_preserves_sequences is only allowed with map_seq_stream")
 
         self._dataset_def = dataset
         self._map_seq = map_seq
         self._map_seq_stream = map_seq_stream
+        self._iterator_preserves_sequences = iterator_preserves_sequences or False
         self._map_outputs = map_outputs
         self._rng = RandomState(self._get_random_seed_for_epoch(0))
         self._seq_list_for_validation: Optional[List[str]] = None
@@ -188,7 +194,7 @@ class PostprocessingDataset(CachedDataset2):
         self._dataset.init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
         self._data_iter = enumerate(self._build_mapping_iter())
         self._seq_list_for_validation = seq_list
-        if self._map_seq_stream is None:
+        if self._map_seq_stream is None or self._iterator_preserves_sequences:
             # If we don't have an iterable mapper we know the number of segments exactly
             # equals the number of segments in the wrapped dataset
             try:
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -232,11 +232,12 @@ class PostprocessingDataset(CachedDataset2):
             try:
                 loaded_seq_idx, tensor_dict = next(self._data_iter)
                 self._data_iter_produced_num_seqs += 1
-                assert (
-                    not self._map_seq_stream_preserves_num_seqs or self._data_iter_produced_num_seqs <= self._num_seqs
-                ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded more seqs than expected"
+                if self._map_seq_stream_preserves_num_seqs and self.num_seqs is not None:
+                    assert (
+                        self._data_iter_produced_num_seqs <= self._num_seqs
+                    ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded more seqs than expected"
             except StopIteration:
-                if self._map_seq_stream_preserves_num_seqs:
+                if self._map_seq_stream_preserves_num_seqs and self.num_seqs is not None:
                     assert (
                         self._data_iter_produced_num_seqs == self._num_seqs
                     ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded {self._data_iter_produced_num_seqs} seqs, while {self._num_seqs} were expected"
@@ -341,13 +342,13 @@ class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
 
     To be composed with any custom data postprocessing logic via :class:`Sequential`.
     """
+    preserves_num_seqs = True
 
     def __init__(self, num_seqs_per_bin: int, length_key: str = "data"):
         """
         :param num_seqs_per_bin: number of segments in a single laplace bin.
         :param length_key: data key to determine the segment length from for ordering.
         """
-        self.preserves_num_seqs = True
         self.length_key = length_key
         assert num_seqs_per_bin > 0
         self.num_seqs_per_bin = num_seqs_per_bin
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -204,10 +204,7 @@ class PostprocessingDataset(CachedDataset2):
             try:
                 self._num_seqs = self._dataset.num_seqs
             except NotImplementedError:
-                # some datasets don't know their num_seqs
-                assert (
-                    not self._map_seq_stream_preserves_num_seqs
-                ), "_map_seq_stream_preserves_num_seqs is True, but wrapped dataset does not know its num_seqs"
+                pass  # some datasets don't know their num_seqs
         return True
 
     def get_current_seq_order(self):
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -88,7 +88,7 @@ class PostprocessingDataset(CachedDataset2):
         map_seq: Optional[Callable] = None,
         map_seq_stream: Optional[Callable] = None,
         map_outputs: Optional[Dict[str, Any]] = None,
-        iterator_preserves_sequences: Optional[bool] = None,
+        map_seq_stream_preserves_num_seqs: Optional[bool] = None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


NeoLegends (commented):
Hah, I implemented the exact same thing about a month ago. Convergent evolution at work!

I don't think that this is a very good idea though, because it invites buggy behavior if your mapper function doesn't actually cohere to the guarantees you made by specifying `iterator_preserves_sequences`. We could catch this using an assert once the underlying dataset has ran out of seqs (comparing actual vs. forwarded `num_seqs`), but in case there is a bug this assertion would trigger only at the very end of a subepoch, which would be quite annoying.

If you simply want to do laplace ordering there is the `LaplaceOrdering` combinator in this very module that you can compose with your postprocessor function using `Sequential` (also in this module). It uses a small buffer to do the ordering instead. It's efficient enough though as long as the buffer size doesn't grow too large.

albertz:
> We could catch this using an assert once the underlying dataset has ran out of seqs (comparing actual vs. forwarded `num_seqs`), but in case there is a bug this assertion would trigger only at the very end of a subepoch, which would be quite annoying.

Yes, we should definitely have such a check.

I don't think it's too annoying. The user should anyway hopefully know whether the `map_seq_stream` has this property, and usually, the user would probably be correct about it. I expect this to be a quite rare case, that the user sets this, but it was wrong. And for this rare case, it's totally fine. On the other side, this is a very simple solution here, and actually even better than the passthrough of `complete_frac` (`epoch_continuous`) (#1673), because the `complete_frac` is more incorrect.

returnn/datasets/postprocessing.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         map_seq: Optional[Callable] = None,
         map_seq_stream: Optional[Callable] = None,
         map_outputs: Optional[Dict[str, Any]] = None,
+        iterator_preserves_sequences: Optional[bool] = None,
```
albertz, line 91: "preserves sequences" is somewhat ambiguous. What exactly is preserved? I think "preserves num seqs" is more clear.
```suggestion
        num_seqs_is_preserved: Optional[bool] = None,
```

returnn/datasets/postprocessing.py
```diff
@@ -121,10 +124,13 @@ def __init__(
             raise ValueError(f"{self}: need to either set map_seq or map_seq_stream")
         if map_seq and map_seq_stream:
             raise ValueError(f"{self}: cannot set both map_seq and map_seq_stream")
+        if map_seq and iterator_preserves_sequences is not None:
```
albertz, line 127:
```suggestion
        if map_seq and iterator_preserves_sequences not in (True, None):
```

albertz (changes requested):
See my comments. Specifically, also add the mentioned check at the end of the epoch. Otherwise good.

returnn/datasets/postprocessing.py
```diff
@@ -188,7 +194,7 @@ def init_seq_order(
         self._dataset.init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
         self._data_iter = enumerate(self._build_mapping_iter())
         self._seq_list_for_validation = seq_list
-        if self._map_seq_stream is None:
+        if self._map_seq_stream is None or self._iterator_preserves_sequences:
             # If we don't have an iterable mapper we know the number of segments exactly
             # equals the number of segments in the wrapped dataset
```
NeoLegends, line 199: Ok, if you want to add this functionality then this comment should also be adapted :)

albertz:
I was wondering, maybe also the `LaplaceOrdering` class can have some class attrib `preserves_num_seqs = True`, and in `PostprocessingDataset`, we can add this:
```python
if map_seq_stream_preserves_num_seqs is None:
    map_seq_stream_preserves_num_seqs = getattr(map_seq_stream, "preserves_num_seqs", None)
```
That way, for some of our predefined postprocessings, we can already tell the dataset that it preserves the num seqs.

returnn/datasets/postprocessing.py
```diff
@@ -187,14 +194,18 @@ def init_seq_order(
...
+            # If we don't have an iterable mapper (or the user explicitly specifies this),
+            # we know the number of segments exactly equals the number of segments in the wrapped dataset
             try:
                 self._num_seqs = self._dataset.num_seqs
             except NotImplementedError:
-                pass  # some datasets don't know their num_seqs
+                # some datasets don't know their num_seqs
+                assert (
+                    not self._map_seq_stream_preserves_num_seqs
+                ), "_map_seq_stream_preserves_num_seqs is True, but wrapped dataset does not know its num_seqs"
```
albertz, line 208: I think `map_seq_stream_preserves_num_seqs` does not necessarily imply that the sub dataset knows the `num_seqs`, so I would not throw an error here.
```suggestion
                pass  # some datasets don't know their num_seqs
```

returnn/datasets/postprocessing.py
```diff
@@ -221,7 +232,15 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
         while True:
             try:
                 loaded_seq_idx, tensor_dict = next(self._data_iter)
+                self._data_iter_produced_num_seqs += 1
+                assert (
+                    not self._map_seq_stream_preserves_num_seqs or self._data_iter_produced_num_seqs <= self._num_seqs
+                ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded more seqs than expected"
```
albertz, line 240:
```suggestion
                if self._num_seqs is not None:
                    assert self._data_iter_produced_num_seqs <= self._num_seqs, (
                        f"{self}: map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded more seqs"
                        f" ({self._data_iter_produced_num_seqs}) than expected ({self._num_seqs})"
                    )
```

returnn/datasets/postprocessing.py
```diff
@@ -221,7 +232,15 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
...
             except StopIteration:
+                if self._map_seq_stream_preserves_num_seqs:
+                    assert (
+                        self._data_iter_produced_num_seqs == self._num_seqs
+                    ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded {self._data_iter_produced_num_seqs} seqs, while {self._num_seqs} were expected"
```
albertz, line 245:
```suggestion
                if self._num_seqs is not None:
                    assert self._data_iter_produced_num_seqs == self._num_seqs, (
                        f"{self}: map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded"
                        f" {self._data_iter_produced_num_seqs} seqs, while {self._num_seqs} were expected"
                    )
```

returnn/datasets/postprocessing.py
```diff
@@ -329,6 +350,7 @@ def __init__(self, num_seqs_per_bin: int, length_key: str = "data"):
         :param num_seqs_per_bin: number of segments in a single laplace bin.
         :param length_key: data key to determine the segment length from for ordering.
         """
+        self.preserves_num_seqs = True
```
albertz, line 353: This can be a class attrib, not an instance attrib.

returnn/datasets/postprocessing.py
dorian-K: I'm a bit torn because as a user, I would explicitly set this option because I want the dataset to set the property num_seqs and expect it to fail when it can't.
Ultimately I agree though, that based on the name of the attribute, this should not be an error

returnn/datasets/postprocessing.py
```diff
@@ -394,3 +416,8 @@ def __call__(self, arg: T, **kwargs) -> T:
         for func in self.funcs:
             arg = func(arg, **kwargs)
         return arg
+
+    @property
+    def preserves_num_seqs(self):
+        """:return: whether the composed functions all preserve the number of sequences"""
+        return all(getattr(f, "preserves_num_seqs", False) for f in self.funcs)
```
albertz, line 423: I wonder a bit, instead of returning `None` when it is unknown, this returns False. But not sure if this is relevant. This code here is certainly easier that way.

returnn/datasets/postprocessing.py
albertz: I can imagine that in some Sisyphus recipe, you would automatically have some code snippet which adds some Laplace (or some other custom user post-proc func) and just always set this flag `map_seq_stream_preserves_num_seqs=True` because the user knows that this post-proc func preserves num seqs, but then it would fail now for datasets where it is not known, so having this check just makes this more complicated and less generic.

returnn/datasets/postprocessing.py
dorian-K: `False` should always be a safe option, so I believe this is fine

albertz:
Black needs to be run again.

returnn/datasets/postprocessing.py
```diff
@@ -221,7 +231,16 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
...
+                self._data_iter_produced_num_seqs += 1
+                if self._map_seq_stream_preserves_num_seqs and self.num_seqs is not None:
+                    assert (
+                        self._data_iter_produced_num_seqs <= self._num_seqs
+                    ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded more seqs than expected"
             except StopIteration:
+                if self._map_seq_stream_preserves_num_seqs and self.num_seqs is not None:
+                    assert (
+                        self._data_iter_produced_num_seqs == self._num_seqs
+                    ), f"_map_seq_stream_preserves_num_seqs is True, but map_seq_stream yielded {self._data_iter_produced_num_seqs} seqs, while {self._num_seqs} were expected"
```
albertz, line 243: See my both comments above on how to change this.

albertz:
Also check PEP8 errors:
```
  returnn/datasets/postprocessing.py:115: WEAK WARNING Pep8CodeStyle: E501: line too long (123 > 120 characters)
  returnn/datasets/postprocessing.py:243: WEAK WARNING Pep8CodeStyle: E501: line too long (175 > 120 characters)
```

returnn/datasets/postprocessing.py
dorian-K: I'm not entirely sure what you are referring to

returnn/datasets/postprocessing.py
dorian-K: But this is wrong, the condition for `_map_seq_stream_preserves_num_seqs` should not be removed. I added `self._num_seqs is not None` in a later commit

returnn/datasets/postprocessing.py
dorian-K: Oh now I see, I didn't notice you also changed the strings

returnn/datasets/postprocessing.py
albertz: No, if `_num_seqs` is known, then it must match.

albertz: No matter what `_map_seq_stream_preserves_num_seqs` is.

dorian-K: Ah, I was thinking about the num_seqs of the wrapped dataset and not of the parent one. I shall add asserts for both

dorian-K:
I have also changed the message to show the actual value of `_map_seq_stream_preserves_num_seqs`, which may not necessarily be True if someone decided to mess with `_num_seqs` without setting the attribute

returnn/datasets/postprocessing.py
```diff
@@ -221,7 +231,20 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
         while True:
             try:
                 loaded_seq_idx, tensor_dict = next(self._data_iter)
+                self._data_iter_produced_num_seqs += 1
+                if self._num_seqs is not None:
+                    assert self._data_iter_produced_num_seqs <= self._num_seqs, (
+                        f"{self}: map_seq_stream yielded more seqs ({self._data_iter_produced_num_seqs}) "
+                        f"than expected ({self._num_seqs}). _map_seq_stream_preserves_num_seqs is set to "
```
albertz, line 238:
```suggestion
                        f"than expected ({self._num_seqs}). map_seq_stream_preserves_num_seqs is set to "
```

albertz: `_map_seq_stream_preserves_num_seqs` is only the internal name, which should not be relevant to the user. The user sets (or controls) `map_seq_stream_preserves_num_seqs`.

returnn/datasets/postprocessing.py
```diff
@@ -221,7 +231,20 @@ def _collect_single_seq(self, seq_idx: int) -> Optional[DatasetSeq]:
...
+                        f"{self._map_seq_stream_preserves_num_seqs}"
+                    )
             except StopIteration:
+                if self._num_seqs is not None:
+                    assert self._data_iter_produced_num_seqs == self._num_seqs, (
+                        f"{self}: map_seq_stream yielded {self._data_iter_produced_num_seqs} seqs, "
+                        f"while {self._num_seqs} were expected. _map_seq_stream_preserves_num_seqs is set to "
```
albertz, line 245:
```suggestion
                        f"while {self._num_seqs} were expected. map_seq_stream_preserves_num_seqs is set to "
```

albertz (approved):

PR #1681, opened 2025-01-26
Fix assert in LaplaceOrdering

dorian-K:
This assert wrongfully throws in an edge case: When the datasets' `num_seqs` is a multiple of `self.num_seqs_per_bin`, then the `seq_buffer` is replaced with an empty list in line 389, and on the next iteration `has_ended` will thus not be set because the for-loop is never entered, and this assert is triggered

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -382,7 +382,7 @@ class LaplaceOrdering(Callable[[Iterator[TensorDict]], Iterator[TensorDict]]):
                     has_ended = True
 
             if len(seq_buffer) < self.num_seqs_per_bin:
-                assert has_ended and not next_seq_buffer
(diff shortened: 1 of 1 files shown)


albertz (changes requested):
I think the new assert makes it more difficult to understand it. Can't we just move the `has_ended = False` out of the loop? I think it's important that we check at the end that at some point, we got a `StopIteration`.

dorian-K:
I agree with you, but on the other hand I don't see a situation where this assert is even useful in the first place. Maybe we should remove it alltogether? From reading the code, I believe it is quite clear that the only situation where the `seq_buffer` has less than `self.num_seqs_per_bin` elements, is when the iterator runs out.

albertz (approved):

albertz:
I think it's good that we check that we got `StopIteration`. It cannot hurt to have this check. I think it's good now.

NeoLegends (commented):
good catch!

PR #1684, opened 2025-02-03
fix build, publish wheels

dimbleby:
You have mixed up build-time and run-time requirements.  Direct invocation of setup.py is [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).  Build and publish wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,6 +82,10 @@ def main():
         packages=["returnn"],
         include_package_data=True,
         package_data={"returnn": package_data},  # filtered via MANIFEST.in
+        install_requires=[
+            "numpy",
+            "h5py",
+        ],
         description="The RWTH extensible training framework for universal recurrent neural networks",
         author="Albert Zeyer",
         author_email="albzey@gmail.com",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -82,6 +82,10 @@ def main():
         packages=["returnn"],
         include_package_data=True,
         package_data={"returnn": package_data},  # filtered via MANIFEST.in
+        install_requires=[
+            "numpy",
+            "h5py",
```
albertz, line 87: Note: h5py should be an optional dependency, not necessarily required. (If it is, consider this a bug.)
```suggestion
```

albertz (approved):
I have not used the `build` module before, but looks fine. Let's just try.
(Our build scripts are a bit custom in the way we determine the version. Let's see if this breaks.)

setup.py
dimbleby: In that case you should want to set it as an extra.

I'm gonna hit "accept" on the suggestion and leave it with you to fix that up.

albertz:
Thanks!

PR #1685, opened 2025-02-04
`DFDataset`: do not pickle sharding info

### LINKED ISSUE #1678: DistributeFilesDataset _num_shards issue
For the latest RETURNN, when I use DistributeFilesDataset, I have this error.

```
File "/nas/models/asr/am/multilingual/16kHz/2024-11-08--jxu-best-rq-pretrain/work/i6_core/tools/git/CloneGitRepositoryJob.LD5f1wKK7LPo/output/returnn/returnn/datasets/basic.py", line 227, in Dataset._create_from_reduce                     
  line: ds = cls(**kwargs)                                                                                                                     
  locals:                                                                                                                              
   ds = <not found>                                                                                                                        
   cls = <local> <class 'returnn.datasets.distrib_files.DistributeFilesDataset'>                                                                                          
   kwargs = <local> {'files': ['/ssd/jxu/nas/data/speech/FR_FR/16kHz/EPPS/corpus/batch.1.v1/hdf-raw_wav.16kHz.split-25/EPPS-batch.1.v1.hdf.15', '/ssd/jxu/nas/data/speech/EN_US/16kHz/NEWS.HQ/corpus/batch.2.NPR.v3/hdf-raw_wav.16kHz.split-261/NEWS.HQ-batch.2.NP
R.v3.hdf.7', '/ssd/jxu/nas/data/speech/IT_IT/16kHz/IT.parli..., len = 25                                                                                               
 File "/nas/models/asr/am/multilingual/16kHz/2024-11-08--jxu-best-rq-pretrain/work/i6_core/tools/git/CloneGitRepositoryJob.LD5f1wKK7LPo/output/returnn/returnn/datasets/distrib_files.py", line 171, in DistributeFilesDataset.__init__               
  line: assert self._num_shards == 1 and self._shard_index == 0, ( # ensure defaults are set                                                                                    
       f"{self}: Cannot use both dataset-sharding via properties _num_shards and _shard index "                                                                                
       f"and {self.__class__.__name__}'s own sharding implementation based on the trainings rank and size."                                                                          
     )                                                                                                                              
  locals:                                                                                                                              
   self = <local> <DistributeFilesDataset 'train' epoch=None>                                                                                                   
   self._num_shards = <local> 8                                                                                                                  
   self._shard_index = <local> 6 
```

The DistributeFilesDataset is inherited from CachedDataset2, which is again inherited from Dataset, the the _num_shards should be set to 1 in the [init function](https://github.com/rwth-i6/returnn/blob/44fde7a53cd7147dbc369809a6a3618c3d626996/returnn/datasets/basic.py#L114). I am not sure how self._num_shards is changed to num of gpus in my case.

NeoLegends:
... but rely on `_distrib_info` instead.

Fixes #1678

@Judyxujj feel free to give this PR a test run.

### DIFF
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -130,6 +130,8 @@ class DistributeFilesDataset(CachedDataset2):
     https://github.com/rwth-i6/returnn/issues/1524.
     """
 
+    _getnewargs_exclude_attrs = CachedDataset2._getnewargs_exclude_attrs.union(("_num_shards", "_shard_index"))
+
     def __init__(
         self,
         *,


albertz:
I don't understand. We actually want that this is pickled/serialized. It doesn't make sense to not serialize this? We want/need this information.

albertz:
How does this work together with #1676?

NeoLegends:
> I don't understand. We actually want that this is pickled/serialized. It doesn't make sense to not serialize this? We want/need this information.

The dataset already handles the pickling of this information on its own via `_distrib_info`. Alternatively we can remove that instead. I think in that case we would also remove the assertion. I went the other way now because it seems like a good-enough fix that keeps the existing assertion intact until #1676 lands.

> How does this work together with https://github.com/rwth-i6/returnn/pull/1676?

#1676 removes the assertion, so the change here won't be necessary anymore.

albertz:
So instead of removing/cleaning up the obsolete/duplicate code/logic with `_distrib_info`, you add another workaround? I think cleaning up makes much more sense than adding yet another workaround?

NeoLegends:
Superseeded by #1676

PR #1692, opened 2025-02-25
RF conv/pool/etc, stft, window, use_mask, new behavior version 23

### LINKED ISSUE #1691: Frontend: masking for more functions, global setting
For example, currently, we have already:

* `matmul`: `use_mask`
* `reduce` and co: `use_mask`
* `moments`: `use_mask`
* `BatchNorm.__init__`: `use_mask`
* `softmax`/`log_softmax`: `use_mask`
* `pad`: `handle_dynamic_dims`
* `reverse_sequence`: `handle_dynamic_dims`

But such an option is missing for, and also not currently used for:

* `Conv1d.__init__` and co, `conv`
* `TransposedConv1d.__init__` an co, `transposed_conv`
* `pool`, `max_pool`, `pool1d` and co
* `stft`
* `window`
* `scatter`
* `sort`, ~`top_k`~ (masking is already handled), `search_sorted`
* `concat`

We have use cases for `conv`/`transposed_conv`/`pool`/`stft`, so it makes sense to implement it for them. `sort`/`top_k`, `search_sorted` is also not so trivial (maybe needs to be similar to `reduce`) but not so important for now.

However, when implementing it, what should be the default? We should not change the behavior of existing setups. So having the default `use_mask=True` would require a new behavior version. But even then, I'm not sure it is really always wanted, as there is some performance hit, and specifically for training, this is maybe not always wanted. So, despite having the default controlled via behavior version, maybe we also should introduce a new global flag? How should this be called? `rf_use_mask`? Other such global options are called eg. `rf_att_dropout_broadcast` and `rf_dropout_broadcast`.

But such `rf_use_mask`, does it make sense to have this shared for all functions? E.g. for `matmul`, `softmax`, `reduce`, `pad`, it's probably never really a good idea? For `BatchNorm.__init__`, it's maybe debatable (in the community, masking is mostly not done). `moments` unclear. For `conv` and co, this option makes sense (in the community. masking is mostly not done).

One problem is maybe, when we forget about one function now, or maybe later introduce another function where we don't properly handle masking, and then again later want to also add some `use_mask` option for this, again we need a new behavior version then to introduce it for this function. But maybe that's not really a problem.

So I would suggest now to add and implement `use_mask` for these:

* `Conv1d.__init__` and co, `conv`
* `TransposedConv1d.__init__` an co, `transposed_conv`
* `pool`, `max_pool`, `pool1d` and co
* `stft`
* `window`
* `scatter`

Basically it just means to add a `copy_masked` somewhere at the end. Maybe we can even do that in the frontend code, such that there is no special handling needed in every backend.

(For the others missing functions, it's not so straightforward. But also, at least I don't really need those at the moment. We should just make sure that there is at least some check for it, which fails with some error then.)

And then we can introduce `rf_use_mask`, which would for now control all those, and then also `BatchNorm.__init__`.

~And then also fix `merge_dims`, as discussed below ([here](https://github.com/rwth-i6/returnn/issues/1691#issuecomment-2684333551)).~ (Fixed independently via #1695.)

(cc @NeoLegends, @dorian-K)

albertz:
`use_mask` option implemented for:

* `Conv1d.__init__` and co, `conv`
* `TransposedConv1d.__init__` an co, `transposed_conv`
* `pool`, `max_pool`, `pool1d` and co
* `stft`
* `window`
* `scatter`

`rf_use_mask: bool` config option to explicitly specify the default for `use_mask`.

If `rf_use_mask` is not specified, for older behavior versions (<=22), we use `use_mask=False` by default, with behavior version >=23, we use `use_mask=True` by default. Thus this introduces the new behavior version 23.

`BatchNorm` `use_mask` is not required to be set now. If it is not explicitly set, it is True by default. (Previously, it was required to set this when there were dynamic dims, so this change does not change the behavior of any existing setup.)

Fix #1691

### DIFF
--- a/docs/configuration_reference/behavior_version.rst
+++ b/docs/configuration_reference/behavior_version.rst
@@ -22,6 +22,21 @@ and not listing legacy/deprecated parameters.
 Version History
 ---------------
 
+Behavior version 23 (2025-02-25)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+RF ``Conv1d.__init__`` and co, ``conv``,
+``TransposedConv1d.__init__`` an co, ``transposed_conv``,
+``pool``, ``max_pool``, ``pool1d`` and co,
+``stft``, ``window`` default changed:
+
+* ``use_mask``: False → True
+
+There is also the global config option ``rf_use_mask: bool`` to overwrite the global default
+for these and also additionally for ``BatchNorm.__init__``.
+
+See issue `#1691 <https://github.com/rwth-i6/returnn/issues/1691>`__.
+
 Behavior version 22 (2025-01-22)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/returnn/frontend/array_.py
+++ b/returnn/frontend/array_.py
@@ -791,6 +791,7 @@ def scatter(
     mode: str = "sum",
     fill_value: Optional[Union[int, float]] = None,
     out_dim: Optional[Union[Dim, Sequence[Dim]]] = None,
+    use_mask: Optional[bool] = None,
 ) -> Tensor:
     """
     Scatters into new zero-tensor.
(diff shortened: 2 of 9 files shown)


albertz:
We need to be careful. It seems the changes introduce some nans (test_conformer, test_e_branchformer). Maybe the masked infs are triggering these? And then also some other seemingly unrelated tests fail now (test_pad_packed_batched).

PR #1697, opened 2025-02-27
PT: add randomization to bucket batching

NeoLegends:


returnn/torch/data/pipeline.py
```diff
@@ -337,17 +337,29 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
...
     ):
         """
         :param dataset: dataset to apply bucket batching to
         :param buckets: Bucket configuration as tuples of seq length and max number of seqs in that bucket.
             Segments longer than the largest size limit configured in the buckets are dropped. To avoid dropping
             any segments make sure your largest bucket allows segments larger than your longest training segment.
         :param length_key: data key to take as length measure
+        :param random_bucket_prob: Probability of putting a segment not into the best-fitting bucket, but into
+            a randomly chosen still-fitting bucket.
+
```
albertz, line 355:
```suggestion
```

returnn/torch/data/pipeline.py
```diff
@@ -383,6 +403,21 @@ def __iter__(self):
...
+        Sets the seed for the next invocation of ``__iter__``, for compatibility with
+        ``torch.utils.data.graph_settings.apply_random_seed``.
+        """
+        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
+        return self
+
+    def reset(self):
+        """resets the internal state of the data pipe"""
+        if self._seed is None:
+            self._seed = int(torch.empty((), dtype=torch.int32).random_().item())
```
albertz, line 417: I wonder, `torch.utils.data.datapipes.iter.Shuffler` uses `int(torch.empty((), dtype=torch.int64).random_().item())`, i.e. 64 bit. Why do you restrict this to 32 bit now? I'm not sure it really matters much for us, but if there is no good reason for this, I would keep it as 64 bit.

returnn/torch/data/pipeline.py
```diff
@@ -337,17 +337,31 @@ class BucketOrderingIterDataPipe(torch.utils.data.IterDataPipe):
...
+            a randomly chosen still-fitting bucket.
+            This increases seq length variation within the buckets at the cost of slighly more padding.
+        :param seed: random seed
         """
         self._dataset = dataset
         self._length_key = length_key
+        assert random_bucket_prob >= 0.0
+        self._random_bucket_prob = random_bucket_prob
+        self._rng = numpy.random.RandomState()
+        self._seed = seed % (2**32) if seed is not None else None
```
albertz, line 364: Same comment here on 32bit vs 64bit. It depends on what `numpy.random.RandomState` expects.

returnn/torch/data/pipeline.py
NeoLegends: Numpy’s RandomState can only be seeded with numbers up to 2^32.

returnn/torch/data/pipeline.py
albertz: Ah I just read, `numpy.random.RandomState` expects a seed between 0 and 2**32 - 1. Ok then.

returnn/torch/data/pipeline.py
albertz: Ok, it's fine because we need it between 0 and 2**32 - 1.

albertz, line 417: We require the seed between 0 and 2**32 - 1.
```suggestion
            self._seed = int(torch.empty((), dtype=torch.uint32).random_().item())
```

returnn/torch/data/pipeline.py
```diff
@@ -383,6 +403,21 @@ def __iter__(self):
...
+        Sets the seed for the next invocation of ``__iter__``, for compatibility with
+        ``torch.utils.data.graph_settings.apply_random_seed``.
+        """
+        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
+        return self
+
+    def reset(self):
+        """resets the internal state of the data pipe"""
+        if self._seed is None:
+            self._seed = int(torch.empty((), dtype=torch.uint32).random_().item())
```
albertz, line 417: Well I did not try this, but now I tried: `module 'torch' has no attribute 'uint32'`
```suggestion
            self._seed = int(2**31 + torch.empty((), dtype=torch.int32).random_().item())
```

NeoLegends: well in that case I'd rather use `abs()`

returnn/torch/data/pipeline.py
```diff
@@ -383,6 +403,21 @@ def __iter__(self):
...
+        Sets the seed for the next invocation of ``__iter__``, for compatibility with
+        ``torch.utils.data.graph_settings.apply_random_seed``.
+        """
+        self._seed = abs(seed) % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
+        return self
+
+    def reset(self):
+        """resets the internal state of the data pipe"""
+        if self._seed is None:
+            self._seed = abs(int(torch.empty((), dtype=torch.int32).random_().item()))
```
albertz, line 417: Now you are restricting yourself to 0 to 2**31, while you could get 0 to 2**32-1 by doing this:
```suggestion
            self._seed = int(2**31 + torch.empty((), dtype=torch.int32).random_().item()))
```

returnn/torch/data/pipeline.py
```diff
@@ -542,13 +577,13 @@ def set_seed(self, seed: int) -> ShufflingDataPipe:
         Sets the seed for the next invocation of ``__iter__``, for compatibility with
         ``torch.utils.data.graph_settings.apply_random_seed``.
         """
-        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
+        self._seed = abs(seed) % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
```
albertz, line 580: No need for the `abs`.
```suggestion
        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
```

returnn/torch/data/pipeline.py
```diff
@@ -542,13 +577,13 @@ def set_seed(self, seed: int) -> ShufflingDataPipe:
...
         return self
 
     def reset(self):
         """resets the internal state of the data pipe"""
         if self._seed is None:
-            self._seed = int(torch.empty((), dtype=torch.int32).random_().item())
+            self._seed = abs(int(torch.empty((), dtype=torch.int32).random_().item()))
```
albertz, line 586:
```suggestion
            self._seed = int(2**31 + torch.empty((), dtype=torch.int32).random_().item())
```

returnn/torch/data/pipeline.py
```diff
@@ -383,6 +403,21 @@ def __iter__(self):
     def __getitem__(self, index):
         raise Exception(f"{self.__class__.__name__}.__getitem__ is not supported")
 
+    def set_seed(self, seed: int) -> BucketOrderingIterDataPipe:
+        """
+        Sets the seed for the next invocation of ``__iter__``, for compatibility with
+        ``torch.utils.data.graph_settings.apply_random_seed``.
+        """
+        self._seed = abs(seed) % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
```
albertz, line 411:
```suggestion
        self._seed = seed % (2**32)  # seed must be within [0, 2**32) for seeding RandomState
```

returnn/torch/data/pipeline.py
albertz: But then you lose again one bit of entropy.

albertz (approved):

albertz:
Can you maybe comment on some good settings here?

PR #1698, opened 2025-03-01
PT: also use `complete_frac` for progress reporting

NeoLegends:
After this PR, the torch engine will use `complete_frac` for progress reporting as well as `num_seqs`.

Example line: 
```
ep 1 train, step 23, ctc_8 10.797, ctc_16 10.016, ce 8.758, fer 0.979, grad_norm:p2 85.874, num_seqs 34, max_size:time 239629, max_size:out-spatial 80, mem_usage:cuda:0 60.0GB, 0.926 sec/step, elapsed 0:00:55, exp. remaining 4:59:46, complete 0.31%
```

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -632,6 +632,7 @@ class Engine(EngineBase):
             step_idx = 0
             num_seqs = None
             last_seq_idx = 0
+            eval_start_time = time.monotonic()
 
             report_prefix = (f"ep {self.epoch} {dataset_name} eval",)
             with torch.no_grad():
@@ -657,6 +658,8 @@ class Engine(EngineBase):
                     )
 
                     self._run_step(extern_data, train_func=True)
+                    step_end_time = time.monotonic()
+
                     train_ctx = rf.get_run_ctx()
 
                     losses_dict = NumbersDict(
@@ -683,6 +686,7 @@ class Engine(EngineBase):
                         num_seqs=num_seqs,
                         seq_idx=last_seq_idx,
                         complete_frac=complete_frac,
+                        start_elapsed=step_end_time - eval_start_time,
                         log_memory_usage_device=self._device if self._log_memory_usage else None,
                     )
                     step_idx += 1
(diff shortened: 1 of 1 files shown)


returnn/torch/engine.py
```diff
@@ -1475,17 +1498,23 @@ def _print_process(
             info += ["%.3f sec/step" % step_duration]
         if start_elapsed is not None:
             info += ["elapsed %s" % hms(start_elapsed)]
-        if num_seqs is not None:
-            assert seq_idx is not None and start_elapsed is not None  # unexpected combination...
+        if complete_frac is not None:
+            complete = complete_frac
+        elif num_seqs is not None:
```
albertz, line 1503: Do we really need to have this logic here? How can it be that `complete_frac` is None but `num_seqs` is known? I think, if `num_seqs` is known, `complete_frac` should never be None. This should be handled properly anywhere in the pipeline, or at least e.g. in `ReturnnDatasetIterDataPipe`.

returnn/torch/engine.py
```diff
@@ -1444,6 +1466,7 @@ def _print_process(
     start_elapsed: Optional[float] = None,
     seq_idx: Optional[int] = None,
     num_seqs: Optional[int] = None,
+    complete_frac: Optional[float] = None,
```
albertz, line 1469: Do we need `num_seqs` at all when we have `complete_frac`?

NeoLegends: Right. I think this logic applies to some more places as well.

returnn/torch/engine.py
```diff
@@ -1663,27 +1627,3 @@ def _get_total_grad_norm(model: torch.nn.Module, p: float) -> float:
             p=p,
         ).item()
     )
-
-
-def _get_num_seqs_last_seq_idx(
```
NeoLegends, line 1639: I completely removed this function, because we only access `num_seqs` once and just for printing it to the log, so I don't think the assertions add enough value to warrant the additional verbosity. Is that ok with you or do you prefer to keep this around?

albertz: It's ok.

albertz (approved):

PR #1704, opened 2025-03-31
Add tensorboard to torch engine

robin-p-schmitt:
This adds tensorboard support to the torch engine. right now, it just uses the default directory name "runs" to dump the data to. also, it just dumps all scores from `eval_info` into the tensorbard. since we might want to discuss these two decisions, i marked this as a draft request.


returnn/torch/engine.py
```diff
@@ -20,6 +20,7 @@
 from torch.utils.data import DataLoader
 from torch import autocast
 from torch.cuda import amp
+from torch.utils.tensorboard import SummaryWriter
```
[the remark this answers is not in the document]
robin-p-schmitt: ah, yes, makes sense. should i just introduce a config option `use_tensorboard` and check whether this is true for all the changes i made to the code?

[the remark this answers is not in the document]
albertz: > should i just introduce a config option `use_tensorboard`

Yes.



> check whether this is true

Just check `if self. _tensorboard_writer` or so.

albertz:
Note: You can rebase to master now. That should fix the CI issues. (But despite that, see also my other comment.)

NeoLegends:
The diff is still huge and contains things already in master, maybe not rebase but merge master again.

robin-p-schmitt:
hey, can i merge this now?

returnn/torch/engine.py
```diff
@@ -695,12 +709,18 @@ def eval_model(self, *, skip_already_evaluated: bool = False):
...
                     step_idx += 1
 
             assert step_idx > 0, f"No data in dataset {dataset_name!r}."
             accumulated_losses_dict = accumulated_losses_dict / accumulated_inv_norm_factors_dict
             accumulated_losses_dict = self._maybe_extend_losses_info(accumulated_losses_dict)
 
+            if self._tensorboard_writer:
+                # write losses/errors to tensorboard
+                for key, val in accumulated_losses_dict.items():
+                    self._tensorboard_writer.add_scalar(f"{dataset_name}/{key}", val, global_step=self.epoch)
```
albertz, line 722: I wonder whether we should use `self.global_train_step` here instead.
```suggestion
                    self._tensorboard_writer.add_scalar(f"{dataset_name}/{key}", val, global_step=self.global_train_step)
```

We did it like that for TF. Also to have it comparable to the train scores. But I was also not so sure about that.

Other opinions? @NeoLegends? @Atticus1806? @JackTemaki? @vieting? @michelwi?

albertz: There was overall not really much feedback on this question. So I would go with `global_train_step` here, to have it consistent to the training.

returnn/torch/engine.py
```diff
@@ -506,6 +516,10 @@ def train_epoch(self):
                     batch_size_info=_get_batch_size_info(extern_data) if self._log_batch_size else None,
                     log_memory_usage_device=self._device if self._log_memory_usage else None,
                 )
+                if self._tensorboard_writer:
+                    # write losses/errors to tensorboard
+                    for key, val in eval_info.items():
+                        self._tensorboard_writer.add_scalar(f"train/{key}", val, global_step=self.global_train_step)
```
albertz, line 522: There was some discussion on the needed disk space (and maybe also IO is an issue). It can become quite large.

Maybe we should do this only for every N steps? So some extra check `and self.global_train_step % N == 0` or so. And `N` maybe configurable?

robin-p-schmitt: yes. should we then have `use_tensorboard` and an additional `tensorboard_opts` in the config?

albertz: Yes. But there maybe should also be a reasonable default for N? E.g. 100 or so? Or 1000?

albertz:
Again (the unrelated) #1770... I will look into that.

Issue #1675, opened 2025-01-16
`FileNotFoundError` when updating mtime of files in file cache


NeoLegends:
Sadly only a part of the error message remains, and the rest of the log is lost:

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/nas/models/asr/[...]/work/i6_core/tools/git/CloneGitRepositoryJob.LD5f1wKK7LPo/output/returnn/returnn/util/file_cache.py", line 479, in run
    os.utime(filename, None)
FileNotFoundError: [Errno 2] No such file or directory
```

First we should find out which file it is failing to touch.

NeoLegends:
@Icemole If this occurs again, kindly post the full log here. :)

albertz:
> First we should find out which file it is failing to touch.

I'm not sure that this really helps us here. But certainly having the full log would be helpful.

Btw, why is `filename` not printed? It should via #1674, as now it should use `better_exchook` in the thread, or not?

NeoLegends:
I think the log came from befire #1674.

hannah220:
I encountered the same issue:
```
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()FileCache: Copy file /nas/data/speech/IT_IT/16kHz/AMARA-crawl/corpus/batch.1.v2/hdf-ffnn_monophone_dense_eow_20250218.alignment.split-5/AMARA-crawl-batch.1.v2.hdf.4 to cache

  File "/nas/models/asr/am/IT_IT/16kHz/2025-03-04-hyoshimochi-conformer-mbw-ffnn-alignments/work/i6_core/tools/git/CloneGitRepositoryJob.LD5f1wKK7LPo/output/returnn/returnn/util/file_cache.py", line 479, in run
    os.utime(filename, None)
FileNotFoundError: [Errno 2] No such file or directory
```
Also, in the same log file, cached training HDF files are being deleted by returnn because there is not enough free disk space, late causing error:
```
[...]

PR #1709, opened 2025-04-11
FileCache: hold lock and refresh mtime during cleanup

NeoLegends:
@hannah220 Can you try out this PR?

Maybe a fix for #1675

### DIFF
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3196,7 +3196,7 @@ class LockFile:
             except OSError as exc:
                 if exc.errno not in [errno.ENOENT, errno.EEXIST]:
                     raise  # raise any other error
-                # Lock is held or directory does not exist.
+                # We did not get the lock.
                 # Remove potential stale lockfile before retrying.
                 self.maybe_remove_old_lockfile()
         return False
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3153,42 +3153,50 @@ class LockFile:
         Acquires the lock.
         """
         import time
-        import errno
 
         wait_count = 0
         while True:
-            # Try to create directory if it does not exist.
-            try:
-                os.makedirs(self.directory)
-            except OSError as exc:
-                # Possible errors:
-                # ENOENT (No such file or directory), e.g. if some parent directory was deleted.
-                # EEXIST (File exists), if the dir already exists.
-                if exc.errno not in [errno.ENOENT, errno.EEXIST]:
-                    # Other error, so reraise.
-                    # Common ones are e.g.:
-                    # ENOSPC (No space left on device)
-                    # EACCES (Permission denied)
-                    raise
-                # Ignore those errors.
-            # Now try to create the lock.
-            try:
-                self.fd = os.open(self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
-                return
-            except OSError as exc:
-                # Possible errors:
-                # ENOENT (No such file or directory), e.g. if the directory was deleted.
-                # EEXIST (File exists), if the lock already exists.
-                if exc.errno not in [errno.ENOENT, errno.EEXIST]:
-                    raise  # Other error, so reraise.
-            # We did not get the lock.
-            # Check if it is a really old one.
-            self.maybe_remove_old_lockfile()
-            # Wait a bit, and then retry.
+            if self.try_lock():
+                break
+            # We did not get the lock. Wait a bit, and then retry.
             time.sleep(min(wait_count * 0.1, 1.0))
             wait_count += 1
             if wait_count == 10:
                 print("Waiting for lock-file: %s" % self.lockfile)
+
+    def try_lock(self) -> bool:
+        """
+        Tries to acquire the lock.
+
+        :return: whether the lock was acquired
+        """
+
+        import errno
+
+        # Try to create directory if it does not exist.
+        try:
+            os.makedirs(self.directory)
+        except OSError as exc:
+            # Possible errors:
+            # ENOENT (No such file or directory), e.g. if some parent directory was deleted.
+            # EEXIST (File exists), if the dir already exists.
+            if exc.errno not in [errno.ENOENT, errno.EEXIST]:
+                # Other error, so reraise.
+                # Common ones are e.g.:
+                # ENOSPC (No space left on device)
+                # EACCES (Permission denied)
+                raise
+            # Ignore those errors.
+        # Remove potential stale lockfile before acquiring
+        self.maybe_remove_old_lockfile()
+        # Now try to create the lock.
+        try:
+            self.fd = os.open(self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
+            return True
+        except OSError as exc:
+            if exc.errno in [errno.ENOENT, errno.EEXIST]:
+                return False  # lock is held or directory does not exist
+            raise  # Other error, so reraise.
 
     def unlock(self):
         """
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3187,16 +3187,19 @@ class LockFile:
                 # EACCES (Permission denied)
                 raise
             # Ignore those errors.
-        # Remove potential stale lockfile before acquiring
-        self.maybe_remove_old_lockfile()
-        # Now try to create the lock.
-        try:
-            self.fd = os.open(self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
-            return True
-        except OSError as exc:
-            if exc.errno in [errno.ENOENT, errno.EEXIST]:
-                return False  # lock is held or directory does not exist
-            raise  # Other error, so reraise.
+
+        for _ in range(2):
+            # Now try to create the lock.
+            try:
+                self.fd = os.open(self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
+                return True
+            except OSError as exc:
+                if exc.errno not in [errno.ENOENT, errno.EEXIST]:
+                    raise  # raise any other error
+                # Lock is held or directory does not exist.
+                # Remove potential stale lockfile before retrying.
+                self.maybe_remove_old_lockfile()
+        return False
 
     def unlock(self):
         """
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -3185,7 +3185,7 @@ class LockFile:
             # Check if it is a really old one.
             self.maybe_remove_old_lockfile()
             # Wait a bit, and then retry.
-            time.sleep(1)
+            time.sleep(min(wait_count * 0.1, 1.0))
             wait_count += 1
             if wait_count == 10:
                 print("Waiting for lock-file: %s" % self.lockfile)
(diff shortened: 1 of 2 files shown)


albertz:
I don't understand: Why is the global lock needed? We already hold the lock for this file.

NeoLegends:
Other processes don‘t see the lock though when they do cleanup, and because they take the mtime pessimistically they can miss updates to it. I wrote down my hypothesis in the issue here: https://github.com/rwth-i6/returnn/issues/1675#issuecomment-2796983096

albertz:
Can you write down more exactly step by step what you think was the process of events? You expect it like this?

Proc 1:

* get_file
* _copy_file_if_needed
* -> "FileCache: using existing file <...>"

Proc 2:

* cleanup
* -> "FileCache: Delete file <...>" (because file is old enough)

If this is the problem, why not update the mtime during inside _copy_file_if_needed, right after "FileCache: using existing file"?

NeoLegends:
Yes, exactly, this is how I think it goes.

> If this is the problem, why not update the mtime during inside _copy_file_if_needed, right after "FileCache: using existing file"?

Because this is what we do already: https://github.com/rwth-i6/returnn/blob/cf50800fd8651be8fa993b47d158f99d97aebfd7/returnn/util/file_cache.py#L130

We update the mtime immediately after we find that we can still use it. It's outside the file lock, but the lock doesn't do anything in this case anyway, as the cleanup affects all files, and not just the one we are having locked.

If you have the following order of events, it currently crashes:

Proc 1: get file a
Proc 2: get file b
Proc 1: -> "FileCache: using existing file A"
Proc 2: cleanup
Proc 2: get mtime of all files
Proc 1: update mtime of file A
Proc 2: delete A, mtime update not seen

And we don't have anything in the code that prevents this order ^ of events to occur.

As I said it might be enough that we do another check of the mtime of the file that one process is just about to delete during cleanup. This is not going to rule out crashes though (because the race condition still exists), just make them less likely again.

This PR therefore introduces the global cleanup lock to prevent that race condition entirely. But it runs into the issue that if two processes try to acquire the cleanup lock at the same time, at least one of them is going to wait for a full second, even if the other process realized it didn't actually have to do any cleanup yet. So we have slowdown. But we could also just make the lock a bit faster/wait a bit less, by waiting for e.g. `min(0.1s * number_of_waits, 1s)` instead of always waiting for one full second. WDYT?

NeoLegends:
Or we use fcntl.flock instead of the manual loop. That could also work a bit faster.

albertz:
> We update the mtime immediately after we find that we can still use it. It's outside the file lock ...

Yes but that's a crucial problem, that this is outside the file lock.

> And we don't have anything in the code that prevents this order of events to occur.

I think it's also a problem that in our cleanup code, we don't try to get the lock for the file that we are about to clean up. If we would have that, it would solve it, right? Of course, within the lock, you would also check again the mtime.
 
But I don't understand why we need a global lock.

returnn/util/file_cache.py
```diff
@@ -324,19 +349,27 @@ def _copy_file_if_needed(self, src_filename: str, dst_filename: str):
...
+                # See https://github.com/rwth-i6/returnn/issues/1675 for discussion.
+                os.utime(dst_filename, None)
+                os.utime(info_file_name, None)
                 return
 
             print(f"FileCache: Copy file {src_filename} to cache")
 
             # Make sure we have enough disk space, st_size +1 due to _copy_with_prealloc
-            self.cleanup(need_at_least_free_space_size=os.stat(src_filename).st_size + 1)
+            self.cleanup(lock_held_for=lock, need_at_least_free_space_size=os.stat(src_filename).st_size + 1)
```
albertz, line 372: I don't understand, why do you pass `lock_held_for`? Why is it specifically relevant what lock this current proc is holding, but irrelevant what other procs are holding? I think in `cleanup`, all locks should be treated the same. The lock of the current proc should not get any special treatment.

NeoLegends: Ah, I realize now we will never be holding a lock for the file that we are trying to fetch here. If it existed, we would have used it. So agree, this code can go away. Fixing...

returnn/util/file_cache.py
```diff
@@ -324,13 +344,21 @@ def _copy_file_if_needed(self, src_filename: str, dst_filename: str):
...
         # Copy the file, while holding a lock. See comment on lock_timeout above.
         with LockFile(
-            directory=dst_dir, name=os.path.basename(dst_filename) + ".lock", lock_timeout=self._lock_timeout
+            directory=lock_dir, name=lock_file, lock_timeout=self._lock_timeout
         ) as lock, self._touch_files_thread.files_added_context(lock.lockfile):
             # Maybe it was copied in the meantime, while waiting for the lock.
             if self._check_existing_copied_file_maybe_cleanup(src_filename, dst_filename):
                 print(f"FileCache: using existing file {dst_filename}")
+                # Update mtime while holding lock, to synchronize with any concurrent cleanup.
+                #
```
albertz, line 358:
```suggestion
```

returnn/util/file_cache.py
```diff
@@ -203,45 +200,64 @@ def cleanup(self, *, need_at_least_free_space_size: int = 0):
...
-            if cur_time - mtime > self._cleanup_files_always_older_than_days * 60 * 60 * 24:
-                delete_reason = f"File is {(cur_time - mtime) / 60 / 60 / 24:.1f} days old"
-            else:
-                reached_more_recent_files = True
-            if not delete_reason and need_at_least_free_space_size > cur_expected_free:
-                # Still must delete some files.
-                if cur_time - mtime > cur_used_time_threshold:
-                    delete_reason = f"Still need more space, file is {(cur_time - mtime) / 60 / 60:.1f} hours old"
+
+            lock_dir, lock_file = self._get_lock_filename(fn)
```
albertz, line 204: I wonder, we iterate here over all files, right? Does this include the lock files themselves? Then this logic here is wrong? We don't need another lock for the lock.

And what about info files, isn't this also wrong for them?

returnn/util/file_cache.py
```diff
@@ -203,45 +200,64 @@ def cleanup(self, *, need_at_least_free_space_size: int = 0):
...
+                # Re-check mtime with lock, could have been updated by another
+                # process in the meantime.
+                #
+                # We do not update the `mtime` variable here, because the code
+                # assumes that the list of files is sorted by mtime to abort
+                # early when enough space has been made.
+                #
+                # Instead, we treat the case where the mtime was updated during
+                # cleanup as an outlier and continue as if no other mtimes had
+                # changed.
```
albertz, line 215:
```suggestion
                # Re-check mtime with lock, could have been updated by another
                # process in the meantime.
                # We do not update the `mtime` variable here, because the code
                # assumes that the list of files is sorted by mtime to abort
                # early when enough space has been made.
                # Instead, we treat the case where the mtime was updated during
                # cleanup as an outlier and continue as if no other mtimes had
                # changed.
                # See for discussion:
                #   - https://github.com/rwth-i6/returnn/issues/1675
                #   - https://github.com/rwth-i6/returnn/pull/1709
```

returnn/util/file_cache.py
NeoLegends: We already exclude the info files, but we should also exclude lock files!

albertz: I think our current solution is both easier and also less error-prone.

NeoLegends:
I'm using this on the big EN model training now, but I think we should also wait for @hannah220's test run because that one actually exceeds local disk limits and has to run cleanups periodically. Unfortunately, the cluster at AppTek is pretty full right now, so we might have to wait for a while.

NeoLegends:
Hm, I think we should also avoid messing with files that are currently being copied.

I think these temporary files won't be affected by the cleanup anyway if all goes well (because their mtime will be fresh), but maybe we should exclude them upfront from any cleanup as well. Then again, if their mtime is stale, these files should be cleaned up.

Maybe, to make this safe, we should detect whether we are trying to clean up a tmp copy file, and try to lock the original file instead of the copy file. (i.e. pick up a lock on `/var/tmp/user/file.hdf` instead of `/var/tmp/user/file.hdf.copy`. Then the cleanup will wait for any outstanding copy processes.

NeoLegends:
There more I think about it, I don't think we need any lock on the temporary copy file. Either its mtime will be really fresh, or it's stale, and then we clean it up.

albertz:
> There more I think about it, I don't think we need any lock on the temporary copy file. Either its mtime will be really fresh, or it's stale, and then we clean it up.

We definitely need a lock. Two RETURNN instances should not copy the file at the same time.

albertz (changes requested):
It seems there are further problems. Let's first resolve those.

returnn/util/file_cache.py
```diff
@@ -203,45 +203,79 @@ def cleanup(self, *, need_at_least_free_space_size: int = 0):
...
-                    break
 
-            if delete_reason:
-                cur_expected_free += size
-                print(
-                    f"FileCache: Delete file {fn}, size {human_bytes_size(size)}. {delete_reason}."
-                    f" After deletion, have {human_bytes_size(cur_expected_free)} free space."
-                )
+            lock_dir, lock_file = self._get_lock_filename(fn)
+            with LockFile(directory=lock_dir, name=lock_file, lock_timeout=self._lock_timeout):
```
albertz, line 208: I realize, we would not want the code like this in `cleanup`. This would potentially hang for all files which are currently copied by other procs, right? The copying could potentially take some time. The `cleanup` code should just skip over those files where there is an active lock.

To do this in a clean way, maybe our `LockFile` API needs to be extended. E.g. there could be some `LockFile.try_lock_ctx` or so.

NeoLegends: I think `try_lock_ctx` is not going to work as we wouldn't want the block contents to be executed when the lock isn't held, but a normal try-finally and an if will do.

returnn/util/basic.py
```diff
@@ -3153,42 +3153,53 @@ def lock(self):
...
-            # We did not get the lock.
-            # Check if it is a really old one.
-            self.maybe_remove_old_lockfile()
-            # Wait a bit, and then retry.
-            time.sleep(1)
-            wait_count += 1
-            if wait_count == 10:
-                print("Waiting for lock-file: %s" % self.lockfile)
+                    raise  # raise any other error
+                # Lock is held or directory does not exist.
```
albertz, line 3199: "Lock is held" is ambiguous. It could also mean that we hold the lock now, which is exactly not the case. You want to say sth like "We did not get the lock." to make it clear. Actually the comment which I had before.

returnn/util/file_cache.py
albertz: > I think `try_lock_ctx` is not going to work as we wouldn't want the block contents to be executed

I was thinking about sth like:
```python
with LockFile(...).try_lock_ctx() as lock_ctx:
    if not lock_ctx.we_hold_the_lock():
        continue
    ...
```
But your code is also fine now.

albertz:
Looks good again.

But again, we should do one or two actual test runs, or at least partial test runs.

NeoLegends:
The new code has done some 23 epochs now on a constrained FS that runs full on every epoch. Previously it would eventually crash after just a handful (e.g. 5-6 epochs), now it's okay. I also have a training run with it running, albeit just with data that fits onto the local disk, so the cleanup is less urgent there.

albertz (approved):

NeoLegends:
FYI, the training that used to crash very frequently has now done more than 100 subepochs. So it seems like the issues are really gone now. 🎉

PR #1711, opened 2025-04-16
Fix incorrect `complete_frac` passed to `_print_process`

dorian-K:
When the dataset doesn't support `complete_frac`, then -1 is written into the data dict (See returnn_dataset_wrapper.py), but this will cause `_print_process` to throw an AssertionError.

This bug could also be fixed inside `_print_process` but I felt that it was better to just not pass a wrong value to it.


albertz:
> This bug could also be fixed inside `_print_process` but I felt that it was better to just not pass a wrong value to it.

I think it makes more sense to fix it inside `_print_process`.
(cc @NeoLegends right?)

NeoLegends:
Agree. Keeps it DRYer.

returnn/torch/engine.py
```diff
@@ -1472,8 +1472,7 @@ def _print_process(
             info += ["%.3f sec/step" % step_duration]
         if start_elapsed is not None:
             info += ["elapsed %s" % hms(start_elapsed)]
-        if complete_frac is not None:
-            assert 1 >= complete_frac > 0, f"{step} step, {complete_frac} complete_frac"
+        if complete_frac is not None and 1 >= complete_frac > 0:
```
albertz, line 1475: I would leave the check in. Specifically, instead of `None`, we also allow `-1` as value to have this disabled. But for anything else, it should be valid, i.e. in the range (0,1].
```suggestion
        if complete_frac not in (-1, None):
            assert 1 >= complete_frac > 0, f"{step} step, {complete_frac} complete_frac"
```

NeoLegends (approved):

albertz (approved):

PR #1717, opened 2025-04-23
`DistributeFilesDataset`: allow specifying files via list file

NeoLegends:
This helps keep the config shorter when the number of files grows large and I vaguely remember Python parsing issues when the (python) files grow too large.

### DIFF
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -242,7 +242,6 @@ class DistributeFilesDataset(CachedDataset2):
         elif isinstance(self.files, (str, os.PathLike)):
             _, ext = os.path.splitext(self.files)
             assert ext, f"{self}: no file extension on file list file {self.files}"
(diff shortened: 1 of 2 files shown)


returnn/datasets/distrib_files.py
```diff
@@ -218,17 +222,52 @@ def _meta_info_cache(self):
...
+                    self._files = [line for line in stripped_lines if not line.startswith("#")]
+            elif ext == ".json":
+                import json
+
+                with open(self.files, "rt") as f:
+                    self._files = json.load(f)
+            elif ext == ".py":
+                with open(self.files, "rb") as f:
+                    self._files = literal_eval(f.read())
+            assert isinstance(self._files, list)
```
albertz, line 258:
```suggestion
            else:
                raise ValueError(f"{self}: type {ext} not supported as file list file")
            assert isinstance(self._files, list)
```

albertz (approved):

PR #1719, opened 2025-05-08
tooling: change formatter to ruff

NeoLegends:
Test PR to see the changes.

@albertz WDYT, switch from black to ruff for formatting?

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -22,4 +22,3 @@ jobs:
         cache-dependency-path: '.github/workflows/ruff.yml'
     - run: pip install ruff==0.11.8
     - run: ruff format --diff .
-    - run: ruff format --check .
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,4 +1,4 @@
-name: black
+name: ruff
 
 on:
   push:
@@ -11,7 +11,7 @@ on:
       - master
 
 jobs:
-  check-black-formatting:
+  check-ruff-formatting:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
@@ -19,7 +19,7 @@ jobs:
       with:
         python-version: 3.8
         cache: 'pip'
-        cache-dependency-path: '.github/workflows/black.yml'
-    - run: pip install black==22.3.0
-    - run: black --diff .
-    - run: black --check .
+        cache-dependency-path: '.github/workflows/ruff.yml'
+    - run: pip install ruff==0.11.8
+    - run: ruff format --diff .
+    - run: ruff format --check .
--- a/demos/demo-tf-search-compiled-graph.py
+++ b/demos/demo-tf-search-compiled-graph.py
@@ -26,7 +26,7 @@ class Hyp:
         :param int idx: hyp idx (to identify it in a beam)
         """
         self.idx = idx
-        self.source_idx: typing.Optional[int]  # source hyp idx = None
+        self.source_idx: typing.Optional[int] = None  # source hyp idx
         self.score = 0.0
         self.seq = []  # label seq
 
--- a/demos/demo-tf-search-compiled-graph.py
+++ b/demos/demo-tf-search-compiled-graph.py
@@ -26,7 +26,7 @@ class Hyp:
         :param int idx: hyp idx (to identify it in a beam)
         """
         self.idx = idx
-        self.source_idx = None  # type: typing.Optional[int]  # source hyp idx
+        self.source_idx: typing.Optional[int]  # source hyp idx = None
         self.score = 0.0
         self.seq = []  # label seq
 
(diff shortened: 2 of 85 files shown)


requirements-dev
```diff
@@ -1,2 +1,3 @@
 black==22.3.0
```
NeoLegends, line 1: I kept black in case we still use it somewhere. Maybe it can be removed, not sure.

albertz:
What about the failing tests?

NeoLegends:
I think after doing one formatting run and then setting the python version appropriately, ruff did not re-format the statements that were incompatible with the older python version. I've re-done the entire formatting run now. There were fewer changes. Let's see what CI says.

requirements-dev
albertz: I don't see the point. With this change, black is not used anymore, or not? So then it should be removed.

NeoLegends:
hm one test remains failing, let's see.

NeoLegends:
AppTek uses 3.10 in the current recipes. That would be ok for us to bump it to, I think.

albertz:
Let's move the discussion on Python version bump to https://github.com/rwth-i6/returnn/issues/1720. Also, that should be an independent PR.

albertz:
Also see https://github.com/rwth-i6/i6_core/pull/597.

pyproject.toml
```diff
@@ -13,5 +13,9 @@ extend-exclude = '''
 )/
 '''
 
+[tool.ruff]
+line-length = 120
+target-version = "py38"  # https://github.com/rwth-i6/returnn/issues/1326
```
albertz, line 18: I don't exactly understand. Is this not used? Why do you have reformatted some files with Python 3.8 incompatible changes?

requirements-dev
albertz: No, it should not be used elsewhere. So remove it.

albertz:
We can also stay with Python 3.8 for now, and only do Python 3.8 compatible formatting. Actually, it looks like your setting file already does this. So sth is broken?

.github/workflows/ruff.yml
```diff
@@ -11,15 +11,15 @@ on:
...
     - uses: actions/setup-python@v5
       with:
         python-version: 3.8
         cache: 'pip'
-        cache-dependency-path: '.github/workflows/black.yml'
-    - run: pip install black==22.3.0
-    - run: black --diff .
-    - run: black --check .
+        cache-dependency-path: '.github/workflows/ruff.yml'
+    - run: pip install ruff==0.11.8
```
albertz, line 23: Why this version? And we should in any case make it consistent across repos.

curufinwe: This is the current version as of today. I would suggest we us this one.

pyproject.toml
NeoLegends: First I did not have this line there, saying the results must be 3.8 compatible. But I formatted. Then I added the line and formatted again. ruff did not re-format the constructs that are invalid in 3.8, though. I did a clean checkout, added the 3.8 config and then did a format, and it outputted 3.8-compatible constructs.

albertz: So can you update this PR then with the correct formatting change?

NeoLegends:
> So sth is broken?

Formatting is no longer broken, just PyCharm disagrees with some of the ruff results it seems.

Seems like ruff does not take comments into account when calculating line length, but PyCharm does. I manually fixed the offending lines, it was mostly `# type: ...`-annotations.

NeoLegends:
Unfortunately, ast-grep does not see comments, therefore an automatic transformation from `var = x # type: T` to `var: T = x` is difficult, and (as of now) there are 489 places where the comment-style type annotation is used.

pyproject.toml
NeoLegends: It is correct now.

michelwi:
see point 3 in https://docs.astral.sh/ruff/rules/line-too-long/:
Ignores line that end with a pragma comment (e.g., # type: ignore or # noqa), as long as the pragma comment starts before the line-length threshold.

demos/demo-tf-search-compiled-graph.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, idx):
         :param int idx: hyp idx (to identify it in a beam)
         """
         self.idx = idx
-        self.source_idx = None  # type: typing.Optional[int]  # source hyp idx
+        self.source_idx: typing.Optional[int]  # source hyp idx = None
```
michelwi, line 29:
```suggestion
        self.source_idx: typing.Optional[int]  = None  # source hyp idx
```
that looks like a bug in ruff formatter

NeoLegends: Nah, it's a bug in my search+replace :D Need to fix them.

albertz:
Why do we get this error in some of the tests now?
```
Exception AttributeError("'Runner' object has no attribute '_horovod_collected_reduce_inputs'") in step 0. (pid 2263)
```

albertz:
Let's keep this as a draft while you are still working on it. There should be no failing test at the end. And only code formatting changes, no source code changes.

.github/workflows/ruff.yml
```diff
@@ -11,15 +11,15 @@ on:
...
+    - run: ruff format --diff .
+    - run: ruff format --check .
```
albertz, line 25: I think this is not needed, `--diff` already does this? Please also check the other repos, and keep it consistent.

albertz:
Are you now replacing all the `# type: ...` things? I'm not sure that's reasonable now for this PR here.

NeoLegends:
In the files that had lines with `# type: ` where they were too long I used a regex search + replace to convert them into formattable type annotations. Some of them I needed to fix manually because my regex didn't capture the variable values correctly when there was more than one annotation in the comment. I'm not converting systematically/all of them.

albertz (approved):

PR #1731, opened 2025-06-20
`FileCache`: avoid race condition in `TouchFilesThread` when deleting files

NeoLegends:
This implementation now uses one lock per file instead of one lock for all files, as was previously implemented in 14590d4.

### DIFF
--- a/returnn/util/file_cache.py
+++ b/returnn/util/file_cache.py
@@ -9,7 +9,7 @@ See https://github.com/rwth-i6/returnn/issues/1519 for initial discussion.
 Main class is :class:`FileCache`.
 """
 
-from typing import Any, Collection, Iterable, List, Optional, Tuple, Union
+from typing import Any, Collection, Dict, Iterable, List, Optional, Tuple, Union
 import errno
 import os
 import pathlib
@@ -20,7 +20,7 @@ from dataclasses import dataclass
 from collections import defaultdict
 from contextlib import contextmanager
 import json
-from threading import Thread, Event
+from threading import Lock, Thread, Event
 from returnn.config import Config, get_global_config
 from .basic import expand_env_vars, LockFile, human_bytes_size
 
@@ -523,6 +523,7 @@ class _TouchFilesThread(Thread):
         super().__init__(daemon=True)
         self.stop = Event()
         self.files = defaultdict(int)  # usage counter
+        self.locks: Dict[str, Lock] = {}
         self.interval = interval
         self.cache_base_dir = cache_base_dir
         self._is_started = False  # careful: `_started` is already a member of the base class
@@ -530,16 +531,18 @@ class _TouchFilesThread(Thread):
     def run(self):
         """thread main loop"""
         while True:
-            all_files = {}  # dict to have order deterministic
-            for filename in self.files.copy():  # copy dict under GIL to avoid modifications during iteration
-                all_files[filename] = True
-                all_files.update({k: True for k in _all_parent_dirs(filename, base_dir=self.cache_base_dir)})
+            # copy dicts under GIL to avoid modifications during iteration
+            all_files = self.files.copy()
+            locks = self.locks.copy()
             for filename in all_files:
-                try:
-                    os.utime(filename, None)
-                except Exception as exc:
-                    print(f"FileCache: failed updating mtime of {filename}: {exc}")
-                    raise
+                with locks[filename]:
+                    if filename not in self.files:
+                        continue
+                    try:
+                        os.utime(filename, None)
+                    except Exception as exc:
+                        print(f"FileCache: failed updating mtime of {filename}: {exc}")
+                        raise
             if self.stop.wait(self.interval):
                 return
 
@@ -556,18 +559,24 @@ class _TouchFilesThread(Thread):
             to_add = [to_add]
         assert isinstance(to_add, Iterable)
         self.start_once()
-        for file in to_add:
+        # we track the parent directories as well and give them their own locks to be
+        # able to synchronize their deletion with the touch thread
+        for file in _files_with_parents(to_add, base_dir=self.cache_base_dir):
             self.files[file] += 1
+            if file not in self.locks:
+                self.locks[file] = Lock()
 
     def files_remove(self, to_remove: Union[str, Iterable[str]]):
         """remove"""
         if isinstance(to_remove, str):
             to_remove = [to_remove]
         assert isinstance(to_remove, Iterable)
-        for filename in to_remove:
-            self.files[filename] -= 1
-            if self.files[filename] <= 0:
-                del self.files[filename]
+        for file in _files_with_parents(to_remove, base_dir=self.cache_base_dir):
+            with self.locks[file]:
+                self.files[file] -= 1
+                if self.files[file] <= 0:
+                    del self.files[file]
+                    del self.locks[file]
 
     @contextmanager
     def files_added_context(self, files: Collection[str]):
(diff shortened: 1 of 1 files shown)


returnn/util/file_cache.py
```diff
@@ -588,3 +597,12 @@ def _all_parent_dirs(filename: str, *, base_dir: str) -> List[str]:
...
     return dirs
+
+
+def _files_with_parents(files: Iterable[str], *, base_dir: str) -> Dict[str, bool]:
+    res = {}  # dict to have order deterministic
+    for file in files:
+        res[file] = True
+        for file in _all_parent_dirs(file, base_dir=base_dir):
+            res[file] = True
+    return res
```
albertz, line 608:
```suggestion
def _files_with_parents(filenames: Iterable[str], *, base_dir: str) -> Dict[str, bool]:
    res = {}  # dict to have order deterministic
    for fn in filenames:
        res[fn] = True
        for fn_ in _all_parent_dirs(fn, base_dir=base_dir):
            res[fn_] = True
    return res
```

returnn/util/file_cache.py
```diff
@@ -523,23 +523,26 @@ def __init__(self, *, interval: float = 1.0, cache_base_dir: str):
...
         """thread main loop"""
         while True:
-            all_files = {}  # dict to have order deterministic
-            for filename in self.files.copy():  # copy dict under GIL to avoid modifications during iteration
-                all_files[filename] = True
-                all_files.update({k: True for k in _all_parent_dirs(filename, base_dir=self.cache_base_dir)})
+            # copy dicts under GIL to avoid modifications during iteration
+            all_files = self.files.copy()
+            locks = self.locks.copy()
             for filename in all_files:
```
albertz, line 537:
```suggestion
            locks = self.locks.copy()
            for filename, lock in locks.items():
```

returnn/util/file_cache.py
```diff
@@ -523,23 +523,26 @@ def __init__(self, *, interval: float = 1.0, cache_base_dir: str):
...
-                try:
-                    os.utime(filename, None)
-                except Exception as exc:
-                    print(f"FileCache: failed updating mtime of {filename}: {exc}")
-                    raise
+                with locks[filename]:
```
albertz, line 538:
```suggestion
                with lock:
```

returnn/util/file_cache.py
```diff
@@ -523,23 +523,25 @@ def __init__(self, *, interval: float = 1.0, cache_base_dir: str):
...
-                all_files[filename] = True
-                all_files.update({k: True for k in _all_parent_dirs(filename, base_dir=self.cache_base_dir)})
-            for filename in all_files:
-                try:
-                    os.utime(filename, None)
-                except Exception as exc:
-                    print(f"FileCache: failed updating mtime of {filename}: {exc}")
-                    raise
+            # copy dicts under GIL to avoid modifications during iteration
+            locks = self.locks.copy()
```
albertz, line 535: I wonder a bit here: When some other thread does `files_extend`, and creates a new lock (`self.locks[file] = Lock()`), there is a small race condition that two lock objects are created for the same `filename`. Here when we do `self.locks.copy`, it might copy the old lock instance, and not the one which survives in the end in `self.locks[filename]`. How much is this a problem? If this is a problem, how can we solve it?

returnn/util/file_cache.py
```diff
@@ -598,10 +598,10 @@ def _all_parent_dirs(filename: str, *, base_dir: str) -> List[str]:
     return dirs
 
 
-def _files_with_parents(filenames: Iterable[str], *, base_dir: str) -> Dict[str, bool]:
-    res = {}  # dict to have order deterministic
+def _files_with_parents(filenames: Iterable[str], *, base_dir: str) -> Dict[str, int]:
```
albertz, line 601: I don't understand. Why is it relevant to return the count, and not just a bool?

returnn/util/file_cache.py
albertz: > Ah, can creating a lock release the GIL?

Maybe, but that's not what I mean. This code:
```python
            if file not in self.locks:
                self.locks[file] = Lock()
```
This code can release the GIL after the `if` check. If two threads concurrently do this, you can end up with this situation.

NeoLegends: Yes. The `Lock()` call will release the GIL, I believe.

albertz: > I think I'll add a lock that guards any dict modifications (i.e. excluding any FS operations)

Yes that should solve it. The thread main loop actually doesn't need it, I think, as all the ops are safe (`self.locks.copy()`, and checking `filename not in self.files`).

albertz: The code:
```python
            if file not in self.locks:
                self.locks[file] = Lock()
```
That are multiple statements. When the GIL gets released *after* `if file not in self.locks:`, but *before* `self.locks[file] = Lock()`, you have the problem. It doesn't even need to get to the `Lock()` call.

NeoLegends: >The thread main loop actually doesn't need it

I'll still add it just to avoid future accidental issues.

albertz:
```
In [6]: compile("if file not in self.locks: self.locks[file] = Lock()", "<source>", "exec")
Out[6]: <code object <module> at 0x1050eff50, file "<source>", line 1>

In [8]: dis.dis(_6)
  1           0 LOAD_NAME                0 (file)
              2 LOAD_NAME                1 (self)
              4 LOAD_ATTR                2 (locks)
              6 CONTAINS_OP              1
              8 POP_JUMP_IF_FALSE       22
             10 LOAD_NAME                3 (Lock)
             12 CALL_FUNCTION            0
             14 LOAD_NAME                1 (self)
             16 LOAD_ATTR                2 (locks)
             18 LOAD_NAME                0 (file)
             20 STORE_SUBSCR
        >>   22 LOAD_CONST               0 (None)
             24 RETURN_VALUE
```

albertz: I mean for example after it executed the op in the bytecode offset 6, or offset 8, or 10, when it releases the GIL at that point. But it would be problematic at all points after 6 before 20.

returnn/util/file_cache.py
```diff
@@ -527,23 +527,26 @@ def __init__(self, *, interval: float = 1.0, cache_base_dir: str):
...
-                all_files[filename] = True
-                all_files.update({k: True for k in _all_parent_dirs(filename, base_dir=self.cache_base_dir)})
-            for filename in all_files:
-                try:
-                    os.utime(filename, None)
-                except Exception as exc:
-                    print(f"FileCache: failed updating mtime of {filename}: {exc}")
-                    raise
+            with self.files_lock:
+                locks = self.locks.copy()
```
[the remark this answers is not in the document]
NeoLegends: Yes, I want to leave it in to prevent future accidental errors. The code has already had enough errors as it was, and I don't think the lock for the quick copy here will hurt.

returnn/util/file_cache.py
albertz: The GIL can be released at any time. (Otherwise pure computational multithreaded code could not work at all.)

returnn/util/file_cache.py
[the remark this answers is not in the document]
albertz: But we shouldn't just put code which is not needed. If you want you can add a comment:
```suggestion
            locks = self.locks.copy()  # locking files_lock not needed
```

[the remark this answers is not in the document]
NeoLegends: How can you be sure this call is actually threadsafe now?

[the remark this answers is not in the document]
albertz: `dict.copy()` is a single Python call (a single `CALL_FUNCTION` op). The GIL can be release before or after that, but not within that (unless the native C code of `dict.copy` does it explicitly - but `dict.copy` does not do that).

albertz (approved):

Issue #1732, opened 2025-06-30
Odd behaviour of load epoch


dorian-K:
I got an exception in my Returnn Forward job, and here is the traceback:
<details>
<summary>Traceback</summary>
<code>
EXCEPTION
Traceback (most recent call last):
  File "/home/tt201262/setups/2024-10-11-denosing_lm/recipe/returnn/rnn.py", line 11, in <module>
    line: main()
    locals:
      main = <local> <function returnn.__main__.main>
  File "/rwthfs/rz/cluster/home/tt201262/setups/2024-10-11-denosing_lm/recipe/returnn/returnn/__main__.py", line 744, in main
    line: execute_main_task()
  File "/rwthfs/rz/cluster/home/tt201262/setups/2024-10-11-denosing_lm/recipe/returnn/returnn/__main__.py", line 589, in execute_main_task
    line: engine.forward_with_callback(
              dataset=data, callback=forward_callback, allow_skipping_seqs=allow_skipping_seqs
          )
    locals:
      engine = <global> <returnn.torch.engine.Engine object at 0x14c43bd4fd40>
      data = <local> <OggZipDataset 'out.ogg' epoch=None>
      forward_callback = <local> <i6_experiments.users.zeyer.recog._returnn_v2_get_forward_callback.<locals>._ReturnnRecogV2ForwardCallbackIface object at 0x14c43a81ecc0>
      allow_skipping_seqs = <local> False
  File "/rwthfs/rz/cluster/home/tt201262/setups/2024-10-11-denosing_lm/recipe/returnn/returnn/torch/engine.py", line 1367, in Engine.forward_with_callback
    line: for extern_data_raw in data_loader:
    locals:
      data_loader = <local> <torch.utils.data.dataloader.DataLoader object at 0x14c43a8af290>
  File "/home/tt201262/pyenvs/py3.12-torch2.5/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 701, in _BaseDataLoaderIter.__next__
    line: data = self._next_data()
    locals:
      self = <local> <torch.utils.data.dataloader._MultiProcessingDataLoaderIter object at 0x14c43a0d3200>
  File "/home/tt201262/pyenvs/py3.12-torch2.5/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 1465, in _MultiProcessingDataLoaderIter._next_data
    line: return self._process_data(data)
    locals:
[...]

PR #1733, opened 2025-06-30
Fix wrong epoch number parsing

dorian-K:
See #1732

PR is in Draft because I still need to test the changes

### DIFF
--- a/returnn/util/basic.py
+++ b/returnn/util/basic.py
@@ -790,6 +790,7 @@ def model_epoch_from_filename(filename):
     # We don't have it in the model, though.
     # For now, just parse it from filename.
     # If TF, and symlink, resolve until no symlink anymore (e.g. if we symlinked the best epoch).
+    file_basename = None
     for potential_ext in [".meta", ".pt"]:
         if not os.path.exists(filename + potential_ext):
             continue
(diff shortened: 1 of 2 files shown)


returnn/util/basic.py
```diff
@@ -790,15 +790,23 @@ def model_epoch_from_filename(filename):
...
+        file_basename = os.path.basename(filename + potential_ext)
         break
-    m = re.match(".*\\.([0-9]+)", filename)
+    if file_basename is None:  # maybe the input filename already has an extension?
+        # noinspection PyBroadException
+        try:
+            file_basename = os.path.basename(filename)
+        except Exception:
+            return None
+    m = re.match(".*\\.([0-9]+)", file_basename)
```
albertz, line 809: I think this is too complicated. Why not just always do `basename` in the end? Why catch the generic `Exception` for `basename`? I think `basename` can newer throw an exception.

```suggestion
    m = re.match(".*\\.([0-9]+)", os.path.basename(filename))
```

returnn/util/basic.py
```diff
@@ -790,15 +790,23 @@ def model_epoch_from_filename(filename):
...
     # If TF, and symlink, resolve until no symlink anymore (e.g. if we symlinked the best epoch).
+    file_basename = None
     for potential_ext in [".meta", ".pt"]:
         if not os.path.exists(filename + potential_ext):
             continue
         while os.path.exists(filename + potential_ext) and os.path.islink(filename + potential_ext):
             fn_with_ext_ = os.readlink(filename + potential_ext)
             assert fn_with_ext_.endswith(potential_ext), "strange? %s, %s" % (filename, potential_ext)
             filename = fn_with_ext_[: -len(potential_ext)]
+        file_basename = os.path.basename(filename + potential_ext)
```
albertz, line 801:
```suggestion
```

PR #1734, opened 2025-07-08
Add "first" phoneme pick strategy to PhoneSeqGenerator

dorian-K:
`i6_core.lib.corpus.ApplyLexiconToCorpusJob` picks the first Lemma from a lexicon by default, this PR adds the same functionality to `PhoneSeqGenerator` (for `LmDataset`).

### DIFF
--- a/returnn/datasets/lm.py
+++ b/returnn/datasets/lm.py
@@ -1178,7 +1178,7 @@ class PhoneSeqGenerator:
         add_extra_begin_lemma: float = 1.0,
         extra_end_lemma: Optional[Dict[str, Any]] = None,
         add_extra_end_lemma: float = 1.0,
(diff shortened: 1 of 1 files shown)


returnn/datasets/lm.py
```diff
@@ -1177,6 +1184,7 @@ def __init__(
         add_extra_begin_lemma: float = 1.0,
         extra_end_lemma: Optional[Dict[str, Any]] = None,
         add_extra_end_lemma: float = 1.0,
+        lexicon_strategy: LexiconStrategy = LexiconStrategy.PICK_RANDOM,
```
albertz, line 1187: Avoid the enum, just use a `str` (makes the Sis config much simpler, also the code here simpler, also more consistent to our other code).

Issue #1724, opened 2025-05-14
PT: No backend type associated with device type cpu


NeoLegends:
Some of my trainings had this crash right after startup. I just want to report this for now, I don't have many details of why this happens yet. The trainings are scheduled rather normally on an A5000 GPU node, and `nvidia-smi` on that node gives sensible results.

```python
EXCEPTION
Traceback (most recent call last):
  File "/home/mgunz/setups/2025-03-07--ws-asr/recipe/returnn/rnn.py", line 11, in <module>
    line: main()
    locals:
      main = <local> <function main at 0x7fe2f2b4cee0>
  File "/home/mgunz/setups/2025-03-07--ws-asr/recipe/returnn/returnn/__main__.py", line 744, in main
    line: execute_main_task()
    locals:
      execute_main_task = <global> <function execute_main_task at 0x7fe2f2b4cdc0>
  File "/home/mgunz/setups/2025-03-07--ws-asr/recipe/returnn/returnn/__main__.py", line 543, in execute_main_task
    line: engine.train()
    locals:
      engine = <global> <returnn.torch.engine.Engine object at 0x7fe2271e7520>
      engine.train = <global> <bound method Engine.train of <returnn.torch.engine.Engine object at 0x7fe2271e7520>>
  File "/home/mgunz/setups/2025-03-07--ws-asr/recipe/returnn/returnn/torch/engine.py", line 256, in Engine.train
    line: self.train_epoch()
    locals:
      self = <local> <returnn.torch.engine.Engine object at 0x7fe2271e7520>
      self.train_epoch = <local> <bound method Engine.train_epoch of <returnn.torch.engine.Engine object at 0x7fe2271e7520>>
  File "/home/mgunz/setups/2025-03-07--ws-asr/recipe/returnn/returnn/torch/engine.py", line 387, in Engine.train_epoch
    line: torch.distributed.all_reduce(_has_data, op=torch.distributed.ReduceOp.MIN)
    locals:
      torch = <global> <module 'torch' from '/usr/local/lib/python3.10/dist-packages/torch/__init__.py'>
      torch.distributed = <global> <module 'torch.distributed' from '/usr/local/lib/python3.10/dist-packages/torch/distributed/__init__.py'>
      torch.distributed.all_reduce = <global> <function all_reduce at 0x7fe226f68550>
      _has_data = <local> tensor[1] i8 [1]
      op = <not found>
[...]

PR #1739, opened 2025-07-18
Fix arg check for backend in distributed training

Stefanwuu:
To provide backend in torch_distributed, we need to get it in this opts check.


albertz (approved):

Stefanwuu:
Yes, otherwise when explicitly providing backend, the check crashes in subprocess.

Stefanwuu:
@albertz I would say it's independent to make providing backend possible. But it's indeed a prerequisite for fixing https://github.com/rwth-i6/returnn/issues/1724

PR #1743, opened 2025-07-24
CI: update PyCharm

### LINKED ISSUE #1742: CI: Update PyCharm
Our CI GitHub Actions `pycharm-inspect` job runs `tests/pycharm-inspect.py` which uses pycharm-community-2020.2.

I usually develop with the latest PyCharm, i.e. currently 2025.1.1.

The differences in code style warnings has so far mostly not been a problem. But now there seem to be such a case: commit c14cd5992e33c707a876880fcfb8e3254935ac81, [CI log](https://github.com/rwth-i6/returnn/actions/runs/16452109598/job/46647294853):
`returnn/frontend/decoder/transformer.py:364: WARNING PyUnresolvedReferencesInspection: Unresolved attribute reference 'transform_encoder' for class 'Module'`

(Btw, maybe also a reminder for #1157. But I don't really think there is any update to that...)

In PR #1743, I was trying to update PyCharm, but in the end reverted back to keeping the 2020.2 version for now, so the PR only contains some other related updates. I think this needs some more work.

I think most crucially, it now finds a couple of valid warnings that were not found before. Those should be fixed.

But then, there are also some more new false positives. We should see what to do about them, if we can easily identify them (but at the same time not ignore huge subsets of potential valid warnings). Some **rare** warn ignores in the code could also be ok, but this should be done very rarely, and only for well understood cases. Better to discuss this here (or in a new draft PR) before this is done too much.

albertz:
Fix #1742

### DIFF
--- a/tests/pycharm-inspect.py
+++ b/tests/pycharm-inspect.py
@@ -11,14 +11,12 @@ See here:
 """
 
 import os
-import stat
 import sys
 import re
 import time
 import shutil
 import subprocess
 import tempfile
-import typing
 from glob import glob
 import argparse
 from xml.dom import minidom
@@ -115,8 +113,7 @@ def install_pycharm():
     print("Install PyCharm into:", pycharm_dir)
     sys.stdout.flush()
 
-    pycharm_version = (2020, 2)
-    name = "pycharm-community-%i.%i" % pycharm_version
+    name = "pycharm-community-2025.1.1"
     fn = "%s.tar.gz" % name
 
     subprocess.check_call(


returnn/datasets/basic.py
```diff
@@ -428,9 +427,9 @@ def get_estimated_seq_length(self, seq_idx):
         """
         raise OptionalNotImplementedError
 
-    def get_num_timesteps(self):
+    def get_num_timesteps(self) -> Union[int, NumbersDict]:
         """
-        :rtype: int
+        :return: how much frames we have in total.
```
NeoLegends, line 432: much -> many

NeoLegends (commented):
lgtm!

PR #1744, opened 2025-07-28
MPD, DFD: use deque as cache impl for better efficiency

NeoLegends:
When the buffer is full and the next seq is loaded from the subproc, the previous list-based implementation needed to move `buf_size - 1` elements in the buffer (O(n)) while for the deque, this operation is O(1). These buffers can be large (I usually use `N` around `max_seqs < N < 10 * max_seqs`), so I believe it's good we use efficient algorithms there.

I decided to postpone also adjusting the implementation of `CachedDataset2`. I believe the cache list in that dataset doesn't grow very large usually (I think at least for torch trainings there is only ever going to be one element in the cache?), so it's okay if it's not as asymptotically efficient (also asymptotically less efficient algorithms can still be efficient on small lists). Also, there are some derived datasets that access and modify this cache (`ChunkShuffleDataset`), and they will not work as well on deques as on a list.

### DIFF
--- a/returnn/datasets/distrib_files.py
+++ b/returnn/datasets/distrib_files.py
@@ -6,6 +6,7 @@ https://github.com/rwth-i6/returnn/issues/1519
 
 from __future__ import annotations
 
+from collections import deque
 from typing import Union, Optional, Any, Callable, Sequence, Tuple, List, Dict
 import os
 import sys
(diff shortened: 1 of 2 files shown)


NeoLegends:
@albertz this just a small PR, do you think you can give it a short review & merge soon? Given the potential impact I don't want to just merge this, even if CI is green.

Issue #1568, opened 2024-07-02
Ignore a single broken gradient


JackTemaki:
In my current language model training I sometimes get "nan" gradients, which break the training. Surprisingly, just restarting the training from the last checkpoint is often enough uncertainty to resume training.

[Here](https://github.com/Lightning-AI/pytorch-lightning/issues/4956) people discussed something like:
```python
        valid_gradients = True
        for name, param in self.named_parameters():
            if param.grad is not None:
                valid_gradients = not (torch.isnan(param.grad).any() or torch.isinf(param.grad).any())
                if not valid_gradients:
                    break
        if not valid_gradients:
            print(f'detected inf or nan values in gradients. not updating model parameters')
            self.zero_grad()
  ```
  
  I think it would be a good idea to have this as a configurable option for the updater. Preferably with a "limit", so that it still crashes after e.g. 5 broken updates.

albertz:
> I sometimes get "nan" gradients, which break the training.

You mean after that, the model parameters itself become nan, i.e. the model is broken?

> Surprisingly, just restarting the training from the last checkpoint is often enough uncertainty to resume training.

Uncertainty? You mean restarting from the last checkpoint solves the issue, i.e. getting nan is non-deterministic, and rare, and after such restart, you are usually lucky that you don't get nan anymore, or only much later?

We could also implement such automatic restart logic. It would be another approach than what you suggest afterwards, i.e. to always check for non-finite grads.

I'm not sure what approach would be better. It probably depends also on how often you get this. (E.g. I personally have never gotten this problem so far.)

JackTemaki:
I do not want such automatic restart logic, because this wastes computation. For testing I now implemented [here](https://github.com/rwth-i6/returnn/blob/master/returnn/torch/updater.py#L180) that it skips the update when the result of the gradient clipping is NaN or Inf. I do this for the `grad_clip_norm_` right now, and I am not sure how `grad_clip_value_` behaves with Inf/NaN values (maybe it automatically corrects them already). 

This specific problem appeared for me only for LSTM-LM training so far, not anywhere else (ASR, TTS). 

In my current training after 2 full epochs, I now get a skip every 40k update steps.

albertz:
[...]

PR #1745, opened 2025-08-05
Add option to check for invalid gradient norms and allow skipping updates

JackTemaki:
Adds "gradient_clip_norm_invalid_gradient_threshold" parameter

Defines a threshold count of allowed sequentially occuring nan/inf gradients until the training stops.

Resolves: #1568

@albertz This is code copied from my RETURNN fork as starting point, so it is not tested with the master branch here. I use this to have successful LSTM-LM training. Please suggest how you want to name the parameter or if you want functional changes, then @Atticus1806 will take over.

### DIFF
--- a/returnn/torch/engine.py
+++ b/returnn/torch/engine.py
@@ -270,7 +270,6 @@ class Engine(EngineBase):
         self._updater.set_current_train_step(
             global_train_step=self.global_train_step, epoch=self.epoch, epoch_continuous=self.epoch - 1
         )
-        self._updater._num_consec_invalid_gradients_steps = 0
 
         self.learning_rate_control.epoch_data[self.epoch].meta.update(
             {
--- a/returnn/torch/updater.py
+++ b/returnn/torch/updater.py
@@ -212,6 +212,8 @@ class Updater:
             if the number of steps per epoch is known in advance.
         """
         self._current_train_step = global_train_step
+        if self._current_epoch != epoch:
+            self._num_consec_invalid_gradients_steps = 0
         self._current_epoch = epoch
         self._current_epoch_continuous = epoch_continuous
         self._update_effective_learning_rate()


returnn/torch/updater.py
```diff
@@ -134,6 +135,9 @@ def __init__(self, *, config, network, device, initial_learning_rate=1.0):
 
         self._grad_clip = self.config.float("gradient_clip", 0.0)
         self._grad_clip_global_norm = self.config.float("gradient_clip_global_norm", 0.0)
+        self._grad_clip_global_norm_invalid_gradient_threshold = self.config.int(
```
NeoLegends, line 138:
```suggestion
        self._grad_clip_global_norm_invalid_gradient_threshold = self.config.typed_value(
```

albertz:
I wonder, should this be per epoch? Or really global as you did here? If global, this is also slightly wrong, as you need to store the number in the checkpoint then.

JackTemaki:
> If global, this is also slightly wrong, as you need to store the number in the checkpoint then.

The counter should only be valid locally, so if there is even a single correct gradient it resets to zero. 

>  you lose an entire set of accumulated and multi-node synchronized step, right?

My implementation was targeted only for single GPU training, I leave the multi-GPU case as exercise for the reader ;)

NeoLegends:
I think for gradient synchronization you would need to first sync a boolean checking if every node has a valid gradient. Otherwise `loss.backward()` and `updater.skip()` are skipped on all nodes. If only the node with the broken gradient skipped `loss.backward()`, the nodes would go out of sync. For parameter averaging/synchronization I don't think there should be any synchronization op involved in `loss.backward()` (correct @albertz?), so for these models you can just skip it (and also skip `updater.step()`).

albertz:
Inspect still fails because `get_total_norm` is not available in Torch 2.0, which is used in the inspect check. I think `get_total_norm` is only available since PyTorch >=2.6. Are you aware of that? Is this ok?

To pass the inspect test, we could bump the Torch version for the inspection to 2.6 (or 2.7, 2.8) (PR: #1746).

returnn/torch/updater.py
```diff
@@ -223,13 +225,35 @@ def step(self, *, grad_scaler: Optional[torch.cuda.amp.GradScaler] = None):
...
+        has_invalid_gradient = False
+        if self._num_allowed_invalid_gradient_steps is not None:
+            if self._grad_clip_global_norm:
+                norm = torch.nn.utils.clip_grad_norm_(self.network.parameters(), self._grad_clip_global_norm)
+            else:
+                norm = torch.nn.utils.get_total_norm(self.network.parameters())
+            has_invalid_gradient = torch.isnan(norm) or torch.isinf(norm)
+            if has_invalid_gradient:
+                self._num_invalid_gradients += 1
+                if self._num_invalid_gradients >= self._num_allowed_invalid_gradient_steps:
```
albertz, line 238:
```suggestion
                if self._num_invalid_gradients > self._num_allowed_invalid_gradient_steps:
```

Atticus1806:
> Are you aware of that? Is this ok?
No I was not. I was just using this function, as you suggested it. For me we bumping or using a different function are both equally fine.

returnn/torch/updater.py
```diff
@@ -223,13 +225,35 @@ def step(self, *, grad_scaler: Optional[torch.cuda.amp.GradScaler] = None):
...
+                    raise RuntimeError(
+                        f"Got {self._num_invalid_gradients} invalid gradients in succession, abort training"
+                    )
+                else:
+                    invalid_grads_left = self._num_allowed_invalid_gradient_steps - self._num_invalid_gradients
+                    print(
+                        f"Invalid gradient in step {self._current_train_step}, skipping. {invalid_grads_left:02d} "
+                        f"subsequent broken steps left until training is aborted."
```
albertz, line 246:
```suggestion
                        f"Invalid gradient in step {self._current_train_step}, skipping. "
                        f"{invalid_grads_left} subsequent broken steps left until training is aborted.",
                        file=log.v2
```

returnn/torch/updater.py
```diff
@@ -223,13 +225,35 @@ def step(self, *, grad_scaler: Optional[torch.cuda.amp.GradScaler] = None):
...
+                    )
+            else:
+                self._num_invalid_gradients = 0
 
         if grad_scaler is not None:
-            grad_scaler.step(self.optimizer)
+            if not has_invalid_gradient:
+                grad_scaler.step(self.optimizer)
+            # updater needs to be called even if we discard the update due to an invalid gradient
```
albertz, line 254:
```suggestion
            # update needs to be called even if we discard the update due to an invalid gradient
```
Why? Any reference? Is this always correct?

albertz:
> I think for gradient synchronization you would need to first sync a boolean checking ...

I don't think this is needed? For grad sync, if there is any inf/nan, then accumulating the grads over the nodes will result in having that inf/nan on every node. So no extra logic is needed.

> For parameter averaging/synchronization I don't think there should be any synchronization op involved in `loss.backward()` (correct @albertz?), so for these models you can just skip it (and also skip `updater.step()`, if it's updater step time).

Yes, I think here, also no extra logic is needed.

returnn/torch/engine.py
```diff
@@ -270,6 +270,7 @@ def init_train_epoch(self):
         self._updater.set_current_train_step(
             global_train_step=self.global_train_step, epoch=self.epoch, epoch_continuous=self.epoch - 1
         )
+        self._updater._num_consec_invalid_gradients_steps = 0
```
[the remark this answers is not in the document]
albertz: (I wonder that our inspection does not warn about this...)

albertz (approved):

PR #1746, opened 2025-08-07
CI inspect: use PyTorch 2.8

albertz:
Some code uses newer features (eg #1745).


albertz:
Failing tests seem to be independent (seems like a newer version of safetensors is not compatible with Torch 2.0, which is still used in the other tests). I will open another issue about that.

Merging anyway now.

Issue #1257, opened 2023-02-07
HuggingFace datasets wrapper


albertz:
We should be able to use [HuggingFace datasets](https://huggingface.co/docs/datasets/index) directly in RETURNN.

dthulke:
> I guess the most canonical way would be to write a RETURNN Dataset for this. Maybe derived from CachedDataset2.

I already implemented this as a custom dataset some time ago (see implementation below).

Things to discuss are maybe how to handle tokenisation and other preprocessing steps.
In the current implementation there is either the option to define a map function (see https://huggingface.co/docs/datasets/about_map_batch) in the config file or to provide a preprocessed dataset stored with `save_to_disk`.

num_outputs and data_type is currently extracted from the dataset features attribute (which may not always be available).

Another point to discuss is how to handle caching for large datasets (currently the default caching mechanism of hf datasets is used which does not fit very well with our setups).

<details>
  <summary>Code</summary>
  
```python
import numpy
from returnn.datasets.basic import DatasetSeq
from returnn.datasets.cached2 import CachedDataset2
from returnn.util.basic import OptionalNotImplementedError


class HuggingfaceDataset(CachedDataset2):

  @staticmethod
  def kwargs_update_from_config(config, kwargs):
    super().kwargs_update_from_config(config, kwargs)
    if 'map_func' in kwargs:
      if isinstance(kwargs['map_func'], str):
        kwargs['map_func'] = config.typed_value(kwargs['map_func'])

  def __init__(self, dataset_opts, map_func=None, map_func_args=None, data_key='data', seq_tag_key='id', features=None,
               **kwargs):
    super(HuggingfaceDataset, self).__init__(**kwargs)

    self._seq_order = None

    self.dataset_opts = dataset_opts

    if isinstance(map_func, str):
      from returnn.config import get_global_config
      config = get_global_config(raise_exception=False)
      map_func = config.typed_value(map_func)

    if map_func_args is not None:
      map_func = map_func(**map_func_args)
    self.map_func = map_func
    self.dataset = None
    self.data_key = data_key
    self.seq_tag_key = seq_tag_key

    self.feature_keys = features

    self.data_dtype = {}

  def initialize(self):
    # Load the dataset
    import datasets
    if isinstance(self.dataset_opts, dict):
      self.dataset = datasets.load_dataset(**self.dataset_opts)
    else:
      self.dataset = datasets.load_from_disk(self.dataset_opts)
[...]

PR #1765, opened 2025-09-19
`PostprocessingDataset`: add support for multiprocessing

### LINKED ISSUE #1701: PostprocessingDataset with multi-processing
This could be an alternative to `MultiProcDataset`. In most cases (`OggZipDataset` etc), the data loading part of the dataset is not really the bottleneck, but any postprocessing is the bottleneck, and the reason to use `MultiProcDataset`.

So, the idea is to have a single source dataset (e.g. `OggZipDataset`), but then only do the post-processing with multi-processing.

For that, we can add such functionality to `PostprocessingDataset`, some `num_workers: int` option. If not used, it would do everything in the process itself, and otherwise, it would spawn the number of worker procs. This should be fairly straightforward. At least for the `map_seq` case. For `map_seq_stream` probably not really, but we don't need that.

My use case would be a standard `OggZipDataset`, but then doing the speed perturbation with `PostprocessingDataset` with multi-processing.

(cc @NeoLegends @dorian-K)

NeoLegends:
Closes #1701 

~~Currently fails due to #1763.~~

### DIFF
--- a/returnn/datasets/postprocessing.py
+++ b/returnn/datasets/postprocessing.py
@@ -475,7 +475,7 @@ class PostprocessingDataset(CachedDataset2):
 
         def _maybe_distrib_seq(*, timeout=0.1):
             assert timeout >= 0.0
-            # do not block indefinetely to periodically check the quit_event
+            # do not block indefinitely to periodically check the quit_event
             ready_conns, _, _ = select.select(worker_conns, [], [], timeout)
             assert len(worker_conns) == len(caches)
             for child_queue, cache in zip(worker_conns, caches):


returnn/datasets/postprocessing.py
```diff
@@ -31,8 +45,14 @@ class PostprocessingDataset(CachedDataset2):
     SpecAugment or speed perturbation into the data loading pipeline.
 
     The integration into the data loading pipeline makes it easy to distribute the
-    data processing work across multiple CPU cores using `MultiProcDataset` and in
-    turn frees the GPU from data preprocessing tasks.
+    data processing work across multiple CPU cores and in turn frees the GPU from
+    data preprocessing tasks.
+
```
hannah220, line 50: If I understand correctly, the former (`MultiProcDatase`) uses multi-processing and the latter (with `num_workers`) uses multi-_threading_, so "Multiprocessing can either be done using ..." is misleading?

NeoLegends: Correct!

NeoLegends:
Okay the test crashes due to some internal detail of `SubProcCopyGlobalConfigPreInitFunc`. Locally I cannot reproduce this error.

returnn/datasets/postprocessing.py
```diff
@@ -205,10 +252,59 @@ def init_seq_order(
             self._num_seqs = 0
             return True
 
-        self._rng = RandomState(self._get_random_seed_for_epoch(epoch=epoch))
-        assert self._dataset is not None
-        self._dataset.init_seq_order(epoch=epoch, seq_list=seq_list, seq_order=seq_order)
-        self._data_iter = enumerate(self._build_mapping_iter())
+        if self._num_workers > 0:
```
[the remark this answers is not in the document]
NeoLegends: @albertz see here for why I made a subclass instead

albertz:
Let's continue the discussion on separate class here (whether to introduce the new `MultiProcPostprocessingDataset`, or whether to add `num_workers` as an option to `PostprocessingDataset`):

> @albertz see [here](https://github.com/rwth-i6/returnn/pull/1765#discussion_r2369367139) for why I made a subclass instead. I did not like how the multiprocessing details bled into the existing, straightforward implementation and made it less straightforward. There were class fields that were only used for one of the two cases but not the other, the data iterator would be dynamically either one or the other variant, etc.

I did not have time to look at the implementation in detail yet.

First, if `MultiProcPostprocessingDataset` derives from `PostprocessingDataset`, that usually makes the code more difficult to follow (you need to read the code of two separate classes to understand what's going on; it's always better to keep code as local as possible, not having too deep nested structures, not too much spreaded code). So, while the code might be cleaner in some aspects, this is one aspect which is not better (in terms of readability).

But I wonder, even if you say this is cleaner, if this is a good argument. I think a higher priority would be that it's simple to use for the user. That the user directly finds this feature. To me, as a user, I would expect that it is just an option for `PostprocessingDataset`. (See for example Torch `DataLoader`: `num_workers` is just an option here. There is no separate `MultiProcDataLoader` to use that feature.)

If you don't fully agree with this argument, maybe let's get the opinion from some others here. E.g. @JackTemaki, @Icemole.

Btw, please restart tests when they fail due to some random hiccup.

NeoLegends:
> It is very suspicious that it's mostly only your newly added test which fails here.

Yeah, I think you are right. I think it's to do with the pickling/unpickling of the queues, which contain semaphores which contain some sort of locking mechanism. The unpickling of that locking mechanism fails. I have since managed to reproduce this error locally.

NeoLegends:
Ah, I guess it's the `NonDaemonicSpawnProcess` which pickles/unpickles the process contents. Maybe I don't even need this here.

NeoLegends:
I found
- https://stackoverflow.com/questions/75193175/why-i-cant-use-multiprocessing-queue-with-processpoolexecutor
- https://stackoverflow.com/questions/75233794/how-is-the-multiprocessing-queue-instance-serialized-when-passed-as-an-argument/75247561#75247561
now working through this. Seems not trivial.

NeoLegends:
For some reason the implementation with pipes just starts to hang at the end of the dataset. I'm trying to figure out why that happens. I'm getting strange KeyboardInterrupts in the worker processes that I don't know where they come from.

albertz:
You can check which cache file it uses (in the CI logs), and delete the corresponding cache file (in the GitHub Actions web interface).

tests/test_Dataset.py
```diff
@@ -1267,6 +1267,34 @@ def _repeat2(input_iter: Iterator[TensorDict], **kwargs) -> Iterator[TensorDict]
...
 
+def _repeat2(input_iter: Iterator[TensorDict], **kwargs) -> Iterator[TensorDict]:
+    for tdict in input_iter:
+        yield tdict
+        yield tdict
+
+
+def test_MultiProcPostprocessingDataset():
+    _demo_txt = "some utterance text that has a few words"
+    with create_ogg_zip_txt_only_dataset_opts(text=_demo_txt) as sub_ds_opts:
```
albertz, line 1278: This dataset only has a single seq, and then via your `_repeat2`, you get 2 seqs. I think for this test, better use some dataset with more seqs.

Actually, the test right now would not even really make use of the second worker at all? It would put the single seq to worker 1, and worker 1 would then call `map_seq_stream`, and worker 2 would do nothing.

NeoLegends:
I can certainly make the postprocessing so fast again that a single proc no longer suffices to load the data from the disk. In this case the user must switch the wrapped dataset to MultiProcDataset to apply some parallelism there as well.

anon-dir:
> I can certainly make the postprocessing (in my case just seq concat + speed perturbation) so fast now that a single proc no longer suffices to load the data from the disk. My feeder process bottlenecks on the HDF dataset internals (shown by py-spy). In this case the user must switch the wrapped dataset to MultiProcDataset to apply some parallelism to the raw data loading as well.

But I wonder, the bottleneck is IO speed then, or not? And if the bottleneck is IO, then MultiProcDataset will not help you there. If the bottleneck is not IO, what is it? What parts of HDF dataset internals? Reading some data from HDF does not really involve much (any?) overhead? Or if it does, maybe we should go away from HDF to Arrow or so? (See also #1257.) But this is maybe all going to be a bit off-topic here for the PR. I guess in `PostprocessingDataset`, the distribution of data from the source dataset to the workers cannot really be improved much more?

NeoLegends:
Yes, I think the concerns around HDF vs arrow are off-topic. I just noticed that the process that holds the feeder thread was close to 100% CPU utilization, and when you ran a py-spy dump those dumps pointed into the HDF routines.

tests/test_Dataset.py
```diff
@@ -1267,6 +1267,45 @@ def _repeat2(input_iter: Iterator[TensorDict], **kwargs) -> Iterator[TensorDict]
...
+
+def test_MultiProcPostprocessingDataset():
+    from test_HDFDataset import generate_hdf_from_dummy
+
+    num_hdfs = 20
+    num_seqs = 23
+    total_num_seqs = num_hdfs * num_seqs
+    total_num_seqs_pp = 2 * total_num_seqs
+    # Create a few HDF files such that we can easily verify the data later.
+    hdf_files = [generate_hdf_from_dummy() for _ in range(num_hdfs)]
```
albertz, line 1284: This is a bit problematic. You will have duplicates then in the seq tags. It might not really matter, but I wonder.

Why to use HDF at all here? Is this relevant for the test? Why not use just sth like this:
```python
{"class": "DummyDataset", "input_dim": 13, "output_dim": 7, "num_seqs": total_num_seqs, "seq_len": 17}
```

NeoLegends:
image

Hm, memory usage is not all that low actually. I expected these workers to be much much lower, but they're at almost 4G each and I just had another case (without screenshot) where they were at 5G. Maybe this feature is not helping at all! The postprocessor does not keep much intermediate state, why is the memory usage so high?

tests/test_Dataset.py
NeoLegends: Yeah, is fine. The test does not care about this 👍

NeoLegends:
Flamegraph of memory over time: [mem-2765784.html](https://github.com/user-attachments/files/22596947/mem-2765784.html)

NeoLegends:
image

I hope I'm reading this right. Most of the memory, as expected, is allocated by the unpickler when it receives data from the feeder thread. This makes sense.

Memory usage is growing over time, until at the 13:32 mark I see a massive drop in heap size. This looks like a `gc.collect()`. So from this I read that there are no leaks, but that maybe this high memory usage is just caused by the GC not collecting often enough? Seems strange to me, but easy to verify if that is the reason.

NeoLegends:
It seems like manually calling `gc.collect()` after every 1000 (just an example number I picked, must not be too low to avoid slowing down the processor too much) segments keeps the resident size at ~2GB. So it is a GC issue in a way.

albertz:
> With buf_size=1 and frequent GC the memory usage is at about 400MB.

This seems still a bit too much to me. Where exactly is it consumed? As said, I would expect more like 40MB.

NeoLegends:
> Buffer size 200 seems way too much. I would expect sth like 1-10 is enough here?

Yeah, maybe. I tend to use larger buffers to make sure the data pipeline never stalls.

> This seems still a bit too much to me. Where exactly is it consumed? As said, I would expect more like 40MB.

Hm, maybe. But I also don't think it's that problematic anymore then, maybe digging even deeper into where this memory is spent is "off topic" for this PR?  This 400MB is the resident size, and as seen is about 2x the heap size here.

Here is the flamegraph: 
image

[mem-2769354.html](https://github.com/user-attachments/files/22598038/mem-2769354.html)

NeoLegends:
https://stackoverflow.com/a/25334520/2000501

Oh, wow. One class instance (without slots) apparently costs at least 500 bytes. I expected python to use _some_ memory, but did not expect it to be that wasteful.

NeoLegends:
I guess we're done here then. No further digging required.

NeoLegends:
wrt. the gc.collection parameter, I am inclined to add a lower bound like e.g. 100 seqs to the value as well. Reason being that if you use buf_size=1, you will end in the pathological case where you gc.collect() after every seq, which is going to be a massive slowdown. I want to make this reasonably easy for the user, and so that they don't have to think much about this config knob at all.

NeoLegends:
Actually, I was thinking about what you said wrt. the process consuming 30MB of memory after having imported numpy. This sounds fine to me, but I realized this does not actually say anything about the memory usage when actually doing work similar to what is being done here.

I think I have a simple reproduction case that shows the behavior is universal to python. If I let it run while checking htop, resident memory size of the parent (!) goes up to ~100MB even though all the script is doing is generating tensors of a fixed size and sending them back and forth:

```python
from collections import deque
import multiprocessing
import numpy as np
import time


vocab_dim = 10_000
sr = 16_000


def main():
    parent, child = multiprocessing.Pipe()
    worker_process = multiprocessing.Process(target=worker, args=(child,))
    worker_process.start()
    child.close()

    start = time.time()

    try:
        while True:
            now = time.time()
            if now - start > 60:
                break

            t_dict = {
                "a": np.random.rand(30 * sr, 1),
                "b": (np.random.rand(30, 1) * vocab_dim).astype(np.int32),
            }
            parent.send(t_dict)
            result = parent.recv()
    except KeyboardInterrupt:
        pass
    finally:
        parent.send(None)
        worker_process.join()


def worker(child_conn):
    try:
        buf = deque()  # simulate worker proc buf
        while True:
            data = child_conn.recv()
            if data is None:
                break

            buf.append(data)
            if len(buf) < 10:
                child_conn.send({})
                continue

            data = buf.popleft()
            child_conn.send(data)
    except KeyboardInterrupt:
        pass


if __name__ == "__main__":
    main()
```

Parent: 
image
Child: 
image

So w/ additional overhead/complexity from passing around tensor dicts containing `rf.Tensor` and doing "actual" postprocessing I think the difference between 100MB and 400MB resident size is easily explained. That would mean this PR is fine and can be merged, and that we should perhaps tighten the GC in the worker subprocesses a little, so that resident size is lower.

albertz:
First of all, let us not do any sort of GC tuning. I think that is a bit meaningless and might have very suboptimal performance. The CPython people already tuned the GC well enough.

> resident memory size of the parent (!) goes up to ~100MB

I don't understand. Why do you care about the parent? We are discussing the memory of the sub processes here, or not?

> I think the difference between 100MB (from this run) and 400MB (from the previous, real RETURNN run) resident size is easily explained

Why? I don't see how it is explained? Also, the 400MB is the mem of the sub proc, and here you say it is the mem of the parent, so the comparison anyway does not make sense?

I think it's good to have your standalone script to play around with. Maybe in the subproc, do sth similar to what you were also testing in RETURNN. (What was it? Speed perturbation via resampy?) And then check the mem of the sub proc. Also, in addition, maybe do the same but without using `multiprocessing`, just direct. I want to see how much it would consume without the additional `multiprocessing` logic.

returnn/datasets/postprocessing.py
```diff
@@ -399,6 +416,495 @@ def _make_tensor_template_from_input(self, data_key: str) -> Tensor:
...
+
+        caches: List[deque[TensorDict]] = [deque() for _ in range(len(worker_conns))]
+
+        def _any_conn_ready() -> bool:
+            ready, _, _ = select.select(worker_conns, [], [], 0)
+            return len(ready) > 0
+
+        def _maybe_distrib_seq(*, timeout=0.1):
+            assert timeout >= 0.0
+            # do not block indefinetely to periodically check the quit_event
```
Icemole:
```suggestion
            # do not block indefinitely to periodically check the quit_event
```

albertz:
(Btw, not saying that we should not go forward with this PR here, but considering GIL-free / free-threaded Python ([since Python 3.14](https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-free-threaded-cpython)), it's maybe possible to do the same thing without multiprocessing, but with multi-threading instead, avoiding all the overhead.)

Issue #1730, opened 2025-06-19
RF PT could improve memory consumption (in general, but see attention with rel.pos.enc. as example)


albertz:
Download the attached zipped pickle file: [dlm_traindata_debug_snapshot_002.pickle.zip](https://github.com/user-attachments/files/20816357/dlm_traindata_debug_snapshot_002.pickle.zip) (via @dorian-K)
Unzip it.
Drag & drop the pickle file to here: https://docs.pytorch.org/memory_viz
That looks like this:
<img width="1197" alt="Image" src="https://github.com/user-attachments/assets/4f702cc3-23aa-4472-b412-addf1bb61e81" />

None of the functions/ops in RF are done inplace. As you can see, this can result in quite heavy memory allocations. E.g. in the attention, the `scores *= self.key_dim_per_head.dimension**-0.5` will make a copy. Or the `pad` in `_rel_pos_enc_shift` allocates a big 4.5GB tensor here.

Unfortunately, I think any solution to improve this problem will be quite involved/complex. We somehow need to figure out where we can safely do inplace. Sometimes we can know this in advance (e.g. check whether grad is required, or so; still not totally trivial, will add some overhead), but often, we can not, and things like `pad` also need further work. Maybe sth like `torch.compile` (or the equivalent in JAX; or sth equivalent on RF side) could do that, but that also would be a huge effort. Frameworks like graph-based TF or JAX probably partly solve this, but not sure if they solve all of it (and if they don't, then it will also be difficult).

PR #1789, opened 2025-11-03
Allow some attention matrices to be garbage collected

dorian-K:
Somewhat related to issue #1730

Explicitly setting these variables to `None` allows the python garbage collector to free the them and thus decreases memory consumption, otherwise the memory would be (unnecessarily) held until the function returns.

Of course this only works when the matrices don't require grad, during training this PR doesn't do anything.

Theoretically this change could be applied to almost all intermediate result variables, but it would also make the code really ugly. Maybe there should be a different, more general fix? Here I only picked those matrices that had by far the biggest memory footprint in my use case

### DIFF
--- a/returnn/frontend/attention.py
+++ b/returnn/frontend/attention.py
@@ -483,7 +483,7 @@ class RelPosSelfAttention(SelfAttentionBase):
         matrix_bd = _rel_pos_enc_shift(matrix_bd, axis, pos_emb_spatial_dim, hist_dim)
 
         scores = matrix_ac + matrix_bd  # (batch, head, time1, time2)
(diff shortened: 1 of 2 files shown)


albertz:
> Explicitly setting these variables to `None` allows the python garbage collector to free the them and thus decreases memory consumption, otherwise the memory would be (unnecessarily) held until the function returns.

Did you really test this? Is this really an issue?

We mostly have memory problems in training. In training, this will have no effect (as you also say).

In recognition, I'm not sure if this really makes any difference. Also, the time that the actual memory is freed might still be delayed (maybe by the GC, or maybe by the PyTorch mem allocator, or so), and there is no difference in the end.

I'm hesitant to add this to all our code when there is no real difference.

PR #1798, opened 2025-12-12
Add torch scaled dot product attention (FlashAttention)

dorian-K:


albertz:
> tracers that use `energy` and `att_weights` are broken now, and there isn't a straightforward way to fix because even if we automatically fall back to the returnn impl once we detect a tracer, that implementation is in the backend class and not in `rf.dot_attention` so current implementations don't find those variables. As far as I can tell this is only used in attention weight analyses and a test in `test_rf_attention.py`, so not huge impact but still annoying

I think it is to be expected that using such tracers can never be reliable and stable. We don't guarantee that, and we don't need to guarantee that. So this is not really an issue.

We should still keep the existing tests. For that, there should be a flag (maybe only internal flag) to disable this and fall back to the current vanilla implementation.

PR #1830, opened 2026-08-03
Amuse optimizer

polycrit:
- Added an implementation of AMUSE optimizer (mirrored from authors' github repository and adapted for RETURNN  as returnn.torch.optim.amuse)
- Strip callables from the saved optimizer_opts metadata in updater.py so optimizer checkpoints are loadable under the torch >= 2.6 weights_only default

### DIFF
--- a/returnn/torch/optim/amuse.py
+++ b/returnn/torch/optim/amuse.py
@@ -1,0 +1,507 @@
+"""
+AMUSE optimizer <https://arxiv.org/html/2605.22432>
+
+Code adapted from https://github.com/kjeiun/amuse/
+"""
+
+import functools
+from typing import Any, Callable, Dict, List, Optional, Sequence
+
+import torch
+from torch.optim.optimizer import Optimizer
+
+UPDATE_TYPES = {"muon", "adamw", "sgd"}
+AUX_UPDATE_TYPES = {"adamw", "sgd"}
+
+
+@torch.no_grad()
+def zeropower_via_newtonschulz5(
+    G: torch.Tensor,
+    steps: int = 5,
+) -> torch.Tensor:
+    assert G.ndim >= 2
+    a, b, c = 3.4445, -4.7750, 2.0315
+
+    X = G.bfloat16()
+    transposed = False
+    if G.size(-2) > G.size(-1):
+        X = X.mT
+        transposed = True
+
+    X = X / (X.norm(dim=(-2, -1), keepdim=True) + 1e-7)
+
+    for _ in range(steps):
+        A = X @ X.mT
+        B = b * A + c * (A @ A)
+        X = a * X + B @ X
+
+    if transposed:
+        X = X.mT
+    return X
+
+
+@torch.no_grad()
+def muon_update(
+    grad: torch.Tensor,
+    momentum: torch.Tensor,
+    beta: float = 0.95,
+    aux_update_type: str = "adamw",
+    nesterov: bool = True,
+) -> torch.Tensor:
+    momentum.lerp_(grad, 1 - beta)
+    update = grad.lerp_(momentum, beta) if nesterov else momentum
+
+    if update.ndim == 4:
+        update = update.view(len(update), -1)
+
+    update = zeropower_via_newtonschulz5(update)
+
+    if aux_update_type == "adamw":
+        # Scaling used in the AdamW-aux AMUSE setting.
+        update *= 0.2 * max(update.size(0), update.size(1)) ** 0.5
+    elif aux_update_type == "sgd":
+        # Muon default scaling used when auxiliary layers are trained by SGD.
+        update *= max(1, update.size(-2) / update.size(-1)) ** 0.5
+    else:
+        raise ValueError(f"Invalid AMUSE aux_update_type: {aux_update_type}. " "Expected one of {'adamw', 'sgd'}.")
+
+    return update
+
+
+class AMUSE(Optimizer):
+    """
+    AMUSE optimizer.
+
+    State convention:
+    - p stores y while training.
+    - eval() converts y -> x using the current beta1.
+    - train() converts x -> y using the current beta1.
+    - state["z"] stores the anchor z.
+
+    Hyperparameters:
+    - beta1: initial y/x interpolation. During warmup beta1 is constant.
+    - rho: controls how quickly beta1 approaches 1 after warmup.
+      Higher rho pushes beta1 toward 1 faster, so y moves closer to x
+      faster. Lower rho keeps y farther from x for longer.
+    - r: polynomial power for the z/x averaging weights.
+    - weight_decay: decoupled decay applied to z for Muon and SGD paths.
+    - weight_decay_at_y: optional decay applied while p is still y.
+
+    Parameter groups:
+    - Matrix hidden-layer weights should use {"use_muon": True}. These use
+      Muon momentum and Newton-Schulz orthogonalization.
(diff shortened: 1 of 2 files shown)


albertz:
So currently we don't even have the much more popular Muon optimizer in RETURNN. I don't think adding AMUSE makes sense when we don't have Muon. But maybe we should add both then.

See the failing code style errors also:
```
  returnn/torch/optim/amuse.py:18: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:19: WEAK WARNING PyPep8NamingInspection: Argument name should be lowercase
  returnn/torch/optim/amuse.py:25: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:28: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:31: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:34: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:35: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:36: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:39: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:44: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:184: WEAK WARNING PyPep8NamingInspection: Variable in function should be lowercase
  returnn/torch/optim/amuse.py:201: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:212: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:223: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:257: WEAK WARNING PyAttributeOutsideInitInspection: Instance attribute beta1 defined outside __init__
  returnn/torch/optim/amuse.py:441: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:443: WARNING PyUnresolvedReferencesInspection: Cannot find reference 'train' in 'None'
  returnn/torch/optim/amuse.py:446: WEAK WARNING PyMissingOrEmptyDocstringInspection: Missing docstring
  returnn/torch/optim/amuse.py:448: WARNING PyUnresolvedReferencesInspection: Cannot find reference 'eval' in 'None'
  returnn/torch/optim/amuse.py:464: ERROR PyUnresolvedReferencesInspection: Unresolved reference 'CodeWrapper'
  returnn/torch/optim/amuse.py:464: ERROR PyUnresolvedReferencesInspection: Unresolved reference 'i6_core'
  returnn/torch/optim/amuse.py:492: ERROR PyUnresolvedReferencesInspection: Unresolved reference 'CallImport'
  returnn/torch/optim/amuse.py:492: ERROR PyUnresolvedReferencesInspection: Unresolved reference 'Import'
  returnn/torch/optim/amuse.py:492: ERROR PyUnresolvedReferencesInspection: Unresolved reference 'i6_core'
```

polycrit:
> So currently we don't even have the much more popular Muon optimizer in RETURNN. I don't think adding AMUSE makes sense when we don't have Muon. But maybe we should add both then.

was also thinking about this, makes total sense, I'll add it as well

albertz:
Re Muon, I always use `i6_experiments.users.zeyer.experiments.exp2024_04_23_baselines.optim_ext.muon.Muon`. Maybe you can copy that code?

polycrit:
> Re Muon, I always use `i6_experiments.users.zeyer.experiments.exp2024_04_23_baselines.optim_ext.muon.Muon`. Maybe you can copy that code?

I just found oud there's an already implemented Muon from pytorch (https://docs.pytorch.org/docs/2.13/generated/torch.optim.Muon.html). So I think maybe we don't need it separately implemented after all? Just a small adapter.

albertz:
Ah Muon seems to exist since PyTorch 2.10. Hm so good question. Maybe you are right.

polycrit:
> Ah Muon seems to exist since PyTorch 2.10. Hm so good question. Maybe you are right.

The problem is that in RETURNN you can't do something like:

```
muon = torch.optim.Muon(hidden_matrix_params, ...)
adamw = torch.optim.AdamW(other_params, ...)

loss.backward()
muon.step()
adamw.step()
```

RETURNN’s PyTorch updater currently manages one optimizer object. So what we could implement there is a compact Muon/AdamW adapter that presents Muon and AdamW as that single object.

albertz:
Doesn't the Torch Muon already handle this automatically? At least `i6_experiments.users.zeyer.experiments.exp2024_04_23_baselines.optim_ext.muon.Muon` does. (Hybrid Muon (2-D hidden weights) + AdamW (everything else))

albertz:
Btw, I think it was discussed maybe in the past whether we somehow should have a possibility to use multiple optimizers for different params. Not sure?

polycrit:
> Doesn't the Torch Muon already handle this automatically? At least `i6_experiments.users.zeyer.experiments.exp2024_04_23_baselines.optim_ext.muon.Muon` does. (Hybrid Muon (2-D hidden weights) + AdamW (everything else))

As far I understand, it doesn't, you have to explicitly call two separate optimizers.

polycrit:
> Btw, I think it was discussed maybe in the past whether we somehow should have a possibility to use multiple optimizers for different params. Not sure?

Can't remember this, but I think it's generally a very good idea, given we want to experiment with such multi-level optimizers. I will look into it.

polycrit:
@albertz I imagine we can create an interface that does something like this:

```
optimizer = {
    "class": "multi",
    "optimizers": [
        {
            "class": "amuse",
            "update_type": "muon",
            "params_filter": make_hidden_matrix_filter(),
            "momentum": 0.95,
            "warmup_steps": 10000,
        },
        {
            "class": "amuse",
            "update_type": "adamw",
            "learning_rate_multiplier": 0.015,
            "warmup_steps": 10000,
        },
    ],
}
```

or:

```
optimizer = {
    "class": "multi",
    "optimizers": [
        {
            "class": "muon",
            "params_filter": make_hidden_matrix_filter(),
            "momentum": 0.95,
            "adjust_lr_fn": "match_rms_adamw",
            "weight_decay": 1e-2,
        },
        {
            "class": "adamw",
            "learning_rate_multiplier": 0.015, 
            "weight_decay": 1e-2,
            "epsilon": 1e-8,
        },
    ],
}
```

`params_filter` defines parameter ownership here, which accepts a predicate function like:

```
def my_filter(*, full_param_name: str, param: torch.nn.Parameter,
              module: torch.nn.Module, rf_module, **_kwargs) -> bool:
    ...
```

similar to RETURNN's existing `weight_decay_custom_include_check` (updater.py:619-623). Example for Muon:

```
def muon_filter(*, full_param_name, param, **_kw):
    lname = full_param_name.lower()
    return param.dim() == 2 and not any(s in lname for s in ("embed", "logit", "head"))
```

albertz:
Yea looks good. Feel free to implement that.

